PR #49, opened 2020-02-19
Add default colors to CSS

tastytea:
`background-color` and `color` are specified in some places, but there is no default. If the fallback colors in the browsers are not what is expected, it looks weird.

Before: screenshot_2020-02-19T22-42-14

After: screenshot_2020-02-19T22-42-04

To reproduce, [configure your browser](https://support.mozilla.org/en-US/kb/change-fonts-and-colors-websites-use#w_change-font-color) to use white text on black background as fallback.

### DIFF
--- a/fedireads/static/format.css
+++ b/fedireads/static/format.css
@@ -6,6 +6,19 @@
     font-family: sans-serif;
 }
 
+html {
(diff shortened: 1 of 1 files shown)


mouse-reeve:
thank you!

PR #62, opened 2020-02-22
Always pass follow/unfollow parameter using POST.

cthulahoops:
Fixes buttons in followers list.

(Cool project!)


mouse-reeve:
thank you so much for fixing this!

PR #65, opened 2020-02-24
Following

cthulahoops:
I managed to get myself very confused with subject vs object vs actor, but I think this is right now.

subject (actor in the activitypub json) should be the person doing the following.
object should be the person being followed.

Hopefully, I'm not still confused. :)


mouse-reeve:
(life has caught up with me a bit but I'm hoping to review this tonight or this weekend! thank you so much for all your contributions)

cthulahoops:
No problem! I've been away, so I haven't had much chance to think more about this either.

I'm still excited about this project and hope to do some more soon. Let me know if there's anything in particular you'd like me to look at. I don't want to overwhelm you with pull requests.

PR #83, opened 2020-03-11
Approval table split

cthulahoops:


cthulahoops:
(Repush was just to get the issues reference right.)

Issue #82, opened 2020-03-10
Approvable Followers


cthulahoops:
I'm trying to figure out the best way to follower approval to work, and not sure of the best way to proceed.

At the moment, we have a ManyToMany field followers, which joins through UserRelationship, but really followers should be a subset of this: relationships where status = 'follows'.

Approaches:

1) I want to create a field relationships and properties (follows, requested follow, blocks, etc.) which contain only a subset. This works for the simple case, but properties can't work quite like fields: you can't add to them, they don't work in complex queries (eg ```Q(user__followers=request.user)```). I might just not know Django well enough.

2) Just have the relationships field, and rewrite everything to explicitly check the relationship status everywhere. Going to bug prone. We'll treat a block as a follow eventually.

3) Split the relationships table into separate tables. It's a bit of a pain moving data between them, and harder to express constraints (a user shouldn't follow and request to follow a user at the same time), but will be very straightforward to implement and fast to query.

Leaning towards option 3. Any thoughts?

jgamble:
Thinking in terms of SQL (my default case), I would go with option three. It's not likely (I'm assuming) that you'd deal with more than one relationship at a time. Llist followers, list followed, list blocked can all be handled by single queries, and if you really need to deal with more than one at a time, the SQL join is simple.

My only caveat is that you will need admin-level sanity check queries. Tables will on occasion get out of sync, and checking for that (X is both followed and blocked) is important. One of my biggest beefs with tag-based apps is that they almost always never have a "list the untagged entries" option.

mouse-reeve:
I think I'm on team option 3 as well. I also want to think forward to blocking and muting.

I don't think think there's any reason to have the UserRelationship table at all given this; I created it thinking it would store all kinds of relationships, but as you said, that just creates very complicated tables. I'm thinking it makes sense to have every kind of relationship be a django-created join table (rather than an explicitly defined through table). Does that make sense?

jgamble:
My knowledge of django is zero, I'm afraid. I gave a look through the code and I kinda sorta get the table creation, but I'm afraid I'm only speak in SQL.

But this is a case where "keep the tables really minimal" (third normal form) really shines. Each type of relationship having its own table makes the most sense to me.

mouse-reeve:
[...]

PR #85, opened 2020-03-13
Follow request acceptance

cthulahoops:
Basic working interface for issue #82 now.

Still need to generate Reject activity. Probably need to store full follow request in the database to populate the Accept/Reject activity.

Really need to get a multi-node test setup to check this really works with multiple nodes.


cthulahoops:
I think that works properly now with rejections crossing between nodes. I'm thinking the next focus needs to be getting some testing with multiple nodes set up.

mouse-reeve:
I've been using my localhost vs fedireads-test.glitch.me for cross-instance testing, fwiw

mouse-reeve:
oh though I guess I'm the only one who can put code there so that's not helpful to anyone besides me, oops

mouse-reeve:
thank you for fixing my fake enums!

cthulahoops:
> oh though I guess I'm the only one who can put code there so that's not helpful to anyone besides me, oops

There's at least some testing I can run with control over one half of things, and I guess I can set up my own node some too. It should be possible to get two nodes both running on localhost, I just haven't really tried yet.

I'd really like to get some sort of automated testing in place. It's a lot of effort to test some of these interactions.

Issue #91, opened 2020-03-21
Local activities shouldn't have to wait on incoming POSTs


mouse-reeve:

mouse-reeve:
Duplicate of #40

Issue #40, opened 2020-02-19
Move blocking processes to background tasts


mouse-reeve:
Some things that should be in the background
- [ ] Broadcasting ActivityPub activities
- [ ] Incoming activities
- [ ] Goodreads import
- [ ] Automatic book data management

mouse-reeve:
this is probably the use case for Celery?

PR #95, opened 2020-03-23
Docker

### LINKED ISSUE #69: Deployment flow
This probably means dockerization??

mouse-reeve:
Fixes #69
Works on #91 and #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,17 +45,40 @@ This project is still in its very early stages, but these are the higher-level f
 But this isn't a set in stone, unchangeable list, so if you have ideas about how this could be tweaked, changed, or improved, please open an issue and start a conversation about it.
 
 ## Setting up the developer environment
+You will need postgres installed and running on your computer.
 
-Install docker and run:
 ``` bash
-docker-compose build
-docker-compose up
-docker-compose exec web python manage.py migrate
+python3 -m venv venv
+source venv/bin/activate
+pip install -r requirements.txt
+createdb fedireads
 ```
+
+Create the psql user in `psql fedireads`:
+``` psql
+CREATE ROLE fedireads WITH LOGIN PASSWORD 'fedireads';
+GRANT ALL PRIVILEGES ON DATABASE fedireads TO fedireads;
+```
+
+Initialize the database (or, more specifically, delete the existing database, run migrations, and start fresh):
+``` bash
+./rebuilddb.sh
+```
+This creates two users, `mouse` with password `password123` and `rat` with password `ratword`.
 
 And go to the app at `localhost:8000`
 
 For most testing, you'll want to use ngrok. Remember to set the DOMAIN in `.env` to your ngrok domain.
+
+
+### Docker
+You can also run the application in a docker container. You'll have to install the Docker and docker-compose:
+
+```bash
+docker-compose build
+docker-compose up
+docker-compose exec web python manage.py migrate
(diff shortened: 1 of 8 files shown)


mouse-reeve:
I'm getting the error:
`ERROR: No matching distribution found for celery==4.4.2 (from -r requirements.txt (line 1))`
when I try to deploy on ubuntu so I would not yet consider it a repeatable deploy.

mouse-reeve:
@cthulahoops I'm still figuring out how to make this run consistently and how to actually use celery, but it's mostly working. since it'll effect everyone's workflow, I welcome however much review/input you have and are interested in giving, but of course no pressure!

cthulahoops:
This looks reasonable to me. I haven't managed to get it build yet. I'll have another look in the morning and see if I can get it working.

In general, I'd prefer to have docker as an option rather than the only way to run the project as I've found it a bit unwieldy in development. I'm not sure how practical that is once we're depending on celery, rabbitmq, etc. I'm not very familiar with celery either.

mouse-reeve:
Agreed on having docker be optional. I don't know for sure yet if I can keep celery as an optional/prod-only dependency

cthulahoops:
> Agreed on having docker be optional. I don't know for sure yet if I can keep celery as an optional/prod-only dependency

Oh yeah! I hadn't thought of that, but it's got to be possible

Similarly for local messages activitypub allows us to bypass https and deliver the messages internally which might make testing easier.

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+version: '3'
+
+services:
+    db:
+        image: postgres
+        env_file: .env
+        volumes:
+          - pgdata:/var/lib/posgresql/data
```
cthulahoops: Should be: ```/var/lib/pos**t**gresql/data```?

Even after fixing I get stuck at:

```
db_1             | running bootstrap script ... ok
db_1             | performing post-bootstrap initialization ... 2020-03-24 14:28:54.434 UTC [40] FATAL:  zero-length delimited identifier at or near """" at character 12
db_1             | 2020-03-24 14:28:54.434 UTC [40] STATEMENT:  ALTER USER ""postgres"" WITH PASSWORD E'"fedireads"';
db_1             | 	
db_1             | child process exited with exit code 1
db_1             | initdb: removing contents of data directory "/var/lib/postgresql/data"
fedireads_db_1 exited with code 1
```

mouse-reeve: after my last commit that fixes this, I re-ran `docker-compose build` and `docker-compose up` and this error went away for me🤞

mouse-reeve: I also changed my database type in .env from sqlite to postgres, which may be related

cthulahoops: Yeah, it's working for me too now. :)

README.md
```diff
@@ -45,19 +45,42 @@ This project is still in its very early stages, but these are the higher-level f
...
 
+### Docker
+You can also run the application in a docker container. You'll have to install the Docker and docker-compose:
+
+```bash
+docker-compose build
+docker-compose up
+docker-compose exec web python manage.py migrate
+```
+
```
cthulahoops: Need instructions to invoke init_db too? (Not sure the best way: ```docker-compose exec web python manage.py shell -c 'import init_db'``` perhaps?)

Issue #92, opened 2020-03-22
Create logged-out friendly views of pages


mouse-reeve:
Right now they throw a bunch of errors

PR #103, opened 2020-03-27
Book & books while not logged in

cthulahoops:
Allow viewing of book page and discover books without logging in.

I don't know if this is the correct fix, or if we should just add a required_login decorator to these views but it doesn't seem unreasonable to allow people to browse books.

### DIFF
--- a/fedireads/templates/book.html
+++ b/fedireads/templates/book.html
@@ -25,6 +25,7 @@
     </div>
 </div>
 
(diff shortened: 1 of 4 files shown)


mouse-reeve:
I think this is the right fix! It should also at some point have an activitypub view, so it can't be behind login

mouse-reeve:
Works on #92

Issue #100, opened 2020-03-25
Implement boost


mouse-reeve:
it's there in the UI, but does not currently do anything

PR #116, opened 2020-03-30
Boosts - handle url, store in database, send, notify.

cthulahoops:
Start of #100


cthulahoops:
(Not sure this is right, I think a boost is a bit more like a reply than a favourite.)

PR #120, opened 2020-03-31
Create celery tasks

### LINKED ISSUE #40: Move blocking processes to background tasts

mouse-reeve:
Fixes #40


fedireads/outgoing.py
```diff
@@ -81,7 +81,7 @@ def handle_account_search(query):
 def handle_follow(user, to_follow):
     ''' someone local wants to follow someone '''
     activity = activitypub.get_follow_request(user, to_follow)
-    errors = broadcast(user, activity, [to_follow.inbox])
+    errors = broadcast.delay(user.id, activity, [to_follow.inbox])
```
cthulahoops, line 84: Rather than rewriting every caller would it be better to leave these calls as is and call the delayed broadcast from the broadcast function?

cthulahoops:
Looks reasonable to me, and I'm excited to see how well this works in practice.

I presume the switch to redis is because it is simpler/more light weight?

mouse-reeve:
Yeah, I got extensive help configuring it from someone at Recurse and switched at his recommendation.

Issue #121, opened 2020-04-01
Reviews without ratings


cthulahoops:
Should it be possible to write a review without providing a numeric rating? If so, should we allow zero star rating, or start the star rating at 1? (The ui could get confusing if we have to distinguish between zero and no rating.)

Looking at my goodreads export, it looks like books I haven't rated show as zero. (I'm hope I'm not accidentally pulling down the average on books just because I didn't rate them.)

cthulahoops:
Likewise ratings without reviews.

mouse-reeve:
I intended 0 stars to mean "not rated," I just didn't catch the UI up with that (so they bring down averages currently, although they shouldn't).  I believe that's what goodreads does too! And I do think you should be able to leave both reviews without ratings and ratings without reviews.

cthulahoops:
Awesome! I suggest we use nulls to represent no rating, then the average will just work. (I think.)

mouse-reeve:
having looked at rating-without-review, I'm starting to think it should be a separate table. `name` is a required field on `Review`, and I feel like without that, the validation and activitypub serialization is going to get gnarly.

cthulahoops:
I thought about that, but I think where both exist there's a strong link between the rating and review: they'll be displayed together, be a single activity, it's the rating for the review... The same table with nullable fields seems sensible.

Is there anything stopping the activitypub fields being optional?

Otherwise, I guess it's two tables, and separate Review and Rating statuses.

mouse-reeve:
I think if the fields are optional, it could lead to display problems across different platforms. But maybe the best solution is just to serialize ratings-only activities differently...

cthulahoops:
That makes sense, and we can easily tweak the serialisation to deal with any problems.

PR #122, opened 2020-04-01
Import reviews

cthulahoops:
Currently only generating reviews where there is both a non-zero rating and a review (waiting on #121). I'm generating activity for every review because I think even imported reviews are interesting enough to show in the activity feed.

### DIFF
--- a/fedireads/goodreads_import.py
+++ b/fedireads/goodreads_import.py
@@ -88,6 +88,14 @@ class GoodreadsItem:
         if self.line['Exclusive Shelf']:
             return GOODREADS_SHELVES[self.line['Exclusive Shelf']]
 
+    @property
+    def review(self):
+        return self.line['My Review']
+
+    @property
+    def rating(self):
+        return int(self.line['My Rating'])
+
     def __repr__(self):
         return "<GoodreadsItem {!r}>".format(self.line['Title'])
 
--- a/fedireads/goodreads_import.py
+++ b/fedireads/goodreads_import.py
@@ -22,8 +22,7 @@ def unquote_string(text):
     match = re.match(r'="([^"]*)"', text)
     if match:
         return match.group(1)
-    else:
-        return text
+    return text
 
 
 def construct_search_term(title, author):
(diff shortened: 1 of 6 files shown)


fedireads/goodreads_import.py
```diff
@@ -36,7 +35,7 @@ def construct_search_term(title, author):
     return ' '.join([title, author])
 
 
-class GoodreadsCsv(object):
```
mouse-reeve, line 39: is this a more modern syntax? or is there a functional difference?

fedireads/status.py
```diff
@@ -23,11 +23,8 @@ def create_review_from_activity(author, activity):
     return review
 
 
-def create_review(user, possible_book, name, content, rating):
+def create_review(user, book, name, content, rating):
```
mouse-reeve, line 26: This seems sensible to me but it will necessitate updating `create_review_from_activity` (and maybe others)

fedireads/goodreads_import.py
cthulahoops: There's no functional difference in Python 3, which is why pylint suggests the change. There is in Python 2, which is why I have a strong habit of always writing (object).

I think pylint is right as we never need to support python 2.

fedireads/status.py
cthulahoops: Oops, I thought I checked for other callers. I'll check again and fix. Thanks!

cthulahoops: I think it was just that one.

cthulahoops:
(Rebase on top of latest master.)

cthulahoops:
Getting there.

Not sure whether I should do something to generate titles (names?) for imported reviews as goodreads only has a reviews, no titles. Currently just providing the empty string (which will cause the non-fedireads formatting to see it as just a rating). Could use the first few words of the review followed by "...", or a placeholder like "imported review".

I note that we sanitize the body of the review but not the title, which I suspect should be fixed.

mouse-reeve:
I'd rather populate a default name, like `Review of "<title>" on Goodreads"

PR #123, opened 2020-04-02
Add model to store reading progress.

cthulahoops:


cthulahoops:
Adds model, gets data from imports, auto-generates entries when shelving books.

Really unsure on naming here. Didn't want to use words like "activity" which have meaning in activitypub, "progress", "completion", etc., only make sense in given states. "read" is has too many meanings. "read through" is sort of right but is somewhat taken from another context.

Names are hard.

mouse-reeve:
this really is a name conundrum, but I think ReadThrough makes sense, and I can't think of anything better :)

PR #129, opened 2020-04-03
Ratings without reviews

### LINKED ISSUE #121: Reviews without ratings

mouse-reeve:
Extremely pleased with myself for writing this without javascript. Fixes #121


mouse-reeve:
Fixes #121

PR #133, opened 2020-04-08
Add css dropdown to select alternative shelves.

cthulahoops:
Adds a little dropdown arrow to the shelf button to select arbitrary shelves.

This is pushing my css abilities a bit, but it basically works. Dropdown doesn't work right with shelf containers at top of feed (requires scrolling), and buttons shouldn't have a fixed width.

### DIFF
--- a/fedireads/templates/snippets/shelve_button.html
+++ b/fedireads/templates/snippets/shelve_button.html
@@ -1,18 +1,18 @@
 {% load fr_display %}
 {% if request.user.is_authenticated %}
 
-<div class="dropdown secondary">
+<div class="pulldown-button">
 <form name="shelve" action="/shelve/" method="post">
     {% csrf_token %}
     <input type="hidden" name="book" value="{{ book.id }}"></input>
     <input type="hidden" name="shelf" value="{% shelve_button_identifier book %}"></input>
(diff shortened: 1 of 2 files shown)


cthulahoops:
Embarrassingly easy to avoid the fixed width... :)

cthulahoops:
It turns out it's much simpler to put the styling on the containing div rather than styling two borders to look like one.

fedireads/templates/snippets/shelve_button.html
```diff
@@ -1,29 +1,22 @@
 {% load fr_display %}
 {% if request.user.is_authenticated %}
-{% if not pulldown %}
 
+<div class="dropdown secondary">
```
mouse-reeve, line 4: I think we can use the existing css classes for the user icon pulldown and this if you use `pulldown-container` instead of `dropdown`, `pulldown` for `dropdown-content` and put the buttons in the form in a `ul` / `li`

mouse-reeve: I can make this change if you prefer, since I'm probably more familiar with the css

cthulahoops: I wouldn't mind having a go at it to get more familiar with the css. The style is bit different, but I think you're right that it should be possible to unify them.

mouse-reeve: awesome, go for it 🙂

cthulahoops:
How does that look?

They don't unify completely because one uses buttons the other links, and because of the control as a whole is a button with pulldown menu to one side.

PR #150, opened 2020-04-20
Async import

cthulahoops:
Draft of #124.

Need to figure out the best way to notify the user of the results.


mouse-reeve:
I don't know how practical this is, but here's a thought:
each csv line is processed in a separate celery task, with a corresponding row in database table with an import id and user, import status (successful, failed, pending), csv line blob, and (when available) shelf and book foreign key. the import data expires and is deleted after 5 days (or whatever), to keep the table from blowing up with irrelevant data.
ui-wise, there's a status page for the import, which shows the progress on each book, which is linked from the upload success page

cthulahoops:
It's interesting.

I think there's two separate questions. Storing the csv in the database with progress reports/etc, and multiple celery tasks. I think the former has a lot of advantages. The latter I'm less convinced by, but is probably a less important detail.

The other advantage is that the user could manually select the correct book if it can't be determined automatically.

I'll have a go at implementing this.

Thank you!

cthulahoops:
I'm not sure I'm happy with everything, but I think this is a big improvement and a reasonable point to merge progress so far.

How does this look to you?

fedireads/goodreads_import.py
```diff
@@ -1,126 +1,61 @@
...
 
-
-# Mapping goodreads -> fedireads shelf titles.
-GOODREADS_SHELVES = {
-    'read': 'read',
-    'currently-reading': 'reading',
-    'to-read': 'to-read',
-}
 # TODO: remove or notify about this in the UI
 MAX_ENTRIES = 20
```
mouse-reeve, line 11: can we remove this limit safely now?

fedireads/models/import_job.py
```diff
@@ -0,0 +1,125 @@
...
+    @property
+    def date_read(self):
+        if self.data['Date Read']:
+            return dateutil.parser.parse(self.data['Date Read'])
+
+    @property
+    def reads(self):
+        return [ReadThrough(
+            # date_added isn't the start date, but maybe better than nothing.
+            start_date=self.date_added,
```
mouse-reeve, line 116: I think that if a book is on any goodreads shelf besides `currently-reading` with an `added_date` but not a `date_read`, it shouldn't be considered a `ReadThrough`? For example, if I see a book I read in the past and shelve it as "read" on goodreads without going through the process of starting and finishing on goodreads. In this case, I wouldn't want it to say that I started reading it on the date added.
Annoying that they don't include the started date in the data, though!

fedireads/templates/import_status.html
```diff
@@ -0,0 +1,66 @@
...
+                    Author
+                </th>
+                <th>
+                    Book
+                </th>
+            </tr>
+            {% for item in items %}
+            <tr>
+                <td>
+                    {% if item.book %}✓{% endif %}
```
mouse-reeve, line 46: this could also show the error message if a book fails to import

fedireads/templates/import_status.html
```diff
@@ -0,0 +1,66 @@
...
+                <th>
+                    Author
+                </th>
+                <th>
+                    Book
+                </th>
+            </tr>
+            {% for item in items %}
+            <tr>
+                <td>
```
mouse-reeve, line 45: I think it would make more sense visually to have this column at the end and the covers column here

fedireads/goodreads_import.py
cthulahoops: I'm not sure. I'm kind of wary of having it unlimited yet. I don't like the idea that someone can upload a million line file and make us spam openlibrary with requests. I might be worrying about nothing as I'm sure there are lots of other ways to cause problems.

The real fix is to rate limit outgoing requests.

I was going to look at this as part of the error handling. How about upping this to 500 for now?

fedireads/models/import_job.py
cthulahoops: It's very annoying. This was too optimistic of me. Fixed.

fedireads/templates/import_status.html
cthulahoops: I thought about writing this code, and even put the data in the database and then somehow failed to actually write it! :/ (Fixed.)

fedireads/templates/import_status.html
cthulahoops: Fixed.

fedireads/goodreads_import.py
mouse-reeve: oh good point! I'll ask Mek about rate limit/politeness settings on Monday

mouse-reeve: okay, I talked to Mek! he says it should be okay to make lots of requests synchronously, but things can get hairy with concurrent requests. He suggested using the data dump rather than querying the API -- I'm not sure how large it is, but it's worth considering I think (although outside the scope of this PR).

cthulahoops: Interesting. I think it might not be reasonable to expect every instance to have a full copy of all the book data, but it's good to have that option.

I think I'd still suggest upping the limit to 500 or a 1,000 so we test it on some real data and make sure that nothing breaks. (Mine export is 267 entries, I think.)

mouse-reeve: 👍 I support trying it with that limit, but let's wait until we've tested it before committing it

cthulahoops: I will do. I was trying it locally but got blocked by the weird id bug.

PR #154, opened 2020-04-26
Don't allow data source to change the primary key of an object.

cthulahoops:
This was a really confusing one to debug!

Openlibrary returns an id, the id gets changed, duplicate error gets raised as django attempts to duplicate the row.

This is the minimal fix for the bug. I think we should work from a whitelist of expected keys as there's lots of exciting things that could get overwritten by unexpected data here (eg: save).

* http://openlibrary.org/works/OL1940021W.json

### DIFF
--- a/fedireads/connectors/abstract_connector.py
+++ b/fedireads/connectors/abstract_connector.py
@@ -70,6 +70,9 @@ def update_from_mappings(obj, data, mappings):
         if not formatter:
             formatter = noop
 
+        if key == 'id':
+            continue
+
         if has_attr(obj, key):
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you! this is a gnarly one. I agree that whitelisting is the best call here.

cthulahoops:
Django's behaviour is really weird/interesting here. It issues the update statement for the new id, spots that the row doesn't exist and then sends the insert. So, if it the id was changed to match an existing object it would happily overwrite the data in the wrong row? That seems less than helpful.

PR #158, opened 2020-04-29
Local search endpoint

mouse-reeve:
- expand book data to make it more readily searchable
 - use postgres full text search
 - query local search connector before openlibrary/other connectors


fedireads/connectors/openlibrary.py
```diff
@@ -80,6 +80,7 @@ def get_or_create_book(self, olkey):
                 edition_data = pick_default_edition(edition_options)
                 key = edition_data.get('key').split('/')[-1]
                 edition = self.create_book(key, edition_data, models.Edition)
+                edition.default = True
```
cthulahoops, line 83: Rather than having default as boolean on edition, should default edition be a foreign key on work? That way we can enforce that each work has only one default edition, and looking up the default edition for a work is the common operation anyway?

cthulahoops:
Looks good!

PR #160, opened 2020-05-03
Separate out local and remote search results

mouse-reeve:
In addition to the UI change for the search page, this should reduce the openlibrary queries in goodreads import by returning a local result immediately from the search endpoint if it's available.

If you run this migration without using the `initdb` script, you *must* set `self_connector` to have `priority=1` and `local=True`.  It also helps to set the `name` field for the `openlibrary` connector to `OpenLibrary`


fedireads/models/import_job.py
```diff
@@ -67,19 +67,19 @@ def get_book_from_db_isbn(self):
 
     def get_book_from_isbn(self):
```
cthulahoops, line 68: Can ```get_book_from_db_isbn``` go away now? This should now search both sources in the correct order?

mouse-reeve: yes

mouse-reeve:
yep, we can add connectors to other instances! I think when a new book is discovered via another instance, that instance should be the canonical source by default.

PR #162, opened 2020-05-04
Fedireads connector

### LINKED ISSUE #54: putting books at the center of the interaction?
Hey! Thanks for you work, really awesome to see it federate already at this early stage!

As I was trying it out, a thought occurred to me of how a different interaction model would fit much better with the federated nature of fedireads. Currently the reviewers and their reviews+comments are actors that are followable across the fediverse. Books on the other hand are not, they are rather passive and even only exist through reference. However what if the book (or maybe a shelf) would also be an object that federates across the fediverse? 

In that model one could search for "http://fedireads-test.glitch.me/book/OL18704688W" through another fediverse client and comment it, review it, share it, without even needing a fedireads account.  Book objects could then have a history of reviews (both from fedireads and from other federated platforms), weighted ratings, and other metadata as part of the 'profile' of the book. 

It is then up to the community behind a particular fedireads instance (let's say specialized in 70s sci-fi) to moderate the comments on that particular book, making sure they are relevant for their context. Or maybe attributing more weight to a particular review because it is very insightful.  

Furthermore, a link could be made between the book profile hosted by community A and the book profile hosted by community B that indicates one is talking about the same book. That way you could have a way of indicating across the fediverse that different communities are talking about the same book, but that '70s sci-fi community' will have a different set of takes on that same book than 'science and technology academic study circle community'

Hope this train of thought is helpful!

mouse-reeve:
this is turning out to be a _lot_, my bad:
 - removes `fedireads_key` as an identifier
 - introduces the idea of `remote_id` to Book, which is a url identifier to a book on another instance
 - builds out the connector to load book data from other instances
 - adds more matching on keys (like isbn) to the book creation process to avoid duplicating books

It should pretty much fix #54


mouse-reeve:
sorry I'm merging this without giving anyone a chance to comment on it, it's just so large that it's hard to do anything else until it's merged

PR #165, opened 2020-05-09
Code cleanup and tests

mouse-reeve:
it turns out that when you write tests you find bugs????? annoying.

### DIFF
--- a/fedireads/connectors/abstract_connector.py
+++ b/fedireads/connectors/abstract_connector.py
@@ -267,17 +267,11 @@ def update_from_mappings(obj, data, mappings):
         if key == 'id':
             continue
 
-        if has_attr(obj, key):
+        try:
+            hasattr(obj, key)
+        except ValueError:
             obj.__setattr__(key, formatter(value))
     return obj
-
-
-def has_attr(obj, key):
-    ''' helper function to check if a model object has a key '''
-    try:
-        return hasattr(obj, key)
-    except ValueError:
-        return False
 
 
 def get_date(date_string):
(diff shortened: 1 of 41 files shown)


rkingett:
Oh yay! Does this mean it's almost ready for people to start hosting? :) By the way, I hope those accessibility changes weren't just for the demo website, but for the whole project. This way, everybody who hosts this will have an accessible product! Server.

mouse-reeve:
I don't know about "almost" ready, there's still a fair amount left to do. But this is definitely getting it a step closer. And yes, the accessibility changes will be present anywhere this software is run.

fedireads/incoming.py
```diff
@@ -128,9 +131,13 @@ def verify_signature(request):
...
 def handle_follow(activity):
     ''' someone wants to follow a local user '''
-    # figure out who they want to follow
-    to_follow = models.User.objects.get(actor=activity['object'])
-    # figure out who they are
+    # figure out who they want to follow -- not using get_or_create because
+    # we only allow you to follow local users
+    try:
+        to_follow = models.User.objects.get(actor=activity['object'])
+    except models.User.DoesNotExist:
```
cthulahoops, line 138: Why don't we just propagate the exception here? The caller doesn't do anything with the return value, so I think it makes debugging easier to just raise the exception.

fedireads/connectors/abstract_connector.py
```diff
@@ -266,25 +267,25 @@ def update_from_mappings(obj, data, mappings):
         if key == 'id':
             continue
 
-        if has_attr(obj, key):
+        try:
+            hasattr(obj, key)
```
cthulahoops, line 271: Is this code right? I don't expect hasattr to raise ```ValueError```

Is this not just:

```
if not hasattr(obj, key):
   setattr(obj, key, formatter(value))
```

mouse-reeve: no, you're right, it doesn't make sense. I've fixed it subsequently (and I also switched update form mappings to be a whitelist of fields, btw)

PR #172, opened 2020-05-13
Signing

cthulahoops:
* Refactor signing code a little for easier testing.
* Prefer public keys already in the database.
* Require that signatures are from the actor in the activity.
* Introduce some tests.

### DIFF
--- a/fedireads/broadcast.py
+++ b/fedireads/broadcast.py
@@ -63,37 +63,38 @@ def broadcast_task(sender_id, activity, recipients):
     return errors
 
 
-def sign_and_send(sender, activity, destination):
-    ''' crpyto whatever and http junk '''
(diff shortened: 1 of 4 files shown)


cthulahoops:
I'm not very convinced that using cached public keys is the right thing to do. It makes it impossible to change your keys, in the best case locking out the new account, in the worst case allowing an attacker to use a compromised key indefinitely. I'm going to see if I can find some info on best practice on this. (I still think this is a reasonable change for the moment, it's easy to undo that part of the change.)

mouse-reeve:
I recall seeing a conversation on socialhub about key rotation: https://socialhub.activitypub.rocks/t/key-rotation-notification/562
https://github.com/tootsuite/mastodon/issues/9658

PR #173, opened 2020-05-14
Refactor signatures into own module.

cthulahoops:


fedireads/incoming.py
```diff
@@ -88,7 +93,6 @@ def get_public_key(key_actor):
     try:
         user = models.User.objects.get(actor=key_actor)
```
mouse-reeve, line 94: as of the PR I just merged about remote_id/absolute_id, this and similar places will have to be:
`user = models.User.objects.get(remote_id=key_actor)`

cthulahoops: Thanks! I'll rebase this on top of that and sort the conflicts.

PR #174, opened 2020-05-14
Fixes remote user code

mouse-reeve:
and tests!

### DIFF
--- a/fedireads/remote_user.py
+++ b/fedireads/remote_user.py
@@ -42,7 +42,7 @@ def get_or_create_remote_user(actor):
 
 def create_remote_user(data):
     ''' parse the activitypub actor data into a user '''
-    actor = data.get('id')
+    actor = data['id']
     actor_parts = urlparse(actor)
 
     # the webfinger format for the username.
--- a/fedireads/remote_user.py
+++ b/fedireads/remote_user.py
@@ -4,6 +4,7 @@ from uuid import uuid4
 import requests
 
 from django.core.files.base import ContentFile
+from django.db import transaction
 
 from fedireads import models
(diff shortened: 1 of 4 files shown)


fedireads/remote_user.py
```diff
@@ -25,37 +26,47 @@ def get_or_create_remote_user(actor):
...
+    user.avatar.save(*avatar)
+
+    if user.fedireads_user:
+        get_remote_reviews(user)
+    return user
+
+
+def create_remote_user(data):
+    ''' parse the activitypub actor data into a user '''
+    actor = data.get('id')
```
cthulahoops, line 45: Should just be:

```actor = data['id']```

(Otherwise, it fails with a hard to debug error trying to handle the none below.)

Comment (cthulahoops) [reply fedireads/remote_user.py:45]: Should we assert that the actor matches the actor we asked for? Otherwise the remote could write use the opportunity to impersonate someone.

(I'm trying to get the signature code to use this and accidentally testing interesting edge cases. :))

Comment (mouse-reeve) [reply fedireads/remote_user.py]: hmm, I'm not sure what to check to verify the actor -- if it's loaded from the database, it's matching the actor string, and if it's created, it's set to whatever was sent. I think this is called after the signature would be verified?

Comment (cthulahoops) [reply fedireads/remote_user.py]: I think in get_or_create_remote_user, we need something like:

```
assert actor == data['id']
```

Otherwise we could fetch the user info for https://bad.example.com/cthulahoops and it could send us fake profile data for https://good.example.com/mouse. I think.

mouse-reeve: oh, do you mean comparing the request url to the id provided in the activitypub json?

PR #176, opened 2020-05-18
Verify http digest is correct if specified in signature.

cthulahoops:


cthulahoops:
Tests pass. :)
Fails in real life. :(

cthulahoops:
My mastodon node doesn't do nodeinfo for some reason. I think returning None is okay as we do so if data is missing anyway.

Issue #141, opened 2020-04-14
tags misuse the "Tag" activitypub concept


mouse-reeve:
they should be a different object type. I think mastodon uses `Hashtag`?

mouse-reeve:
this was fixed in #177

PR #177, opened 2020-05-19
Use dataclasses to define activitypub (de)serialization

mouse-reeve:
apparently there's stuff in python 3 besides using parens with print??

all the activitypub code becomes a set of `dataclass`es that describe the expected shape of the activity they represent. They can be given external activity json, make it into an activity object, and validate it, or take keyword arguments from a model, make it into an activity object, and serialize it into json.

models that are mapped into activitypub use the `ActivitypubMixin` to get serialization functionality, and must define which `dataclass` they use, and how to map the model fields to the activitypub dataclass fields.

this should streamline translation from activitypub json to models and vice versa, make a sensible space for activitypub validation, and simplify model serialization.

---

My first goal right now (august 2020) is to finish this dang PR.
What's remaining:

Activity objects
- [ ] Books
- [x] Authors
- [x] Boosts
- [x] Favorites
- [x] Tagging (#141)

Collections
- [ ] Editions list
- [x] following/followers lists

Activity verbs
- [x] Create (mostly for posting statuses)
- [x] Update
- [x] Add/remove (for shelving)
- [x] Follow/unfollow
- [x] Undo (currently only used for unfavorite)

I think this'll just be a squash commit because yikes.


mouse-reeve:
this has become _such_ a stream of consciousness commit log I think I'll re-commit the changes once I'm done in a more coherent way.

cthulahoops:
This is cool, btw. I haven't really used dataclasses, and it looks like it's going to really make things cleaner.

mouse-reeve:
My first goal right now is to finish this dang PR.
Serializers remaining:
- [ ] Books
- [ ] Authors
- [ ] Editions list
- [ ] Create
- [ ] following/followers lists
- [ ] Shelve actions
- [ ] Follow actions
- [ ] Boosts
- [ ] Favorites
- [ ] Tagging

I think this'll just be a squash commit because yikes.

mouse-reeve:
Current status of this PR: it's not totally ready to merge, but it'd be great if anyone who is interested in looking at it and giving feedback did.

It breaks the unit tests for Signatures (sorry @cthulahoops !) which I'm still thinking about how to fix.
It also breaks the assumptions that the fedireads data connector makes about how the incoming book data will be structured.

Overall, it's got some kludginess that I'd like to iron out, but I've been spinning my wheels on it for a while.

fedireads/incoming.py
```diff
@@ -129,21 +130,22 @@ def handle_follow(activity):
         # Duplicate follow request. Not sure what the correct behaviour is, but
         # just dropping it works for now. We should perhaps generate the
         # Accept, but then do we need to match the activity id?
-        return
+        return True
```
cthulahoops, line 133: I'm not a fan of this pattern of returning True/False to communicate success on the handle methods. Can we use return nothing on success and use exceptions to communicate failure?

fedireads/incoming.py
```diff
@@ -118,7 +119,7 @@ def handle_follow(activity):
     # figure out who the actor is
     user = get_or_create_remote_user(activity['actor'])
     try:
-        request = models.UserFollowRequest.objects.create(
+        relationship, _ = models.UserFollowRequest.objects.get_or_create(
```
cthulahoops, line 122: I don't know if the get_or_create is correct here. (If it is, the code and comment below handling the duplicate needs to change.)

cthulahoops (commented):
There's a lot here, I'm not really sure how to review it. So, I've just noted a couple of things jumped out at me while looking through it.

I'm quite sure why my signature test is causing issues. I see you've replaced my mock database object with a real user object.

I'd think it'd be best to get something merged, and sort out the details in further PRs. Let me know if there's anything I can do to help with this. If you want input on anything in particular, or to talk through bits of it or whatever?

fedireads/incoming.py
mouse-reeve: yeah, I think that's a good idea. I was addressing the linter wanting the methods to have return statements that are all the same type but I think raising vs returning none is a better solution

fedireads/incoming.py
mouse-reeve: you're right! this should stay as `create`

PR #178, opened 2020-05-22
Use fetch_remote_user and key refreshes.

cthulahoops:
* Use fetch_remote_user to get public_key for authentication.
* Refresh user data and retry on authentication failure.

Btw, would ```fetch_remote_user``` be better as a class method on the User class? (```User.fetch_remote``` and ```User.refresh_remote``` perhaps?)

### DIFF
--- a/fedireads/incoming.py
+++ b/fedireads/incoming.py
@@ -54,7 +54,7 @@ def shared_inbox(request):
         key = get_public_key(key_actor)
 
         signature.verify(key, request)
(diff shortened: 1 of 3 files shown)


cthulahoops:
It's very pleasing when a rabbit hole of changes allows a two line patch to fix the issue that started it all.

mouse-reeve:
I think you're right about making fetch_remote a class function 🤔

Issue #58, opened 2020-02-20
Instance settings


mouse-reeve:
there should be a json file with local settings with (non-exhaustively):

- instance name
- domain
- instance description
- code of conduct
- allow registration vs invite only
- suggested book tags

mouse-reeve:
I think the remaining work before this is close-able is to create the admin panel to set these. Uploading logos and themes would be the next level, for some other issue.

PR #182, opened 2020-06-01
Add instance settings.

cthulahoops:
Addresses: #58

Currently implemented:

* Instance name
* Instance description
* Code of conduct
* Allow registration

I decided to store this in the database so that settings can be easily
changed at runtime through the web interface. (This web interface does
not exist.)


fedireads/migrations/0042_sitesettings.py
```diff
@@ -0,0 +1,23 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='SiteSettings',
+            fields=[
+                ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
+                ('name', models.CharField(default='wyrms.cthulahoops.org', max_length=100)),
+                ('instance_description', models.TextField(default='This instance has no description.')),
+                ('code_of_conduct', models.TextField(default='Add a code of conduct here.')),
+                ('allow_registration', models.BooleanField(default=True)),
```
mouse-reeve, line 20: can we add an optional logo field here as well?

cthulahoops: Sure! Storing image data in the database?

I figure there's lots more settings that can be added to this, so this is mainly to get the structure in.

mouse-reeve: I think that works -- thanks so much for building out this feature!

mouse-reeve:
this is reminding me of my vague intention of setting up django admin and superusers. it wouldn't be as nice as a custom interface but it would probably be comparatively quick 🤔

cthulahoops:
> this is reminding me of my vague intention of setting up django admin and superusers. it wouldn't be as nice as a custom interface but it would probably be comparatively quick thinking

Yeah, I wondered about whether to do it that way. It would certainly be quick and easy.

I'm currently working on invites for when registration isn't allowed which might be a bit painful to have to manage via the admin interface. But it'd work.

cthulahoops:
Adding invite code support onto this as it depends on it lots.

fedireads/migrations/0042_sitesettings.py
cthulahoops: I might add the logo in separate PR - there's enough in this already.

You're welcome. It's fun to do a feature again after all the hidden auth work!

mouse-reeve: no need to prioritize logo, it just occurred to me

cthulahoops: Cool! Though it might be fun to work out how to store the image data.

mouse-reeve:
sorry for the delay on this, things have been *waves vaguely at current events*

cthulahoops:
No problem. Things very much are.

PR #185, opened 2020-06-17
Tweaks needed to get running with docker-compose locally

loganmhb:
Hi! This looks like a very cool project. I ran into a couple of small problems running it locally with docker-compose, so I thought I'd put the fixes up here in case you want to incorporate them. The problems were 1) I failed to initialize the `.env` file before running `docker-compose up` for the first time, so the postgres role wasn't auto-created, and 2) the webapp was not accessible on localhost.

(Not sure if the port exposure here matches how the dev setup is supposed to work, feel free to suggest otherwise!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,13 @@ Since the project is still in its early stages, not everything here is fully imp
     - Control which instances you want to federate with
 
 ## Setting up the developer environment
+
+Set up the environment file:
+
+``` bash
+cp .env.example .env
+```
+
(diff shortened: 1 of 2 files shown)


mouse-reeve:
thank you! you can access localhost through port 1336 via nginx, but I think this change makes it more usable for dev regardless. this is a good reminder to create separate dev and prod versions, I think :)

loganmhb:
Ah, I knew I must be missing something! Thanks.

PR #186, opened 2020-06-17
Add Author.display_name to decide which name to display

loganmhb:
When poking around the app locally, I noticed that author names weren't showing up because the templates look for `Author#name` but the authors were getting stored in my db with `first_name` and `last_name`.

It seemed to me (knowing very little about about either Django or the data model :) ) that the easiest solution is to display `first_name last_name` if `name` doesn't exist. So I implemented that as a property on the author model (not sure if this is the preferred approach?).

If there is some other strategy that would make more sense let me know!

### DIFF
--- a/fedireads/models/book.py
+++ b/fedireads/models/book.py
@@ -184,3 +184,14 @@ class Author(FedireadsModel):
     @property
     def activitypub_serialize(self):
         return activitypub.get_author(self)
+
+    @property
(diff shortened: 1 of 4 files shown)


mouse-reeve:
yes! I think this is a good fix. I added an issue (#187) for correcting how this data gets loaded as well.

PR #189, opened 2020-07-04
Look up book by remote_id when adding tag

cincodenada:
This is what downstream expects, I think this is the right level to convert. This fixes tag adding, which was previously failing because it couldn't find the book to tag


mouse-reeve:
thanks so much!

Issue #191, opened 2020-08-10
Mastodon rejecting our messages?


cthulahoops:
I think last time I checked mastodon was rejecting our messages due to issues with the digest header. (You can get it to work by commenting out the digest header code.) I'm not sure why it's broken, and I won't have time to look at it until next week.

I thought I should get it out of my head and into an issue as I know you're going to be starting active development again soon.

PR #193, opened 2020-08-19
Use the correct digest algorithm. (Fixes: #191)

cthulahoops:
Oops. :)

This one was... interesting... to debug. I ended up having to add debug logging to my mastodon node. I guess this means our requests to other bookwyrm instances were being rejected too. I'll check and confirm this.


cthulahoops:
And there was an anti-bug to go with the bug. We weren't verifying digests correctly, so we didn't notice that we signing requests wrongly.

mouse-reeve:
whew, thank you for catching and fixing this!

cthulahoops:
I guess this might be why people write tests.

PR #241, opened 2020-10-21
Deleted date should be null-able

### LINKED ISSUE #240: null value in column "deleted_date" of relation "bookwyrm_status" violates non-null constraint error
**I see:**
A few variations of the following fullscreen Django error:
```
IntegrityError at /<the-route-hit>/
null value in column "deleted_date" of relation "bookwyrm_status" violates non-null constraint error
```

**When I:**
- toggle the state of the "start reading" toggle
- click post on a comment
- click post on a quote
- click post on a review

**My Current Theories / Questions**
My guess is that the DB is expecting `not null` on `is_deleted` but the state of the book doesn't default to `is_deleted` as `false`. I need to dig more into the data model to understand which part of the schema is having this issue. It could also be that a migration didn't fully run? Wanted to flag this before I dig into it in case it's a small thing that I'm missing since I'm new to the code

**Context**
This happens on a local env installed from the Docker instructions.

mouse-reeve:
Fixes #240


ibrand (approved):
Thanks for addressing this! I checked out your branch to test it and it fixes all the conditions I listed 🎉

Issue #265, opened 2020-10-31
Flower error in production


mouse-reeve:
Flower not working in production on bookwyrm.social. Is is working in `main` and `production` branches for me locally.

This issue matches the error, but I'm using celery 4.4.2 and flower 0.9.4, so the fix doesn't help: https://stackoverflow.com/questions/64180054/importerror-cannot-import-name-command-from-celery-bin-base

```
flower_1         | ImportError: cannot import name 'Command' from 'celery.bin.base' (/usr/local/lib/python3.9/site-packages/celery/bin/base.py)
flower_1         | Traceback (most recent call last):
flower_1         |   File "/usr/local/bin/flower", line 5, in <module>
flower_1         |     from flower.__main__ import main
flower_1         |   File "/usr/local/lib/python3.9/site-packages/flower/__main__.py", line 4, in <module>
flower_1         |     from flower.command import FlowerCommand
flower_1         |   File "/usr/local/lib/python3.9/site-packages/flower/command.py", line 17, in <module>
flower_1         |     from celery.bin.base import Command
flower_1         | ImportError: cannot import name 'Command' from 'celery.bin.base' (/usr/local/lib/python3.9/site-packages/celery/bin/base.py)
```

jimfingal:
I can replicate this locally while on main with the basic .env config, I'm taking a look.

mouse-reeve:
it works now 😭 thank you!

PR #309, opened 2020-11-08
Run docker-compose flower in main container

jimfingal:
# Description

This PR switches from using the dockerhub version of flower, to using the flower that is run within the main container. This ensures that the flower run in docker is always pinned to whatever is in requirements.txt

# How I tested

I tried running it, and it looked like it worked and I could see the celery worker. I haven't done much with bookwyrm yet so haven't got much further than that!


mouse-reeve:
Fixes #265

PR #319, opened 2020-11-08
Move initdb into a management command

jimfingal:
This is a minor python / django cleanup item: it's generally nice to not have the importing of a file execute actions like create database objects, since it's hard to test.

This PR takes the init_db.py file that previously was imported to initialize the database, and turns it into a [django management command](https://docs.djangoproject.com/en/3.1/howto/custom-management-commands/). In theory this would make it easier to write tests against the functions there (though I'm still not that familiar with the codebase so I don't do this yet).

These in theory could also be loaded as [fixtures](https://docs.djangoproject.com/en/3.1/howto/initial-data/), though I didn't attempt that.


jimfingal:
I'm going to close this and roll it into a larger PR with other tooling!

Issue #56, opened 2020-02-20
Add Ability to Track Progress in a Book


bcj:
I would like to be able to keep track of where I am in a book as I am reading it (e.g., page number or equivalent for an ebook).

It would be nice if this kind of update optionally created a status for it (I may want to post a status if I got to a pivotal part in a book or finished a book in a collection, but not want to spam my followers if I am just updating my place in a book I'm reading)

mouse-reeve:
definitely should be a feature. I'm not sure if it should go on the ShelfBook table or be stored separately from that.

cincodenada:
#354 was just merged, which accomplishes the first part of this - you can now track your progress through a book as you go.

It won't post anything to your feed yet, but optionally posting an update is next on my list :)

mouse-reeve:
I think this is resolved with #774! Feel free to reopen if there are any further changes needed

PR #354, opened 2020-11-19
Keep track of progress through books over time

cincodenada:
This works towards implementing #56, which is one of my personal must-haves for record-keeping purposes :)

This is very much an early draft that's very open for feedback, I'm not attached to any details of the implementation here.

What I've done so far:
 - Created a new ProgressUpdate model/table/migration to store progress updates, which are associated with a ReadThrough
 - Added a "current page" input to the readthrough input which populates the existing `pages_read` field and creates a `ProgressUpdate` record on save
 - Added basic UI to the book page to display reading progress, hidden by default and shown with a query param

Base functionality:

- [x]  Test coverage (and fixes, if I broke anything, getting testing going now)
- [x] Add ability to delete a progress update
- [x] Allow specifying progress as a percentage, mainly for ebooks (the model supports this already, but the UI/logic is not there)
- [ ] Figure out what to do if a book doesn't have a page count populated - should we allow users to update the book record, or store a per-user page count that overrides/fills in for any page count on the book level? What does that UI look like?

Improvements:
- [ ] Add option to add a comment along with the progress udpate
- [ ] Optionally post progress updates on ActivityPub as a new type of status (there's some code working towards this that's not used yet)
- [ ] The ActivityPub statuses will be pretty duplicative of the ProgressUpdate model. We could possibly have them be the same thing, that would just require having some, statuses (private maybe?) that never actually get posted to ActivityPub or possibly not even show up in your private feed, which may be more trouble than it's worth.

Nice to have but not essential:
 - [x] Add a quick update on the feed sidebar
 - [ ] Make it look prettier
 - [ ] Add (optional) Javascript for showing the progress updates w/o reloading the page
 - [ ] Show a chart of progress over time
 - [ ] Get clever and convert when switching between pages/percent

I think that's most of it, although I'll probably realize later I forgot something. Right now this is just on my local dev instance, but I'll probably put it up on my staging instance once I get it running again. Happy to hear any feedback on style/approach/etc!


mouse-reeve:
wow this rules, I'm so stoked!! I'm going to take a closer look tomorrow but rad, thank you

mouse-reeve (changes requested):
here's a question I have as someone who doesn't use this feature -- would you want to post updates to your feed as you update your progress, or would you assume/prefer that it's just for your personal tracking?

bookwyrm/views.py
```diff
@@ -562,6 +562,9 @@ def book_page(request, book_id):
             book=book,
         ).order_by('start_date')
 
+        for readthrough in readthroughs:
+            readthrough.progress_updates = readthrough.progressupdate_set.all().order_by('date')
```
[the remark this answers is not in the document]
cincodenada: Ah yes, thanks! And please continue to be nitpicky :)

cincodenada:
> would you want to post updates to your feed as you update your progress, or would you assume/prefer that it's just for your personal tracking?

It kinda depends - I'd say it's primarily for my personal tracking, but I might occasionally want to post an update. A feature that I haven't added yet is the ability to write a short note along with the progress update (e.g. "oh dang, this is getting good!"), and I'd be more likely to want to post those to my feed.

mouse-reeve:
ah yeah that makes a lot of sense! My first thought is that the simplest route would be to associate a progress update with a `Comment` status, maybe with an optional foreign key from the status to the update. It would be pretty straightforward to include the page/percent value in the ActivityPub serialization without having to federate the update object itself.

rkingett:
Wish the bot gave more information, but it looks like you're trying to make a div be an edit field. Using native HTML elements is better because, not only will it work across browsers effortlessly, but you get a dozen accessibility properties for free.

cincodenada:
accesslintbot seems a little...overexcited here :laughing: I don't see an issue tracker for it, I was thinking of just emailing their support with a link to this issue?

rkingett:
I'm not seeing any issues. I think the bot is happy it's the weekend. I mean, I could be nit picky and say you should allow for changing of numerals using the arrow keys, up or down, but having just a plain edit field with a proper label  is fine.

cincodenada:
@rkingett What I'm talking about is that the bot correctly recognized the fix the first time, and now is posting "You fixed the issue(s)! Great work." after every commit I push, despite those commits not changing anything re: accessibility. It seems like there's a bug somewhere where it failed to make note that it's already congratulated me or something.

mouse-reeve:
sorry about all the merge conflicts! I kinda pulled the rug out from under you with the models. I'd be happy to resolve them if that's easiest

cincodenada:
Super no worries! That's what I get for leaving a PR hanging :smile: I'll take a look this weekend and check back in if there's anything I have questions about :+1:

cincodenada:
Okay, merged in the latest, some of my new tests are broken but I'll get to those soon :+1:

mouse-reeve:
I wonder if re-requesting a review will make accesslint chill out on posting? I'm going to try it.

rkingett:
I'm thinking accessbot doesn't have a default message for good code. I can try to look for a, well, more polished version? But nothing out is jumping out at me, code wise yet.

cincodenada:
I did send an email to AccessLint's support email and haven't heard back yet, I'll follow up there. Seems like a straightforward bug, where some flag that says "the last commit had issues" didn't get properly cleared.

cincodenada:
As for the PR itself, I just merged in the latest changes (love the reorganization!) and I think this is in an okay place to merge - there's still stuff I'd like to add, but I think it's reasonably complete and useful in its current state. I've rearranged the checklist accordingly.

On the last item, I think we can punt on that for now, and just show the total if we have it, and not if we don't, and I think that's fine.

cincodenada:
Okay, actually I'll leave most of the migrations in place, cause I don't want to mess with them. There was just one I hadn't added yet (0037), and so I added that and in place of the 0036 merge, since that's a pretty simple substitution.

bookwyrm/tests/actions/test_readthrough.py
```diff
@@ -0,0 +1,74 @@
...
+            local=True, localname='cinco')
+
+        self.client.force_login(self.user)
+
+    def test_create_basic_readthrough(self):
+        """A basic readthrough doesn't create a progress update"""
+        self.assertEqual(self.edition.readthrough_set.count(), 0)
+
+        self.client.post('/start-reading/{}'.format(self.edition.id), {
+            'start_date': '2020-11-27',
```
mouse-reeve, line 33: this might have the side effect of creating a "started reading" status and then triggering a `broadcast` task, which could be why the tests are hanging. you can use `with patch('bookwyrm.broadcast.broadcast_task.delay')` if that's what's happening. Stopping celery and redis when you run the tests will show these errors locally. I have learned the hard way!

cincodenada: Oh rad, that's very helpful, thanks! I'll take a peek at that, hopefully the logs will confirm once they time out again in a couple hours :upside_down_face:

mouse-reeve:
I made a bunch of merge conflicts for you _again_! As before, I'd be happy to resolve them for you if that would be easier.

cincodenada:
Those merge conflicts were :cake: :smile:, barely had to tell my diff tool anything, yay!

I poked around at the `patch()` stuff and implemented it by mocking it for the whole test class, since there's no case in this test where we want to have it unpatched. Let me know what you think - I'm happy to switch it up if it's better to be consistent with the others.

mouse-reeve:
oh patching for the whole class is great! I'm looking forward to testing tomorrow :)

bookwyrm/templates/feed.html
```diff
@@ -47,7 +47,17 @@ <h2 class="title is-5">Your books</h2>
                 </div>
             </div>
             <div class="card-content">
-                {% include 'snippets/shelve_button.html' with book=book %}
+                <div class="columns is-gapless">
```
mouse-reeve, line 50: I think this works better just flat, rather than in columns -- so this markup minus all the added divs? it was too long for my screen as is

bookwyrm/templates/snippets/progress_update.html
```diff
@@ -0,0 +1,25 @@
...
+        {% if readthrough.progress_mode == 'PG' %}
+        on page
+        {% else %}
+        currently at
+        {% endif %}
+    </div>
+    <div class="control">
+        <input
+         aria-label="{% if readthrough.progress_mode == 'PG' %}Current page{% else %}Percent read{% endif %}"
+         class="input is-small" type="text"
```
mouse-reeve, line 14: this type should be `number`, rather than `text` for better input validation

bookwyrm/models/readthrough.py
```diff
@@ -24,3 +31,26 @@ def save(self, *args, **kwargs):
...
+        if self.progress:
+            return self.progressupdate_set.create(
+                user=self.user,
+                progress=self.progress,
+                mode=self.progress_mode)
+
+class ProgressUpdate(BookWyrmModel):
+    ''' Store progress through a book in the database. '''
+    user = models.ForeignKey('User', on_delete=models.PROTECT)
+    readthrough = models.ForeignKey('ReadThrough', on_delete=models.PROTECT)
```
mouse-reeve, line 45: When I try to delete a readthrough that has an associated progress update, I get an error because of the protected relationship. I think this is a case where CASCADE would be better than PROTECT?

bookwyrm/models/readthrough.py
```diff
@@ -24,3 +31,26 @@ def save(self, *args, **kwargs):
...
+    progress = models.IntegerField()
```
mouse-reeve, line 46: I think this should have a min value validator to prevent negative values here

mouse-reeve (changes requested):
I agree that this is at a good point to merge, after these couple small model and display changes. I really appreciate all your work on this!

bookwyrm/models/readthrough.py
cincodenada: Good call, if a user is deleting a readthrough there's no reason to keep the progress updates. There's already a confirmation dialog for readthroughs, I bet I can add a "...and its X progress updates" to that just to make it clear.

bookwyrm/models/readthrough.py
cincodenada: I added validators here and on ReadThrough.progress (and tests to make sure they work), but didn't actually use them anywhere. I'm not sure they'll have any effect on ProgressUpdate, but might on the ReadThrough depending on how we use them?

In any case, having the validators is better than not

bookwyrm/templates/feed.html
cincodenada: I ended up flattening this and then totally reworking the update snippet. Curious to hear what you think, I'm...mostly happy with it I think :sweat_smile:

mouse-reeve:
Whew, thank you so much for bearing with this, I'm stoked to have the feature. I think the next step is making it possible to post status update Generated Notes, and associated progress points with statuses

mouse-reeve (approved):

mouse-reeve:
I've been posting periodic updates on patreon with a list of newly added features. If you're okay with it, what's the best way to credit you for this? The default would just be like, First Name Last Name with a link to your website, but whatever feels best to you within reason is good for me

cincodenada:
Oh, sure, thanks! Joel Bradshaw is good, with a link to...uhh https://cincodenada.com I suppose. Or just my Mastodon account, whatever seems best!

PR #356, opened 2020-11-19
Rename devtools, make it work with non-default database names

cincodenada:
My database is named `bookwyrm` - this commit alters fr-dev to pull the dbname/username from the .env instead of hardcoding.

This does require the .env to be a valid shell script, hence the updates to the example `.env` - specifically, the secret key has to be quoted, apparently Python's env loader is more tolerant of special characters than bash is.

~~I'm not sure what the best way to warn people about this is, I'm not sure if there's a good way to catch that error in the bash script. Hmm.~~ It will give a (hopefully) helpful error message now if loading the `.env` fails

I also took this chance to rename `fr-dev` to `bw-dev`, and left behind an alias for finger memory's sake :)

For reviewing it's easiest to look at the individual commits rather than the overall diff, because the renamed file makes it impossible to see the changes to the script.


mouse-reeve:
This is such a well needed change! thank you

PR #361, opened 2020-11-21
Fix navbar on mobile, make search an addon

### LINKED ISSUE #305: Top bar too wide on mobile
At least on my phone, the top bar creates horizontal scrolling and makes the hamburger button hard to find

cincodenada:
This should fix #305, I think, just make the search box container a proper column that takes up a third of the space.

Also, in reading up on Bulma I found the has-addons option which is a nice touch, I think - if you prefer the previous way feel free to put it back :)

### DIFF
--- a/bookwyrm/templates/layout.html
+++ b/bookwyrm/templates/layout.html
@@ -25,7 +25,7 @@
         <a class="navbar-item" href="/">
             <img class="image logo" src="/static/images/logo-small.png" alt="Home page">
         </a>
-        <form class="navbar-item column is-third-mobile" action="/search/">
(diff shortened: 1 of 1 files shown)


cincodenada:
Huh, so I originally thought I was setting the width to fix this but it turns out I typo'd the width, and just adding `column` to the middle column fixes it. Trying to make the other two `column`s as well, even with `is-narrow`, doesn't work right, so this turned out to be is a little strange structurally, but I think I get why it works out anyway.

mouse-reeve:
omg thank you this was bugging me so much. the has-addons is such an improvement to the display of the search button as well.

PR #368, opened 2020-11-28
Replaces naive datetimes with aware ones

cincodenada:
This replaces various timezone-naive datetimes with timezone-aware ones, using `django.util.timezone`. This should fix everything that's covered in tests, but there may be other usages that don't have test coverage, although I searched through the codebase for keywords and didn't find any.


mouse-reeve:
tests running without a million warnings, this is my dream!

PR #422, opened 2020-12-24
Fix typo in Patreon link

AndrewRadev:
Seems like the "Sponsor this project" patreon link is mistyped -- this patch should fix it.


mouse-reeve:
omg, oops, thank you for catching this!

PR #425, opened 2020-12-27
Fixes goodreads import failing on every book

mouse-reeve:


rkingett:
Should we actually start deprecating this because Goodreads is shutting down their API?

mouse-reeve:
This just uses a csv download of a user's library on goodreads, which I think is not part of the functionality being removed? it uses OpenLibrary and other BookWyrm instances for looking up actual book data.

but now that you mention it, I'm not 100% sure they won't remove the option to download your data as a csv! my longer-term plan is to expand import to take more formats (like a librarything export) using the same functionality

PR #431, opened 2020-12-29
WIP Allow for a default user post privacy

### LINKED ISSUE #253: let user set default post privacy
Right now, a user can set the privacy level of a status when they post the status. The user should also be able to set their preferred default post privacy in the user settings (this is the same as the edit profile page for now).

What I think this will ential:
 - adding a default post privacy setting to the User model (there's already an enum defined with the options in `bookwyrm/models/base_model.py`), with the default value to `Public`
- in the status creation template, make the default post privacy setting the correct one for the user (`bookwyrm/templates/snippets/create_status.html`)
- when privacy isn't set explicitly in a template (boosts, generated statuses come to mind), pass along the user's default

ibrand:
# Fixes #253 

## Description
Allow users to set a post privacy default in their profile. Not quite sure how we'd want this to work with boosts and also not feeling solid on what a "generated status" is so I thought I'd start with a WIP to get a conversation going.


fix_migrations.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
ibrand, line 1: is this worth committing? I can remove it. It's from @cincodenada 's proposed fix for my migrations problem.

bookwyrm/templates/edit_user.html
```diff
@@ -42,6 +42,13 @@ <h1 class="title">Profile</h1>
                     {{ form.manually_approves_followers }}
                 </label>
             </p>
+            <p class="block">
+                <label class="label" for="id_default_post_privacy">
+                    Default post privacy:
+                    {% include 'snippets/privacy_select.html' with no_label=False current=user.default_post_privacy %}
```
ibrand, line 48: Considered actually making a change to `privacy_select.html` itself, but think this is a reasonable way to handle this. Open to feedback.

rkingett:
If a post is set to unlisted or followers only can we disable boost? Or were you talking about setting post defaults for boosts? Like allow boosts or block all boosts?

fix_migrations.sh
mouse-reeve: this is great! I think it would make the most sense to have this available separately with the original issue (in a gist or in a comment body) rather than in this PR, since it shouldn't effect people who clone the repo any time after the merge migration was created.

bookwyrm/models/user.py
```diff
@@ -10,7 +10,7 @@
 from bookwyrm.models.status import Status
 from bookwyrm.settings import DOMAIN
 from bookwyrm.signatures import create_key_pair
-from .base_model import ActivityMapping, OrderedCollectionPageMixin
+from .base_model import ActivityMapping, OrderedCollectionPageMixin, PrivacyLevels
```
mouse-reeve, line 13: I moved some things around in how models are structured, so you'll have to import PrivacyLevels from `.fields` instead of  `.base_model` -- sorry about that! (https://github.com/mouse-reeve/bookwyrm/commit/b6907f39e97529211f53ee175ed657f999261ecf)

bookwyrm/templates/edit_user.html
mouse-reeve: I think it would make sense to edit the `privacy_select.html` snippet so that it could have `name="default_post_privacy"` here and `name="privacy"` when it's used in the create status snippet, that way you wouldn't have to change `bookwyrm/view_actions.py` at all (I refactored it a bit in https://github.com/mouse-reeve/bookwyrm/pull/418/commits/e6105c6cb0000a49c8aeb76411f57124a80947c1 to use django's form.save() functionality, rather than setting the fields manually, which I didn't realize was a thing when I started coding this 😂 )

mouse-reeve (changes requested):
wow I'm stoked about this, thank you for all your work!

mouse-reeve:
> If a post is set to unlisted or followers only can we disable boost? Or were you talking about setting post defaults for boosts? Like allow boosts or block all boosts?

This wouldn't directly effect what can be boosted, since it just lets you set the default privacy on your statuses. I did make a fix to disable the boost button on followers only statuses quite recently, although unlisted is still boostable. I think boost ought to inherit the privacy of the boosted status, so unlisted statuses should be unlisted boosts, which is outside the scope of this, but I'll make a separate issue for it.

Does that answer your question?

mouse-reeve:
> ... also not feeling solid on what a "generated status" is

yeah generated status is a weird one. it just means a status that was automatically created when someone marked a book as "want to read" or whatever. They are triggered in `bookwyrm/outgoing.py` and created in `bookwyrm/status.py`

ibrand:
Closing in favor of #623 which is a clean start off of main to deal with merge conflicts

Issue #429, opened 2020-12-29
Improve Mastodon integration by using the `application/activity+json` type


renatolond:
I'm not sure if this should be a bug, sorry if it's the wrong type!

**Is your feature request related to a problem? Please describe.**
When you follow someone from Mastodon, you don't see previous posts. One of the ways you can see them, however, is by pasting the URL of their status, which allows Mastodon to fetch that activity and associate with a profile. Currently, doing so with BookWyrm doesn't fetch the status.

**Describe the solution you'd like**
I'd like to be able to look for comments and profiles by pasting the URL in Mastodon search bar.

**Describe alternatives you've considered**
I think BookWyrm should use the `content-type` in the api request responses. This would make it so that Mastodon treat the responses as ActivityPub things and parses them correctly.
This however, raises another issues. It seems the serializer used in the views is serializing using `pure=False`, which gives a `Comment`, instead of a `Note`, which is ignored by Mastodon.
Since it probably would be interesting to still return BookWyrm-specific serializations, one way I see for this is to add a user-agent to bookwyrm requests and differentiate them in the views code.

I'd be happy to submit a PR about this if this is the way you think this should go :)

mouse-reeve:
Yeah! I was assuming that the only time mastodon would receive statuses is when they are sent from bookwyrm, and it didn't occur to me that they should be fetch-able from mastodon as well. Bookwyrm instances definitely need the custom activity types (Comment, Review, Quote, etc) to each other, and "pure" versions of those statuses to mastodon, so I think your idea of using a user agent makes a lot of sense. So tl;dr you're right and I think your approach sounds correct. Thank you!

mouse-reeve:
I'm getting a 500 error when I search for statuses in friend.camp that weren't posted to it. It looks like the bookwyrm server is responding with the "pure" activity as expected, but mastodon fails trying to load it.

mouse-reeve:
okay, this turned out to be unrelated! false alarm.

PR #434, opened 2020-12-30
Create a ActivitypubResponse class and use it in the views

renatolond:
This is linked to #429. This does not change the responses themselves, but changes the content-type of the activitypub-linked responses by introducing an ActivitypubResponse class, which mimics JsonResponse, but uses ActivityEncoder and sets the content-type accordingly.

### DIFF
--- a/bookwyrm/activitypub/response.py
+++ b/bookwyrm/activitypub/response.py
@@ -1,0 +1,18 @@
+from django.http import JsonResponse
+
+from .base_activity import ActivityEncoder
+
+class ActivitypubResponse(JsonResponse):
+    """
+    A class to be used in any place that's serializing responses for
+    Activitypub enabled clients. Uses JsonResponse under the hood, but already
+    configures some stuff beforehand. Made to be a drop-in replacement of
+    JsonResponse.
+    """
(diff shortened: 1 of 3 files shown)


bookwyrm/activitypub/response.py
```diff
@@ -0,0 +1,17 @@
+from django.http import JsonResponse
+
+from .base_activity import ActivityEncoder
+
+class ActivitypubResponse(JsonResponse):
+    """
+    A class to be used in any place that's serializing responses for Activitypub enabled clients.
+    Uses JsonResponse under the hood, but already configures some stuff beforehand
+    Made to be a drop-in replacement of JsonResponse
```
mouse-reeve, line 9: sorry, this is such a nitpicky change, but I have a tiny laptop screen - can you line wrap at 80 characters?

mouse-reeve:
this is great! it's a big improvement, thank you. Happy to merge after tiny stylistic changes (I can also fix it myself if you prefer)

bookwyrm/activitypub/response.py
renatolond: Of course :)
Done!

PR #443, opened 2020-12-31
Template bugs and fixes

mouse-reeve:


rkingett:
I know this has been fixed, but, I'd just like to reiterate that HTML elements are best rather than trying to wrap everything in DIVS. If it can be done via HTML, use HTML.

But, then again, I'm half asleep.

mouse-reeve:
hm I'm not sure what you mean? the accesslint complaint here in particular was that the reply textarea didn't have a label. are you saying that are more nested divs are harder to navigate than have fewer nested divs, or that semantic html elements are easier to navigate that divs?

rkingett:
Yes, symantic HTML semantic html elements are easier to navigate than divs, always. LOL! Much better on browsers, especially mobile browsers too. Always use symantic HTML elements over divs. Your users will thank you for it.

PR #460, opened 2021-01-01
Review rate

mouse-reeve:
this is _gnarly_, it deletes all rating-type reviews and re-creates them as a separate model. it absolutely does fuckery over any external reference to existing ratings.


bookwyrm/migrations/0033_reviewrate.py
```diff
@@ -0,0 +1,72 @@
+# Generated by Django 3.0.7 on 2021-01-05 05:32
+
+from django.db.models.fields.reverse_related import ManyToOneRel
+from django.db.models import Q
+from django.db import migrations, models
+import django.db.models.deletion
+
+def convert_review_rating(app_registry, schema_editor):
```
cincodenada, line 8: So, since ReviewRating is a subclass of Review and thus the database implementation is literally just inserting a pointer to the relevant `review` in `bookwyrm_reviewratings`, I think if we wanted to get fiddly with it we could basically just `INSERT INTO bookwyrm_reviewrating(review_ptr_id) SELECT id FROM bookwyrm_review` and directly convert those reviews into reviewratings.

I'm not a huge fan of the direct database modifications, but when it's a simple database operation and the alternative is also gnarly...blah. I poked around and found a 13-year-old(!) ticket asking for the ability to do this properly in Django (create a child model from an existing parent model) and it is still open  https://code.djangoproject.com/ticket/7623

bookwyrm/migrations/0033_reviewrate.py
```diff
@@ -0,0 +1,72 @@
...
+        rating = review_ratings.objects.create(
+            user=review.user, rating=review.rating, book=review.book)
+        print('-----')
+        print(rating, review)
+        for field in review._meta.get_fields():
+            if isinstance(field, ManyToOneRel) or field.name in ['status_ptr', 'id', 'remote_id']:
+                continue
+            value = getattr(review, field.name)
+            try:
+                setattr(rating, field.name, value)
```
cincodenada, line 26: This all looks perfectly reasonable to me, but in my tests it isn't copying over the `name` field, and I haven't figured out why :confused:

cincodenada (commented):
This is gnarly indeed. See comments for a possibly fiddly bit, I also still wonder if there's a way to twist Django's arm into doing it for us, but it is late and I gotta put a pin in this :)

bookwyrm/activitypub/note.py
```diff
@@ -59,7 +66,8 @@ class Review(Comment):
...
 
 @dataclass(init=False)
-class Quotation(Comment):
-    ''' a quote and commentary on a book '''
-    quote: str
-    type: str = 'Quotation'
+class Rating(Comment):
+    ''' a full book review '''
+    rating: int = None
+    content: str = None
```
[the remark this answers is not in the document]
cincodenada: I was reading up on Multi-table inheritance and see now that this removes the field from `Rating`, so ignore this comment!

bookwyrm/models/status.py
```diff
@@ -258,6 +258,22 @@ def pure_content(self):
...
+    ''' a subtype of review that only contains a rating '''
+    def save(self, *args, **kwargs):
+        if not self.rating:
+            raise ValueError('Rating object must include a numerical rating')
+        return super().save(*args, **kwargs)
+
+    @property
+    def pure_content(self):
+        #pylint: disable=bad-string-format-type
+        return 'Rated "%s": %d stars' % (self.book.title, self.rating)
```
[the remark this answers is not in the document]
mouse-reeve: oh that's good to know! I should really get on board with the new string formatting syntax. it wasn't registering as an int to the linter I think because it's a custom model field wrapping IntegerField.

bookwyrm/migrations/0033_reviewrate.py
```diff
@@ -0,0 +1,72 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+def convert_review_rating(app_registry, schema_editor):
+    ''' take reviews with no content and turn them into ratings '''
+    db_alias = schema_editor.connection.alias
+    reviews = app_registry.get_model('bookwyrm', 'Review')
+    review_ratings = app_registry.get_model('bookwyrm', 'ReviewRating')
+    ratings = reviews.objects.using(db_alias).filter(
+        Q(content__isnull=True) | Q(content=''))
```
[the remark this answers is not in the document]
mouse-reeve: oh I _think_ I've always used __isnull, but I've never formed an opinion on it...

bookwyrm/migrations/0033_reviewrate.py
```diff
@@ -0,0 +1,72 @@
...
+        print('-----')
+        print(rating, review)
+        for field in review._meta.get_fields():
+            if isinstance(field, ManyToOneRel) or field.name in ['status_ptr', 'id', 'remote_id']:
+                continue
+            value = getattr(review, field.name)
+            try:
+                setattr(rating, field.name, value)
+            except TypeError:
+                getattr(rating, field.name).set(value.all())
```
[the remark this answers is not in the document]
mouse-reeve: yep!

bookwyrm/migrations/0033_reviewrate.py
mouse-reeve: I'm not sure I'm understanding you/my own code correctly, but I think ratings (new form or old form( shouldn't have names set at all?

bookwyrm/migrations/0033_reviewrate.py
mouse-reeve: oh I love this, it really is gnarly in a wholly different way. and for this specific case, yeah, it seems like a sensible fit

mouse-reeve:
@cincodenada  thank you! I really needed an affirmation that this wasn't a totally hairbrained idea, and I'm super appreciative of the feedback.

bookwyrm/activitypub/note.py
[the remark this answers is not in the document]
mouse-reeve: this makes the default value for content (which shouldn't be set on ratings) `None`, which means the field will be removed from the activity when it's serialized. But really it just shouldn't inherit from `Comment`, because it's silly to add the field and then hope it never gets used and cull it later instead of just not having it present. And rating shouldn't have a `None` default value.

bookwyrm/migrations/0046_reviewrating.py
```diff
@@ -0,0 +1,59 @@
...
+    with connection.cursor() as cursor:
+        for review in reviews:
+            cursor.execute('''
+INSERT INTO bookwyrm_reviewrating(review_ptr_id)
+SELECT status_ptr_id FROM bookwyrm_review
+WHERE status_ptr_id={:d}'''.format(review.id))
+
+def unconvert_review_rating(app_registry, schema_editor):
+    ''' undo the conversion from ratings back to reviews'''
+    # TODO: this does not work
```
mouse-reeve, line 27: @cincodenada any chance you know how to write the reverse migrations for this? I gave up for the day

cincodenada: An excellent TODO 😄 I do have some ideas here that I have not actually tried, so I will give this a go hopefully tonight, if not tomorrow morning 👍

cincodenada: Okay, so I just opened #652 which is a PR from my repo into this PR, and...it's the silliest thing, but literally all we need to do is just let Django drop the table, since that table is the only thing that makes those records ReviewRatings instead of Reviews :shrug:

There's also a couple other things in there that I described better in the description :)

mouse-reeve:
@cincodenada I think this is ready to go! any last thoughts?

PR #462, opened 2021-01-02
Goodreads tests

mouse-reeve:
- removes limit on goodreads csv length
- adds tests for `goodreads_import.py`


rkingett:
Seriously, where is this import? I've been looking for days! Like, does it exist in the profile settings or is this a feature actively being developed? Either way, I hope this won't break the server if the limitation is removed.

mouse-reeve:
it's in the top bar menu: there's your username and avatar, and then links to direct messages, profile, settings, and then, the one you're looking for, "import books." it's a dropdown but it works fine in my screen reader

I got the okay a while back from folks at openlibrary to remove the limit, so it should be okay, and they know where to find me if it causes problems 😂

PR #468, opened 2021-01-03
Use user-agent to determine whether to show pure representation

### LINKED ISSUE #429: Improve Mastodon integration by using the `application/activity+json` type

renatolond:
Together with #434 and #435, this fixes #429. Use the user-agent to determine if the call is coming from a BookWyrm instance or not. If it's not, give a pure activitypub representation for the status. Otherwise give a BookWyrm one, to allow for a complete integration between instances.

### DIFF
--- a/bookwyrm/tests/test_views.py
+++ b/bookwyrm/tests/test_views.py
@@ -10,7 +10,7 @@ from django.test.client import RequestFactory
 
 from bookwyrm import models, views
 from bookwyrm.connectors import abstract_connector
-from bookwyrm.settings import DOMAIN
+from bookwyrm.settings import DOMAIN, USER_AGENT
 
 
 # pylint: disable=too-many-public-methods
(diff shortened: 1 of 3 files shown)


mouse-reeve:
beautiful! and there's a test! I'm so happy

PR #519, opened 2021-01-12
Overhauls views/view_actions/outgoing

mouse-reeve:
Moving all the code from these three files into class based views


mouse-reeve:
there were a bunch of ignore csrf decorators that I ignored, and they might be important or they might not! also this is going to make every other pr in existence have merge conflicts 😬

Issue #518, opened 2021-01-12
Login form on mobile is too wide 


cincodenada:
On mobile, the login form is all on one line and thus extends off the right hand of the screen, requiring horizontal scrolling. It should wrap to two lines on mobile.

(This is mostly a note to self 🙂)

PR #521, opened 2021-01-14
Make login form wrap nicely on mobile

cincodenada:
Fix for #518. Open to feedback here, changing a form to columns from a grouped form item isn't my favorite, but I think it's okay, and it gets us better mobile behavior for free, which is the goal.

### DIFF
--- a/bookwyrm/templates/layout.html
+++ b/bookwyrm/templates/layout.html
@@ -127,17 +127,19 @@
                     <div class="column">
                         <form name="login" method="post" action="/user-login">
                             {% csrf_token %}
-                            <div class="field is-grouped">
-                                <div class="control">
+                            <div class="columns is-variable is-1">
+                                <div class="column">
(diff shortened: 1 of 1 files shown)


bookwyrm/templates/layout.html
```diff
@@ -127,17 +127,19 @@
                     <div class="column">
                         <form name="login" method="post" action="/user-login">
                             {% csrf_token %}
-                            <div class="field is-grouped">
-                                <div class="control">
+                            <div class="columns is-variable is-1">
```
cincodenada, line 130: Narrowing the column gap here isn't totally necessary (and will only work on some browsers anyway), but with columns the gap was wider than it used to be, and now it's narrower than it used to be. I thought narrower was a little better, but if wider is better we can drop this and use the default gap.

mouse-reeve:
this looks great!!

PR #523, opened 2021-01-14
Add order for shelf books to satisfy warning

cincodenada:
I decided to hunt down the "collection_queryset must be ordered" errors in the test suite, and it ended up being caused by `Shelf.collection_queryset()` not being ordered (it was just `self.books.all()`).

I think I managed to make it sort by order they were added to the shelf (aka creation date of the ShelfBook), but I haven't actually tested that. I should add some tests to make sure.

Also I added a RuntimeError if a `collection_queryset()` is not ordered, which now that I'm writing this I realize would probably be better placed directly in `to_ordered_collection()` maybe? Or, may not be worth having as an error at all.

### DIFF
--- a/bookwyrm/models/base_model.py
+++ b/bookwyrm/models/base_model.py
@@ -225,6 +225,9 @@ class OrderedCollectionPageMixin(ActivitypubMixin):
     def to_ordered_collection(self, queryset, \
             remote_id=None, page=False, **kwargs):
         ''' an ordered collection of whatevers '''
+        if not queryset.ordered:
+            raise RuntimeError('queryset must be ordered')
+
         remote_id = remote_id or self.remote_id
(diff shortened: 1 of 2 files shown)


mouse-reeve:
I'm glad you found this, I was wondering where it was coming from

mouse-reeve:
I think it makes more sense to put the error in `to_ordered_collection`, since some models use that functionality without using the `to_activity` in `OrderedCollection`

cincodenada:
This is not actually related to this particular PR, but I'm curious why these functions take `**kwargs`, I don't see where they're using them if they are.

cincodenada:
Okay, I decided to answer my own question by trying to pull them out and am learning about Signal receivers. So, never mind on that question!

PR #538, opened 2021-01-19
Polling to update notification count

mouse-reeve:
I considered adding a live region but I think it would be too annoying (at least without also adding settings to disable it)?

### DIFF
--- a/bookwyrm/static/js/shared.js
+++ b/bookwyrm/static/js/shared.js
@@ -34,23 +34,25 @@ window.onload = function() {
 
     // polling
     document.querySelectorAll('[data-poll]')
-        .forEach(t => setInterval(function () { polling(t); }, 10000));
+        .forEach(el => polling(el));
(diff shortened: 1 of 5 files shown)


rkingett:
How about making it a default browser notification instead of a live region? I'm not too wild about the live region because I am worried it would interrupt NVDA. That being said, what about a default browser notification toggle?

mouse-reeve:
hm, desktop notifications seem aggressive for an application that never reasonably needs to grab anyone's attention. But there's not any harm in making them opt-in. Do you prefer them to notification sounds?

rkingett:
I actually would prefer notification sounds rather than browser notifications, now that I think about it, but yes. Making these opt in would be best.

mouse-reeve:
rad, that's my preference as well. I'm going to merge this change and add that in a separate PR once I find a ding I like.

PR #544, opened 2021-01-19
Sets minimum goal

mouse-reeve:


mouse-reeve:
something went wrong with this branch's relationship to my local???

PR #552, opened 2021-01-20
Add docker-compose up to README

bibliotechy:
Dev docks current left out a step I needed to get the app serving at `https://localhost:1333`


mouse-reeve:
good catch, thank you!

PR #553, opened 2021-01-22
Makes pages/percents toggle-able in sidebar for progress readthrough

mouse-reeve:


cincodenada:
Oh rad! I had considered doing this but you beat me to it, thanks! Also, thanks for catching those migrations, I didn't think about those changes to the model needing to be migrated :+1:

PR #557, opened 2021-01-25
Fix migration for if db has multiple empty emails

cincodenada:
If the database has multiple users with an empty email column, this
migration will fail because multiple empty strings break the unique
constraint.

A fresh database won't have this problem because it won't have any
legacy users with empty strings instead of NULL, but for existing
databases we need to convert the empty strings to NULL so they don't run
awry of the unique constraint.

### DIFF
--- a/bookwyrm/migrations/0037_auto_20210118_1954.py
+++ b/bookwyrm/migrations/0037_auto_20210118_1954.py
@@ -2,6 +2,15 @@
 
 from django.db import migrations, models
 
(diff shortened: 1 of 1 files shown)


cincodenada:
I ran into an error about emails not being unique in migrating to the latest main, I believe this should resolve the issue. To reproduce, start up an instance before the commit introducing migration 0037, search for at least two remote users to get their entries in the user database, and then try to migrate.

mouse-reeve:
thank you! this is a much better migration

PR #585, opened 2021-02-03
Inline form

mouse-reeve:


rkingett:
Oh good! No access bot flags. I admit, when I saw form in the email subject line I went, oh no! DIVS and unlabeled edit fields everywhere! I haven't checked the code yet.

mouse-reeve:
there shouldn't be any unlabelled fields! this change makes it so that all the edit/create forms use the same html template, so things will be a lot more consistent across different places. It will slightly change editing and creating shelves and lists, but barely at all

PR #586, opened 2021-02-04
List tweaks

mouse-reeve:


rkingett:
Will there be a URL I can share to people to say, here, go add books or suggest books? Or would just sharing the address bar URL work even if I give it to others?

mouse-reeve:
Sharing the address bar will work. Any logged in user who has permission to view the list will be able to suggest or add books.

PR #623, opened 2021-02-14
Add user default privacy setting to the privacy_select and user prefe…

ibrand:
…rences

Revisiting #431 after life settled down enough to sit down with it!

This should address edits around moving the changes into `privacy_select.html` and definitely keeps the code simple and has the user default as the fallback whenever a `current` isn't passed in.


bookwyrm/templates/preferences/edit_user.html
```diff
@@ -43,6 +43,12 @@
             {{ form.manually_approves_followers }}
         </label>
     </div>
+    <div class="block">
+        <label class="label" for="id_default_post_privacy">
+            Default post privacy:
+            {{form.default_post_privacy}}
```
ibrand, line 49: doing it this way means that instead of `private` the dropdown says `direct` because that's what's on the model. I was having trouble setting it up to use the privacy_select instead because I need to read more about this `form.save()` stuff, but my guess is that that's what's right here

ibrand:
Noting that the test that's failing is the `test_edit_user` test with:

```
  File "/app/bookwyrm/tests/views/test_user.py", line 158, in test_edit_user
    self.assertEqual(delay_mock.call_count, 1)
AssertionError: 0 != 1
```

So it's definitely related to this change. I'm a little surprised because the email and display names are saving when I hit save locally. I'll circle back to investigate and will mark as a draft til I can fix it!

bookwyrm/templates/preferences/edit_user.html
mouse-reeve: to use the `privacy_select.html` snippets, you'll need to have it take an optional parameter for the `name` of the form element (right now it's `<select name="privacy" id="privacy-{{ uuid }}">`), so that you can pass `default_post_privacy` in this instance. Otherwise the form will get the wrong name for that field and won't be able to find it on the model. I know my complex snippets setup is kind of a headache, so it's also completely fine to leave this as is, and I'll go in and modify the markup later.

bookwyrm/templates/snippets/privacy_select.html
```diff
@@ -5,16 +5,16 @@
...
-        <option value="unlisted" {% if current == 'unlisted' %}selected{% endif %}>
+        <option value="unlisted" {% if user.default_post_privacy == 'unlisted' or current == 'unlisted' %}selected{% endif %}>
             Unlisted
         </option>
-        <option value="followers" {% if current == 'followers' %}selected{% endif %}>
+        <option value="followers" {% if user.default_post_privacy == 'followers' or current == 'followers' %}selected{% endif %}>
             Followers
         </option>
-        <option value="direct" {% if current == 'direct' %}selected{% endif %}>
+        <option value="direct" {% if user.default_post_privacy == 'direct' or current == 'direct' %}selected{% endif %}>
```
mouse-reeve, line 17: I think this is going to lead to cases where multiple options will be selected when the `default_post_privacy` is set and the `current` value is set to something different. It should go with `current` if it's set, and use `default_post_privacy` if not

mouse-reeve (changes requested):
this is great, thank you for following back up on it :)

mouse-reeve:
I just went in and fixed merge conflicts - this is very close to ready to go, and the linting issues can be resolved by running `./bw-dev black`

PR #624, opened 2021-02-15
Show lists on book page

mouse-reeve:


todrobbins:
Rad! This will be really helpful to discovery!

PR #625, opened 2021-02-16
Inbox refactor

mouse-reeve:
remaining work off the top of my head:
 - [x] fix unfollow view
 - [x] handle tombstones
 - [x] ambiguous serializers for various kinds of Add
 - [x] clean up calls to to_model
 - [x] test inbox Add to list and tag
 - [ ] set last updated by value in book data models
 - [x] make sure deletion/tombstones are working correctly (w/ tests)


bookwyrm/activitypub/verbs.py
```diff
@@ -29,57 +35,106 @@ class Delete(Verb):
...
 
 @dataclass(init=False)
 class Undo(Verb):
     ''' Undo an activity '''
     type: str = 'Undo'
 
+    def action(self):
+        ''' find and remove the activity object '''
+        # this is so hacky but it does make it work....
+        # (because you Reject a request and Undo a follow
```
mouse-reeve, line 64: tf is this, I think this could be better solved with a function on the model for unfollowing?

bookwyrm/activitypub/verbs.py
```diff
@@ -29,57 +35,106 @@ class Delete(Verb):
...
+        # we want to related field that isn't the book, this is janky af sorry
+        model = [t for t in type(target)._meta.related_objects \
+                if t.name != 'edition'][0].related_model
+        self.to_model(model=model)
+
 
 @dataclass(init=False)
 class AddBook(Add):
     '''Add activity that's aware of the book obj '''
     object: Edition
```
mouse-reeve, line 137: this is an unnecessary subclass

PR #629, opened 2021-02-22
Update "running in production" section of Readme

jimfingal:
**Context:**

I was reading that section and was trying to follow along but was confused at the absence of some of the commands and files referenced, so I thought that would be helpful!

**Change Description:**

This updates the "running in production" section of the Readme with:

* A few minor spellchecks and cleanups
* Generalizing a "mailgun" account to an SMTP account such as mailgun you can use to send emails
* Adds a small paragraph contextualizing the significance of the `production` branch and how it differs from the `main` branch


mouse-reeve:
this is great, thank you!!

PR #630, opened 2021-02-22
Proof of concept adding `black` command

jimfingal:
This is more of a proof of concept of installing a runnable version of the `black` auto-formatter -- I've found this tool very useful in my work projects to support a generally-agreed-upon python style. The first time you run it, it makes a lot of changes, but then it's less dramatic as time goes on. There are plugins for IDEs such as VSCode that run this command every time you save a file, which allows you to just not think about some standard formatting tradeoffs.

You can see the changes this would apply by checking out this branch and running `make itblack`.

Take it or leave it w/r/t actually applying this to the project but thought you might be interested!

Ref: 

- https://github.com/psf/black
- https://github.com/psf/black/blob/master/docs/the_black_code_style.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,0 @@
-.PHONY: itblack
-
-itblack:
-	docker-compose run --rm web black celerywyrm
-	docker-compose run --rm web black bookwyrm
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,5 @@
+.PHONY: itblack
+
+itblack:
+	docker-compose run --rm web black celerywyrm
+	docker-compose run --rm web black bookwyrm
--- a/bw-dev
+++ b/bw-dev
@@ -35,6 +35,10 @@ function initdb {
     execweb python manage.py initdb
 }
 
+function makeitblack {
+    runweb black celerywyrm bookwyrm
+}
+
 CMD=$1
 shift
 
@@ -96,7 +100,10 @@ case "$CMD" in
     clean)
         clean
         ;;
+    black)
+        makeitblack
+        ;;
     *)
-        echo "Unrecognised command. Try: build, clean, up, initdb, resetdb, makemigrations, migrate, bash, shell, dbshell, restart_celery, test, pytest, test_report"
+        echo "Unrecognised command. Try: build, clean, up, initdb, resetdb, makemigrations, migrate, bash, shell, dbshell, restart_celery, test, pytest, test_report, black"
         ;;
 esac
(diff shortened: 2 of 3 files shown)


mouse-reeve:
oh this is so good but it wants me to use double quotes instead of single 😵. but I _suppose_ consistency is better than the quote I like, thank you!

jimfingal:
I *think* it may be possible to turn that off (later versions of black are less absolutist than earlier versions and offer a few configuration values), description here: https://github.com/psf/black/blob/master/docs/the_black_code_style.md#strings

*Technically* docstrings should have triple-double quotes [according to PEP-257](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) but there is no enforcement, and it's your project =)

jimfingal:
This is probably more a hassle during the startup phase if you have a groove going and are doing most of the development -- may be helpful in the future if there start to be multiple maintainers and stylistic matters are something people start to have debates about. Will close for now to remove any implicit pressure to respond or merge, was just something that occured might be helpful one day as I read through code.

mouse-reeve:
wait don't close it! I run into stylistic differences with contributors all the time and this would be a great way to avoid them altogether!

mouse-reeve:
the `make itblack` command isn't working for me, but `docker-compose exec web black bookwyrm` is, so I think it's a difference with how `run --rm` works? maybe because I'm also running the docker images simultaneously.

jimfingal:
Yeah, I think that's right -- `exec` runs within existing containers and `run` expects to run in a new container. It's possible it's  a no-op if the container is already running. I tend to use `run --rm` when I am trying to execute a stateless command from my local machine (spin up the container, run a thing, spin it down and remove it).

ref. https://medium.com/@zhao.li/how-to-understand-the-difference-between-docker-composes-up-vs-run-vs-exec-commands-a506151967df

mouse-reeve:
Okay, I'm as emotionally ready as I'll ever be to let go of single quotes. My one preference would be making this a `bw-dev` command rather than adding a Makefile, unless there's a particular benefit to that that I'm not seeing?

jimfingal:
@mouse-reeve that seems reasonable since that's your main entrypoint! Did that in 33cf03d.

Do you have a preference for `exec` over `run` for this particular operation? My inclination would be to keep the `run` syntax -- since this will be formatting the code in place, that would be an atomic action that should be able to run outside of the world of a running bookwyrm instance -- it feels like you shouldn't need the webserver running to run the code formatter.

mouse-reeve:
yes, using `run` is better!

PR #652, opened 2021-02-27
ReivewRate fixes

cincodenada:
PRs into PRs! What a world

This makes the downward migration work in the silliest way, which is...just doing nothing, so that Django will drop the table, which accomplishes what we want. Sheesh!

Also I dig the new string formatting, but I replaced it with SQL parameters, which are safer and the way to go when you're messing with direct queries.

I also added second commit with some even more SQL-y code, to insert all the values at once, that you can take or leave. I tried to do it the way I advised in the original PR and ran into probably the same thing you did, which is that the `content` column is actually on some other parent model's table, which could be worked around with JOINs, but I decided I didn't want to mess with that, but did the next best thing, which is using the query we already had, gathering all the ids we want together, and inserting them with one big `INSERT INTO...VALUES` statement. I even found a fancy psycopg helper that makes it nicer to work with VALUES lists, whee!

I tested these and made sure they got the same results as the original code (and reversed cleanly) by dumping the database before migrating, after migrating up, and after migrating back down, and all looked as expected.

### DIFF
--- a/bookwyrm/migrations/0046_reviewrating.py
+++ b/bookwyrm/migrations/0046_reviewrating.py
@@ -4,6 +4,7 @@ from django.db import migrations, models
 from django.db import connection
 from django.db.models import Q
 import django.db.models.deletion
+from psycopg2.extras import execute_values
 
 def convert_review_rating(app_registry, schema_editor):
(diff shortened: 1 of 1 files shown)


mouse-reeve:
whew, what an exciting journey in the depths of sql

PR #672, opened 2021-03-02
[i18n:fr_FR] Update locale

arkhi:
Round 2. :)

- I updated both .po and .mo based on the changes you made in #666; they seem to work well (for French, at least ;) ).
- I also misinterpreted some contexts (_commented on_, etc.); this should now be fixed.

I’m sure there will be some changes needed once it is tested extensively but that should be a _good enough_ base. It might be worth it to ping French people you know to give their feedback as a form of proof‑reading.


arkhi:
Hold that thought. I’m going to push more changes.

arkhi:
I understand _Post_ and _Status_ as being the same; do they need to be called differently?

arkhi:
All right… I made some updates that should prevent French readers to bleed from their eyes when reading statuses. :roll_eyes: 

The current translation read something like _John Doe with reviews Moby Dick_.

mouse-reeve:
> I understand _Post_ and _Status_ as being the same; do they need to be called differently?

"Post" in the context of line 1332 is being used as a verb (as in "click here to post a status"), but in the context of "post privacy" and "delete post", it is being used interchangeably with "status". Definitely in the latter case, I should change the wording in the app to "delete status"

mouse-reeve:
I'm merging this and #673 so I can make some template fixes

arkhi:
> > I understand _Post_ and _Status_ as being the same; do they need to be called differently?
> 
> "Post" in the context of line 1332 is being used as a verb (as in "click here to post a status"), but in the context of "post privacy" and "delete post", it is being used interchangeably with "status". Definitely in the latter case, I should change the wording in the app to "delete status"

Thanks for the explanation. That’s what I understood, so we’re all good on my side. :)

PR #673, opened 2021-03-02
[i18n:zh_CN] Add Simplified Chinese translation

GuDzpoz:
Hello there! I just did some Chinese translation and wonder if you mind adding it to bookwyrm.
I won't say it's perfect but at least I tried to keep consistency of the translation of Mastodon and some popular goodreads-like websites in China.

By the way, I might have a tiny suggestion for the original text. To start with, it is really difficult to translate the following string into Chinese because it has the "by who" part fixed at the end of the string. 
```po
#: bookwyrm/templates/lists/list.html:35
msgid "Added by"
```
Most ideally, I might suggest something as follows, and I suppose there are other languages that want `%(user)s` in the middle.
```po
msgid "Added by %(user)s"
msgstr "由 %(user)s 添加"
```
As I don't understand the code part, I doubt whether it is actually possible. But if it is, it can really helps.

Anyway, thank you for the amazing work!


mouse-reeve:
definitely doable! thank you!!

arkhi:
真厉害你会翻译了 !谢谢。:)

PR #675, opened 2021-03-02
Added German translation

tofuwabohu:
I copied the English .po file and translated the strings to German. In some cases, there's no German translation because we use the English word or it's the same (eg. ISBN), here, I left `msgstr` empty. I tried keep the "spirit" during the translations rather than doing everything word by word and took some creative freedom when I was unsure or there was no direct translation. I think, everything should be understandable if I got the context of the strings correctly.


mouse-reeve:
this rules! thank you so much! Would you like me to tag you in ongoing changes to i18n? If you're interested in following changes (with zero pressure to maintain this locale, I'm supper happy for whatever amount of time and energy you're willing to share), I'd be happy to

tofuwabohu:
Yes please :)

PR #678, opened 2021-03-03
Reduce Dockerfile size and the number of RUN steps

Flameborn:
I've modified the Dockerfile when experimenting with creating a standalone Bookwyrm-only set of images.

This reduces the image size by about 80 MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,15 +2,12 @@ FROM python:3.9
 
 ENV PYTHONUNBUFFERED 1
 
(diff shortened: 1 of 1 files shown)


mouse-reeve:
thank you! I'll remove this line form the production dockerfile entirely as well, since it's just used for compiling the locale files

Flameborn:
I'm glad to help. I sent the PR against *main* in case you think it needs more testing before merged into *production*.

PR #682, opened 2021-03-04
Spanish translation and one (1) typo fix

reesporte:
Not sure if my typo fix will result in needing to edit and recompile the other translations, but I figured that'd be better for another pull request.


mouse-reeve:
yay! thank you so much

PR #688, opened 2021-03-04
Add .editorconfig files and global linting Github Workflow.

arkhi:
I was thinking this kind of stuff would be better done early with as little changes and people involved as possible.

I added the .editorconfig files and updated the files until I had no more error.

Let me know if I should change anything, thanks!

I’ll make another PR with eslint and stylelint using my default configuration once this one is merged; let me know if you have any preference instead as I’m likely to refactor a bit around with a large trout. :)


mouse-reeve:
I love the editorconfig and I agree with the whitespace changes! With regard to the linter, I wonder if getting #630 into the deployment flow would be easier on contributors than having a linting check? But since that is only a python formatter, it would be good regardless to have an es linter.

arkhi:
OK. I’ll let you check my last two PR and will move forward later. :)

arkhi:
@mouse-reeve: I spent a non trivial amount of time fixing this PR after the latest merges on main. The longer the discussion on linters will be post-poned, the less likely this PR will be merged because the more work it’s going to become.

mouse-reeve:
thank you for all this work!

arkhi:
> thank you for all this work!

My pleasure; thanks for the merge. :)

Don’t hesitate to ping me if this linter causes any head-scratching, though. I will to help if possible.

PR #689, opened 2021-03-04
Removes fuzzy tag from spanish translations

mouse-reeve:
It turns out that any string associated with the "fuzzy" keyword doesn't get translated when the messages are compiled. I think all the unfuzzy-ing and my modest additions are correct, but I defer to you, @reesporte. The one string I noticed that isn't translated is "To Read" on line 482


reesporte:
looks good to me! i'll submit a pull request to translate the missing line

PR #691, opened 2021-03-05
very minor es translation tweak

reesporte:
changed a line. very productive commit


mouse-reeve:
unironically a productive commit though. actually, would you mind removing the fuzzy flag on line 479 while you're in there?

reesporte:
got it!

Issue #467, opened 2021-01-03
Use openlibrary isbn api


mouse-reeve:
since I wrote the openlibrary connector initially, an isbn api has been added (or at least, documented) -- this might improve goodreads import quality, which currently does a general search on the isbn, but could instead use the isbn api.

https://openlibrary.org/dev/docs/api/books

PR #697, opened 2021-03-06
ISBN search

mmai:
This is a proposition to fix the #467 issue and to allow other connectors beyond OpenLibray to  have a  custom ISBN search method.
The global search now try first a ISBN search if the query looks like a isbn10 or isbn13, and fallback to regular search if no result is found.

For this I made a new migration which creates a 'isbn_search_url' field in the Connector table,  and  I've updated the initdb.py script with the values for OpenLibrary and Bookwyrm instances.


mouse-reeve:
This is great! I can't find anything to quibble about, thanks for your thorough work and tests :)

PR #701, opened 2021-03-07
Screen reader improvements

Flameborn:
I've added some improvements for screen reader users:

* The currently active page is now indicated on the home/local/federated status pages for their respective links.
* The avatar image is now completely hidden on status pages. There is a link for the user whose status is displayed, so in this case the image does not serve a purpose to screen reader users and it was quite chatty while navigating through headings.
* I've also noticed a small typo, which I fixed.

Hopefully my non-existent Django templating knowledge is not a problem ☺️. This should not break anything visually, but please let me know if it does.


mouse-reeve:
beautiful! thank you

PR #714, opened 2021-03-09
Turn on Automatic backups for the DB

bcj:
This PR modifies the DB container to start `cron` at startup so that the automatic backup script will run.

It also adds a script for automatically pruning the backups in the DBs backup directory. The script is turned off by default but instructions are included for enabling it.


mouse-reeve:
backup code: judicious

PR #715, opened 2021-03-09
Reporting

mouse-reeve:
Let users flag posts and users
- [x] Reports models
- [x] Button for making a report
- [x] Reporting form modal
- [x] Admin reports list view
- [x] Admin reports detail view
- [x] Moderation actions
- [x] Resolve reports
- [x] Comments on reports
- [x] Screen reader review


rkingett:
Bad bot. I did spot a missing label in it's quote, though, but will need to check again when I'm more awake.

PR #732, opened 2021-03-13
Search info

### LINKED ISSUE #718: Preview Books Before Importing

When importing a book that is not yet in Bookwyrm that has multiple OpenLibrary entries, there is no easy way to decide which version to choose without importing all of them or searching on OpenLibrary and comparing there.

E.g., searching Illiad gives `Illiad by Homer (2009)`, `The Illiad by Homer (1997)`, and `Homers Illiad by Unknown (1994)` (along with several results that aren't The Illiad)

I would like the title of the entry to be a link to the OpenLibrary page (like how the already-imported books link to their Bookwyrm pages), and, ideally, a way to see at least some of the other metadata that would be imported without leaving the page.

If I could import by posting an OpenLibrary link (which I guess is related to #296), it could partially solve this since some books may not be distinguishable prior to import just from title/author/year.

Being able to directly add a book to Bookwyrm by supplying the information myself would also be useful but is significantly more work and adds ongoing moderation work.

mouse-reeve:
Fixes #718 
A search result preview where the title is a link and a cover is visible


mouse-reeve:
openlibrary shows a cover preview that isn't necessarily the cover of the default edition, which is weird, but I think whether a cover is available is present is a useful indicator when selecting which search result to import.

PR #784, opened 2021-03-23
Uses redis to manage activity streams

mouse-reeve:
Work:
- [x] Update README
- Administration
  - [x] Add second redis image for streams
  - [ ] Configure password protected redis instance for prod
  - [x] Management command to populate streams
- Ongoing stream updates
  - [x] Push status to streams on create
  - [x] Trim streams
  - [x] Remove status on delete
  - [x] Add statuses on follow events
  - [x] Remove statuses from feeds on block, unfollow events
  - [x] Track unread statuses for respective feeds
- Populating streams
  - [x] Functionality for populating streams wholesale
  - [x] Populate streams on account creation
- Tests 
  - [x] Add mocks to existing tests
  - [x] Test stream classes
  - [x] Test signals
- User interface
  - [x] Show unread notice


mouse-reeve:
I'm tagging folks who may be running instances because this change requires special handling! Feel free to ignore if you're doing other things, or bug me if there are things I need to clarify: @cincodenada @tofuwabohu @bcj 

This change creates a new redis image which handles all the user timelines, so updating with this changes requires:
- update `.env` with new environment variables added to `.env.example` and generate an additional redis password
- running `docker-compose build` to re-build everything to handle the new image and renaming the existing redis image
- running `docker-compose exec web python manage.py rebuild_feeds` to initialize the user feeds on your instance (otherwise, they'll be blank)

(also if you want to add your instance to https://github.com/mouse-reeve/bookwyrm/blob/main/instances.md you're welcome to but no pressure)

Flameborn:
After executing the steps above, it seems like my feed contains updates up to the 20th of November last year. Newer status updates via Bookwyrm are missing.

Search also returns a server error without anything in the logs, though I'm not sure if this is related.

mouse-reeve:
> After executing the steps above, it seems like my feed contains updates up to the 20th of November last year. Newer status updates via Bookwyrm are missing.

Thank you! This should fix it: https://github.com/mouse-reeve/bookwyrm/pull/790

mouse-reeve:
I'm not sure what's causing the search error, but it might be because of the bookwyrm.social downtime earlier. Is search still having errors now that b.s is back up?

Flameborn:
Good catch, search seems to be working now.

Do you think it'd be possible to add a search type, e.g. local, federated, open library only, etc? Just an idea so that search would not stop functioning in case the main instance is down ☺️

mouse-reeve:
yeah it definitely shouldn't stop working when one of the connectors is down!

Flameborn:
Also, https://github.com/mouse-reeve/bookwyrm/pull/790 seems to have fixed the status issue. Thank you!

PR #800, opened 2021-03-26
716 books can be on multiple shelves, but only on one shelf for reading status

SavinaRoja:
https://github.com/mouse-reeve/bookwyrm/issues/716

There are some tests which are not passing, but I'm not certain if they were introduced in this branch. I need to improve my familiarity with Django and the testing framework.

If Shelves are either Custom or Reading Status ("to-read", "reading", "read"), are each of these transitions valid for a Change Shelf action? The present implementation accepts them all.

 - Custom -> Custom | :ok: 
 - Reading Status -> Reading Status | :ok: 
 - Custom -> Reading Status | :question: 
 - Reading Status -> Custom | :question:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 *.pyc
 *.swp
 **/__pycache__
+.local
 
 # VSCode
 /.vscode
@@ -17,3 +18,6 @@
 
 # Testing
 .coverage
+
+#PyCharm
+.idea
--- a/bookwyrm/models/shelf.py
+++ b/bookwyrm/models/shelf.py
@@ -11,6 +11,12 @@ from . import fields
 class Shelf(OrderedCollectionMixin, BookWyrmModel):
     """ a list of books owned by a user """
 
+    TO_READ = "to-read"
+    READING = "reading"
+    READ_FINISHED = "read"
+
+    READ_STATUS_IDENTIFIERS = (TO_READ, READING, READ_FINISHED)
+
     name = fields.CharField(max_length=100)
     identifier = models.CharField(max_length=100)
     user = fields.ForeignKey(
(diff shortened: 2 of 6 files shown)


SavinaRoja:
Cool. The tests are on CI. I will review and circle back.

SavinaRoja:
I haven't figured out the bit that is alarming the CodeQL check here yet. What I was aiming to do was to have the page return to whatever page submitted the post and borrowed some code from another view. Will follow up on that.

mouse-reeve:
Yeah, it doesn't like that because the user can set the referrer header to anything, but there's are a ton of views I wrote that do it, so it's not, in my opinion, a problem that needs to be fixed in this PR :)

mouse-reeve:
hm, I put a shelf on a custom shelf, and when I pressed the "want to read" button it was removed from my custom shelf. I think books should stay in place and be added to the to-read shelf in this case? I'm not totally sure what behavior makes the most sense

SavinaRoja:
Here are some scenarios I considered and tested:

1.  Setting a shelf from a book's page. This seems like it would be a simple addition of a shelf (unless a different reading status is already selected, then change it).

https://user-images.githubusercontent.com/454944/112738080-19df4280-8f36-11eb-9b3d-48cce87bdec1.mp4

https://user-images.githubusercontent.com/454944/112738203-126c6900-8f37-11eb-9a43-23fdc1fd86ca.mp4

2. Unshelve from a book's page. This should just be a removal of the selected shelf.

https://user-images.githubusercontent.com/454944/112738185-e2bd6100-8f36-11eb-8b8d-8513a02771b1.mp4

3. Change a shelf from a book's page. This should replace the original shelf with the new shelf?

https://user-images.githubusercontent.com/454944/112738160-b43f8600-8f36-11eb-85fe-71940d09e5c8.mp4

4. Change shelf from the shelves page. This should replace the original shelf with the new shelf?

https://user-images.githubusercontent.com/454944/112738213-32039180-8f37-11eb-8fee-0dc3088a7328.mp4

SavinaRoja:
Let me know what you think here. And if I missed any other user interaction points with the shelving of books.

mouse-reeve:
This looks good to me! Thank you for the videos :)

PR #802, opened 2021-03-26
Adds "suggested follows" panel

mouse-reeve:


rkingett:
Could this be toggled on and off? Just wondering! But this will help me browse other YA readers!

mouse-reeve:
I'm going to add a way to, yes. Maybe tomorrow :)

PR #812, opened 2021-03-28
357 Users set preferred timezone, used for readthroughs

SavinaRoja:
https://github.com/mouse-reeve/bookwyrm/issues/357

Adds a user setting for preferred timezeone with a default to UTC:

image

and a translation for the `es` locale.

Updating a readthrough now uses this timezone and converts it to UTC for persistence.

image


bookwyrm/templates/preferences/edit_user.html
```diff
@@ -61,6 +61,10 @@
         {% url 'directory' as path %}
         <p class="help">{% blocktrans %}Your account will show up in the <a href="{{ path }}">directory</a>, and may be recommended to other BookWyrm users.{% endblocktrans %}</p>
     </div>
+    <div>
+        <label class="dropdown label" for="id_preferred_timezone">{% trans "Preferred Timezone: " %}</label>
+        {{ form.preferred_timezone }}
+    </div>
```
mouse-reeve, line 67: for stylistic consistently, this will work best if the `select` element itself on line 66 is wrapped in a div with class `select` (https://bulma.io/documentation/form/select/), and the div on line 54 gets the class `block`. I believe `dropdown` is intended for html menus, not for select form elements.

It's fussy, sorry! I can also fix it later if you prefer not to get into the weeds on markup.

bookwyrm/templatetags/bookwyrm_tags.py
```diff
@@ -139,7 +139,6 @@ def time_since(date):
     if not isinstance(date, datetime):
         return ""
     now = timezone.now()
-    delta = now - date
```
mouse-reeve, line 142: oops! thank you

bookwyrm/views/reading.py
```diff
@@ -141,16 +149,18 @@ def update_readthrough(request, book=None, create=True):
     start_date = request.POST.get("start_date")
     if start_date:
         try:
-            start_date = timezone.make_aware(dateutil.parser.parse(start_date))
-            readthrough.start_date = start_date
+            readthrough.start_date = load_date_in_user_tz_as_utc(
```
mouse-reeve, line 152: I wonder if this is the best place to intervene with the timezone? If I'm understanding this right, the user's preferred timezone is treated as utc, and stored accordingly in the database. As a result, when the template to select read dates is rendered, "now" still appears as a UTC date, and on save, the date that gets stored would be inaccurate because of the translation between timezones.

If the timezone gets configured in the template, I think the user can select the date in their preferred timezone, and it will be correctly annotated to save (https://docs.djangoproject.com/en/3.1/topics/i18n/timezones/#selecting-the-current-time-zone). 

That said, I'm fuzzy on how all this works and might be completely off base!

mouse-reeve (changes requested):
Thank you for this! I'm excited to be able to one day leave UTC

bookwyrm/templates/preferences/edit_user.html
SavinaRoja: Front end is my weak point. Thanks for this note. I can read up on this and try to address it or I'll ask you for help.

mouse-reeve: it's one of those things that's easy if you already know the css framework, and super non-obvious when you don't

bookwyrm/views/reading.py
SavinaRoja: What I have learned in my experience is that you want to be super consistent in how you persist. UTC everywhere in the database is a great, that way you have a clear target for converting to and from. So regardless of the user's preferred timezone, we should convert to UTC for storage; when retrieving the value from storage, we should convert it to the member's preferred timezone for display.

While I have covered the input from the user here, I think that I have overlooked the output...

I think that we could employ a middleware to curry the request context with the correct timezone for the user or fall back to UTC if not authenticated.

mouse-reeve: yeah agreed! I'm still trying to get my head inside what the conversion does so I think I just misunderstood it. Middleware seems like a good approach to me :)

bookwyrm/templates/preferences/edit_user.html
SavinaRoja: Looks better!

image

bookwyrm/timezone_middleware.py
```diff
@@ -0,0 +1,17 @@
...
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.user.is_authenticated:
+            timezone.activate(pytz.timezone(request.user.preferred_timezone))
+        else:
+            timezone.activate(pytz.utc)
+        response = self.get_response(request)
+        timezone.deactivate()
+        return response
```
SavinaRoja, line 17: This should be a Django-friendly timezone middleware https://docs.djangoproject.com/en/3.1/topics/i18n/timezones/

I am not 100% sure if the explicit call to deactivate is necessary. What I read led me to believe that that is performed after the view is called for the request and response, and I thought it might be safe to clean up. https://docs.djangoproject.com/en/3.1/topics/http/middleware/#writing-your-own-middleware

Even if Django recycles threads I think that this might be superfluous, though.

mouse-reeve:
looks good to me! thank you

PR #831, opened 2021-03-31
Add linters for CSS and JS files

arkhi:
This PR adds linters for static files (css, js, jsx, ts, tsx) to Github Actions:

It avoids tracking some files I understood as _vendor files_. Ideally, it would be great to move those files in a _vendor_ directory of some sort. I understand this would simplify automatisation and understanding of what should not be updated manually.

I kept the rules as very minimal to avoid having to refactor all files and be more flexible for now. Rules can (should?) be added as the project evolves, but I can propose my own recommendations in another PR if you wish.

I also added minimal hacks and their @todo to prevent some errors that would need refactoring if not with the hacks. I’m thinking about the rating stars, for example.

Let me know. :)


mouse-reeve:
Rad! Let's have _all_ the linters

PR #842, opened 2021-04-01
Wording consistency and locales compiled for getting started views

mouse-reeve:


mouse-reeve:
not sure why Black is failing here but it doesn't seem related to the PR.

PR #844, opened 2021-04-01
Invite request sorting

mouse-reeve:


mouse-reeve:
seems like github actions are flaking out. I ran Black and test locally.

PR #862, opened 2021-04-03
Valid html solution to disabling pagination links

mouse-reeve:
Uses a class to disable pagination links, rather than the `disabled` attribute (cc @arkhi )

In my screenreader (mac OSX VoiceOver), a disabled pagination link reads as plain text, and an active link reads as a link, which seems fine to me.

### DIFF
--- a/bookwyrm/static/css/format.css
+++ b/bookwyrm/static/css/format.css
@@ -25,14 +25,13 @@ html {
     min-width: 75% !important;
 }
 
-
 /* --- "disabled" for non-buttons --- */
 .is-disabled {
     background-color: #dbdbdb;
     border-color: #dbdbdb;
     box-shadow: none;
     color: #7a7a7a;
-    opacity: .5;
+    opacity: 0.5;
     cursor: not-allowed;
 }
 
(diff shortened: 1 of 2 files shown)


arkhi:
Thanks. :)

I am wondering if disabled elements should be read by screen readers at all. Maybe @rkingett or @Flameborn would have a better idea on this topic?

Flameborn:
The tricky part is figuring out whether the disabled element would add any meaningful information to the page.

For example, with pagination, if there is a *next page* link when there is only one page, hiding it might be a better idea than showing it with a disabled state, because when these controls are inactive, they do not convey any information, nor they can be interacted with.

Having, however, a book on the *read* shelf is meaningful information, even though the shelf selector is disabled.

*Aria-hidden* is also useful in a situation like this.

mouse-reeve:
Thanks, yeah! I guess the only reason I can think not to use aria-hidden is that it lets you know that pagination is generally available for that page, even when there aren't enough items to warrant it. But I don't think everyone is super interested in hypothetic pagination, so hiding the links is probably better 😄

PR #869, opened 2021-04-04
Simplify rating form

arkhi:
I stumbled on the CSS for this feature and was a bit confused so I thought I could work on it to understand and maybe improve it a bit. Here’s the result.

- Create a snippet to regroup similar code.
- Reduce and document tricky logic of CSS.
- Add i18n strings.


mouse-reeve:
wow this is great, thank you for untangling this one!

PR #873, opened 2021-04-04
772 Use gettext_lazy for form classes

SavinaRoja:
https://github.com/mouse-reeve/bookwyrm/issues/772

`gettext` at import time will be static, so use `gettext_lazy` instead.


mouse-reeve:
Ah thank you! I was totally flummoxed about this one!

Issue #713, opened 2021-03-08
Performance improvements


Flameborn:
I am currently running a production Bookwyrm on a Raspberry Pi 3, with both the Postgres database and Bookwyrm being on a hard drive. I'm using Caddy as a web server. My current install is without using Docker.

My feed of about 1000 imported books takes about 10 seconds to open on the local network, the same is true for a user who has about 100 imported books. The page does not have statuses displayed from other users.

I am getting similar results for the status page or a 1000-book import.

My read shelf contains about 600 books or so, this absolutely refuses to load via Safari under MacOS, and in addition, I receive process timeouts and reboots in the logs. My CPU usage is around 1.0, which is about a quarter of the total CPU capacity. Memory also does not seem to be an issue, about 300 MB is used with other services running as well.

If you don't mind sharing, what's the performance of Bookwyrm like on bookwyrm.social?

Would it be possible to improve this experience, e.g. via paging?

mouse-reeve:
Those could and should definitely be paged! I haven't been tracking performance on bookwyrm.social

Flameborn:
The new pagination seems to have sped up loading times quite significantly in most cases.

The feed, however, takes about 20 seconds to load now on a Raspberry Pi 3.

I have as much data as possible on a hard drive, such as the database, Python packages, covers, etc.

As a result, Gunicorn worker processes often time out and restart, resulting in an empty page, when Bookwyrm is opened in Safari for a logged in user.

I'll likely need to investigate how profiling works for Django apps to provide more info on this.

I realize that this issue is partly out of scope, as neither Python or Bookwyrm were designed to run on low-end hardware (an ARM Cortex A53 in this case), so I am only posting updates regarding this, in case some more performance improvements can be made in the future.

mouse-reeve:
yikes! Is that a worse load time than before? there may be some very inefficient queries behind this. 

While it's true that it isn't intended to run on the Pi, performance improvements are improvements for every platform, so I appreciate you keeping an eye on this

Flameborn:
Unfortunately, yes, about double the time of what it used to be.

Profiling this is going to be difficult on a Pi, I've just tried *Cprofile* and *Django Silk*, however, the overhead is so huge that the worker processes keep timing out when I try and access the feeds page, no matter what browser I use.

mouse-reeve:
[...]

PR #880, opened 2021-04-05
Uses redis for storing suggested users

mouse-reeve:
Computing suggested users is too heavy an operation to do on the fly, so this would store and update a list of suggestions for each use in redis. It probably also needs to be turned into a background task or debounced in some way because the update operations are costly.

Works on #713 

- [x] Create redis stores for user suggestions
- [x] Populate stores from empty when local user is created
- And and remove users from stores on the fly
  - [x] Add newly discoverable users to everyone's suggestions
  - [x] Remove newly non-discoverable users from all suggestions
- Update relative suggestion rankings
  - [x] When a book is shelves or unshelved
  - [x] When someone follows or unfollows someone else
  - [x] When a block takes place
- User redis to serve suggested follows
  - [x] In timeline
  - [x] In onboarding
- [x] Management command to re-build suggestions
- [x] Update tests

Remaining work:
- [x] Make sure change password isn't broken
- [x] Manually test feature:
  - [x] Inset in feed
  - [x] Get started
  - [x] Directory
- [ ] Screen reader review


mouse-reeve:
@bookwyrm-social/instance-admins You'll need to run `./bw-dev populate_suggestions` to get this working. This should help with speed when loading the feed view

PR #892, opened 2021-04-06
Adds inventaire connector

### LINKED ISSUE #771: Let admins set preferred default edition language
Right now, the default edition of a book prefers english-language editions. An instance admin should be able to set it to pick whatever language is appropriate to the instance.

### LINKED ISSUE #937: Potential problem with the "main" version of a book
1. A list of books is imported from a GoodReads export, using the "Import books" feature.
2. Some of these books don't have an ISBN, so all their version are imported, and the one with the most info is used as a "main" edition (sometimes it is the audio cd version instead of print).
3. My imported reviews are applied on the "main" edtion.
4. I go see the list of editions, and switch to a new one.
5. My "Read" shelf displays the new edition (yay).
6. My review appears at the bottom of the page for that new edition.
7. The stars rating does not appear below the cover on that new edition.
8. Adding a rating on the new edition counts as a new review for the book.
9. When looking at an author's page, the list of books shows the old "main" edition, instead of the new one that's in my shelf.

**Expected behavior**

Steps 1-8 seem logical to me. Step 9 seems illogical. Step 4 should perhaps move my review to the new selected edition.

I don't know how the "main" edition is selected as "main", but IMO the list should show the edition that's in the most shelves.

**Screenshots**

On this screenshot you can see that the last item on the list is not the edition that's in my shelves (I'm the only user on my instance, so that edition is in no other shelf). I would prefer to see the edition that's in my shelf (mainly because audio cds have square covers and it's not very harmonious)

image

mouse-reeve:
This will allow users to import book data from inventaire.io as well as openlibrary.

- [x] Add connector to available list
- [x] Initialize new servers with the connector
- [x] Search inventaire in free text mode
- [x] Search inventaire in isbn mode
- [x] Load already-available edition/work metadata
- [x] Load edition/work metadata that requires lookup
- [x] Connect editions and works
- [x] Load and serialize authors
- [x] Expand book data
- [x] Re-think default editions
- [x] Tests
- [x] Fix links to original source books

Fixes #771
Fixes #937


fenimore:
I know OpenLibrary accepts CSV imports from Goodreads -- does this connector involve something like that?

arkhi:
I was having a quick chat with people on inventaire. They also have an [Issue about implementing ActivityPub](https://github.com/inventaire/inventaire/issues/187) Could this provide cross‑functionality between both project?

mouse-reeve:
> I was having a quick [chat with people on the [matrix] inventaire chan](https://matrix.to/#/!sRmiUrdcokwhhFTmxY:matrix.org/$G5g0ToY2qrdrK1o2hQl-RDURLIMvmjMgmcrj5vF-r74?via=matrix.org&via=t2bot.io&via=talk.puri.sm). They also have an [Issue about implementing ActivityPub](https://github.com/inventaire/inventaire/issues/187). Could this provide cross‑functionality between both project?

This work largely flattens the heavily interlinked wikidata style format the Inventaire uses into a flatter schema, and pulls out specifics pieces of data while dropping others. I'm not sure that it's directly applicable to getting them on the protocol, but I'm excited to see what direction their work heads in.

mouse-reeve:
@bookwyrm-social/instance-admins In order for this connector to be available, you have to add it to the database. It gets added to new instances when the `initdb` command is run.

You can add it from the Django Admin, `/admin/bookwyrm/connector/`, with the following values:
identifier: `inventaire.io`
name: `Inventaire`
connector_file: `inventaire` (from the dropdown)
base_url: `https://inventaire.io`
books_url: `https://inventaire.io/api/entities`
covers_url: `https://inventaire.io`
search_url: `https://inventaire.io/api/search?types=works&types=works&search=`
isbn_search_url: `https://inventaire.io/api/entities?action=by-uris&uris=isbn%3A`
priority: `3`

Alternatively, from the python shell (`./bw-dev shell`), you can run:
```python
from bookwyrm.models import Connector
Connector.objects.create(
        identifier="inventaire.io",
        name="Inventaire",
        connector_file="inventaire",
        base_url="https://inventaire.io",
        books_url="https://inventaire.io/api/entities",
        covers_url="https://inventaire.io",
        search_url="https://inventaire.io/api/search?types=works&types=works&search=",
        isbn_search_url="https://inventaire.io/api/entities?action=by-uris&uris=isbn%3A",
        priority=3,
)
```
Which will do the same thing.

### Optional, if you want to get into the weeds:
The `priority` field controls which search result will be tried first, both in regular search and when books are added in a goodreads import. As long as you don't mess with the "self" connector at priority `1` (which searches the local database), you can tweak the priorities of different connectors. 

By default, other bookwyrm instances (which are added automatically) are priority 2, and OpenLibrary and inventaire are priority 3

Issue #886, opened 2021-04-06
Make user handle search case insensitive


tofuwabohu:
**Is your feature request related to a problem? Please describe.**
I'm not sure if this is intentional, but currently, you can only look up users if you spell their name witht he correct cases. This is also the case when you try to find BW users from Mastodon. I don't think case sensitivity makes much sense here.

**Additional context**
Mastodon hase case insensitive user names as well.

One can also register `Tofuwabohu` when there is a `tofuwabohu` already. I think that should be prevented before instances get opened?

PR #908, opened 2021-04-08
Makes localname case insensitive

mouse-reeve:
This makes the uniqueness of local users' usernames case insensitive.
Works on #886


mouse-reeve:
If there are any existing case-insensitive duplicate usernames, this migration will fail. There aren't any on bookwyrm.social, but it's theoretically possible. I can't think of any programmatic way to resolve this, I think it would require having a conversation and reaching a compromise with the affected users. But hopefully no one runs into this issue!

tofuwabohu:
I have one (that I created to test if it's possible). It's a dummy account, what's the easiest way to delete one of them?

mouse-reeve:
if it doesn't have any posts or friends, you should be able to truly delete it from the database using django admin (`/admin/bookwyrm/user/` is a list of users `/admin/bookwyrm/user/<user_id>` if you know the id for that user). You can also change the localname in the django admin if you can't delete

tofuwabohu:
:exploding_head: I always thought you meant the bookwyrm admin settings when you spoke of the django admin interface, never seen the /admin thingy! Guess I have a lot to explore now :D

tofuwabohu:
Alright, couldn't delete it due to the auto-generated shelves but renaming worked :)

PR #914, opened 2021-04-09
Ensure `EMAIL_USE_TLS` is read as a bool

DeeUnderscore:
This change makes environs read `EMAIL_USE_TLS` as a bool, because otherwise it reads it as a string (so `"false"` ends up being True in boolean contexts, and TLS is always enabled)


mouse-reeve:
@tofuwabohu not sure if this is relevant to you

tofuwabohu:
I'll have a look at that, maybe I should do the same with `EMAL_USE_SSL` and then test again with the other mailserver. :)

PR #915, opened 2021-04-09
Adds SSL parameter

tofuwabohu:
With these settings, it worked on my local instance with the SSL connected server.

I'm assuming the second parameter in ` EMAIL_USE_SSL = env.bool("EMAIL_USE_SSL", False)` is the default parameter, but I'm not sure about that.


tofuwabohu:
A test failed and complained that `EMAIL_USE_SSL` is not set, so I set it to `false` in the example config in the hope that the deploying works.

tofuwabohu:
My next attempt would be to set a default `False` value in `celerywyrm/settings.py` too but I don' want to do too many unnecessary commits so I'll wait for now.

mouse-reeve:
I think setting the default values in celerywyrm/settings.py will fix it. I need to re-evaluate why there's a separate settings file there though, I think it's unnecessary.

Issue #84, opened 2020-03-13
HTML views should support microformats


bekopharm:
Consider adding [microformats](http://microformats.org/wiki/microformats2) to HTML views.

jalcine:
Seconded! Using Microformats will:

- make it easy for most Web tools to both present and parse
- keep it extensible with a battle-tested format
- make it friendly to presenting in the poster's preferred way without having to create alternative formats to show the same thing.

mouse-reeve:
thank you both for the suggestion! @jalcine can you clarify your third point? I'm not sure I understand that use case.

jalcine:
With a Microformats parser, you can define something in HTML (mostly) anyway you'd want but still get useful structured information (similar to how one would parse JSON).

An example is my own personal site! 

https://jacky.wtf to https://ruby.microformats.io/microformats?utf8=%E2%9C%93&url=https%3A%2F%2Fjacky.wtf

jalcine:
I'd be happy to brainstorm some ideas on how to define the static pages of fedireads with microformats with you!

mouse-reeve:
I see, that example is super helpful, thank you! I took a look a while back and as I recall it seems like there are a lot of natural fits around serializing book data, so brainstorming would be greatly ppreciated

bekopharm:
Mayhap I should have mentioned https://microformats.io/ too. It's probably a better example to get the idea.

arkhi:
There are existing definitions for [structured data for books](https://schema.org/Book). Same goes for [Rating](https://schema.org/Rating), [Person](https://schema.org/Person), etc.

I can work on implementing this in the [book template](https://github.com/mouse-reeve/bookwyrm/blob/main/bookwyrm/templates/book/book.html) unless anybody has any objection. :)

arkhi:
A useful browser addon to have is http://osds.openlinksw.com/.

arkhi:
An interesting problem is with the [publisher property](https://schema.org/publisher) which needs to be either an Organization or a Person.

mouse-reeve:
This would just be the `publishers` field that already exists in the book data, right?

arkhi:
>This would just be the `publishers` field that already exists in the book data, right?

I saw an included template with publisher details and the name of the publisher; is the type (person or organisation) available too?

mouse-reeve:
There's no additional metadata on that, I think it's safe to treat it as an organization.

arkhi:
[...]

PR #919, opened 2021-04-09
Add structured data to Book

arkhi:
This PR is a proposal for the [Book](/mouse-reeve/bookwyrm/blob/main/bookwyrm/templates/book/book.html) part of #84.

You can visualise the [diff without space changes](/mouse-reeve/bookwyrm/compare/main...arkhi:structured-data?expand=1&w=1).

I did what I could to avoid long lines for legibility. This means sometimes simplifying includes so that they are wrapped within `{% with %}` statements instead of inlined in the `{% include %}`.

### Potential Issues and questions

- I moved the reviews within the main `.block` in _book.html_ and saw no adverse consequence, but I might be missing something.
- The [bookFormat] property is limited to a list of values whereas the book edition is free text. I assigned the property as free text and added a `@todo`.
- [Author] and [Publisher] are supposed to be [Oganization] or [Person]. I used [Thing] which is the most generic schema in that branch and added a `@todo`.
- Is it possible to get the maximum rating value (5) in the templates instead of hard-coding it?
- About `keywords` vs. `about` vs. `genre`, I opted for `about` based on an [existing issue](https://github.com/schemaorg/schemaorg/issues/347).

### Resources

- [OpenLink Structured Data Sniffer Addon](http://osds.openlinksw.com/)
- [Google Structured Data Testing tool](https://search.google.com/structured-data/testing-tool#)

[Author]: https://schema.org/Author
[bookFormat]: https://schema.org/bookFormat
[Organization]: https://schema.org/Organization
[Person]: https://schema.org/Person
[Publisher]: https://schema.org/Publisher
[Thing]: https://schema.org/Thing


mouse-reeve:
There's isn't a way to get the max rating score programmatically, so hard-coding it is the best optional available right now. I think it would be fine to use Organization for Publisher and Person for Author, since that's the usual value - I guess it's just a tradeoff between but largely more accurate while occasionally being clearly wrong, vs being less specific but never _wrong_. So basically, I don't feel strongly about it and I think using `Thing` is just fine, but I can see the merits of either approach.

All in all looks great! Thank you

arkhi:
I like your very practical thinking and the fact you’re explaining your thoughts.

Thanks for the merge. :)

PR #926, opened 2021-04-12
update links in README and {bookwyrm,celerywyrm}/settings.py files

void-witch:
the link to celery's homepage has been dead for at least several months
activitypub.rocks now supports tls (yay!)
the links to django's docs were for older versions of django than what bookwyrm uses


mouse-reeve:
thanks so much!

void-witch:
of course c:

void-witch:
just gonna drop this here as context for the celery site in case anyone needs additional context in the future: https://github.com/celery/celeryproject/issues/47

PR #935, opened 2021-04-14
French locale: improve gender-neutral language

joachimesque:
This PR proposes improvements to gender-neutral language in French:

- Use neutral terms when possible
- Use the `·` type abbreviation (_point médian_) when neutral terms are unavailable

It also fixes a couple of bugs in the French translation.

Should I also generate the updated `django.mo` in the PR, or is it taken care at time of build? (Django noob here)


mouse-reeve:
Thank you! The _point médian_ is so interesting to me. Tagging @arkhi because I speak zero french and I figure it would be helpful for you both to be on the same page 🙂

arkhi:
Salut @joachimesque. Our roads cross again, old rascal! :)

Thanks for the ping @mouse-reeve!

The _point médian_ is indeed an interesting concept if we tag each occurrence with `aria-label` so that screen readers can actually read them properly.

I am not a good reader and find this punctuation unbearable as it really distracts me from the message, so I would be happy not to see any. That’s why I used _abonnement_ which is gender neutral, along with other terms. The strings I used were also based on the ones from Mastodon, in part for consistency and because, as far as I know, the community there is quite sensitive to the question of gender neutrality too.

So a big :-1: from me about the _point médian_ because it hurts me directly and it most likely hurts screen readers, but people’s mileage may vary and I would be happy to be proven wrong.

About the other changes, I’ll come back later but knowing @joachimesque, I’m sure we’ll find the best outcomes; I’m quite happy someone took the time to proof‑read what I did. :)

joachimesque:
Thanks for your answer!

To be honest, I only started the PR because of the very confusing sentence _« 6 abonnements, 5 abonnements »_ for "6 followers, 5 following" on the profile page, and it sort of evolved into these changes :)

The _point médian_ is a bad solution, I agree, but IMO it's the best solution [in a handful of bad ones](http://romy.tetue.net/feminiser-au-point-median). I won't insist in getting rid of paretheses if you think they are more legible, even though they are seen as devaluing the feminine. I think parentheses are more legible to us because we've seen them around for a long time, and I see no reason why we couldn't get used to the _point médian_ (only works if we give them a chance).

So as I said I can revert the `·` to `()`. As for the screen readers, I just learned that the _point médian_ [is better to be avoided](https://www.lalutineduweb.fr/ecriture-inclusive-accessibilite-numerique-jetdv2020/), though I don't know if the parentheses fare better—but I have to try (I can only test with VoiceOver though). That's why I think the best course of action first is to see which epicene words can be used in place of gendered ones. 

Mastodon uses _« 6 abonnements, 5 abonné·e·s »_ for "followers", and I don't think it's the right solution (I won't get into why `é·e·s` is worse than both `é·es` and `é(e)s`). Pleroma uses _« 6 suivis, 5 vous suivent »_ which is not good either, as the problem is now on _suivis_ instead of _abonnés_.

A longer way would be to use _« 6 comptes suivis, 5 comptes abonnés »_ but I'll be the first to say it's cumbersome, plus it's depersonalizing—you follow a person, not an account. And using _« 6 personnes suivies, 5 personnes abonnées »_ is more cumbersome.

An out-of-the-box proposal : _« ${name} suit 6 personnes, 5 suivent ${name} »_. More verbose than we're used to on social networks, but… well… it's for readers… so we should not be afraid of sentences, should we? 😅 

What do you think?

Naturally, the question that can be asked is _is this really necessary to display someone's following_? But let's say it's out of scope for this PR 😄 

---

@mouse-reeve: most of the material on French inclusive language is in French, sadly, but I found these two articles in English about the whole debate.

- https://languagelog.ldc.upenn.edu/nll/?p=34912
- https://www.theatlantic.com/international/archive/2017/11/inclusive-writing-france-feminism/545048/

Of course it's very politicized on both sides. French people love institutions, and the _Académie Française_ is the one that's in charge of the language. It's actually responsible for erasing some job names in the feminine form in the 17th century (mainly labor of the mind vs labor of the body: _autrice_ wasn't included in their encyclopedia but _actrice_ was—a woman couldn't be a writer, but could be an actress). The _académie_ is quite against any change to the language, especially changes made to make other genders included. Interestingly there's been 723 men and 10 women inducted in the academy in its history, which might explain _some_ things. The rule about job names was overturned a couple of years ago. The _point médian_ abbreviation technique has been described as "a mortal peril to the French language", so I guess we'll wait a couple of centuries more to see any rule accepted? I hope not to wait that long.

arkhi:
Please bear with me, I’ll focus on the discussion (so much to say…) more than your actual proposals in this comment. I’ll come back for more tomorrow; I’m cutting Internet at 22:00. :)

> To be honest, I only started the PR because of the very confusing sentence « 6 abonnements, 5 abonnements » for "6 followers, 5 following" on the profile page, and it sort of evolved into these changes :)

That’s precisely why I’m happy you’re submitting this PR. :) I’m sure I missed a lot considering the volume to translate. Thank you also for fixing some typos!

> it's depersonalizing—you follow a person, not an account.

I disagree. One follows an account, not a person. A person is a human whereas an account is the representation of a person in a certain context. As an example, defining an account as a person leads to Facebook using _Friends_ to make it harder to leave their platform. More on that right after. :)

---

I had started to make a list of some terms I intended to use for translating:

- _Reading Goal_ is most often translated by _défi lecture_.
- _Remove_ can be _Retirer_ or _Supprimer_. _Retirer_ means the object does not disappear from the database but does not belong to its parent anymore. _Supprimer_ means the object is being delete from the database (now or scheduled for deletion).
- _User_ is most often translated by _Compte_, closer to _Account_ and with a defined gender. This allows to not deal with human genders while keeping a maybe more accurate meaning. The human terminogy _Utilisateur ou Utilisatrice_, _lecteur or lectrice_ is used when specifically targetting a physical person.

I don’t know if this kind of thing could go into a section of the documentation; @mouse-reeve? Do you have an opinion on that, should we start implementing [Software-localization tools](https://en.wikipedia.org/wiki/Category:Software-localization_tools), or…?

---

For the sake of the discussion because I think it’s good to discuss it, and to express my own sensibilities on the topic:

If we’re talking about getting used to changes in the language, then I’d be really happy to demote the default form from being designated as _male_ to being _neutral_. That would have the advantage of simplifying the language instead of complexifying it. Language is a code and the more complexity you have, the more bugs come along. Now… We’re in 2021 and there is still people claiming _Étudiante_ should be written _Etudiante_, so there is a lot of work ahead either way.

This is definitely a complex issue and when Romy writes “Il ne freine pas la lecture.”, this is very opinionated (and somehow inconsiderate). This also renders her whole argument moot to me because I can totally feel the opposite, not because I’m strictly against and I didn’t try, but simply because I’m a slow reader and breaking words makes reading any text even more straining and non‑accessible (to me). I might also have some non diagnosed reading disorder, I don’t know; I decipher the words I’m reading and usually bug on those words. :)

I don’t like parenthesis for the same reason, but it’s already used and understood. I find it strange that `⋅` is fine when parenthesis are not as both are about singling the feminine form out anyway. I do understand the reasoning behind _mettre entre parenthèses_ and the fact `⋅` is somehow agnostic for now, but it feels more like an _afterthought argument_ instead of a practical one; I’d personally rather write a feminine form to describe a group (and maybe use one or the other interchangeably) than use either punctuations (`(…)` or `⋅…⋅`), but that again would depend on a global consensus on the evolution of our language.

Another issue I’m wondering about is: why write the masculine form first? _utilisateur⋅rice_ or _utilisatrice⋅eur_ ? I don’t see one of them being harder to pronounce than the other, so how do we decide? Should we? There is an endless questionning about how to proceed. :) If the goal is to push forward a group that is considered oppressed by a patriarchal society, I would find it more fair to put the feminine first. “Ladies first”, after all… but that would then be yet another very patriarchal bias.

So back to the first paragraph, I would prefer to call for people to define the simplest form _neutral_ or _multi‑gender_ so that it actually encompasses everyone, including non‑binary people or future unknown gender identifications.

But that just me. :)

---

For our current Pull Request, I would be happy to find terms where we do not not have to use parenthesis or median points while still giving a chance to everyone to feel at home.

I’m gonna have a closer look at the actual change in the Pull Request and hope we can both find something that satisfies everyone. :notes: _Dream o…n!_

mouse-reeve:
thank you both for this conversation (and for the articles, @joachimesque) - I love language politics!

> I had started to make a list of some terms I intended to use for translating:
> 
> * _Reading Goal_ is most often translated by _défi lecture_.
> * _Remove_ can be _Retirer_ or _Supprimer_. _Retirer_ means the object does not disappear from the database but does not belong to its parent anymore. _Supprimer_ means the object is being delete from the database (now or scheduled for deletion).
> * _User_ is most often translated by _Compte_, closer to _Account_ and with a defined gender. This allows to not deal with human genders while keeping a maybe more accurate meaning. The human terminogy _Utilisateur ou Utilisatrice_, _lecteur or lectrice_ is used when specifically targetting a physical person.
> 
> I don’t know if this kind of thing could go into a section of the documentation; @mouse-reeve? Do you have an opinion on that, should we start implementing [Software-localization tools](https://en.wikipedia.org/wiki/Category:Software-localization_tools), or…?

I'm not sure what you mean specifically about implementing software-localizations tools - Gettext is what is currently in use. I definitely support adding things to the documentation, though!

When "remove" is used, it's in the sense of "does not belong to its parent anymore" not in the sense of "deleted from the database". 

The principals I'm using when I think about languages choices are:
1. The wording should be as gender neutral and inclusive of non-binary people as possible,
2. It should be clear and easy to understand,
3. Avoid making unnecessary assumptions on behalf of the user (for example, as you mentioned @arkhi, Facebook using the language of "friends")

I think the distinction between "person" and "account" is a lot less presumptuous than between "friend" and "account", so I could go either way on that if one more clearly communicates what the behavior is. 

So if there's a gender neutral way of saying things that is accessible and isn't so out of standard usage that it would confuse people, I prefer that. Introducing an entire new neutral grammatical gender and inventing declensions for it, for example, would probably be a step too far 😉.

But obviously with French, it's not as easy as me picking "they" over "he/she" in English, or this would be a much simpler but less interesting conversation.

---

> Naturally, the question that can be asked is _is this really necessary to display someone's following_? But let's say it's out of scope for this PR 😄

I'd be happy to have a deeper conversation on this in a separate issue! My short answer is, I find it handy to know how socially connected to someone I am when evaluating how comfortable I feel with strangers.

arkhi:
> I'm not sure what you mean specifically about implementing software-localizations tools - Gettext is what is currently in use. I definitely support adding things to the documentation, though!

I mean an online tool to specifically collaborate on localising a software, like Transifex, for example.

> When "remove" is used, it's in the sense of "does not belong to its parent any more" not in the sense of "deleted from the database".

Yep. Actually _Remove_ should be _Retirer_ and _Delete_ should be _Supprimer_.

> So if there's a gender neutral way of saying things that is accessible and isn't so out of standard usage that it would confuse people, I prefer that. Introducing an entire new neutral grammatical gender and inventing declensions for it, for example, would probably be a step too far wink.

Aaaah… You’re breaking my heart! :) I was going to propose some work to come up with proper neologisms based on greek roots. This is mentioned in @juliemoynat’s [excellent article](https://www.lalutineduweb.fr/ecriture-inclusive-accessibilite-numerique-jetdv2020/) (in French) @joachimesque linked, and I can see the benefit of creating new words or grammar constructions over “demoting” from male to neutral since those words would always keep a male scent for generations anyway.

arkhi:
> Mastodon uses « 6 abonnements, 5 abonné·e·s » for "followers" […]. Pleroma uses « 6 suivis, 5 vous suivent » […].
> 
> A longer way would be to use « 6 comptes suivis, 5 comptes abonnés » but I'll be the first to say it's cumbersome, plus it's depersonalizing—you follow a person, not an account. And using « 6 personnes suivies, 5 personnes abonnées » is more cumbersome.
> 
> An out-of-the-box proposal : « ${name} suit 6 personnes, 5 suivent ${name} ». More verbose than we're used to on social networks, but… well… it's for readers… so we should not be afraid of sentences, should we? sweat_smile

:)

I was thinking we could actually use something  more graphical in nature, with a more verbose label. I like the idea of finding [specific words for each interaction](https://mastodon.cgx.me/@cgx/106073881035075471) too (although maybe not those ones :) ).

```django
<span aria-label="
    {% blocktranslate with name=name %}
        {name} suit 6 personnes, 5 suivent {name}
    {% blocktranslate %}
}">
    6 ≺ {{ username }} ≺ 5
</span>
```

`≺` means _precedes_ (and might be OK, it was for pure mathematical fun). :)

The previous code would also need some tweak for `rtl` writing directions.

> I disagree. One follows an account, not a person.

Interestingly enough, I marked usernames as https://schema.org/Person with microdata. :)

---

For _utilisateurs_ or _utilisatrices_, would _bibliophiles_ work? It’s an a‑gender aproximation of “bookworm” which, I think, expresses the addiction to books enough in the context of BookWyrm. The gender issue would still occur with adjectives and articles, but that’s an option… :)

With the previous example, that would give _${name} suit 6 bibliophiles. 5 suivent ${name}_.

juliemoynat:
Hi,

This code will not work:

```html
<span aria-label="
    {% blocktranslate with name=name %}
        {name} suit 6 personnes, 5 suivent {name}
    {% blocktranslate %}
}">
    6 ≺ {{ username }} ≺ 5
</span>
```

`aria-label` on a non-semantic tag is not working. By the way, if you use this trick for a button or a link (for example), it will not be accessible.
Actually, `aria-label` is totally replacing the visible text label so, for persons using speech recognition applications, they will not be able to tell "click on this link / button with that name" because they will tell the visible name but the visible name is replaced by an invisible one (in `aria-label`). [See WCAG "label in name" criterion](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html)

**About the words in French…**
I do not recommend to invent a new word for "follower" in French because it exists but it has two forms: masculine and feminine. So, if you want to invent something, in my opinion, it should be a neutral ending.
As an example, in French, words with "aire" ending are gender neutral ("gestionnaire", "fonctionnaire"…). We should reproduced this functioning.

Alpheratz has thought about it. You can read about it on [Simonæ (French)](https://simonae.fr/articles/nommer-exister-alpheratz-troisieme-genre) and there are tables on [Alpheratz website (French)](https://www.alpheratz.fr/linguistique/genre-neutre/).
I particularly appreciate the "æ" ending for words ending by "é" / "ée". "Abonné" and "Abonnée" are read the same way in oral form. The form is changing only on writing. The "æ" ending ("abonnæ") is not changing pronunciation but it's removing the gender mark.

In the book "[Tirons la langue, plaidoyer contre le sexisme dans la langue française](https://www.librairiesindependantes.com/product/9782919160228/)", Davy Borde gives a table at the end of the book with some propositions such as :
- using the "ë" to end words with only a variation on writing: exploitéë, ravië, inconnuë
- "nouveaulle" for "nouveau" and "nouvelle"
- "agriculteurice" for "agriculteur" and "agricultrice"
- "actifive" for "actif" and "active"
- "chômeurze" for "chômeur" and "chômeuse"
- "copaine" for "copain" and "copine"
- "conquérande" for "conquérant" and "conquérante"
- "patiende" for "patient" and "patiente"
- etc.

I don't like all the propositions (I like those with "ande" and "ende" endings) but this is a beginning to think about gender neutral words. 

I hope this can help your thinking.

arkhi:
Thanks a lot for your feedback, @juliemoynat.

> `aria-label` on a non-semantic tag is not working. By the way, if you use this trick for a button or a link (for example), it will not be accessible.

I didn’t know about that and could not find anything on the topic but the [specs that say](https://www.w3.org/TR/wai-aria/#aria-label) “Used in Roles: All elements of the base markup”. I understand “base markup” as being any tag but you most likely have resources or experience on the topic. If you can share any of them, I’d be happy to learn more on the topic.

> […] they will not be able to tell "click on this link / button with that name" because they will tell the visible name but the visible name is replaced by […] `aria-label`.

Good Lord; thank you! “With great power, etc.”

arkhi:
> As an example, in French, words with "aire" ending are gender neutral ("gestionnaire", "fonctionnaire"…). We should reproduced this functioning.

We could reproduce this functioning indeed. I like _abonnaire_, if only because it would be close to _débonnaire_. :)

Any attempt to create new orthographs would still create a neologism though.

> I particularly appreciate the "æ" ending for words ending by "é" / "ée".

I do love this _ending_ particle as well, plus it reminds me of the spanish `@` used for _tod@s_ to address whatever gender.

> Davy Borde gives a table at the end of the book with some propositions

I share your dislike and that’s why I strongly prefer what you’re mentioning in your article, about using or creating new or unused words (like _adelphe_ for “siblings”) instead of creating “Frankenstein monsters”.

I would love the use of `æ` which seems to be the less invasive change to me and somehow in line with changes in other languages. @joachimesque and @mouse-reeve: Do you have preferences or dislikes? Shall we experiment on this or other proposals?

juliemoynat:
I'm sorry, I've just tested the "æ" ending and it's unfortunately not read by the NVDA screen reader. So… it's not usable for now. So bad.

---

`aria-label` is used to give an accessible name to elements. A non-semantic HTML element doesn't have an accessible name because it doesn't have a role. The specification tells it can be used even if there is no role but… the compatibility with screen readers is not really there (it doesn't work either on `p` elements whereas its role is "paragraph"). Each time you use ARIA, you will need to test your code with the most used screen readers (NVDA, JAWS, VoiceOver (desktop and mobile), Talkback) to be sure it's actually working.

`aria-label` works better on interactive elements but it's a dangerous attribute because it replaces the visible label.

I think you would like to also know that [`aria-label` does not translate](https://adrianroselli.com/2019/11/aria-label-does-not-translate.html) (for automatic browser translation).

arkhi:
>  Alpheratz website (French)

I like the (x|z) forms and how some of the forms are actually going back to sounds close to [patois](https://en.wikipedia.org/wiki/Patois). Thanks a lot for the educational links; I’ve [added the books](https://bookwyrm.social/list/29), since… well… That’s what BookWyrm is about!

---

> I'm sorry, I've just tested the "æ" ending and it's unfortunately not read by the NVDA screen reader. So… it's not usable for now. So bad.

/me sad.

Just for the sake of playing with my previous example with agnostic tags, not implying the translation is even desirable, should we resolve to things like a full version and a summarised version in some cases?

```html
<span>
    <span class="is-sr-only">joachimesque suit 6 personnes, 5 suivent joachimesque</span>
    <span>6 ≺ joachimesque ≺ 5</span>
</span>
```

---

Thanks for the detailed feedbacks about `aria-label`! Orange a11y is always a sturdy resource. :)

locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -2594,14 +2594,14 @@ msgstr "Livres"
 #: bookwyrm/templates/user/user_preview.html:13
 #, python-format
 msgid "Joined %(date)s"
-msgstr "Enregistré(e) %(date)s"
+msgstr "Membre depuis %(date)s"
```
arkhi, line 2597: I’m usually browsing in English and just switched to French for testing…

Just raising the question that `%(date)s` will be replaced by `il y a [n] (jour(s)|mois|année(s)]`, so _Membre depuis_ is going to be a bit weird (to me) although correct (I think). I don’t know what kind of tweaking is possible for the formatting of `date` in that case but maybe @mouse-reeve knows.

arkhi (commented):
Here you go! I don’t think there are any real issues beside the first comment. Anything else is, I think, quite subjective.

joachimesque:
> ```
> <span aria-label="
>     {% blocktranslate with name=name %}
>         {name} suit 6 personnes, 5 suivent {name}
>     {% blocktranslate %}
> }">
>     6 ≺ {{ username }} ≺ 5
> </span>
> ```
> 
> `≺` means _precedes_ (and might be OK, it was for pure mathematical fun). :)
> 
> The previous code would also need some tweak for `rtl` writing directions.

That's an interesting solution, but quite circonvoluted IMO 😄 

> For _utilisateurs_ or _utilisatrices_, would _bibliophiles_ work? It’s an a‑gender aproximation of “bookworm” which, I think, expresses the addiction to books enough in the context of BookWyrm. The gender issue would still occur with adjectives and articles, but that’s an option… :)
> 
> With the previous example, that would give _${name} suit 6 bibliophiles. 5 suivent ${name}_.

Ooh yes I like that a lot!

----

> 
>     * using the "ë" to end words with only a variation on writing: exploitéë, ravië, inconnuë
> 
>     * "nouveaulle" for "nouveau" and "nouvelle"
> 
>     * "agriculteurice" for "agriculteur" and "agricultrice"
> 
>     * "actifive" for "actif" and "active"
> 
>     * "chômeurze" for "chômeur" and "chômeuse"
> 
>     * "copaine" for "copain" and "copine"
> 
>     * "conquérande" for "conquérant" and "conquérante"
> 
>     * "patiende" for "patient" and "patiente"
> 
>     * etc.
> 
> 
> I don't like all the propositions (I like those with "ande" and "ende" endings) but this is a beginning to think about gender neutral words.
> 
> I hope this can help your thinking.

Someone on Twitter linked [this article](https://naath.medium.com/de-lusage-du-e-inclusif-comme-accord-de-genre-neutre-59f8e08d2c99) about the letter `ė`, which would give _« 5 suiveurės, 6 suiviės », « 5 abonnéės, 6 abonnements »_. I'm not sure I'm a big fan, but it's the same kind of thought process as with the list above

----

I'll get on to the rest of the conversation when time allows it 😄

locale/fr_FR/LC_MESSAGES/django.po
joachimesque: > Just raising the question that `%(date)s` will be replaced by `il y a [n] (jour(s)|mois|année(s)]`

Ah yes, in my mind I thought I had removed `il y a`, so _« Membre depuis 2 jours »_ would have worked. I don't know if we'll be able to use a different translation for the `|naturaltime` filter

arkhi: Yep… I thought the same when translating. :)

I do like _Membre_ better though, I wonder if @mouse-reeve would have anything against updating the logic and use something less verbose like `Member since: {2021-04-17}` or `Member for {3 months}`, for example.

I would guess the logic is to easily read for how long the account has been a member of the instance; some sort of trust factor that is easier to read than computing a date difference. Then again… Just an assumption.

joachimesque: In French we'd like to use `date|timesince` instead of `date|naturaltime`. Do you know if it's possible to give the choice of formats to the translator?

mouse-reeve: I don't think it's possible to change it just for one locale, but I have no objection whatsoever to using `timesince` across the board here

juliemoynat:
> Just for the sake of playing with my previous example with agnostic tags, not implying the translation is even desirable, should we resolve to things like a full version and a summarised version in some cases?
> 
> ```
> <span>
>     <span class="is-sr-only">joachimesque suit 6 personnes, 5 suivent joachimesque</span>
>     <span>6 ≺ joachimesque ≺ 5</span>
> </span>
> ```

@arkhi, yes, it's a better version. In case of a text like this (not inside a link or button), you could even add an `aria-hidden="true"` attribute on the second `span` inside because the sr-only text ([if well used](https://kittygiraudel.com/2016/10/13/css-hide-and-seek/)) is the alternative text for that piece of text. I don't like to hide visible text for assistive technologies like this but the "6 ≺ joachimesque ≺ 5" will not be understood.

And finally… I should tell you that I think "6 ≺ joachimesque ≺ 5" is not understandable for anyone. This is not a screen reader problem to solve here :-)

---

@joachimesque About the ė character, after testing, it is not read by NVDA either and could cause some problems to understand sentences. I don't like the fact that examples in the article are using the masculine form of words to create the neutral form. 

Example: "l’hôtė d’accueil est unė prêtrė." This sentence will be read "l'hôte d'accueil est un prêtre". Is it really neutral?

This is really really complicated… but interesting.

locale/fr_FR/LC_MESSAGES/django.po
joachimesque: I updated the file and the French translation. Will that cause a problem for other translators?

mouse-reeve: hm, I realized I actually misread the initial problem here with `naturaltime` - sorry, I've been kinda scatterbrained the for the last week. It would cause problems for other locales to switch to `timesince`, but would `naturalday` work instead? There's no need as far as I can see to be precise about how many hours ago someone joined, I think `naturalday` doesn't add the "ago"/"il y a", and seems to be compatible with the existing translations?

joachimesque: Here's the current situation

image

`naturaltime` is what's used right now, and my problem was that it adds "ago"/"il y a", so my proposal to change _« Enregistré(e) il y a 3 jours »_ (= "Joined 3 days ago") to _« Membre depuis 3 jours »_ (= "Member for 3 days") won't work, it'll give us _« Membre depuis il y a trois jours »_ (= "Member for 3 days ago").

Do you think that we can specify how the `naturaltime` filter works for French? I don’t think that filters can be applied inside translation files.

joachimesque: (I proposed to change the current translation because using a verb in French requires to adapt it to the gender of the user, or use a possibly inaccessible solution with an _(e)_ ending. "Member for 3 days" is the best neutral way to convey the meaning)

mouse-reeve: right, I think `naturalday`, as opposed to `naturaltime`, will _not_ add `ago"/"il y a"`. 

With `naturaltime`: "Joined today"
With `naturalday`: "Joined an hour ago"

With `naturaltime`: "Joined 2 weeks, 1 day ago"
With `naturalday`: "Joined April 6, 2021"

I think that solves the issues with "ago" in French without changing other locales (it at least doesn't require any wording changes in English). I hope I'm understanding this correctly?

joachimesque: I hadn't read `naturalday`, I thought it was `naturaltime`, sorry about that 🤦

joachimesque: After testing a bit and reading the source code for `naturalday`, it won't work well with French. Before dates you need the article _« le »_, but before relative times (_« hier »_, "yesterday") you don't need it. `naturalday` alternates between the two, without adding the article, so it won't work.

What could work well in French would be `naturaldelta` from Python Humanize:

> This is similar to `naturaltime`, but does not add tense to the result.

https://python-humanize.readthedocs.io/en/latest/time/#humanize.time.naturaldelta

Sadly it's unavailable, and I think it would cause the same problems as using directly `timesince`.

Back to square one, I guess :)

mouse-reeve: dang! too bad, thank you for looking into it.

arkhi:
@joachimesque:

> > `≺` means _precedes_ (and might be OK, it was for pure mathematical fun). :)
> > The previous code would also need some tweak for `rtl` writing directions.
> 
> That's an interesting solution, but quite circonvoluted IMO smile

> [from @juliemoynat] And finally… I should tell you that I think "6 ≺ joachimesque ≺ 5" is not understandable for anyone. This is not a screen reader problem to solve here :-)

On that we agree. Throwing ideas, as stupid or complicated they might be can inspire others, so… :)

> > For _utilisateurs_ or _utilisatrices_, would _bibliophiles_ work? […]
> > With the previous example, that would give _${name} suit 6 bibliophiles. 5 suivent ${name}_.
> 
> Ooh yes I like that a lot!

Noted. :)

The added benefit is that it’s also an english word.

> Someone on Twitter linked [this article](https://naath.medium.com/de-lusage-du-e-inclusif-comme-accord-de-genre-neutre-59f8e08d2c99) about the letter `ė`, which would give _« 5 suiveurės, 6 suiviės », « 5 abonnéės, 6 abonnements »_. I'm not sure I'm a big fan, but it's the same kind of thought process as with the list above

I still think it would be more retro-compatible and easier for everyone, those for or against inclusive speech, to use the simplest form for neutral and add diacritics (modifying the pronounciation or not) when there is a will or need to specify a gender. That could seem unfair from a historic perspective but far more pragmatic for everyone speaking, writing, reading or learning French, emphasising on the neutral form instead of trying to adapt any gender based on what is now defined as _masculine_.

A silly example (with whatever symbols):

- un   étudian   heureu   (_a happy student_, a-gendered)
- un~  étudian~  heureu   (_a happy female student_)
- un⋅  étudian⋅  heureu⋅  (_a happy male student_)
- des  étudians  heureus  (_happy students_, a-gendered)
- des~ étudians~ heureus~ (_happy female students_)
- des⋅ étudians⋅ heureus⋅ (_happy male students_)

There would be of course the question about how to identify gender when spoken, as other proposals.

---

@juliemoynat:

> > ```html
> > <span>
> >     <span class="is-sr-only">joachimesque suit 6 personnes, 5 suivent joachimesque</span>
> >     <span>6 ≺ joachimesque ≺ 5</span>
> > </span>
> > ```
>
> @arkhi, yes, it's a better version. In case of a text like this (not inside a link or button), you could even add an aria-hidden="true" attribute on the second span inside because the sr-only text (if well used) is the alternative text for that piece of text. I don't like to hide visible text for assistive technologies like this […]

My intention was to add the `aria-hidden="true"` indeed; it got lost in the process; thanks! :)

Agreed on the fact that using `aria-hidden` for code like this is more likely to be the symptom of a problem than a real solution. :+1:

> I don't like the fact that examples in the article are using the masculine form of words to create the neutral form.
> 
> Example: "l’hôtė d’accueil est unė prêtrė." This sentence will be read "l'hôte d'accueil est un prêtre". Is it really neutral?

It’s not with the current grammatical rules in French, definitely. I’ll loop on my silly example above which strips letters intead of adding some, redefining a gendered word into something else to avoid the natural bias we have, when thinking about that old rule that states “masculine form prevails”, when we could consider masculine is taking the neutral form in French (and some other languages?).

This is marketing, I agree, but any solution will require a change of perspective. :/

locale/fr_FR/LC_MESSAGES/django.po
arkhi: > […] It would cause problems for other locales to switch to timesince, […]

@mouse-reeve: Do you have specific problems in mind?

I’m having in mind a less verbose form like `Member for: {2021-04-22}`, for example. There would be less possible variations and the string would focus more on translating the text than the generated form of the date. That way, I understand `naturalday` would work as intended.

Does it seem too easy of a solution that I missed something or is someone against lists? :)

arkhi: Ping… :)

arkhi (changes requested):
@joachimesque: Thanks again for the PR. It seems something somehow went wrong with some strings since the PR was opened. Sorry this PR has been sitting here for that long.

The discussion about _Membre depuis_ has never found a closure (here, at least) so I’ll let you check it out as well.

locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1875,6 +1950,27 @@ msgstr "Nombre de fois utilisée"
...
+#: bookwyrm/templates/settings/server_blocklist.html:26
+#: bookwyrm/templates/snippets/goal_progress.html:5
+msgid "Success!"
+msgstr "Bravo !"
+
+#: bookwyrm/templates/settings/server_blocklist.html:30
+#, fuzzy
+#| msgid "Successfully imported"
+msgid "Successfully blocked:"
+msgstr "Importation réussie"
```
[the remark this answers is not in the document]
mouse-reeve: Yes, this has been bugging me about how I've worded things. For English, I think "instance" is preferable, since that's what mastodon uses and it doesn't clash with the technical definition of a server

locale/fr_FR/LC_MESSAGES/django.po
```diff
@@ -1654,115 +1660,184 @@ msgstr "Aucune liste trouvée pour « %(query)s »"
...
+#: bookwyrm/templates/settings/edit_server.html:6
+#: bookwyrm/templates/settings/edit_server.html:20
+#: bookwyrm/templates/settings/federation.html:9
+#: bookwyrm/templates/settings/federation.html:10
+#: bookwyrm/templates/settings/server_blocklist.html:3
+#: bookwyrm/templates/settings/server_blocklist.html:20
+#, fuzzy
+#| msgid "Add cover"
+msgid "Add server"
+msgstr "Ajouter une couverture"
```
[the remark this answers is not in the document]
mouse-reeve: Yes, those lines have to be removed for the translation to be active

joachimesque:
> @joachimesque: Thanks again for the PR. It seems something somehow went wrong with some strings since the PR was opened. Sorry this PR has been sitting here for that long.

I've been taken by a book this last week so I didn't have the time to fix this PR. I'll get on it as soon as I can :)

PR #936, opened 2021-04-14
Restrict list creation form to authenticated users

joachimesque:
When visiting the `/list` route on an instance while unauthenticated, the button to toggle the List creation form is shown. This PR aims to hide the button to unauthenticated visitors.


mouse-reeve:
woops! thank you

PR #940, opened 2021-04-15
Set body minimum height

joachimesque:
This PR sets a min-height for the `body` element, allowing for the footer to always be at the bottom of the window.

## before

Screenshot_2021-04-15 BookWyrm

## after

Screenshot_2021-04-15 BookWyrm(1)

### DIFF
--- a/bookwyrm/static/css/bookwyrm.css
+++ b/bookwyrm/static/css/bookwyrm.css
@@ -3,6 +3,12 @@ html {
     scroll-padding-top: 20%;
 }
 
+body {
+    min-height: 100vh;
+    display: flex;
+    flex-direction: column;
+}
+
 .image {
     overflow: hidden;
 }
--- a/bookwyrm/templates/layout.html
+++ b/bookwyrm/templates/layout.html
@@ -22,165 +22,169 @@
     <meta name="twitter:image:alt" content="BookWyrm Logo">
 </head>
 <body>
-<nav class="navbar container" aria-label="main navigation">
-    <div class="navbar-brand">
-        <a class="navbar-item" href="/">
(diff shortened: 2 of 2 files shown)


mouse-reeve:
beautiful! thank you!

bookwyrm/static/css/bookwyrm.css
```diff
@@ -3,6 +3,12 @@ html {
     scroll-padding-top: 20%;
 }
 
+body {
+    min-height: 100vh;
+    display: flex;
+    flex-direction: column;
```
mouse-reeve, line 9: what's function are the flex attributes serving here?

joachimesque: If I only define a `min-height` at `100vh` ( = 100% of the height), the Body will have the size of the window, but the children will be stacked at the top.

What I want is to have the Navbar to be at the top and the Footer at the bottom, both only taking what space they need but no more. Then, the main Section should be able to take whatever space is between the Navbar and the Footer, and more if needed.

So I set the Body as a flexbox, and I set its direction as a `column`. Now, nothing has changed visually: the three children are still stacked at the start of the column. What I want is for the Section to be able to grow as needed.

By default, flexbox children can not grow. So I added  the Bulma helper class `is-flex-grow-1` to the Section, telling it that whatever space it needs, it can take. 

Here's the **before**, with Firefox FlexBox highlighting:

Screenshot_2021-04-15 BookWyrm(2)

The unused space is the hatched zone.

Here's the **after**:

Screenshot_2021-04-15 BookWyrm(3)

Now the content section takes up all the available space

Here's the **after**, with a lot of content:

Screenshot_2021-04-15 BookWyrm(4)

The content section takes up all the space it needs, and the Navbar and the Footer always have the same height.

mouse-reeve: got it, thank you for the detailed explanation!

PR #946, opened 2021-04-16
Gracefully handle expect but unsupported activities

mouse-reeve:
This will prevent errors being thrown when `Question` activities are received (we don't support polls yet), but still throw errors when other unknown types show up, as they may be things we _should_ handle.

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -52,10 +52,14 @@ def naive_parse(activity_objects, activity_json, serializer=None):
         if activity_json.get("publicKeyPem"):
             # ugh
             activity_json["type"] = "PublicKey"
+
+        activity_type = activity_json.get("type")
         try:
-            activity_type = activity_json["type"]
             serializer = activity_objects[activity_type]
         except KeyError as e:
+            # we know this exists and that we can't handle it
+            if activity_type in ["Question"]:
+                return None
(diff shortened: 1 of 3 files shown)


bookwyrm/activitypub/base_activity.py
```diff
@@ -52,10 +52,14 @@ def naive_parse(activity_objects, activity_json, serializer=None):
...
             # ugh
             activity_json["type"] = "PublicKey"
+
+        activity_type = activity_json.get("type")
         try:
-            activity_type = activity_json["type"]
             serializer = activity_objects[activity_type]
         except KeyError as e:
+            # we know this exists and that we can't handle it
+            if activity_type in ["Question"]:
```
arkhi, line 61: Just mentioning: Could/Should/… this be in an environment variable, so that the list of unsupported features is “documented”?

mouse-reeve: While it _is_ weird to hard code it like this, I thinkputting it in as an environment variable would suggest that changing it there would have a meaningful effect on the how the app functions. If someone running the site changed the environment variable, it would just cause a bunch of unexpected errors.

mouse-reeve:
merging this because it's activity causing errors, not to close the conversation -- I'd be happy to approach this differently

arkhi:
I was just raising the question thinking things can always be overlooked; no worries on my side. :)

PR #860, opened 2021-04-03
Using Poetry for dependency management and adjustments for docker

SavinaRoja:
https://github.com/mouse-reeve/bookwyrm/issues/793

Brought the branch up to date. Give this a good review. I think I did not fully update `bw-dev`. It looks like it is used basically like a Makefile?

PR #947, opened 2021-04-17
Fix editorconfig greediness 

arkhi:
- Ignore computer generated files when linting.
- Include .mo files for consistency.
- Use multiline yaml syntax for `run` commands.

This PR should avoid [linting issues in some current or future PR](https://github.com/bookwyrm-social/bookwyrm/pull/860/checks?check_run_id=2260841186#step:4:13). It should also close https://github.com/SavinaRoja/bookwyrm/pull/1.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -20,7 +20,7 @@ block_comment_end = */
 [{bw-dev,fr-dev,LICENSE}]
 max_line_length = off
 
-[*.{csv,json,html,md,mo,po,py,svg,tsv}]
+[*.{csv,json,html,md,po,py,svg,tsv}]
 max_line_length = off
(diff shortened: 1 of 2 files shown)


mouse-reeve:
What's the motivation for adding `.mo` files, rather than including them in the list of generated files that will be ignored?

arkhi:
> What's the motivation for adding `.mo` files, rather than including them in the list of generated files that will be ignored?

The idea was that .mo files were sort of special because the other files, although computer generated, are not compiled. But after you asking, it probably won’t hurt to apply the `insert_final_newline` rule to the former group, yes. I’ll push a commit reflecting this.

PR #959, opened 2021-04-18
Display book identifiers in the list of Editions

arkhi:
When searching for the edition of a book, it can be really useful to search it by ISBN or others.

I started refactoring the editions list but gave up because I was caught between the time it would spend to do it without Bulma and the fact I couldn’t see how to do what I wanted with Bulma. Changes are then minimal, which is not necessarily a bad thing. :)

- Add template for identifiers.
- Remove `space-between` that adds too much space on narrow views.
- Apply the `content` class only on publisher infos to avoid applying automatic styles to the book identifiers description list.

### DIFF
--- a/bookwyrm/templates/book/book.html
+++ b/bookwyrm/templates/book/book.html
@@ -67,31 +67,16 @@
             </div>
             {% endif %}
 
-            <section class="content is-clipped">
-                <dl>
-                    {% if book.isbn_13 %}
-                    <div class="is-flex is-justify-content-space-between is-align-items-center">
-                        <dt>{% trans "ISBN:" %}</dt>
-                        <dd itemprop="isbn">{{ book.isbn_13 }}</dd>
+            <section class="is-clipped">
+                {% with book=book %}
+                    <div class="content">
+                        {% include 'book/publisher_info.html' %}
                     </div>
-                    {% endif %}
 
-                    {% if book.oclc_number %}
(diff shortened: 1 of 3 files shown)


mouse-reeve:
The only thing I'm not sure about here is the `content` class only applying to one of the columns - the different line spacing between two adjacent blocks of text is tripping me up

arkhi:
And if you consider four blocks of different information laid out in two columns? :}

I don’t think the `content` class is appropriate here as the markup is not user generated. It’s better employed with elements containing parsed markdown that can’t be styled.

That being said, having margins between blocks helps when the content is laid on multiple lines, especially when the blocks are not related.

I can push changes like individual classes or we can improve later; I don’t mind. Your call. :)

mouse-reeve:
My preference is to remove `content` on line 31. Unless I'm misunderstanding you, neither the publisher info nor the identifiers are user generated markup?

arkhi:
> My preference is to remove `content` on line 31. 

Done!

> neither the publisher info nor the identifiers are user generated markup?

They are user generated data that is not parsed from markdown to html, that’s correct. Just that `content` is used in book pages on _publisher info + identifiers_ and caused issues with extra spacing on `<dd>`.

Ultimately, I think it would be better to only use `content` on markup that is generated by markdown, like quotes or descriptions. This would avoid unwanted side effects.

PR #960, opened 2021-04-18
Use object ids with base path of current instance for follow accepts or rejects

DeeUnderscore:
This changes the object ids on `Accept` and `Reject` activities that BookWyrm emits to have URIs belonging to the instance sending the activity. The current behavior uses the domain of the instance that the follow request was sent from, so if we're accepting a follow request from example.com, the id ends up being something like `https://example.com/users/SomeUser#accepts/10`, even though *we* are not example.com. 

Using object `id` URIs that the originating server controls seems to be what [the spec says to do](https://www.w3.org/TR/activitypub/#obj-id). Current behavior does not seem to create problems with Mastodon or BookWyrm, but it does cause Accepts to get ignored by Pleroma.

### DIFF
--- a/bookwyrm/models/relationship.py
+++ b/bookwyrm/models/relationship.py
@@ -50,11 +50,10 @@ class UserRelationship(BookWyrmModel):
             ),
         ]
 
-    def get_remote_id(self, status=None):  # pylint: disable=arguments-differ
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you!

PR #969, opened 2021-04-20
More compact display of statuses

### LINKED ISSUE #964: Design: Thinking of the density and hierarchy of feed elements
I was browsing profiles, to discover books and other like-minded readersn and that experience gave me a few remarks:

_("should" is to be taken as "the author of this issue would like if/would find a use in")_

- Feed activities should be filtered by types. I may want to see only comments or critiques.
- Feed activities should be displayed differently depending on the type.      
   Some types ("started reading a book", "finished reading a book") don't need to take much space, while other types ("wrote a comment", "wrote a critique") should be more prevalent (compared to other types), and star ratings can be easily confused with other types due to the lack of text
- Books without a description take up a whole lotta space for nothing

**Here's some rules I set:**

- We should understand the type of update at first sight
- Book previews are most useful where some context is needed: in response to a critique or a comment
- Comments and critiques should not be cut (unless they run _really_ long)
- All the actions should be grouped together

**Here's what I have in mind:**

Desktop HD

Of course, it's just a very basic proposal, and can be amended. Looking at it, I'm thinking that a "Write a comment" link could be inserted at the end of the critique/comment text, so the "reply" flow is more visible.

If time allows I'll open a PR to better show what I have in mind.

mouse-reeve:
Works on #964


tofuwabohu:
This looks really nice!

PR #971, opened 2021-04-21
Move some status footer elements to the header

### LINKED ISSUE #964: Design: Thinking of the density and hierarchy of feed elements

joachimesque:
Closes #964.

An implementation for some ideas in #964

image


bookwyrm/templates/snippets/fav_button.html
```diff
@@ -3,18 +3,17 @@
 {% with status.id|uuid as uuid %}
 <form name="favorite" action="/favorite/{{ status.id }}" method="POST" class="interaction fav-{{ status.id }}-{{ uuid }} {% if request.user|liked:status %}is-hidden{% endif %}" data-id="fav-{{ status.id }}-{{ uuid }}">
     {% csrf_token %}
-    <button class="button is-small" type="submit">
+    <button class="button is-small is-white" type="submit">
         <span class="icon icon-heart" title="{% trans 'Like status' %}">
-            <span class="is-sr-only">{% trans "Like status" %}</span>
         </span>
+        <span>{% trans "Like status" %}</span>
```
mouse-reeve, line 9: Seeing this displayed, it feels more consistent to just have this read "Like" rather than "Like status", and ditto for "Boost". I think it's sufficiently clear from context that it refers to the status?

mouse-reeve (changes requested):
I super appreciate all your thought and work on the UI! Thank you!

bookwyrm/templates/snippets/status/status_header.html
```diff
@@ -1,72 +1,90 @@
...
-
-    {% endwith %}
-{% endif %}
-
-{% if status.book %}
-    {% if status.status_type == 'GeneratedNote' or status.status_type == 'Rating' %}
-        <a href="/book/{{ status.book.id }}">{{ status.book.title }}</a>{% if status.status_type == 'Rating' %}:
+    <div class="media-content">
+        <p class="is-size-7">
+            <a href="{{ status.remote_id }}">{{ status.published_date | post_date }}</a>
```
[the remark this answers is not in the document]
joachimesque: Ah yes, my PR was based on an older branch, and because I'm on a forked repo I'll have to find out how to update from `bookwyrm-social:main`

joachimesque:
Desktop:

image

Mobile:

image

bookwyrm/templates/snippets/status/status_header.html
```diff
@@ -1,72 +1,94 @@
...
 
-        <span itemprop="name">{{ status.user.display_name }}</span>
-    </a>
-</span>
+<div class="media">
+    <figure class="media-left">
+        <a class="image is-48x48" href="{{ status.user.local_path }}">
+            {% include 'snippets/avatar.html' with user=status.user ariaHide="true" medium="true" %}
+        </a>    
+    </figure>
```
arkhi, line 10: Moving the avatar here means moving it out of of the _Person_ scope, which might not be ideal. We’re also duplicating links.

bookwyrm/static/css/bookwyrm.css
```diff
@@ -30,6 +30,30 @@ body {
     min-width: 75% !important;
 }
 
+/** Utils
```
arkhi, line 33: Is there a more specific word than _Utils_ in this case? :)

Maybe a reason why classes are named this way would be helpful for others, for example.

joachimesque: What do you think of `Utilities not covered by Bulma`?

bookwyrm/templates/snippets/status/status_header.html
joachimesque: What could be done to affect the image to the _Person_ scope?

bookwyrm/static/css/bookwyrm.css
arkhi: That would give more hint already about why classes are named this way, yep. :)

bookwyrm/templates/snippets/status/status_header.html
arkhi: (This is a summary of our chat on [matrix])

In the not so distant past, I couldn’t manage to make the microdata `itemref` to work consistently.

In our case, I would probably end up using this kind of solution:

```html
{% with user=status.user %}
    <div class="media">
        <figure class="media-left" aria-hidden="true">
            <a class="image is-48x48" href="{{ user.local_path }}">
                {% include 'snippets/avatar.html' with medium="true" %}
            </a>
        </figure>

        <div class="media-content">
            <h3 class="has-text-weight-bold">
                <span
                    itemprop="author"
                    itemscope
                    itemtype="https://schema.org/Person"
                >
                    {% if user.avatar %}
                        <meta itemprop="image" content="/images/{{ user.avatar }}">
                    {% endif %}

                    <a
                        href="{{ user.local_path }}"
                        itemprop="url"
                    >
                        <span itemprop="name">{{ user.display_name }}</span>
                    </a>
                </span>

                […]
            </h3>
        </div>
    </div>
{% endwith %}
```
Changes in this code are:

- abstracting `status.user` with a `{% with user=status.user %}` to reduce code a bit and declare an intention for the block of code.
- moving the `aria-hidden` from the `img` to the `figure` to avoid redundancy of contiguous links with different labels leading to the same URL.
- adding the `<meta>` to keep the image property within its scope. This also means we can serve a bigger image for the avatar to clients using microdata, and a smaller one for people with visual capabilities.

I’ll let you pick what you want from my proposal, is any. :)

joachimesque: Changes will be coming later today :)

bookwyrm/templates/snippets/status/status_header.html
```diff
@@ -1,72 +1,107 @@
...
-    {% blocktrans with username=parent_status.user.display_name user_path=parent_status.user.local_path status_path=parent_status.local_path %}replied to <a href="{{ user_path }}">{{ username}}'s</a> <a href="{{ status_path }}">review</a>{% endblocktrans %}
-    {% elif parent_status.status_type == 'Comment' %}
-    {% blocktrans with username=parent_status.user.display_name user_path=parent_status.user.local_path status_path=parent_status.local_path %}replied to <a href="{{ user_path }}">{{ username}}'s</a> <a href="{{ status_path }}">comment</a>{% endblocktrans %}
-    {% elif parent_status.status_type == 'Quotation' %}
-    {% blocktrans with username=parent_status.user.display_name user_path=parent_status.user.local_path status_path=parent_status.local_path %}replied to <a href="{{ user_path }}">{{ username}}'s</a> <a href="{{ status_path }}">quote</a>{% endblocktrans %}
-    {% else %}
-    {% blocktrans with username=parent_status.user.display_name user_path=parent_status.user.local_path status_path=parent_status.local_path %}replied to <a href="{{ user_path }}">{{ username}}'s</a> <a href="{{ status_path }}">status</a>{% endblocktrans %}
```
mouse-reeve, line 38: It looks like this getting cut is why the tests aren't passing and replies don't render. If you add 
```html
               {% blocktrans with username=parent_status.user.display_name user_path=parent_status.user.local_path status_path=parent_status.local_path %}replied to <a href="{{ user_path }}">{{ username}}'s</a> <a href="{{ status_path }}">status</a>{% endblocktrans %}
               {% endwith %}
           {% endif %}
```

on line 43

and remove
```html
                        {% endwith %}
                    {% endif %}
```
on line 55 and 56, you'll be set.

You should be able to replicate the problem by loading a reply, or running the tests (`./bw-dev pytest`)

bookwyrm/templates/snippets/fav_button.html
```diff
@@ -3,18 +3,17 @@
 {% with status.id|uuid as uuid %}
 <form name="favorite" action="/favorite/{{ status.id }}" method="POST" class="interaction fav-{{ status.id }}-{{ uuid }} {% if request.user|liked:status %}is-hidden{% endif %}" data-id="fav-{{ status.id }}-{{ uuid }}">
     {% csrf_token %}
-    <button class="button is-small" type="submit">
-        <span class="icon icon-heart" title="{% trans 'Like status' %}">
-            <span class="is-sr-only">{% trans "Like status" %}</span>
+    <button class="button is-small is-light is-transparent" type="submit">
+        <span class="icon icon-heart m-0-mobile" title="{% trans 'Like' %}">
```
mouse-reeve, line 7: These are still showing the text on mobile for me:
Screen Shot 2021-04-24 at 11 25 28 AM

mouse-reeve (changes requested):
Just a couple small tweaks (I wonder if the text showing again on mobile is a merge regression?) and we're ready to go. I'm so excited to have this live, it looks great

bookwyrm/templates/snippets/status/status_header.html
joachimesque: Oooh yes thanks! I handled the merge from main and the conflicts resolution in GitHub and was now trying to troubleshoot it locally…

bookwyrm/templates/snippets/fav_button.html
joachimesque: Strange, that's how it's displayed for me (tested on the latest Chromium, Firefox and Safari on macOS)

Capture d’écran 2021-04-25 à 20 37 52

mouse-reeve: Oh that's my bad! I didn't run `collectstatic`. It looks perfect, nevermind :)

PR #972, opened 2021-04-21
Design and a11y fixes on Create Status forms

joachimesque:
- Use `field` CSS class to better respect Bulma styles
- Fix buggy `id` on content field, causing a11y problems
- Fix progress control & select styles
- Various small style fixes


mouse-reeve:
Beautiful! Thank you!

Issue #949, opened 2021-04-17
Menu to move the last Book in a list is cropped


arkhi:
### Describe the bug

On the list of Books for a user, the menu to move the last book of the list is cropped.

### Steps to reproduce the behavior

1. Log in BookWyrm.
1. Go to your [Your books](https://bookwyrm.social/user/arkhi/books). (mine for the example)
2. Scroll (or press <kbd>End</kbd>) to reach the last Book listed on the page.
3. Click on the menu “Move book”.
4. The menu is cropped at the limit of the table, as illustrated by the attached picture.

### Expected behavior

The last menu is not cropped.

### Hint

The `.scroll-x` class on an ancestor of those menus is not allowing elements to overflow. It is defined in [bookwyrm.css:24](https://github.com/bookwyrm-social/bookwyrm/blob/797c3a12c87fcfe7167074cbab789711653d0d08/bookwyrm/static/css/bookwyrm.css#L24-L27).

### Resources

![The last “Move Book” menu is cropped](https://user-images.githubusercontent.com/277773/115113072-b373a000-9f88-11eb-9e95-867ea602fbd4.png)

arkhi:
@mouse-reeve: Do you know why the `.scroll-x` class was added to an ancestor?

mouse-reeve:
It's so that if the table is bigger than the screen (which is always the case on mobile), it will add a horizontal scroll bar fjust or the table, rather than making the entire app super wide.

arkhi:
OK, thanks. Do you mind a proposal replacing the table with something more responsive?

mouse-reeve:
Not at all! That page has needed some love for a long time.

joachimesque:
Here's a simple proposal: #974

arkhi:
This has been fixed by #974.

PR #974, opened 2021-04-21
Add mobile styles for book preview table

joachimesque:
Mobile:

image

Desktop:

(vertical align)

image

### DIFF
--- a/bookwyrm/static/css/bookwyrm.css
+++ b/bookwyrm/static/css/bookwyrm.css
@@ -260,6 +260,10 @@ body {
 /* Book preview table
  ******************************************************************************/
 
+.book-preview td {
+    vertical-align: middle;
+}
+
 @media only screen and (max-width: 768px) {
     table.is-mobile, table.is-mobile tbody {
         display: block;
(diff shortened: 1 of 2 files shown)


arkhi:
That’s neat and fixes #949. Thanks. :)

The following is just food for thoughts; I have nothing against a merge.

My preference would have been a description list to avoid having to deal with exceptions for the table. Not sure what would be the best from a flow perspective. A `dl` would also reduce the `:empty` CSS logic since empty elements would not have to exist and could be dealt with from the template and HTML. That would also allow (maybe) to have only one link for the cover and the title which seem redundant with a screen reader. :)

I tried Orca on Ubuntu and it read the `content: [data-title]` logic well. Did you have the chance to try with other screen readers?

Anyhow… :+1:

joachimesque:
I decided to keep the `table` structure so it works well out of the box, witout too much work. A list of lists might be more malleable but perhaps less maintainable? 

VoiceOver reads it correctly enough, but I've noticed that we have the attribute `lang="en"` on the `html` tag even in French, so I'll fix that (in another PR).

PR #975, opened 2021-04-21
spanish translation updates

reesporte:
heres some more spanish translations

i did leave some comments for you to handle where i was unsure, this will tell you where 
`grep "TODO cc @mouse" -n < locale/es/LC_MESSAGES/django.po`

only one requires you to make edits (line 3139) 

the other is a maybe edit i think


locale/es/LC_MESSAGES/django.po
```diff
@@ -3282,892 +3023,853 @@ msgid ":?.!"
...
 msgid_plural "Please submit %d or more forms."
-msgstr[0] ""
-msgstr[1] ""
+msgstr[0] "Por favor, enviar %d o más formularios."
+msgstr[1] "Por favor, enviar %d o más formularios."
 
 #: venv/lib/python3.8/site-packages/django/forms/formsets.py:379
 #: venv/lib/python3.8/site-packages/django/forms/formsets.py:386
+#, fuzzy
+# TODO cc @mouse is this a verb or noun
```
mouse-reeve, line 3139: wait I have no clue! I don't think these should be in here, and I think it's because I put `venv3` in the ignore instead of `venv`.

https://github.com/bookwyrm-social/bookwyrm/blob/main/bw-dev#L93:
`runweb django-admin makemessages --no-wrap --ignore=venv3 $@`

reesporte: lol oops! well based on [this](https://github.com/django/django/blob/main/django/forms/formsets.py) i think it's a verb and also a completely different one, like order in the sense of arranging based on a rule, not order in the sense of you order food at a restaurant  

so i'd have to change it if it's to stay, but if i understand you right it really shouldn't be in there anyway?

mouse-reeve: I think all these are translated internally in Django already 😬

reesporte: alright  i changed it to venv and then rebuilt

mouse-reeve:
I think the test that failed is failing nondeterministically. At any rate, it appears totally unrelated to this pr

PR #984, opened 2021-04-22
Accessibility: add language to `lang` attr on `html` tag

joachimesque:
I only use the first subtag of the language string given by `get_language()`, because `get_language()` returns an all-lowercase string, and I don't know if it'll be considered valid by browsers.


mouse-reeve:
This is so much better than my screen reader just saying "ideograph ideograph" over and over the Chinese locale :)

Thank you!

PR #994, opened 2021-04-23
Refactor code for covers

arkhi:
The goal of this PR is to reduce the complexity of the code for covers, and try to use a component-style approach because it is used in many places.

I might be going a bit heavy handed here, but that will help to actually understand the project better.

One of the goals is to remove classes like `.is-large` and use responsive images instead.

This PR will probably conflict with #971, but that’s fine when it will. :)

I will be using this script to generate screenshots with [pageres](https://github.com/sindresorhus/pageres-cli/) along the way, to make sure (as much as possible) I am not breaking anything. I identified some URLs where a book cover would appear and I hope I am not missing too many. Let me know if that’s the case.

```sh
#! /bin/sh
#
# Usage: sh .pageres http://localhost:1333 [VALUE_OF_SESSION_COOKIE]

base=$1
cookie=$2

if [ -z "$base" ]; then
  echo "Please try again with a base URL. 👋"
  exit 0;
fi

  # /redraft-status/ID <=> post?status_type=TYPE&book=ID
if [ -n "$cookie" ]; then
  pageres \
    "$base/" \
    "$base/author/63" \
    "$base/book/361" \
    "$base/book/392" \
    "$base/book/990" \
    "$base/book/356/editions" \
    "$base/book/361/edit" \
    "$base/list" \
    "$base/list/1" \
    "$base/list/1/curate" \
    "$base/search/?q=kittens" \
    "$base/local" \
    "$base/federated" \
    "$base/discover" \
    "$base/get-started/books?query=3+secondes" \
    "$base/user/arkhi/" \
    "$base/user/arkhi/goal/2021" \
    "$base/user/arkhi/books/to-read" \
    "$base/post?status_type=comment&book=361" \
    320x568 \
    768x1024 \
    1024x768 \
    1216x960 \
    1920x1280 \
    --cookie="sessionid=$cookie" \
    --filename='logged-in-<%= date %>-<%= url %>-<%= size %><%= crop %>' \
    --format=jpg \
    --overwrite

  pageres \
    "$base/import/1" \
    320x9999 \
    768x9999 \
    1024x9999 \
    1216x9999 \
    1920x9999 \
    --crop \
    --cookie="sessionid=$cookie" \
    --filename='logged-in-<%= date %>-<%= url %>-<%= size %><%= crop %>' \
    --format=jpg \
    --overwrite
fi

pageres \
  "$base/" \
  "$base/local" \
  "$base/author/125" \
  "$base/book/361" \
  "$base/book/392" \
  "$base/book/990" \
  "$base/book/356/editions" \
  "$base/book/391/editions?page=2" \
  "$base/search/?q=kittens" \
  "$base/list" \
  "$base/list/1" \
  "$base/user/arkhi/" \
  "$base/user/arkhi/goal/2021" \
  "$base/user/arkhi/books/to-read" \
  320x568 \
  768x1024 \
  1024x768 \
  1216x960 \
  1920x1280 \
  --filename='logged-out-<%= date %>-<%= url %>-<%= size %><%= crop %>' \
  --format=jpg \
  --overwrite
```


bookwyrm/templates/discover/small-book.html
```diff
@@ -1,12 +1,22 @@
...
+
 {% if book %}
-<a href="{{ book.local_path }}">{% include 'snippets/book_cover.html' with book=book %}</a>
-{% include 'snippets/stars.html' with rating=book|rating:request.user %}
+    {% with book=book %}
+        <a href="{{ book.local_path }}">{% include 'snippets/book_cover.html' %}</a>
 
-<h3 class="title is-6"><a href="/book/{{ book.id }}">{{ book.title }}</a></h3>
-{% if book.authors %}
-<p class="subtitle is-6">{% trans "by" %} {% include 'snippets/authors.html' with book=book %}</p>
```
mouse-reeve, line 9: I realize this is a WIP, so feel free to disregard if this isn't helpful context -- the only reason I'm specifying `book=book` here is to be super explicit about the scope variables passed into the snippet. it would work the same without stating that, since the scope is inherited from the parent template. So adding `{% with book=book %}` isn't necessary either - it's just a matter of if you think it makes it clearer

arkhi:
Thanks @mouse_reeve. I do like the fact we know what variables we’re working with in a given file, so having the context repeated in the file seems good to me.

I do dislike the fact DTL does not allow for line-breaks and space management though, so I tend to not repeat the same variable in different places in the same file, hence the `{% with… %}` blocks. I also like to think about those blocks as a form of statement that says “Here, I want to work with that”.

bookwyrm/static/css/bookwyrm.css
```diff
@@ -153,26 +153,25 @@ body {
...
-    height: 100%;
-    object-fit: scale-down;
-}
+    display: block;
+    width: 100%;
 
-.no-cover {
-    position: relative;
-    white-space: normal;
+    /* Usweful when stretching under-sized images. */
```
joachimesque, line 162:
```suggestion
    /* Useful when stretching under-sized images. */
```

joachimesque:
In your list of pages to screenshoot, I'd add `http://localhost:1333/discover`: to my knowledge this is the only page where the book cover component is used with `size="large"`. See https://lire.boitam.eu/discover, https://bookwyrm.social/discover

(in `/bookwyrm/templates/book/book.html`, the book cover snippet is called with `size=large` instead of `size="large"` so it doesn't work)

mouse-reeve:
Yep, `/discover` is the only place `"large"` is used. `/discover` is the same as `/` when the user is logged out. Here's a deep cut, though, the standalone edit status page: `http://localhost:1333/post?book=445&status_type=comment`

arkhi:
> I'd add http://localhost:1333/discover

Thanks! I was playing with the logged out `/` which uses the same partial, but it won’t hurt to check this too. :)

> (in `/bookwyrm/templates/book/book.html`, the book cover snippet is called with `size=large` instead of `size="large"` so it doesn't work)

Yep! I’ve fixed this while updating templates… I’ll see what broke when I get to this page and if I need to do anything.

> Here's a deep cut, though, the standalone edit status page: http://localhost:1333/post?book=361&status_type=comment

Added, thanks!

bookwyrm/templates/book/book.html
```diff
@@ -48,10 +48,9 @@ <h2 class="subtitle">
 
     <div class="columns">
         <div class="column is-one-fifth">
-            <div class="is-clipped">
-                {% include 'snippets/book_cover.html' with book=book size=large %}
-                {% include 'snippets/rate_action.html' with user=request.user book=book %}
-            </div>
+            {% include 'snippets/book_cover.html' with book=book cover_class='has-height is-h-medium-mobile' %}
```
joachimesque, line 51: What do you think about inserting the `book_cover` button on the cover itself, when the cover is `large` or `medium` only?

mouse-reeve: do you mean the one to add the cover? I've been thinking for a while that it might be nice to have an "enlarge cover" button on the cover, which could be the add cover button if the cover is absent

joachimesque: ooh yes, the dual mode button is a good idea

arkhi: I’ll give it a shot. :)

arkhi:
I think I updated all templates containing covers and they seem to be fine, from the screenshots I took and the tests I did.

I’ll have another round of reading after a night of sleep, then I’ll submit the Review.

bookwyrm/templates/discover/small-book.html
arkhi: Thinking about it, would it be helpful to have a global `with` wrap listing the variables that are expected to be available in the template?

Would there be adverse consequences?

arkhi: Thinking about it, would it be helpful to have a global `with` wrap listing the variables that are expected to be available in the template? A sort of documentation… :)

Would there be adverse consequences?

bookwyrm/templates/book/book.html
arkhi: Actually not… This PR is already quite big, so I will create a new Issue for that.

arkhi: Done: https://github.com/bookwyrm-social/bookwyrm/issues/1009

bookwyrm/templates/get_started/book_preview.html
```diff
@@ -1,8 +1,8 @@
 {% load i18n %}
-<div class="column is-narrow is-clipped has-text-centered">
-    {% include 'snippets/book_cover.html' with book=book %}
-    <label class="label" for="id_shelve_{{ book.id }}">
```
arkhi, line 4: I could not see any use for this unclosed `<label>`, let me know if I missed something.

bookwyrm/templates/lists/curate.html
```diff
@@ -2,49 +2,71 @@
...
     {% else %}
-    <table class="table is-striped">
-        <tr>
-            <th></th>
-            <th>{% trans "Book" %}</th>
-            <th>{% trans "Suggested by" %}</th>
-            <th></th>
-        </tr>
+
+    <dl>
```
arkhi, line 18: @joachimesque: I implemented the list I mentioned in the past. I think it works fine but I can replace it with a table if need be.

arkhi (commented):
I think this PR is ready enough. I’m opening it to review; let’s see… :)

Better view the [whitespace-free version](https://github.com/bookwyrm-social/bookwyrm/pull/994/files?w=1).

Here is my [goodreads_library_export.csv](https://github.com/bookwyrm-social/bookwyrm/files/6392130/goodreads_library_export.csv), in case you’d just want to run the script for screenshots with a fresh DB.

I tried to understand what the current rules about the book-cover were doing, then I tried to make it a more resilient component. It is not perfect; maybe the dimensional constraints could be handled by `flex-direction` and `flex` instead of fixed dimensions, but I think the changes work well enough for now.

- The container to be flexible enough by composing with other classes, to align its content inside it if necessary, or just adapts to its content based on its context.
- The column following a `.is-cover` column will regain access to an automatic `flex-basis`, allowing it to resize its main axis.
- Generally, I think the `<a>` around the covers should better be within the container. For example, adding a button in the container is impossible with the current DOM tree.

There were a couple of existing issues that also prompted this PR:

- Discover was sort of broken.
- Book was cropping covers that are not vertical.
- The content beside the cover was not always aligned if the covers are not all the same width.
- Spacings sometimes felt odd. (Some still do. :) )
- Columns were using a lot of extra padding.
- `.content` is used in many places for content that is not markdown-generated, causing unwanted side-effects.

### Some Before / After

#### Author with one book

 

#### Author with multiple books

 

#### Editions with varying sized or missing covers

 

#### Lists with varying sized covers

 

#### Cropped content

 

#### Reading process

 

#### Clipped covers on Book

 

#### Funky Discover

 

---

 

#### Book editing

 

---

I took the opportunity to fix the horizontal scrolling for the getting started modal:

 

### A few other untouched weirdnesses

Getting started on a tiny screen:



Import needs some table love on narrow screens:



What @joachimesque did on the User Books works. :)

---

There… That’s a lot. Thank you. :)

bookwyrm/templates/get_started/book_preview.html
mouse-reeve: you didn't miss anything, that shouldn't be there

bookwyrm/templates/discover/small-book.html
mouse-reeve: I'd prefer using comments to document rather than variable assignments, I think

mouse-reeve:
This is a bug improvement! My only qualm is with the display of book previews on reviews in mobile:
Screen Shot 2021-04-28 at 9 22 29 AM

I'd prefer a large cover preview, but maybe the easiest fix here is to move the rate action and stars below the cover/snippet so that it doesn't constrain the width of the cover column?

arkhi:
> I'd prefer a large cover preview, but maybe the easiest fix here is to move the rate action and stars below the cover/snippet so that it doesn't constrain the width of the cover column?

I’d prefer that too. :)

I actually started doing that and, considering the increasing size of the PR, I went back to try to change what was necessary more than all I think I could improve. I sometimes got carried away, especially at the beginning.

I think it’s better if this specific layout can be solved in another smaller PR. Do you agree?

Also, an update on my last commit and a screenshot of the even more visible issue in French; I can only imagine the problem when German will come into play:

bookwyrm/templates/discover/small-book.html
arkhi: Sure. :)

I’ll try to move those `with` variables in a comment at the beginning of each template from now on. That will also reduce the number of indentations.

kopischke:
> I can only imagine the problem when German will come into play

From my experience doing both German and French localisation, total localised text length is mostly smaller in German than in French (German, a language devoid of most niceties, can be pretty terse; compare „Rauslehnen verboten“ with « Interdiction de se pencher »). The biggest issue are the bloodylongcompositewords German contains, which cause problems with reflow (you’d need automated hyphenation) and truncation (there are rules for correct truncation in German, but again, no automated support AFAIK).

mouse-reeve:
@arkhi agreed! apologies for the merge conflicts, otherwise I'm happy to merge

arkhi:
> From my experience doing both German and French localisation, total localised text length is mostly smaller in German than in French (German, a language devoid of most niceties, can be pretty terse; compare „Rauslehnen verboten“ with « Interdiction de se pencher »). The biggest issue are the bloodylongcompositewords German contains, which cause problems with reflow (you’d need automated hyphenation) and truncation (there are rules for correct truncation in German, but again, no automated support AFAIK).

@kopischke: I had been seeing some of those _bloodylongcompositewords_ issues during my (very quick) study of German and working for a German website, so I was taking this case into account more basing my comment on any accurate knowledge of the language. I hope you didn’t get offended as mockery was not my intention; I’m mostly trying to think about the worst case scenarii if possible and German composite words were what popped in my mind.

Thanks for the feedback; it’s good to learn a bit more. :)

In our specific case, I don’t believe line truncation on buttons would simply not really work (I’d go for hiding text overflow and ellipsis), but a rework of the interface could be a first step to avoid hiding any text at all.

kopischke:
> I hope you didn’t get offended as mockery was not my intention.

No worries there; I very much appreciate the BookWyrm project’s awareness of these issues (the discussion on gender neutral French forms I found illuminating in how it mirrors many of our own – I was pretty much only aware of the _point médian_ until then).

arkhi:
> @arkhi agreed! apologies for the merge conflicts, otherwise I'm happy to merge

No need to apologise. Part of the conflicts came from me formatting templates further than I should; my responsibility. :)

I’m very happy about the inclusion of inventaire.io in the mix and the simplified display of _author[ (date)]_!

Issue #948, opened 2021-04-17
Localised date is displayed in a weird format


arkhi:
When visiting the [page for a book](https://bookwyrm.social/book/118780), the publication date seems off.

### Steps to reproduce the behavior:

1. Set the preferred language of your browser to French.
1. Go to the book [Le léopard des neiges ](https://bookwyrm.social/book/118780).
2. Search for “Publié”.
4. The string reads _Publié Avr 23rd 1991 par Gallimard._

### Expected behavior:

I would expect to read _Publié le 23 Avril 1991 par Gallimard._

The _le_ part could be solved with the localisation, but the ordinal `rd` and the truncation of the month are not something I would expect in French.

### Screenshots

![image](https://user-images.githubusercontent.com/277773/115110447-c59b1180-9f7b-11eb-9874-340eeeb28b2a.png)

Cc: @joachimesque

arkhi:
The simplified Chinese also has the issue. I think it should read _1991年04月23日_ or simply 1991-04-23? Does @GuDzpoz concur? :)

![image](https://user-images.githubusercontent.com/277773/115110919-faa86380-9f7d-11eb-87bd-7bc8ffb070ec.png)

GuDzpoz:
Yes, either will do but probably not _Apr 23rd 1991_ :P

![image](https://user-images.githubusercontent.com/14026120/115231065-a8457f00-a104-11eb-980d-195889d28042.png)
![image](https://user-images.githubusercontent.com/14026120/115232134-e8f1c800-a105-11eb-8381-42c8bbc7d9ce.png)
By the way, the date under each status doesn't seem yet localized. Could it be of the same problem?

mouse-reeve:
Using Django's `naturalday` filter (https://docs.djangoproject.com/en/3.1/ref/contrib/humanize/) gives me, for example:

`在 Oct. 31, 2016 由 Princeton University Press 出版。` 
`Publié 31 octobre 2016 par Princeton University Press.`

It looks like the month isn't getting localized for Chinese? Which I believe is a bug in Django?

The date on statuses is a separate issue, I wrote a hyper-concise version of the `timesince` filter, but didn't think about localization. I replaced with timesince, and can live with the verbosity for the time being.

arkhi:
> Using Django's `naturalday` filter (https://docs.djangoproject.com/en/3.1/ref/contrib/humanize/) gives me, for example:
> 
[...]

PR #997, opened 2021-04-24
Locale `zh-cn` should be `zh-hans`

GuDzpoz:
According to [a StackOverflow comment](https://stackoverflow.com/questions/7728977/django-how-to-add-chinese-support-to-the-application#comment64616838_7729311) and [a Django release note](https://docs.djangoproject.com/en/3.2/releases/1.9/), the `zh-cn` language code has been removed and one should be using `zh-hans` instead. 
So dates are not getting localized as is in #948 becuase I have done it wrong in the very first step by running `./bw-dev makemessages -l zh_CN`. After renaming the directory and editing `settings.py` accordingly, it now seems to work on my side.
image


mouse-reeve:
thank you for keeping on top of this!

PR #999, opened 2021-04-25
update lines for spanish translations

reesporte:
i was thinking about adding a git hook that runs ./bw-dev makemessages and ./bw-dev compilemessages before pushing so that they're mostly up to date, i can look into this unless you have a better option in mind? not sure how annoying it would be to wait for those two scripts to run before pushing


mouse-reeve:
I think it would be cumbersome as a hook, but I wonder if there's a way to check in a github action to see if they should have been called 🤔

PR #1000, opened 2021-04-25
as far as i can tell, fr-dev is literally exactly the same as bw-dev

reesporte:
maybe someone still uses it but it's probably best in the long run to have one and i chose bw-dev because bookwyrm


reesporte:
pr 1000! wow

mouse-reeve:
it was just an alias back from when the name changed. Good to delete it, thank you!

PR #1001, opened 2021-04-25
Directory: Homogenize cards heights

joachimesque:
When the content of a profile card stretches it in height, the grid is not homogenous. This PR adds a CSS class that'll display cards as stretchable columns. The card content will be able to grow, which should always place the card footer at the bottom of the card.

### Before

Screenshot_2021-04-25 lire boitam eu

### After

Screenshot_2021-04-25 lire boitam eu(1)


mouse-reeve:
this is great! thank you

PR #1016, opened 2021-04-29
List style fixes

joachimesque:
This PR moves the counter on the Books List cards

image


mouse-reeve:
I love the little tab numbers! While the list position change looks great on desktop, for mobile, the input box is getting too squished to use:
Screen Shot 2021-04-29 at 1 30 40 PM

joachimesque:
> I love the little tab numbers! While the list position change looks great on desktop, for mobile, the input box is getting too squished to use:

Ah yes, I should have tested on mobile, the card footer is useless, too. I'll try an addon to Bulma styles so you can stack them horizontally on mobile with a specific class.

bookwyrm/static/css/bookwyrm.css
```diff
@@ -346,3 +346,33 @@ body {
         display: none;
     }
 }
+
+/* Book list
+ ******************************************************************************/
+
+ol.ordered-list {
+    list-style: none;
+    counter-reset: books;
```
arkhi, line 355: _books_ seems more specific than the name of the class implies. How about _items_ or having a more specific class name?

joachimesque:
@mouse-reeve what do you think of this look on mobile?

image

PR #1022, opened 2021-04-30
Let instance admins add a column of whatever they want in the footer

mouse-reeve:
Original:
Screen Shot 2021-04-30 at 11 06 02 AM

No additional info provided:
Screen Shot 2021-04-30 at 11 04 19 AM

With additional info:
Screen Shot 2021-04-30 at 11 04 28 AM
 (cc @bookwyrm-social/instance-admins)


mouse-reeve:
Feedback extra welcome on this, since i'm not sure what other folks would want to have in their instance footers

bookwyrm/management/commands/initdb.py
```diff
@@ -7,12 +8,14 @@
...
 
 def init_groups():
+    """permission levels"""
     groups = ["admin", "moderator", "editor"]
     for group in groups:
         Group.objects.create(name=group)
 
 
 def init_permissions():
+    """permissin types"""
```
joachimesque, line 18:
```suggestion
    """permission types"""
```

bookwyrm/models/site.py
```diff
@@ -19,19 +19,28 @@ class SiteSettings(models.Model):
...
+    # images
     logo = models.ImageField(upload_to="logos/", null=True, blank=True)
     logo_small = models.ImageField(upload_to="logos/", null=True, blank=True)
     favicon = models.ImageField(upload_to="logos/", null=True, blank=True)
+
+    # footer
     support_link = models.CharField(max_length=255, null=True, blank=True)
     support_title = models.CharField(max_length=100, null=True, blank=True)
     admin_email = models.EmailField(max_length=255, null=True, blank=True)
+    footer_item = models.TextField(null=True, blank=True)
```
joachimesque, line 43: Is the field sanitized before storage? (I don't have experience with the framework)

joachimesque (commented):
I like that feature! It'll be useful for admins who need to link elsewhere.

bookwyrm/models/site.py
mouse-reeve: It isn't, the admin has to make sure whatever html is in there is valid.

tofuwabohu:
I like this! I tried messing with the donation thing at one point but it was a bit to strict (wanted to set it to the bw patreon instead of some instance specific thing but it always includes the instance name, so this is a better way I think. Also some general purpose every here and then allows some nice theme-related stuff!

PR #1033, opened 2021-05-01
Search changes

### LINKED ISSUE #1028: [Discussion] Re-organize the search (results)
This is a bit of loud thinking, but I thought about this every time I'm searching anything so maybe others have opinions on this too?

The user usually knows whether they are looking for a book, a user or a list. The results of what they are not looking for are not necessary.
Searches take pretty long currently. I assume this is because many different searches are handled at the same time, even though in most cases, only some of them are necessary.

Ask the user what they are looking for, before the search is executed. I'm not sure what a good ui solution would look like here. My first idea is to have 3 search buttons in a row or a dropdown where the appropriate mode is chosen.
A mockup where the bookwyrm search is modified to have 2 different search buttons, one with a user and one with a book
Seeing this, I don't like the idea of having 3 buttons anymore, also starting the search with `Enter` wouldn't work anymore in a clear way. But you get the idea.

Maybe the tabbed search in Tusky could be an inspiration.

I thought maybe the context can get detected by the search input, but probably there are books out there containing an email address that should be found by searching "user@domain.tld" or users choose a name that looks like ISBN-13, so that's not an option I guess.

**Additional context**
I assume that once there are more bw instance connectors, (let's say bw.social is connected to 100 other instances), the current model will reach it's limits.
One failing search will also cause the result page to crash (I think), so only running necessary searches might help with that, even though it's a workaround.

A bit of another issue, but it'd be great to be able to search for known remote users by just entering their name without the instance.

mouse-reeve:
Fixes #1028 
This is _way_ faster.

Book search:
Screen Shot 2021-04-30 at 8 06 07 PM

Book search with remote results:
Screen Shot 2021-04-30 at 8 06 32 PM

User search:
Screen Shot 2021-04-30 at 7 59 52 PM

List search:
Screen Shot 2021-04-30 at 8 00 18 PM

Logged out search:
Screen Shot 2021-04-30 at 8 08 46 PM

Remaining work:
- [x] Tests
- [x] Result pagination
- [x] Screenreader review


tofuwabohu:
Just updated this, amazingly fast! Big UX improvement. :)

PR #1036, opened 2021-05-01
pagination list navigation using a range of pages such as provided by get_elided_page_range

SavinaRoja:
Updated the snippet for pagination to now handle a variable of `page_range`. If there are other pages, and if this variable is present, it creates a `<ul class="pagination-list">...` per documentation here: https://bulma.io/documentation/components/pagination/

The paginated book list view now uses this enhancement, providing the results of `Paginator.get_elided_page_range` (added in Django 3.2) to the templating.


mouse-reeve:
This looks great! Thank you

PR #1037, opened 2021-05-02
Fixes error on empty search results

mouse-reeve:


mouse-reeve:
The linting issue is due to a Black update, so I'm ignoring it for the PR

PR #1049, opened 2021-05-04
Datepicker

### LINKED ISSUE #1039: Date pickers display incorrectly to screen readers
When using a screen reader, such as VoiceOver under MacOS, the date pickers do not display as an editable spin control, but as a group when using the *tab* key.

When navigating via the arrow keys, the date displays as text, and there's no feedback when changing the value, either via *up/down* or by typing in a new value.

This can be seen when adding a new reading progress for a book.

Tested via VoiceOver/Safari, using b3e369cdba823413b81fec0b4f506a1909520a09

### LINKED ISSUE #1044: Date picker is invisible to screen readers when starting/finishing a book
After pressing *Start Reading* or *Finish Reading* on a book's page, a dialog pops up with the current date. This date picker is invisible to screen readers, nor can be focused/edited via the keyboard.

This might be due to an *ARIA-hidden* flag in the template.

mouse-reeve:
This switches from using the browser native datepicker, which isn't supported by Safari, to the [Duet datepicker](https://github.com/duetds/date-picker). It works well for me in VoiceOver, and should be well supported by other screen readers. And as a bonus it looks nice visually. (cc @rkingett)

Fixes #1039 
Fixes #1044

### DIFF
--- a/bookwyrm/templates/author.html
+++ b/bookwyrm/templates/author.html
@@ -39,3 +39,6 @@
 </div>
 {% endblock %}
 
+{% block scripts %}
+{% include 'snippets/datepicker_js.html' %}
+{% endblock %}
--- a/bookwyrm/templates/book/book.html
+++ b/bookwyrm/templates/book/book.html
@@ -333,4 +333,5 @@
 
 {% block scripts %}
 <script src="/static/js/vendor/tabs.js"></script>
+{% include 'snippets/datepicker_js.html' %}
 {% endblock %}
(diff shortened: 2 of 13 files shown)


Flameborn:
I really like this.

The date picker in Safari before 14.1 used to just include an edit field, which I liked, since I could type in the YYYY-MM-DD format. This adds this feature back, plus more, with an accessible date picker window.

Tested this using Safari and Brave via VoiceOver, as well as via NVDA under Windows Brave and Firefox.

PR #1070, opened 2021-05-10
Top bar

mouse-reeve:
This implements some changes to what links go in the top bar and the top bar pulldown suggested by @binyamin.

The top bar after ("Directory" is now in the main links, "Your Books" is in the pulldown):
Screen Shot 2021-05-10 at 4 12 17 PM

"Import Books" link is now in the "Your Books" page (I'm a little iffy on this link placement, tbh):
Screen Shot 2021-05-10 at 4 04 03 PM

@arkhi, this seems like the kind of change you might have opinions on 🙂


mouse-reeve:
Thank you! I guess I was thinking about this from a personal, pragmatic perspective, rather than a philosophical one, so I really like your framing. And by "personal, pragmatic" I mean, I use directory more than the "Your Books" view. I'm curious if you have a different experience in practice?

arkhi:
Mmm… I implied more than I stated; here is how I use BookWyrm:

I am detail oriented in general and will use BookWyrm to have and give the most accurate data I can get around books I read or plan to read; cleaning my own place first, basically. This means I use “Your Books” and “Your Lists” a lot to see what I can improve data-wise.

Although I love people in general and I have met a lot of amazing people in virtual spaces, I tend to more and more focus on the people in my immediate physical space (which is quite ironic during a pandemic, I know. :) ). Although I love books, I am also not a great reader, so I can’t add books endlessly, especially with the idea of keeping my library clean and manageable. That means Feeds, Notifications and Directory are places I will barely use if I want to keep a sense of sanity. Input overload can be real and I sometimes cut all notifications on my devices to take a breather.

That’s my experience: I love sharing with accuracy (the whole [Postel principle](https://en.wikipedia.org/wiki/Robustness_principle)), because I think it’s respectful to others and I need to limit an overwhelming input I receive one way or another.

I understand my situation is not everyone’s and probably not the best one to push for the use of BookWyrm as a social platform. :)

I don’t think pleasing everybody is possible, but for the sake of the discussion, I will develop a potential solution…

In the case of BookWyrm which, if I understood well, is about creating multiple communities through small interconnected instances, and if I’m transposing what I wrote before into software development, that could mean menu order depends on how a specific instance is used by its community or individuals.

An implementation could be menu customization so that specific communities can define their focus. That could be either by allowing menu entries to be reordered, or by giving pre‑established layout options (social interactions, data, maybe others…).

The downside is that the BookWyrm interface would change from one instance to another, but since interactions are done through the filter of one’s instance, that might not be a problem?

joachimesque:
I like this proposal. What's in the top bar is what's updated without my input, what might be new when I visit it.

Your Books and Settings only change when you update them. Arguably Direct Messages shouldn't be in this menu, but new messages show up in the Notifications.

mouse-reeve:
Thank you both! I think I'm going to merge this as for now just so it doesn't get hopelessly out of date, but I'm open to adjusting it.

While I do think of bookwyrm as being social-first insofar as that sets it apart from services like librarything that forefront cataloguing, I don't see book-focussed views and interaction-focussed views as a clear binary. Lists feel more social to me than "Your Books" because lists are about the curation of books by people in your network, and I envision it having more interactivity in the future.

binyamin:
> Lists feel more social to me than "Your Books" because lists are about the curation of books by people in your network, and I envision it having more interactivity in the future.

@mouse-reeve how about an "explore" or "discover" page? Besides for people's lists, it might display various popular books along with a couple comments and ratings. I specifically chose a vague name for this page, since that would provide plenty of room for adding stuff.

Issue #1088, opened 2021-05-17
Many BookWyrm services error when running on non-standard ports


bcj:
**Describe the bug**
If running BookWyrm on a server that hosts other non-BookWyrm services, those services may already be running on ports BookWyrm requires.
With the `db`, `redis_activity`, and `redis_broker` services, errors occur even after editing `.env`, `docker-compose.yml`, and `redis.conf`.

**To Reproduce**
For `db`:
1. In `docker-compose.yml`: update the `db` service's ports to `5433:5433` and add `-p 5433` to the command
2. follow the [install instructions](https://docs.joinbookwyrm.com/installing-in-production.html) up to initializing the db
3. attempt to run `./bw-dev initdb`
4. See the following error:
    ```
    Is the server running on host "db" (172.19.0.3) and accepting
    TCP/IP connections on port 5432?
    ```
For `redis_broker`:
1. In `.env` update `REDIS_BROKER_PORT` to 6380
2. In `redis.conf` update `port` to  6380
3. In `docker-compose.yml`: update the `redis_broker` service's ports to `6380:6380`
4. Run `docker-compose up --build`
5. Watch the `redis_broker` fail to come up with the following error:
   ```
    [2021-05-16 06:24:09,911: ERROR/MainProcess] consumer: Cannot connect to redis://:**@redis_broker:6380/0: Error 111 connecting to redis_broker:6380. Connection refused..
    ```

For `redis_activity`:
1. In `.env` update `REDIS_ACTIVITY_PORT` to 6380
2. Create a copy of `redis.conf` and update its `port` to  6380
3. In `docker-compose.yml`: update the `redis_activity` service's ports to `6380:6380` and update the volume to point at your modified conf file
4. follow the [install instructions](https://docs.joinbookwyrm.com/installing-in-production.html) up to initializing the db
5. Attempt to visit your instance and see a Server Error message that is being caused by an inability to connect to `redis_activity`

**Expected behavior**
It should be possible to run any BookWyrm service on a non-standard port and it should only require edits to `docker-compose.yml`, your `.env`, and the config file for that service.

**Desktop (please complete the following information):**
    - OS: macOS Big Suer 11.3.1
    - Browser Firefox 89b

bcj:
I have a work-in-progress PR for this, but I'm stuck on getting Redis working (details in PR)

PR #1089, opened 2021-05-17
Allow BookWyrm Services on Non-Standard Ports

bcj:
My attempt at resolving: https://github.com/bookwyrm-social/bookwyrm/issues/1088


bcj:
I think once this is done, a related PR would be needed to modify the `.dev` versions of files in the `main` branch. Is that correct?

bcj:
I'm a little stuck on the Redis fixes.
It looks like a bug in `docker-compose.yml` was stopping either redis service from seeing the `redis.conf` file.
Even with that fixed, connections seem to fail when the port isn't `6379`.
Celery does look for Redis on the right port, so it seems like the issue is that Redis ignores the port in the config.
Trying to manually access Redis from my host machine align with this (I'd need to double-check but I think the broker also wasn't accessible on `6379`).

bcj:
Alright, yeah. Just confirmed I can run two instances side-by-side. Non-standard ports work

mouse-reeve:
> I think once this is done, a related PR would be needed to modify the `.dev` versions of files in the `main` branch. Is that correct?

It should all be fine to do in one PR - the prod and main branches are much more in sync now, config-wise

bcj:
dev updated. this should be good to go

mouse-reeve:
Fixes #1088

mouse-reeve:
shoot, I just realized this merged directly into prod - could you make the same PR into main? I realize that sounds like the opposite of what I said before, but the workflow is that I merge into main and then merge main into prod (not obvious! I should have checked!)

bcj:
sure thing

PR #1097, opened 2021-05-19
Allow BookWyrm to run on Non-Standard Ports

bcj:
port of #1089 for the `main` branch.

`docker-compose.yml` is enough different that it wasn't a straight cherry-pick so two PRs makes sense (even if they were done in the wrong order)


docker-compose.yml
```diff
@@ -42,13 +42,9 @@ services:
     image: redis
     command: ["redis-server", "--appendonly", "yes"]
     env_file: .env
-    ports:
-      - 6378:6378
     networks:
       - main
     restart: on-failure
-    volumes:
-      - redis_activity_data:/data
```
bcj, line 51: oops, this is wrong! these should probably be saving to a named volume still.

docker-compose.yml
```diff
@@ -58,8 +54,6 @@ services:
     networks:
       - main
     restart: on-failure
-    volumes:
```
bcj, line 61: Ditto. I'll need to fix these before it's safe to merge

bcj: dropped the commit that made this change. should have watched a bit more closely when cherry-picking

PR #1123, opened 2021-05-23
Add max upload size config

eatsleepdeploy:
This PR should:
* Increase maximum file size for all file uploads to 10 MB in nginx.
* Add a UI warning if the image about to be uploaded is greater than 10 MB, and disable the offending form (via JavaScript). This is used for all custom image fields.
* Allow email backend to be settable from env for dev convenience.


mouse-reeve:
Thank you! I'll take a more thorough look at this next week.
You can automatically fix the python linting by running `./bw-dev black`.

mouse-reeve:
Thank you -- this looks good!

PR #1131, opened 2021-05-23
Author page microformats

joachimesque:
Start work on Author page microformats

ping @arkhi, you know more than me

### DIFF
--- a/bookwyrm/templates/author/author.html
+++ b/bookwyrm/templates/author/author.html
@@ -27,10 +27,10 @@
 
     {% if author.aliases or author.born or author.died or author.wikipedia_link or author.openlibrary_key or author.inventaire_id %}
     <div class="column is-two-fifths">
-        <div class="box">
+        <div class="box py-2">
             <dl>
                 {% if author.aliases %}
-                <div class="is-flex is-flex-wrap-wrap mb-2">
+                <div class="is-flex is-flex-wrap-wrap my-1">
                     <dt class="has-text-weight-bold mr-1">{% trans "Aliases:" %}</dt>
                     {% for alias in author.aliases %}
                         <dd itemprop="alternateName" content="{{alias}}">
@@ -39,14 +39,16 @@
                     {% endfor %}
                 </div>
                 {% endif %}
+
                 {% if author.born %}
-                <div class="is-flex mb-2">
+                <div class="is-flex my-1">
                     <dt class="has-text-weight-bold mr-1">{% trans "Born:" %}</dt>
                     <dd itemprop="birthDate">{{ author.born|naturalday }}</dd>
                 </div>
                 {% endif %}
+
                 {% if author.died %}
-                <div class="is-flex mb-2">
+                <div class="is-flex my-1">
                     <dt class="has-text-weight-bold mr-1">{% trans "Died:" %}</dt>
                     <dd itemprop="deathDate">{{ author.died|naturalday }}</dd>
                 </div>
@@ -54,7 +56,7 @@
             </dl>
 
             {% if author.wikipedia_link %}
-            <p class="mb-1">
+            <p class="my-1">
                 <a itemprop="sameAs" href="{{ author.wikipedia_link }}" rel=”noopener” target="_blank">
                     {% trans "Wikipedia" %}
                 </a>
@@ -62,7 +64,7 @@
             {% endif %}
 
             {% if author.openlibrary_key %}
-            <p class="mb-1">
+            <p class="my-1">
                 <a itemprop="sameAs" href="https://openlibrary.org/authors/{{ author.openlibrary_key }}" target="_blank" rel="noopener">
                     {% trans "View on OpenLibrary" %}
                 </a>
@@ -70,7 +72,7 @@
             {% endif %}
             
             {% if author.inventaire_id %}
-            <p class="mb-1">
+            <p class="my-1">
                 <a itemprop="sameAs" href="https://inventaire.io/entity/{{ author.inventaire_id }}" target="_blank" rel="noopener">
                     {% trans "View on Inventaire" %}
                 </a>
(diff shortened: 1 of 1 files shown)


arkhi:
(on it!)

bookwyrm/templates/author/author.html
```diff
@@ -22,42 +22,59 @@ <h1 class="title">{{ author.name }}</h1>
...
             <dl>
                 {% if author.aliases %}
                 <div class="is-flex">
                     <dt class="mr-1">{% trans "Aliases:" %}</dt>
-                    <dd itemprop="aliases">{{ author.aliases|join:', ' }}</dd>
+                    <dd>
+                        {% for alias in author.aliases %}
+                        <span itemprop="alternateName">{{alias}}</span>{% if not forloop.last %},{% endif %}
+                        {% endfor %}
+                    </dd>
```
arkhi, line 39: I understand you used the `<span>` to avoid including the `,` in the value, which is great. I think the markup could be simplified to the following and would be slightly more semantic regarding the description list. What do you think?
```html
                    {% for alias in author.aliases %}
                        <dd itemprop="alternateName" content="{{ alias }}">
                            {{ alias }}{% if not forloop.last %},{% endif %}
                        </dd>
                    {% endfor %}
```

bookwyrm/templates/author/author.html
```diff
@@ -22,42 +22,59 @@ <h1 class="title">{{ author.name }}</h1>
     </div>
 </div>
 
-<div class="block content columns">
-    {% if author.aliases or author.born or author.died or author.wikipedia_link %}
+<div class="block content columns" itemscope itemtype="https://schema.org/Person">
```
arkhi, line 25: We could take the opportunity to remove the [`content`](https://bulma.io/documentation/elements/content/) which generates styles that are problematic when not used on markdown-generated code.

This would avoid adding extra margins to `<dd>` elements if we decide to use one for each alias.

bookwyrm/templates/author/author.html
```diff
@@ -22,42 +22,59 @@ <h1 class="title">{{ author.name }}</h1>
...
             </dl>
 
             {% if author.wikipedia_link %}
-            <p><a href="{{ author.wikipedia_link }}" rel=”noopener” target="_blank">{% trans "Wikipedia" %}</a></p>
+            <p>
+                <link itemprop="sameAs" href="{{ author.wikipedia_link }}" />
+                <a href="{{ author.wikipedia_link }}" rel=”noopener” target="_blank">
+                    {% trans "Wikipedia" %}
+                </a>
+            </p>
```
arkhi, line 62: I think the `href` is the data that matters here, not the content being marked up, so the following code should work all the same:

```django

            <p class="mt-1">
                <a href="{{ author.wikipedia_link }}" rel=”noopener” target="_blank" itemprop="sameAs">
                    {% trans "Wikipedia" %}
                </a>
            </p>
```

I added a `mt-1` in case the `content` is removed.

arkhi: The same logic goes for other links if you agree.

bookwyrm/templates/author/author.html
```diff
@@ -22,42 +22,59 @@ <h1 class="title">{{ author.name }}</h1>
...
                 {% endif %}
                 {% if author.born %}
                 <div class="is-flex">
                     <dt class="mr-1">{% trans "Born:" %}</dt>
-                    <dd itemprop="aliases">{{ author.born|naturalday }}</dd>
+                    <dd itemprop="birthDate">{{ author.born|naturalday }}</dd>
                 </div>
                 {% endif %}
-                {% if author.aliases %}
+                {% if author.died %}
```
arkhi, line 48: I stumbled on a bug I didn’t fill about a `None` value displaying the absence of a date of death. We can either fix that later or now at your own convenience. :)

(example: https://bookwyrm.social/author/18686)

arkhi (commented):
Thanks a lot for that PR. :)

I commented <i>on-the-topic</i> code that, I think, could be simplified and small off topic <i>easy-to-fix</i> stuff. Feel free to ignore off-topic stuff, of course!

bookwyrm/templates/author/author.html
joachimesque: I forgot about multiple `dd`!

bookwyrm/templates/author/author.html
joachimesque: Already done in #1130 (I just have to update this branch with `main`)

bookwyrm/templates/author/author.html
```diff
@@ -22,42 +22,74 @@ <h1 class="title">{{ author.name }}</h1>
...
+            {% if author.librarything_key %}
+            <p class="mb-1">
+                <a itemprop="sameAs" href="https://www.librarything.com/author/{{ author.librarything_key }}" target="_blank" rel="noopener">
+                    {% trans "View on LibraryThing" %}
+                </a>
+            </p>
+            {% endif %}
+            
+            {% if author.goodreads_key %}
+            <p class="mb-1">
```
arkhi, line 89: Mmm… Not a blocker, just a comment:

I don’t have a strong opinion about `mb-x` or `mt-x` but if a `mb-x` happens on the last item, it will add an extra vertical space at the bottom of the surrounding `.box`. That’s why I had proposed `mt-x` on the successive `<p>`: there would already be the `dl` sibling before the first `p.mt-x`, allowing to style its following `p` consistently without adding extra space.

joachimesque: I added the `mb-x` to all the elements for consistency. I don't know which elements will be present in the box. It may be just a Wikipedia link, or just two definitions, or the whole enchilada. What I could do, is apply `my-2` to all the line elements (`p` and `dl > div`) and removing the `.box` vertical padding (with `py-0`). The only problem would be the `dl` keeping margins to be collapsed. I'll try it out.

joachimesque: image

Here's what I got with b72580d

arkhi: I was feeling like `mt-x` on all `p` would work for whatever number of `p` without any other change, but the `dl` might be empty, OK.

arkhi: The styling is out of the scope of this PR anyway. All’s good for me.

Issue #933, opened 2021-04-14
Storing images on the cloud


joachimesque:
I just installed Bookwyrm on a small server (congrats on the documentation, very straightforward). I don't have a lot of free space, and I expect the database to take a big chunk of it. I'd like to be able to store all the images on a cloud storage service, ideally S3-compatible, as many cloud providers offer storage buckets compatible with AWS's offer.

That feature saved my Mastodon instance when I had severe storage problems.

arkhi:
I find it ironical to create an alternative to Goodreads and host our files on Amazon servers. ;p

I don’t think S3 would go against the license, but still… :)

joachimesque:
You’ll notice I wrote “S3-compatible”, not “AWS S3” 😅

the S3 system has become a de facto industry standard, and is implemented on various other cloud providers… notably on Scaleway, which I use for my Mastodon instance. I personally discourage using AWS infrastructure but I don’t know if there’s any other system as widely used as S3 for Cloud file storage

seniorm0ment:
It could maybe be setup in a way that you can mount another server over SSH for media.
That being said no matter how this is implemented, it would be appreciated to have any remote content proxied through the instance server, instead of served directly to the end users.

PR #1138, opened 2021-05-24
Generate thumnails for books

joachimesque:
This PR adds a web performance improvement: Bookwyrm will now generate thumbnails for book covers.

One dependency is added, `django-imagekit`. No modifications have been made to the data model, all the calls for thumbnail generation are in the `snippets/book_cover.html`.

Sizes are passed through the arguments `size` and `size_mobile` in the snippet call for `snippets/book_cover.html`, if none of these args are provided there will be no alternative source, and the original cover image will be displayed.

The thumbnail sizes are modelled after the sizes already defined in the CSS file (xs to xxl). Thumbnails are generated in Wepb and JPEG. Variants are defined with different `<source>` inside a `<picture>`, the main `` carries the original cover image as a fallback. For more info about this technique, see [MDN: Responsive images](https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#use_modern_image_formats_boldly)

I do not know the intricacies of Django (and i'm not a back-end developer), so I'll appreciate any help to make this PR better!


mouse-reeve:
I'm finding that when I import a new book, the cover image file is present, but doesn't load correctly on the book page. 
Search result is working:
Screen Shot 2021-06-05 at 12 27 43 PM

Book page not working:
Screen Shot 2021-06-05 at 12 27 55 PM

joachimesque:
whoa, that's strange! after a few tests, I get the same result. The images are not generated. The urls are created, but the image files don't exist. The only way I could generate image was to add the 12 image sizes in the Model as `ImageSpecField` and send the command `./bw-dev generateimages`. Apart from that, updating an image or importing a book don't update the image. I'll have to see if the merge from main has changed anything. Perhaps the caching system have been changed? It's a part of Django I don't know anything about. I'll continue my inquiries

mouse-reeve:
huh! I did upgrade django from 3.2.0 to 3.2.1 for a security patch.

joachimesque:
that's not it :( I've been going up the chain of commits and it still doesn't work. At one point it did, I'm sure of it, but, well, I broke something along the way 😬 This PR status is back to draft

joachimesque:
The PR is open once again!

# Important!

When ALL the formats for ALL the books are generated, the `/images/thumbnails/covers` size is 4 to 5 times the size of the `/images/covers` dir. From my local install, with all the books imported from my Goodreads export (~400 books), `/images/covers` is 122M and the size of generated files in `/images/thumbnails/covers` is 565M (with ~2300 dirs created?).

This is a concern for instances with limited disk space, and if you prefer @mouse-reeve I can see if I can disable thumbnail generation by default and enable it via an env variable.

mouse-reeve:
I think having this disabled by default is a good idea - especially with #933 pending

joachimesque:
> I think having this disabled by default is a good idea - especially with #933 pending

I updated the PR to add an env var to enable/disable the feature. I suggest waiting for #1163 before merging the branch.

bw-dev
```diff
@@ -123,6 +123,9 @@ case "$CMD" in
     populate_streams)
         runweb python manage.py populate_streams
         ;;
+    generateimages)
```
mouse-reeve, line 126: To avoid confusion with `generate_preview_images`, would `generate_thumbnails` be clearer here?

joachimesque: You're right, it'll be less confusing that way.

The command `generateimages` comes from `django-imagekit` (https://github.com/matthewwithanm/django-imagekit/#management-commands), but it'll be more consitent internally (and less confusing) to use `generate_thumbnails`, I'm changing it rignt now.

bookwyrm/templates/snippets/book_cover.html
```diff
@@ -2,41 +2,136 @@
...
+
+            <img
+                class="book-cover"
+                itemprop="thumbnailUrl"
+                {% if book.cover %}
+                src="{% if img_path is None %}{% get_media_prefix %}{% else %}{{ img_path }}{% endif %}{{ book.cover }}"
+                {% else %}
+                src="{% static "images/no_cover.jpg" %}"
+                alt="{% trans "No cover" %}"
+                {% endif %}
```
joachimesque, line 118: I just saw that there's already a condition on the exitence of `book.cover` (L.7), so I think we can simplify this to:

```suggestion
                src="{% if img_path is None %}{% get_media_prefix %}{% else %}{{ img_path }}{% endif %}{{ book.cover }}"
```

The case for no `book.cover` is below, L.125

bookwyrm/templates/snippets/book_cover.html
```diff
@@ -2,41 +2,136 @@
...
-        title="{{ book.alt_text }}"
-    {% endif %}
->
-    <img
-        class="book-cover"
+                {% if size_mobile == 'xsmall' %}
+                    <source
+                        media="(max-width: 768px)"
+                        type="image/webp"
+                        srcset="{{ book.cover_bw_book_xsmall_webp.url }}"
```
mouse-reeve, line 24: I ran into a FileNotFound error here in cases where there was a cover in the database that was missing from static storage. In theory this shouldn't happen, but in practice.....

The best workaround that I could think of is to create a filter (maybe in /bookwyrm/templatetags/utilities.py`) that calls the url attribute wrapped in a try/catch. Something vaguely like:

```python3
@register.filter(name="cover_xsmall_webp")
def get_cover_small_webp(book):
    try:
        return book.cover_bw_book_xsmall_webp.url
    except OSError:
        return
```

Then you can call `{{ book|cover_xsmall_webp }}` in the template instead of invoking imagekit directly.

It would be cool to serve the `no_cover.jpg` asset here, but I couldn't figure out how to do that quickly and it's really out of scope anyway.

bookwyrm/templates/snippets/book_cover.html
mouse-reeve: This looks like something I janked up in the merge conflict. Thanks!

bookwyrm/templates/snippets/book_cover.html
joachimesque: Ok, I'll check that out in the next two days, I should have the time to get back into that PR. I think that the filter is a good idea. I think we can serve the right URL with the help of Django's Static utilities: https://github.com/django/django/blob/4b6208ffdd93d5d379d247ec094d567ef208effd/django/templatetags/static.py#L166

joachimesque: I fixed it in ce4553b367beb905fcf81d2efab9e7d265c52402

joachimesque:
> Since there's a function that takes size as a variable, I think you can pass `size_mobile` directly into `get_book_cover_thumbnail`, which would mean you don't need all these if/else statements

Thanks, I hadn't seen the obvious way I could refactor that code :)

.gitignore
```diff
@@ -29,4 +29,4 @@
 nginx/default.conf
 
 #macOS
-**/.DS_Store
+.DS_Store
```
mouse-reeve, line 32: Is this a merge oddity?

bookwyrm/templatetags/utilities.py
```diff
@@ -50,3 +51,13 @@ def truncatepath(value, arg):
     except ValueError:  # invalid literal for int()
         return path_list[-1]  # Fail silently.
     return "%s/…%s" % (path_list[0], path_list[-1][-length:])
+
+
+@register.simple_tag(takes_context=False)
+def get_book_cover_thumbnail(book, size, ext):
+    """Returns a book thumbnail at the specified size and extension, with fallback if needed"""
```
mouse-reeve, line 58: I'm hitting error when `size` is undefined here (I'm not sure where that's coming from) - I think a reasonable solution would be to use `'medium'` as a default value? Something like `size = size or "medium"`?

mouse-reeve (changes requested):
Sorry it took me so long to follow up on this! It looks great, just one minor change for error handling. Thanks so much :)

bookwyrm/templatetags/utilities.py
joachimesque: ah yes, good idea!

joachimesque: Would something like this work for you, or do you prefer positional args?

```diff
diff --git a/bookwyrm/templatetags/utilities.py b/bookwyrm/templatetags/utilities.py
index 1be66049..bcb1ec4e 100644
--- a/bookwyrm/templatetags/utilities.py
+++ b/bookwyrm/templatetags/utilities.py
@@ -54,7 +54,7 @@ def truncatepath(value, arg):
 
 
 @register.simple_tag(takes_context=False)
-def get_book_cover_thumbnail(book, size, ext):
+def get_book_cover_thumbnail(book, size="medium", ext="jpg"):
     """Returns a book thumbnail at the specified size and extension, with fallback if needed"""
     try:
         cover_thumbnail = getattr(book, "cover_bw_book_%s_%s" % (size, ext))
```

mouse-reeve: that works! that's better than what I suggested imo

mouse-reeve:
woo!

mouse-reeve:
@bookwyrm-social/instance-admins a note for when this gets merged into production: sometimes docker doesn't seem to pick up new dependencies for pip, so you may need to re-build the `celery_worker` and `web` images to make sure `django-imagekit` gets installed and `Pillow` gets updates

PR #1148, opened 2021-05-26
Update subtitle behaviour

arkhi:
When a title ends with a punctuation like `?`, `!` or similar, the whole title + subtitle is odd.

Furthermore, the `:` has different rules depending on the locale (narrow no-break space before in French, for example).

I revamped the view a bit to make the subtitle less part of the title itself and have a better separation of concerns.

image

image


joachimesque:
I think that the authors should still have the `subtitle` class, and the subtitle should have the `title is-5` class.

arkhi:
> I think that the authors should still have the `subtitle` class, and the subtitle should have the `title is-5` class.

Do you mind explaining why? :)

joachimesque:
Title and subtitle share a same role, one is the complement of the other. Author has another role. With that styling, title and subtitle dont share the same font weight. Am I looking at a subtitle or some other type of info? And now the author is not as big as it was before, I think it was not necessary to diminish its importance.

bookwyrm/templatetags/utilities.py
```diff
@@ -25,7 +25,7 @@ def get_title(book):
         return ""
     title = book.title
     if len(title) < 6 and book.subtitle:
-        title = "{:s}: {:s}".format(title, book.subtitle)
+        title = "{:s} ({:s})".format(title, book.subtitle)
```
mouse-reeve, line 28: Using parens for `Title (Subtitle)` scans oddly to me for English-language titles. I think it would be better to treat this as a translation using `gettext`, with "Title: Subtitle" as the default translation?

bookwyrm/templates/book/book.html
```diff
@@ -8,28 +8,35 @@
...
-            <h1 class="title">
-                <span itemprop="name">
-                    {{ book.title }}{% if book.subtitle %}:
-                        <small>{{ book.subtitle }}</small>
+            <h1 class="title" itemprop="name">
+                {{ book.title }}
+            </h1>
+
+            {% if book.subtitle or book.series %}
+                <p class="subtitle">
```
mouse-reeve, line 16: I agree with @joachimesque about the font weight - I think the subtitle should be visually close to the title, rather than the author.

mouse-reeve (changes requested):
I've been unsure about this one because using a colon seems to be very standard in how English-language titles are formatted, and it's unexpected to see them formatted as title/subtitle even though logically that makes sense?

Taking a look at how other platforms handle this:
- [GoodReads](https://www.goodreads.com/book/show/43704836-que-faire-des-cons-pour-ne-pas-en-rester-un-soi-m-me?from_search=true&from_srp=true&qid=CuV5PCEPfh&rank=1): `Que faire des cons? Pour ne pas en rester un soi-même` - displayed on one line, joined with a space
- [LibraryThing](https://www.librarything.com/work/23305040) - `Que faire des cons ? : Pour ne pas en rester un soi-même` - one line, joined with French-style colon
- [Amazon](https://www.amazon.fr/faire-cons-pour-rester-soi-m%C3%AAme/dp/2081452073) `Que faire des cons ?: Pour ne pas en rester un soi-même` - one line, joined with English-style colon
- [Publisher](https://editions.flammarion.com/que-faire-des-cons/9782081452077) - two lines

So in conclusion apparently there's no standard way of doing this!

joachimesque:
I am not convinced that a colon is the best solution for all books. It's to be avoided when there's another punctuation mark as the main goal of the PR states, but it's also not necessary in many cases: a colon suggests that the subtitle will explain the title, which is not always the case. It might be an expectation in English litterature that the subtitle explains the title, but from what I've seen in French books, it's more often a complement than a definition. And I've not seen any book (or publisher page for a book) sporting a colon to separate title from subtitle.

IMO, subtitles should be separated from the title by a new line and be smaller than the title, while keeping the same typographic context. I haven't looked if there are best practices semantically for book subtitles, but I think that it's best to keep the subtitle in the same element type (`hn` title) and making it less important, visually and in the code (via `<br><small>{subtitle}</small>`).

bookwyrm/templates/book/book.html
arkhi: I’m confused with this comment because that’s what the PR does. :)

arkhi:
Thanks for your feedbacks!

## Author

The whole idea of using a `subtitle` class for something that is not a subtitle is a [code smell] in itself. I really don’t think the `subtitle` class should be used on the author. With `subtitle`, the differentiation is not only operated based on the font, but also based on the surrounding white space.

Overall, the importance is not diminished; it is not “augmented”. Considering the information structure visually, it’s hard to not see or not identify the author as being the author.

Sematically, the author is not a structural heading, which is why I removed the `<h2>`.

## Subtitle

[English Wikipedia][wp-subtitle] shows the following:

> In printing, subtitles often appear below the title in a less prominent typeface or following the title after a colon.

Removing cultural behaviours by clearly splitting title and subtitle being the point of this PR, and with no clear‑cut information about how to display subtitles, like the discussion about punctuation shows, I’m more inclined to consider something more semantic like what @joachimesque proposed (`<br> + <small>`), although this destroys the logic between `title` and `subtitle`.

I’ll give it a shot in the coming days. :heart: 

[code smell]: https://en.wikipedia.org/wiki/Code_smell
[wp-subtitle]: https://en.wikipedia.org/wiki/Subtitle_(titling)

bookwyrm/templatetags/utilities.py
arkhi: I agree with the gettext but I won’t take the time to look into it at this point. Do you prefer I revert my changes or keep it like this?

arkhi:
I made some changes; here is a summary of the situation… Let me know your thoughts.

The current status:

 

With the current PR:

 

With the following code:

```html
<h1 class="title" itemprop="name">
    TITLE
    <br><small>
        <meta itemprop="alternativeHeadline" content="SUBTITLE">
        SUBTITLE
    </small>
</h1>

<div>
    par <a href="/author/70" class="author" itemprop="author" itemscope itemtype="https://schema.org/Thing"><span itemprop="name">Author<span></span></span></a>
</div>
```

 

---

With the intended class for the subtitle:

```html
<h1 class="title" itemprop="name">
    TITLE
    <br><small class="subtitle">
        <meta itemprop="alternativeHeadline" content="SUBTITLE">
        SUBTITLE
    </small>
</h1>

<div>
    par <a href="/author/70" class="author" itemprop="author" itemscope itemtype="https://schema.org/Thing"><span itemprop="name">Author<span></span></span></a>
</div>
```

 

With the following, tweaking the PR visually but keeping the title as the title only:

```html
<h1 class="title" itemprop="name">
    TITLE
</h1>

<div class="subtitle is-5 mb-1">
    <meta itemprop="alternativeHeadline" content="SUBTITLE">
    <span class="has-text-weight-bold">
        SUBTITLE
    </span>
</div>

<div>
    par <a href="/author/70" class="author" itemprop="author" itemscope itemtype="https://schema.org/Thing"><span itemprop="name">Author<span></span></span></a>
</div>
```

`is-5` follows [Bulma’s rule of thumb about sizes for titles](https://bulma.io/documentation/elements/title/#sizes).

 

### Document structure

There is also the consideration of what the structure of the document is.

The current status, the PR and the tweaking with `<hr><small>`:





- I fixed the `<h3>` on the ratings which is not semantically correct.
- We can see that semantically including the subtitle within the title can make it very long. It also means we either have an issue with the punctuation in case of multiple signs, or the absence of a punctuation if no sign at all.

arkhi:
(I think we can also see the benefits of consistently wrapping the title within quotes when it is used in a sentence; it makes it easier to scan what is the title in the sentence.)

bookwyrm/templatetags/utilities.py
arkhi: Scratch my previous comment. ’Not being in a good place at the moment and I needed a productive distraction.

I added the string and the filter can now take a variable from the template, like so:
This could actually be a variable that can be set for each language, as limits are likely to differ from one to another.


```django
<a href="{{ status.book.local_path }}">{{ status.book|book_title:10 }}</a>
```

mouse-reeve: I like it!

bookwyrm/templates/book/book.html
mouse-reeve: Sorry for being unclear! I just mean that when the subtitle has a lighter font weight (as here: https://user-images.githubusercontent.com/277773/119652123-03664200-be26-11eb-85ec-a25899ada82d.png) it looks visually similar to the author, as opposed to with a heavier font weight (as here:https://user-images.githubusercontent.com/277773/121321562-813e4900-c90e-11eb-820a-3d2cc21ae520.png), where the title and subtitle are more visually similar.

arkhi:
Rejoice: you’ll both be happy to learn `.subtitle` is back on the author. I think it’s semantic heresy but it needs to be there for now, modulo a simple tweak to deal with margins created by Bulma.

@mouse-reeve: Ready when you are.

joachimesque:
IMO, what's important about “semantics” in HTML only applies to tags themselves, not to class names. The class name `subtitle` is dictated by Bulma, it if we want to benefit from its styles we have no choice but use even if it's used for an author, not a subtitle.

What we can do when we'll be able to compile SASS server-side, is to make an alias of `subtitle` called `author-name`, to keep the class name relevant to the content. We should also be able to make a `book-subtitle` class for the book subtitle styles.

Using multiple utility classes for size, weight, margin, could also be done, but would be way less practical when we'll be able to modify SASS variables and make themes, so I wouldn't encourage it.

arkhi:
In a World where time and energy do not matter, I’d be happy to get rid of Bulma altogether.

mouse-reeve:
thank you for the work and thought on this!

PR #1154, opened 2021-06-02
Front-end: Better forms

joachimesque:
Some fixes and enhancements for many forms. 

- Use `class="field"` when possible for more consistent spacing between form fields
- Only allow `input[type="checkbox"]` or `input[type="checkbox"]` inside `<label>`
- Truncate file URL in clearable_file_input
- Place checkboxes and radios before label text
- Disallow interactive elments (like `a`) inside `label`


mouse-reeve:
this is such a big improvement! thank you!

Issue #1098, opened 2021-05-19
Redis Broker Unreachable if Password Contains Characters That Need URL Encoding


bcj:
**Describe the bug**
The `CELERY_BROKER` and `CELERY_RESULT_BACKEND` environment variables use the redis broker password without urlencoding it, so any unsafe characters will result in the broker being unreachable (and may erroneously be listed as IPv6 errors by flower)

**To Reproduce**
1. set the redis broker password to something including a `:`
2. attempt to create a new server using the production instructions
3. See errors when running `./bw-dev up`

**Expected behavior**
No errors occur and everything works as accepted (or: note in .env.prd.example that the password needs to only use `[A-Za-z0-9~._-]`.

**Screenshots**

**Desktop (please complete the following information):**
    - OS: macOS Big Sur
    - Browser N/A
    - Version `production`

**Smartphone (please complete the following information):**

**Additional context**

bcj:
sorry for the slightly rushed issue info.
I can tackle this later this week, In all the python code it should be easy to construct the URL manually and safely from port and password, need to check on how it will be easiest on the flower service

tofuwabohu:
This is probably related: When the smtp password contains such characters and is not set in `""`, `bw-dev update` will fail. However, when the passwort is set in `""`, emails can't be sent. Currently I switch this back and forth at every update and it's not too bad, but it may be the same thing?

PR #1157, opened 2021-06-03
URL-Encode password in broker_url

bcj:
Related to #1098. The `main` branch doesn't actually use redis passwords but these are the structural changes needed to only use url-encoded versions of passwords when passing over HTTP.

I've tested this against the `production` branch and confirmed it works, but the changes to `docker-compose.yml` result in an ugly merge conflict.
I know you've said that `main` and `production` are close enough to just target `main` but I'm fine sending separate PRs for both if that's easier (or even if you just want to see if it's easier).

Also, oops. I've been sitting on this for 2 weeks because I didn't want to debug an issue that turned out to be I was using an invalid flower password.


Dockerfile
```diff
@@ -9,5 +9,3 @@ WORKDIR /app
 COPY requirements.txt /app/
 RUN pip install -r requirements.txt --no-cache-dir
 RUN apt-get update && apt-get install -y gettext libgettextpo-dev && apt-get clean
-
-COPY ./bookwyrm ./celerywyrm /app/
```
bcj, line 13: Deleted because Flower _had_ been the only container that wasn't setting up the working directory as `/app` and now it's actively using the Celery config so it seemed to make sense to have it use the same volume that's live updated as the Celery workers

mouse-reeve:
yay! thank you

mouse-reeve:
The merge was no problem, it's on both branches now

PR #1163, opened 2021-06-07
Use S3 as static and media storage

### LINKED ISSUE #933: Storing images on the cloud

joachimesque:
Closes #933

TODO

- [x] Write some documentation for non-AWS, S3-compatible services (https://github.com/bookwyrm-social/documentation/pull/14)
- [x] Write some documentation for the CORS (why fonts won't load without a command) (https://github.com/bookwyrm-social/documentation/pull/14)
- [x] Make a management command to move existing media to the new S3 bucket
- [ ] See if Nginx can reverse-proxy the S3 bucket in order to keep the files accessible through the domain name
  (as per https://github.com/bookwyrm-social/bookwyrm/issues/933#issuecomment-828702871, perhaps by [following this guide](https://www.scaleway.com/en/docs/setting-up-object-proxy-object-storage/))


.env.dev.example
```diff
@@ -42,6 +42,19 @@ EMAIL_HOST_PASSWORD=emailpassword123
 EMAIL_USE_TLS=true
 EMAIL_USE_SSL=false
 
+# Set this to true when initializing certbot for domain, false when not
+CERTBOT_INIT=false
```
mouse-reeve, line 46: this looks like a merge regression, I think lines 45-46 should be deleted relative to `main`

.env.prod.example
```diff
@@ -42,6 +42,19 @@ EMAIL_HOST_PASSWORD=emailpassword123
 EMAIL_USE_TLS=true
 EMAIL_USE_SSL=false
 
+# Set this to true when initializing certbot for domain, false when not
+CERTBOT_INIT=false
```
mouse-reeve, line 46: Here too

bookwyrm/templates/book/book.html
```diff
@@ -321,5 +326,5 @@ <h2 class="title is-5">{% trans "Lists" %}</h2>
 {% endblock %}
 
 {% block scripts %}
-<script src="/static/js/vendor/tabs.js"></script>
+<script src="{% static "js/vendor/tabs.js" %}"></script>
```
mouse-reeve, line 329: this is great! I didn't realize this was a thing

bookwyrm/context_processors.py
```diff
@@ -11,10 +11,7 @@ def site_settings(request):  # pylint: disable=unused-argument
     return {
         "site": models.SiteSettings.objects.get(),
         "active_announcements": models.Announcement.active_announcements(),
-        "static_url": settings.STATIC_URL,
-        "media_url": settings.MEDIA_URL,
-        "static_path": settings.STATIC_PATH,
-        "media_path": settings.MEDIA_PATH,
```
joachimesque, line 17: (Most of these were not used)

bookwyrm/settings.py
```diff
@@ -58,6 +58,7 @@
 
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = env.bool("DEBUG", True)
+USE_HTTPS = env.bool("USE_HTTPS", False)
```
joachimesque, line 61: I'm defaulting `USE_HTTPS` to False, but it might not be best for people who are updating their instance in Production and haven't updated their `.env`. What's your opinion?

bw-dev
```diff
@@ -110,7 +121,40 @@ case "$CMD" in
...
+    set_cors_to_s3)
+        awscommand "$(pwd):/bw"\
+            "s3api put-bucket-cors\
+            --bucket ${AWS_STORAGE_BUCKET_NAME}\
+            --endpoint-url ${AWS_S3_ENDPOINT_URL}\
+            --cors-configuration file:///bw/$@"
+        ;;
     *)
-        echo "Unrecognised command. Try: build, clean, up, initdb, resetdb, makemigrations, migrate, bash, shell, dbshell, restart_celery, test, pytest, test_report, black, populate_streams, generate_preview_images"
+        set +x
```
joachimesque, line 138: Here's the result:

image

bookwyrm/settings.py
mouse-reeve: hm, I think nginx should automatically redirect to https in these cases? I wonder if it will cause a problem for social media previews 🤔

joachimesque: Ah yes, I forgot about nginx auto redirection! I don't think it'll be a problem.

.env.prod.example
```diff
@@ -42,6 +43,16 @@ EMAIL_HOST_PASSWORD=emailpassword123
 EMAIL_USE_TLS=true
 EMAIL_USE_SSL=false
 
+# S3 configuration
+# Commented are example values if you use Scaleway instead of AWS
+USE_S3=false
+# AWS_ACCESS_KEY_ID=
+# AWS_SECRET_ACCESS_KEY=
```
mouse-reeve, line 50: Are these optional, even for AWS? if not, I think they should be uncommented for clarity

.env.dev.example
```diff
@@ -42,6 +43,16 @@ EMAIL_HOST_PASSWORD=emailpassword123
 EMAIL_USE_TLS=true
 EMAIL_USE_SSL=false
 
+# S3 configuration
+# Commented are example values if you use Scaleway instead of AWS
+USE_S3=false
+# AWS_ACCESS_KEY_ID=
```
mouse-reeve, line 49: Would it make more sense to name these variables `S3_...` since they aren't necessarily AWS?

bookwyrm/settings.py
```diff
@@ -179,19 +181,51 @@
...
+
+USE_S3 = env.bool("USE_S3", False)
+
+if USE_S3:
+    # AWS settings
+    AWS_ACCESS_KEY_ID = env("AWS_ACCESS_KEY_ID")
+    AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY")
+    AWS_STORAGE_BUCKET_NAME = env("AWS_STORAGE_BUCKET_NAME")
+    AWS_S3_CUSTOM_DOMAIN = env("AWS_S3_CUSTOM_DOMAIN")
+    AWS_S3_REGION_NAME = env("AWS_S3_REGION_NAME")
```
mouse-reeve, line 210: I wasn't sure what value I'd use for region name (it isn't obvious in DO settings), so I added a default value here with `env("AWS_S3_REGION_NAME", "")` to make it optional in the .env file and it seemed to work fine without it.

mouse-reeve (commented):
This is so cool and easy to use! It's unfamiliar territory for me so these are very genuinely just questions.

mouse-reeve:
Sorry about the merge conflict! I renamed `bookwyrm/templates/feed/feed_layout.html` to `bookwyrm/templates/feed/layout.html`

.env.dev.example
```diff
@@ -42,6 +43,16 @@ EMAIL_HOST_PASSWORD=emailpassword123
...
+# AWS_SECRET_ACCESS_KEY=
+# AWS_STORAGE_BUCKET_NAME=        # "example-bucket-name"
+# AWS_S3_CUSTOM_DOMAIN=None       # "example-bucket-name.s3.fr-par.scw.cloud"
+# AWS_S3_REGION_NAME=None         # "fr-par"
+# AWS_S3_ENDPOINT_URL=None        # "https://s3.fr-par.scw.cloud"
```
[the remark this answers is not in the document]
joachimesque: If I remember well, here's how it works:

The S3 interface code is handled by the AWS SDK for Python, named Boto3. These `AWS_S3_*` env vars are used by Boto3.

If you use Amazon AWS S3, you only need `AWS_STORAGE_BUCKET_NAME` and `AWS_S3_REGION_NAME`. Boto3 will create an Amazon AWS compatible URLs with these two, for both the endpoint and the server domain.

If you use an alternative S3 compatible service, like DigitalOcean or Scaleway, Boto3 won't know how to make the endpoint URL (for storing) or the domain (for serving), so you'll have to specify both.

`AWS_S3_CUSTOM_DOMAIN` is also useful if you serve whatever is on your bucket through a CDN, like Cloudfront. It can also be used if you want to serve your images through the same domain as your instance (some Nginx config will be useful there)

.env.dev.example
joachimesque: Sadly, these variables are expected by Boto3 (through `django-storages`), so I don't think it can be done. Perhaps we could define `S3_ACCESS_KEY_ID` in `.env` and then "translate" it to `bookwyrm/settings.py` by defining `AWS_ACCESS_KEY_ID = env("S3_ACCESS_KEY_ID")`, but IMO it adds some unwanted complexity.

bookwyrm/settings.py
joachimesque: I think that the region name is only used in case you use AWS S3 and no `AWS_S3_CUSTOM_DOMAIN` or `AWS_S3_ENDPOINT_URL`.

bookwyrm/settings.py
```diff
@@ -179,19 +181,51 @@
...
+    STATIC_URL = "https://%s/%s/" % (AWS_S3_CUSTOM_DOMAIN, STATIC_LOCATION)
+    STATICFILES_STORAGE = "bookwyrm.storage_backends.StaticStorage"
+    # S3 Media settings
+    MEDIA_LOCATION = "images"
+    MEDIA_URL = "https://%s/%s/" % (AWS_S3_CUSTOM_DOMAIN, MEDIA_LOCATION)
+    MEDIA_FULL_URL = MEDIA_URL
+    DEFAULT_FILE_STORAGE = "bookwyrm.storage_backends.ImagesStorage"
+    # I don't know if it's used, but the site crashes without it
+    STATIC_ROOT = os.path.join(BASE_DIR, env("STATIC_ROOT", "static"))
+    MEDIA_ROOT = os.path.join(BASE_DIR, env("MEDIA_ROOT", "images"))
```
joachimesque, line 225: If you have an opinion on these lines, I don't remember clearly why I wrote that comment, but I seem to remember that `STATIC_ROOT` are only used for routes in dev?

mouse-reeve: Huh, I have no idea. But if it crashes without it, then keep it a guess 😂

mouse-reeve: Huh, I have no idea. But if it crashes without it, then keep it I guess 😂

mouse-reeve:
🎉

Issue #1099, opened 2021-05-19
Remove user information from "Yours books" page


cats-shadow:
The "yours books" page displays the same userinfo just like the user profile. But the user already knows everything about himself (he filled out a profile). :) 
As a result, to access their "shelves" every time you have to scroll down the page, which is inconvenient. 
In the screenshot the information that is offered to be removed is framed.

![2021-05-19_17-24-12](https://user-images.githubusercontent.com/16667173/118830035-4e80d200-b8c7-11eb-805a-36447a15db4b.png)

mouse-reeve:
Ah yeah, this really makes a big difference for people with long profiles. I think there does need to be a clear indication of who's books page is being shown, since this same view is used to show any users' shelves, not just your own. But it certainly doesn't need to be the entire profile and bio!

cats-shadow:
Perhaps it makes sense to show this in the header?
I mean, instead of "Yours books", in my example, a link to the user's profile "<a href='https://books.cats-home.net/user/cats-shadow'>Michael [Кошак] Skolsky</a> books", that would save space without losing information.

shilpeePrasad:
Hi, I would like to work on this issue. Thanks

mouse-reeve:
Go for it @shilpeePrasad 😊

shilpeePrasad:
> Go for it @shilpeePrasad 😊

Hi could you please let me know the file that I should refer for removing the bio.

mouse-reeve:
This is the template for the shelf page: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/user/shelf/shelf.html

It uses a user page layout (https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/user/layout.html#L23), which contains the bio. Since a number of pages where it does make sense to show the bio all use that template, it might make sense to use the base template (https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/layout.html) for the shelf page instead.

shilpeePrasad:
> This is the template for the shelf page: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/user/shelf/shelf.html
> 
[...]

PR #1195, opened 2021-06-22
Fixed issue #1099 (Remove bio from Your Books)

shilpeePrasad:
Fixed issue #1099


bookwyrm/templates/user/layout_without_bio.html
```diff
@@ -0,0 +1,58 @@
...
+<header class="block">
+    {% block header %}
+    <h1 class="title">
+        {% trans "User Profile" %}
+    </h1>
+    {% endblock %}
+</header>
+
+
+{% block tabs %}
```
mouse-reeve, line 24: Since this layout is only being used by one template, I think it makes more sense to have `bookwyrm/templates/user/shelf/shelf.html` inherit directly from `templates/layout.html` and remove this entirely.

shilpeePrasad: Hi if we directly use templates/layout.html then tabs All books, To Read, Currently Reading, Read are no displayed please find screenshot below.


image

mouse-reeve: Sorry, I should have been more specific. The child template has to call the blocks that are defined in the parent template. In the user layout template, the two important blocks are the `tabs` block, which has a default set of tabs that can be overwritten, and the `panel` block for the main body of the page. This layout uses a grandparent layout (sorry, it's so complicated!), `templates/layout.html` that has a `content` block, which contains the tabs and panel blocks.

To use `templates/layout.html` directly, you have to call the `content` block, and include the html that is currently wrapped in the `tabs` and the `panel` in that block. 

Does that make any sense at all?

PR #1199, opened 2021-06-23
More efficient search index

### LINKED ISSUE #1196: Impossible to search for "There There" by Tommy Orange
When you search for "There There" (just the title, not including the author) absolutely nothing turns up. The book is most certainly in the database, but something in the postgres search algorithm (maybe it's filtering repetition or common words?) is making it un-findable.

mouse-reeve:
This commit changes the way search is computed by creating a SearchVectorField (https://docs.djangoproject.com/en/3.2/ref/contrib/postgres/search/#searchvectorfield) and using a psql trigger (https://www.postgresql.org/docs/current/textsearch-features.html#TEXTSEARCH-UPDATE-TRIGGERS) to populate the search index column. 

Fixes #1196


bookwyrm/migrations/0077_auto_20210623_2155.py
```diff
@@ -0,0 +1,73 @@
...
+                fields=["search_vector"], name="bookwyrm_bo_search__51beb3_gin"
+            ),
+        ),
+        migrations.RunSQL(
+            sql="""
+                CREATE FUNCTION book_trigger() RETURNS trigger AS $$
+                begin
+                    new.search_vector :=
+                        coalesce(
+                            NULLIF(setweight(to_tsvector('english', coalesce(new.title, '')), 'A'), ''),
```
mouse-reeve, line 43: This line tries to use the "english" dictionary to create the search vector, and uses the "simple" dictionary as a fallback if the result of the first vector is empty.

For "There There", this takes the vector from:
"'orange':2C 'tommy':1C"
to
"'orange':4C 'there':1A,2A 'tommy':3C"

You might think this would get the "there there" query working, but so far it is still producing no results.

mouse-reeve: using F() when passing the search vector field into SearchRank fixed this

PR #1208, opened 2021-07-03
Add backend parameter to Celery initializer

cincodenada:
This fixed an error when viewing imports for my setup. I was getting the error mentioned in [this StackOverflow post](https://stackoverflow.com/questions/23215311/celery-with-rabbitmq-attributeerror-disabledbackend-object-has-no-attribute) when trying to access `/import/`. StackOverflow suggested adding the `backend` argument to Celery's initializer, so I did that, and it seemed to do the trick.


mouse-reeve:
I've run into this as well, thank you!

PR #1216, opened 2021-07-07
Fixes detection of links with certain characters

### LINKED ISSUE #1207: full links not recognized
**Describe the bug**
Only the first part of a link with a /@ in it is being linked so https://mastodon.social/@username only auto links https://mastodon.social and strips '@username' from being a link

**To Reproduce**
here is an example https://bookwyrm.social/user/wakest/comment/30867

**Expected behavior**
When a link is in a comment, the whole link should be turned into a link not just part of it

**Screenshots**
image

**Instance**
https://bookwyrm.social

mouse-reeve:
Fixes #1207

### DIFF
--- a/bookwyrm/views/status.py
+++ b/bookwyrm/views/status.py
@@ -150,7 +150,7 @@ def find_mentions(content):
 def format_links(content):
     """detect and format links"""
     return re.sub(
-        r'([^(href=")]|^|\()(https?:\/\/(%s([\w\.\-_\/+&\?=:;,])*))' % regex.DOMAIN,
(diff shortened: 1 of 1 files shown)


wakest:
wow it was really as simple as deleting @# from the regex?! thats cool!

mouse-reeve:
I was just thinking it seemed janky, so I'm glad it seems cool!

wakest:
will probably be fine for now but might be useful to another projects link detector library. There are a surprising amount of different link/uri types that people want to be able to link to. Like magnet: and gemini: links were a hugely requested feature on mastodon that got implemented.

Issue #1174, opened 2021-06-14
Incorrect shelving dates after goodreads import


emiljacobsen:
**Describe the bug**
The shelving dates are all wrong after importing my library from goodreads. For example, looking at goodreads, I can see that I added Slaughterhouse Five on July 26, 2016. In bookwyrm on the other hand, the shelving date is October 30, 2020. In general, it seems all the shelving dates on bookwyrm are quite recent (but many still prior to the date I joined the website). I see no other pattern to what dates were chosen.

**To Reproduce**
1. Import library from goodreads.
2. Compare shelving dates in bookwyrm to added dates in goodreads.
3. Note the difference.

**Expected behavior**
I expected the dates to either coincide, or for all the bookwyrm shelving dates be the date I imported the library.

**Screenshots**
![bookwyrm-sh5](https://user-images.githubusercontent.com/1376211/121950228-a50fed80-cd59-11eb-9947-2c67c6399dfe.png)
![goodreads-sh5](https://user-images.githubusercontent.com/1376211/121950234-a5a88400-cd59-11eb-8e02-1252d45d5aea.png)

mouse-reeve:
Have you added read dates retroactively to goodreads? For example, is it possible that you read _Slaughterhouse Five_ in 2016, but added it to goodreads in 2020?

Sometimes goodreads doesn't include the reading start date, and we use the date added instead.

emiljacobsen:
No, sorry, I should have been more specific. The book has no read dates, it's only ever been on the 'to read' shelves, and I can't associate any other dates to it than July 26. In fact, I was only looking at books in this situation when I noticed the discrepancy.

I'm now comparing dates for books in my 'read' shelf, and these too differ between the sites. Here's an interesting example: A People's History of the United States:

On goodreads:
- added July 26, 2016
- started reading July 26, 2016
- finished reading March 9, 2017

On bookwyrm:
- shelved October 30, 2020
- started reading July 26, 2016
- finished reading March 9, 2017

Note: not all the books have these two dates 26.6.2016 and 30.10.2020... I just happened to pick two that did. It seems to not always be the case. On goodreads, I added Masters of Doom and The Waves on March 1, 2017. On bookwyrm, they're shelved November 8, 2020 and January 4, 2021, respectively.

mouse-reeve:
Would you be comfortable attaching the csv here or emailing it to me (mousereeve@riseup.net)?

emiljacobsen:
[...]

PR #1221, opened 2021-07-14
Add shelved date, populate from Goodreads import

cincodenada:
`shelved_date` defaults to current datetime for new books, and uses the "Date Added" date from Goodreads imports. The shelf page now uses this field for its Shelved Date column, and shelfbooks are sorted by it first.

This may be part of a larger question, but I'm torn on what to do about timezones - I considered making shelved_date a DateField instead of DateTimeField, which I think it should be properly, as that's all we have on imports, but then when folks add a book in Bookwyrm, we'd tag it with whatever date it happened to be in UTC, which isn't necessarily what users would want.

That said, I'm not sure how we're displaying dates, so we may display that date anyway - but not storing the datetime a book was added makes it so we can't recover the "correct" date even if we do (at any point) have user timezone preferences.


bookwyrm/tests/importers/test_goodreads_import.py
```diff
@@ -13,6 +15,10 @@
 from bookwyrm.settings import DOMAIN
 
 
+def make_date(*args):
+    return datetime.datetime(*args, tzinfo=pytz.UTC)
```
cincodenada, line 19: I don't remember if we like free functions like this, but it seemed just laborious enough to use one to me. I'm not wedded to it tho.

bookwyrm/tests/importers/test_goodreads_import.py
```diff
@@ -130,22 +136,25 @@ def test_handle_imported_book(self):
...
         self.assertEqual(readthrough.book, self.book)
-        # I can't remember how to create dates and I don't want to look it up.
-        self.assertEqual(readthrough.start_date.year, 2020)
-        self.assertEqual(readthrough.start_date.month, 10)
-        self.assertEqual(readthrough.start_date.day, 21)
-        self.assertEqual(readthrough.finish_date.year, 2020)
-        self.assertEqual(readthrough.finish_date.month, 10)
-        self.assertEqual(readthrough.finish_date.day, 25)
+        self.assertEqual(readthrough.start_date, make_date(2020, 10, 21))
+        self.assertEqual(readthrough.finish_date, make_date(2020, 10, 25))
```
cincodenada, line 146: I was feeling ambitious and looked it up :smile:

bookwyrm/views/shelf.py
```diff
@@ -69,7 +69,8 @@ def get(self, request, username, shelf_identifier=None):
             reviews = privacy_filter(request.user, reviews)
 
         books = books.annotate(
-            rating=Subquery(reviews.values("rating")[:1])
+            rating=Subquery(reviews.values("rating")[:1]),
+            shelved_date=F("shelfbook__shelved_date"),
```
cincodenada, line 73: I'm not sure if is the best way to get at the `shelved_date`, I tried accessing it via `book.shelfbook__shelved_date` in the template to no success.

cincodenada:
I think this should resolve #1174

bookwyrm/tests/importers/test_goodreads_import.py
mouse-reeve: I haven't formulated a general opinion on it (seems fine?) but for tests I see absolutely no problem whatsoever. It's handy!

bookwyrm/tests/importers/test_goodreads_import.py
mouse-reeve: Yessssss thank you for being less lazy than me

bookwyrm/views/shelf.py
mouse-reeve: I think this is good, I don't think the templates can do that kind of advanced lookup, and it's cleaner to do it in the view anyway, imo

PR #1231, opened 2021-07-28
Automatically search remote endpoints if a local search is empty

mouse-reeve:
I've heard from users that they are confused about how to load books when they aren't in the catalog, so this change includes remote results by default if there are no results from a remote query. It's slow, but I think the clarity is worth it.


mouse-reeve:
@tofuwabohu I don't know if you have a feeling on this - you had good input earlier on doing the remote search less frequently.

Issue #1225, opened 2021-07-21
[Bug] Linebreak Not Recognized In Bio


seniorm0ment:
Single linebreaks are not recognized in profile descriptions, and acts as one paragraph.
If the user double linebreaks, it shows up as a double line break (as exoected).

mouse-reeve:
Statuses and bios are formatted with markdown, which treats single line breaks as ignorable whitespace, and double line breaks as paragraph breaks. I think this can be fixed by changing how css handles whitespace for the element.

PR #1234, opened 2021-07-31
Recognize profile summary linebreaks + whitespace

EvilDrPurple:
Fix for #1225 
Uses inline CSS to change the white-space attribute of the text. Code is injected around the string during the HTML sanitation process so that it avoids being detected and striped by the HTMLParser, while still preserving the integrity of the Parser. By using the "data" tag it ensures the code will remain in place, even if the Parser determines any HTML tags in the original string should be removed.

Please let me know if there's a better way to do this, if I've missed any unforeseen consequences by making this change, or if there's a potential risk that this may be abused. Thank you :purple_heart:

### DIFF
--- a/bookwyrm/static/css/bookwyrm.css
+++ b/bookwyrm/static/css/bookwyrm.css
@@ -76,6 +76,10 @@ body {
     white-space: pre-wrap !important;
 }
 
+.display-inline p {
+    display: inline !important;
+}
+
 /** Shelving
  ******************************************************************************/
 
--- a/bookwyrm/static/css/bookwyrm.css
+++ b/bookwyrm/static/css/bookwyrm.css
@@ -72,6 +72,10 @@ body {
     flex-grow: 1;
 }
 
+.preserve-whitespace p {
+    white-space: pre-wrap !important;
+}
+
 /** Shelving
  ******************************************************************************/
 
--- a/bookwyrm/templates/directory/user_card.html
+++ b/bookwyrm/templates/directory/user_card.html
@@ -18,7 +18,7 @@
             </div>
         </div>
 
-        <div class="preserve-whitespace">
+        <div class="display-inline">
         {% if user.summary %}
             {{ user.summary|to_markdown|safe|truncatechars_html:40 }}
         {% else %}&nbsp;{% endif %}
--- a/bookwyrm/templates/directory/user_card.html
+++ b/bookwyrm/templates/directory/user_card.html
@@ -18,7 +18,7 @@
             </div>
         </div>
 
-        <div>
+        <div class="preserve-whitespace">
         {% if user.summary %}
             {{ user.summary|to_markdown|safe|truncatechars_html:40 }}
         {% else %}&nbsp;{% endif %}
--- a/bookwyrm/templates/directory/user_card.html
+++ b/bookwyrm/templates/directory/user_card.html
@@ -20,7 +20,7 @@
 
         <div>
         {% if user.summary %}
-            {{ user.summary|to_markdown|safe|truncatechars_html:81 }}
+            {{ user.summary|to_markdown|safe|truncatechars_html:40 }}
         {% else %}&nbsp;{% endif %}
         </div>
     </div>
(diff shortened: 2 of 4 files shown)


bookwyrm/templates/directory/user_card.html
```diff
@@ -20,7 +20,7 @@
 
         <div>
         {% if user.summary %}
-            {{ user.summary|to_markdown|safe|truncatechars_html:40 }}
+            {{ user.summary|to_markdown|safe|truncatechars_html:81 }}
```
EvilDrPurple, line 23: I had to increase the truncatechars_html allowance, since on the profile cards that would display in the user search, the truncatechars_html was counting about 41 characters of the CSS bloc (despite it not being visible). This caused it to cut off the entire profile summary preview. Increasing the cap to 81 preserves the original preview limit of 40.

mouse-reeve:
I'm not sure it makes sense to modify the html on the fly, as opposed to changing the static css (https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/static/css/bookwyrm.css) that applies to the elements that wrap the injected html for statuses and bios.

EvilDrPurple:
Okay, I redid it and edited the CSS this time!
Also sorry for the mess of commits, I'm kinda brand new to Github :dizzy_face:

bookwyrm/static/css/bookwyrm.css
```diff
@@ -72,6 +72,10 @@ body {
     flex-grow: 1;
 }
 
+.preserve-whitespace p {
+    white-space: pre-wrap !important;
```
EvilDrPurple, line 76: Still unsure if it would be best to use pre-wrap here or break-spaces instead

mouse-reeve: I had to look up what the difference was! I think pre-wrap is better

bookwyrm/templates/directory/user_card.html
```diff
@@ -18,7 +18,7 @@
             </div>
         </div>
 
-        <div>
+        <div class="preserve-whitespace">
```
mouse-reeve, line 21: In this case, it makes sense to force-wrap the whitespace, since it's a limited preview in a card that is meant to be a standard size. A bio with a lot of line breaks could produce results like:

Screen Shot 2021-07-31 at 10 41 29 AM
 instead of:
Screen Shot 2021-07-31 at 10 41 49 AM

Since it's a preview, I think it's worth sacrificing the accuracy of the bio display.

bookwyrm/templates/user/layout.html
```diff
@@ -28,7 +28,7 @@ <h1 class="title">
         </div>
 
         {% if user.summary %}
-        <div class="column box has-background-white-bis content">
+        <div class="column box has-background-white-bis content preserve-whitespace">
             {{ user.summary|to_markdown|safe }}
```
mouse-reeve, line 32: The only tweak this needs is the `spaceless` django tag - that will remove the system-generated whitespace between html tags, which would artificially pad out the expected whitespace in the user bio. You'll just need to add `{% spaceless %}` below line 30, and `{% endspaceless %}` below line 33.

As is:
Screen Shot 2021-07-31 at 10 53 13 AM

With `spaceless`:
Screen Shot 2021-07-31 at 10 53 45 AM

mouse-reeve (commented):
Very minor changes, thank you for your work on this! And zero worries about having a clean commit history

bookwyrm/templates/user/layout.html
EvilDrPurple: Done! :smiley:

bookwyrm/templates/directory/user_card.html
EvilDrPurple: I ended up adding a new CSS tag "display-inline" since putting multiple blank lines creates a new <p> tag, adding an additional newline character. This could be abused in the current implementation as well, you can see here: https://bookwyrm.social/search?q=EvilDrPurple&type=user
By forcing the user cards to display inline, it will fix this workaround in the current implementation.

mouse-reeve: oh good point!

mouse-reeve:
Heck yeah

PR #1247, opened 2021-08-04
Add get_absolute_url to ImageField

### LINKED ISSUE #1246: External Storage: Image preview in status update has wrong URL
Following this message:

https://social.apreslanu.it/users/tk/statuses/106693318429554036

The image is unavailable, because its URL is `https://lire.boitam.euhttps//lire-boitam-eu.s3.fr-par.scw.cloud/images/covers/15c8e1f6-4df9-4045-ae6a-8248226a1791.jpeg`

It looks like `https://DOMAIN` has been added to the full image URL.

I think it might originate in this code, though I don't know what went wrong:

https://github.com/bookwyrm-social/bookwyrm/blob/6d82bafdc01d9bad856b41129a78b4fa92a15755/bookwyrm/models/fields.py#L352-L360

joachimesque:
Fixes #1246

I used Django's own code, if that's OK.

https://github.com/django/django/blob/910ecd1b8df7678f45c3d507dde6bcb1faafa243/django/core/files/storage.py#L341-L344


bookwyrm/models/fields.py
```diff
@@ -387,7 +387,8 @@ def set_activity_from_field(self, activity, instance):
         activity[key] = formatted
 
     def field_to_activity(self, value, alt=None):
-        return image_serializer(value, alt)
+        url = self.get_absolute_url(value)
+        return activitypub.Document(url=url, name=alt)
```
mouse-reeve, line 391: Okay I think I figured out the issue - `image_serializer` was only creating the `activitypub.Document` object if the image url exists. In this function, it's created even if the image is `None`. If you check whether the url is a falsey value and return None, I think that'll fix the issues.

bookwyrm/models/fields.py
```diff
@@ -423,6 +424,20 @@ def formfield(self, **kwargs):
             }
         )
 
+    # pylint: disable=no-self-use
+    def get_absolute_url(self, value):
+        """returns an absolute URL for the image"""
+        name = getattr(value, "name")
+        if name is None:
```
mouse-reeve, line 431: I think this is too limited a check - it looks like `name` may be an empty string. Maybe `if not name:`?

mouse-reeve (changes requested):

PR #1267, opened 2021-08-07
Link author in search

### LINKED ISSUE #1197: Search results should link the author's name to the author's page
### Summary
When I search for a book, I receive a list of titles. Each title is accompanied by the author's name. I actually assumed that the name would link to the author page, and I was surprised when it didn't.

### Considerations
For the sake of visual hierarchy, the link should not be styled like the book title. I don't think the normal link color would be good either, because it would look to similar to the title. The book title is more relevant, so it should draw attention to itself.

### Additional Context
the bookwyrm user interface for search results
###### The bookwyrm user interface for search results

justinabrahms:
Perhaps this should be a snippet given it's in two places? Not sure how important the structured data pieces are to you.

Fixes #1197

### DIFF
--- a/bookwyrm/templates/snippets/authors.html
+++ b/bookwyrm/templates/snippets/authors.html
@@ -19,7 +19,7 @@
         itemtype="https://schema.org/Thing"
     ><span
         itemprop="name"
-        >{{ author.name }}<span></a>{% if not forloop.last %}, {% elif remainder_count > 0 %}, {% blocktrans trimmed count counter=remainder_count %}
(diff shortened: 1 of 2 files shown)


justinabrahms:
note: I haven't tested this locally. Just pattern matched what was going on in the other file.

mouse-reeve:
This is a totally sensible fix based on context, but for complex, un-wise reasons, the "author" in the search panel is a plain text string rather than an object, so using `{{ author.name }}` won't work in this case. I need to refactor the way search results are returned to fix it. 

I would love to merge your span close tag fix though! Would you be able to just PR that change?

justinabrahms:
Just noticed this now that I have a working dev environment. :)   Will make another PR for the span fix.

PR #1335, opened 2021-08-27
Make webfinger match usernames in case insensitive manner

DeeUnderscore:
Currently, Bookwyrm serves webfinger data by looking up the username in a case sensitive manner. This means that when you look up a webfinger address without using an account's canonical case, it will not return the account's data. This PR makes these lookups case-insensitive. 

I'm not sure if they're *supposed* to be case-insensitive. [RFC7565](https://datatracker.ietf.org/doc/html/rfc7565) says this, which maybe applies to this situation: 

> If an application needs to compare two 'acct' URIs (e.g., for purposes of authentication and authorization), it MUST do so using case normalization and percent-encoding normalization as specified in Sections 6.2.2.1 and 6.2.2.2 of RFC3986.

In practice, however, Mastodon, Pleroma, and Misskey all serve lookups in a case-insensitive manner, and Misskey lowercases all outgoing lookups (which means that right now it can't fetch Bookwyrm accounts with usernames that contain capitals)

### DIFF
--- a/bookwyrm/tests/views/test_wellknown.py
+++ b/bookwyrm/tests/views/test_wellknown.py
@@ -51,6 +51,16 @@ class UserViews(TestCase):
         data = json.loads(result.getvalue())
         self.assertEqual(data["subject"], "acct:mouse@local.com")
 
+    def test_webfinger_case_sensitivty(self):
+        """ensure that webfinger queries are not case sensitive"""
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Sounds to me from RFC that they should be case insensitive - I think that's what "case normalization" means here? At any rate, I think we all agree that this is a good change _and_ has a test :) Thank you!

PR #1343, opened 2021-08-30
fix link formatter

reesporte:
resolves issue #1207

Instead of using regex to test for links and then replace them, I split the `content` by whitespace, then use the [django.core.validators.URLValidator](https://github.com/django/django/blob/3219dd3388c437b4bd869b76ddd43c9cdad05090/django/core/validators.py#L65) to test if the content is a link. If it is, I then use [urllib.parse.urlparse](https://docs.python.org/3/library/urllib.parse.html#url-parsing) to format it into a pretty link. I also made a function `_wrapped` which tests if the link is wrapped in parentheses, square brackets, or curly brackets, so it can wrap the link in the appropriate symbols during formatting. 

I also added some more test cases hitting those links (or similar ones) that @wakest put in the issue. 

`_wrapped` could maybe use some regex magic, but I'm no regex wizard

### DIFF
--- a/bookwyrm/tests/views/test_status.py
+++ b/bookwyrm/tests/views/test_status.py
@@ -307,14 +307,14 @@ class StatusViews(TestCase):
         self.assertEqual(
             views.status.format_links(url), '<a href="%s">tech.lgbt/@bookwyrm</a>' % url
         )
-        url = "users.speakeasy.net/~lion/nb/book.pdf"
+        url = "https://users.speakeasy.net/~lion/nb/book.pdf"
         self.assertEqual(
             views.status.format_links(url),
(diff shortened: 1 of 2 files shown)


mouse-reeve:
aaa thank you!!

reesporte:
okay so i was doing it a little wrong, i can work on this more later

the issue was i wasn't actually calling the validator, i've fixed that, but now i need to handle the wrapped links again

mouse-reeve:
sounds good, can you mark it as ready for review when you're ready?

mouse-reeve:
(or just comment! whatever's easier

reesporte:
ready for review capn

PR #1363, opened 2021-09-05
Create and view quotation page numbers

### LINKED ISSUE #1189: Quotes with page numbers
When I add a quote, I would like to be able to add page number(s).

I tried adding it to the quote itself, then to the comment, but none of it looks good. Example:

Quote with page number added as comment.

I would like the page number to be part of the source reference, like this:

> — In Our Clients' Shoes by Stephen E. Finn (p. 20)

mouse-reeve:
Fixes #1189


mouse-reeve:
@bookwyrm-social/translators this and an earlier PR from today have a couple small locale additions

PR #1388, opened 2021-09-08
Save status drafts in localstorage

mouse-reeve:
Fixes #475 
Fixes #259

- [x] Javascript utility to save input content
- [x] Incorporate appropriate text elements
- [x] Clear draft on post
- [x] Update shelve buttons
- [x] Show pending and error states

PR #1392, opened 2021-09-10
Re-adds error handling code written a while back

mouse-reeve:
@arkhi I'm re-adding the css and javascript error handling code you wrote a few months ago which was ahead of it's time. Do you feel good about me listing you as a co-author for the commits for this, or would you prefer to be credited in some other way?


arkhi:
Thanks for asking, however is easier for you, really. :)

Sorry I'm not following the activity much these days; is there any specific reason that made you change your mind?

mouse-reeve:
I'm using it a little differently than where it was originally - this is building on #1388, which will make posting a status happen in an ajax request, and it has a meaningful pending and error state, and needs feedback in the ui to indicate that a status was posted, otherwise it looks like the form just reset itself

PR #1448, opened 2021-09-22
Correctly use custom postgres ports

### LINKED ISSUE #1447: POSTGRES_PORT isn't respected during server setup
**Describe the bug**
`POSTGRES_PORT` isn't respected - `docker-compose up --build` starts the postgres instance on port 5432 regardless.

**To Reproduce**
1. Follow the [documentation for setting up a new instance](https://docs.joinbookwyrm.com/installing-in-production.html)
2. In `.env`, set `POSTGRES_PORT` to `5433` (and make the corresponding changes in `docker-compose.yml`)
3. Run `docker-compose up --build`

**Expected behavior**
The database is started on port 5433

**Actual behavior**
The database is started on port 5432

**Additional context**
I also tried setting the fallback to `5433` in `bookwyrm/settings.py` in case something was going wrong with the env propagation, but that didn't help either.
I was finally able to workaround by additionally setting `PGPORT` in `.env`

---

**Server:**
    - OS: Raspberry Pi OS (raspbian) 10.4
    - docker 20.10.8
    - docker-compose 1.29.2

mouse-reeve:
Fixes #1447


docker-compose.yml
```diff
@@ -20,8 +20,6 @@ services:
       - pgdata:/var/lib/postgresql/data
     networks:
       - main
-    ports:
-      - 5432:5432
```
Tak, line 24: You don't need to keep these, like in the redis services?

mouse-reeve: I found that it didn't have any impact on which port was used when I tested it.

PR #1481, opened 2021-10-01
Improve UI for adding authors to books

bibliotechy:
Update the book creation/edit form to allow users to enter authors in multiple input fields instead of comma separated single input.

Screen Shot 2021-09-30 at 9 35 10 PM

Adds a little bit of vanilla JS to dynamically add extra author fields when the use clicks on the "Add Anther Author" help text below the initial author field. 

Screen Shot 2021-09-30 at 9 35 04 PM

Updates the view to handle the multiple field values for persistence.

If you like the approach, I could try to generalize it for the other fields. Or at very least apply the same approach for them.


mouse-reeve:
yesss I was literally thinking about how to do this today! I've been working on this area of the app with https://github.com/bookwyrm-social/bookwyrm/pull/1477, which might cause some merge weirdness. It would be great to generalize this, language and publisher are also list fields, and there are one or two others that I'd like to turn into list fields in the future

mouse-reeve:
I think this is the right approach, and you could probably deep copy the html element rather than creating a new one, so that it naturally handles different input types and attributes?

bibliotechy:
Yup, cloning the target sounds perfect. I'll work on generalizing it for a few fields.

mouse-reeve:
Thank you, let me know if you run into issues with merge conflicts -- I moved the templates

bibliotechy:
@mouse-reeve  Sorry about the long delay in getting back to this. Work got busy and I just didn't have the spare cycles.

Refactored the javascript into a  generic function so that one can pass in an input's id and it clones that input inside the same div.

One addition I made is to add a screenreader-only label for each input that is dynamically added. WAVE and Firefox accessibility tools complain if each input doesn't have a label, and you can't re-use a label for multiple inputs.

Will try to extend this to Publishers and Languages in another PR, but they were different enough from Authors that it felt worth it to split out.

bookwyrm/templates/book/edit/edit_book_form.html
```diff
@@ -119,10 +119,16 @@ <h2 class="title is-4">{% trans "Authors" %}</h2>
...
+                    <label class="label">{% trans "Add Authors:" %}</label>
+                    {% for author in add_author %}
+                    <label class="label is-sr-only" for="id_add_author{% if not forloop.first %}-{{forloop.counter}}{% endif %}">{% trans "Add Author" %}</label>
+                    <input class="input" type="text" name="add_author" id="id_add_author{% if not forloop.first %}-{{forloop.counter}}{% endif %}" placeholder="{% trans 'Jane Doe' %}" value="{{ author }}" {% if confirm_mode %}readonly{% endif %}>
+                    {% empty %}
+                    <label class="label is-sr-only" for="id_add_author">{% trans "Add Author" %}</label>
+                    <input class="input" type="text" name="add_author" id="id_add_author" placeholder="{% trans 'Jane Doe' %}" value="{{ author }}" {% if confirm_mode %}readonly{% endif %}>
+                    {% endfor %}
                 </div>
+                <span class="help"><a id="another_author_field" onclick="BookWyrm.duplicateInput('id_add_author')">{% trans "Add Another Author" %}</a></span>
```
mouse-reeve, line 131: I think this makes more semantic as a `<button type="button" class="button is-small" ...>` than an `a`

mouse-reeve: It would be more consistent with the rest of the javascript to use a data attribute or class to indicate that this is a duplication button, and add the listener in the javascript file, rather than using `onclick`. Something like:

``` javascript
document.querySelectorAll('[data-duplicate]')
    .forEach(t => t.addEventListener('click', this.duplicateInput.bind(this)));
```

mouse-reeve (changes requested):
This works so well! Just some small changes for internal consistency. Thanks so much :)

bookwyrm/templates/book/edit/edit_book_form.html
bibliotechy: That makes total sense. I can get to it in the next couple of days.

bibliotechy:
@mouse-reeve OK, updated the JS as per request, and switched to a button.

mouse-reeve (approved):
Yay! This works beautifully

Issue #1318, opened 2021-08-21
List curation option - "Limited"/"Group"


hughrun:
**Is your feature request related to a problem? Please describe.**

_Scenario:_

Alex wants to create a list that both Alex and Elham can add books to directly, but nobody else. Currently the only option available is `curated`, but this allows _anyone_ to suggest books and does not allow Elham to add them directly.

**Describe the solution you'd like**
I would like a fourth option called "Limited" or "Group" or something similar, which allows the list owner to add one or more other users as allowed to add books, and prevents anyone else. Basically it's like `public` for the selected users and like `Closed` for everyone else.

**Describe alternatives you've considered**
A possible alternative would be an option to allow anyone a user is following or any followers to add items to a list, however that makes assumptions about the relationship between followers and followees.

**Additional context**

hughrun:
I'd like to work on this. I'm new to Django but I think I can see how this would be done.

mouse-reeve:
Go for it! Let me know if you have any questions

mouse-reeve:
Wow, thank you for all this thought and detail! This rules.

The first thing that comes to mind for this is whether/how to support cross-instance teams. For a team to include a user on a remote instance, there would need to be an activitypub representation of the list members (maybe an OrderedCollection associated with a field on the list object?). In the bigger picture, this would be good, but for now it would be a lot simpler keep the scope of this work just to teams within an instance and not have to handle federation. 

As to questions:
> Q1: is there a better way to design this dropdown, or a better place to put it on the user profile page?

I do think this makes sense to have on the profile page when the viewer has team-managed lists, yeah. I'm not sure where the perfect right place is, but I think this is basically solid

> Q2: Currently after adding a user to a list, the active user is redirected to the list - should they be returned to the user page they were on?

I think redirecting to the list is good.

> Q3: should this be placed in a different location on the List page? & Q4: should it be placed on a different page?

I like your idea to have a separate page from the list itself that shows the team members and has the interface for removing members

> Q5: should team members be allowed to delete books from the list?

I'm leaning to yes, let any member remove a book from the list, but it's an opinion weakly held :)

> Q6: is "Team" the right word to use for this?

Ah, wording, my number one nemesis. My one concern is that it'd be good to keep the phrasing as consistent as possible. Right now both "list members," "team member," and "list team" are in use. I'm not sure if I like "list team" or "list members" better, but I guess "members" is a little more vague?

mouse-reeve:
[...]

Issue #170, opened 2020-05-12
Let users create groups


mouse-reeve:
Idea: A user can create a public or private/invite-only group of other users and post content only to that group. Content posted to the group would show up in the feeds of group members, and in the group's dedicated page.

Use cases:
 - a book club or group read
 - a small circle of friends who want to share content privately with each other
 - a public reading challenge (like, read a book a week, or whatever)
 - a public interest group that wants to curate their content in one place or have less visibility by not showing up in the public feed

loganmhb:
I would be interested in working on this if you are looking for help! (I'm particularly interested in the private-book-club version, but it makes sense to design them together probably.)

rkingett:
I'm interested in this as well.

mouse-reeve:
definitely looking for help on this, go for it :)

loganmhb:
I'm reading up on ActivityPub to try to get a sense of how this would work -- sketching out some ideas here to see if they make sense.

I see that there is an ActivityPub "Group" actor type which it probably makes sense to use for this. Then activities which are conducted in the context of a group (which could be any type of activity I think? review, comment, reading update, etc -- though there are plenty of questions about how the UI for this should work) can be directed to the group's members via a `members` collection.

(Or maybe it makes sense to simply use the standard-defined `followers` collection, at least for public groups?)

So a reading update in the context of a group could look something like this:
```
{
  '@context': 'https://www.w3.org/ns/activitystreams',
  'id': 'https://your.domain.here/user/logan/status/3/activity',
  'type': 'Create', 'actor': 'https://your.domain.here/user/logan',
  'published': '2020-06-18T13:18:50.288697+00:00',
  'to': ['https://your.domain.here/groups/my_private_bookclub/members'],
  'object': {
    ...
  }
}
```

(I am not sure how verifying group membership in order to access content would work across federated instances. I suppose when a user requests a post that is private to a group it would be necessary to ask the server on which the group exists if they are a member, or something along those lines?)

I think this data model would support use cases like "mark this book as read, tell my followers, and also publish it to this group I'm in", though I'm not sure how you would make this work smoothly in the UI -- it would probably be clunky to ask for a context every time a user creates an activity.

mouse-reeve:
this all sounds right to me (especially the part about the UI being challenging). I think you're running into a lot of the same questions around post privacy, which hasn't been implemented yet -- "post to group" is, I think, essentially in the same list of options as "public" and "followers only"

[...]

PR #1490, opened 2021-10-03
Bookwyrm groups

hughrun:
Wow, this turned out to be bigger than I expected. Apologies, if I make any in future I'll try to break the changes up a bit.

This PR relates to:

#1318 
#170

This PR creates:

1. `Group`s that a user can join via a `GroupMembership`
2. A `group` curation option for Book Lists
3. A `group` value for lists, so that lists can be associated with a group for curation purposes

Groups do not currently interact with ActivityPub, but could. I am not confident to implement this functionality at this stage, hence not including it, but that's the only reason. I'd be happy to look into this for a later PR. Hopefully I haven't accidentally inluded any AP functionality where it shouldn't be.

## Outstanding problems/questions to resolve/things to note

1. I can't work out how to adjust `privacy_filter` so that Group members can see groups they wouldn't otherwise see (i.e. groups with privacy levels of `direct`, or of `followers` when the viewer isn't a follower), whilst ensuring that the privacy filter nevertheless works in all other cases (e.g. when a user is blocked). Currently therefore `privacy_filter` is **NOT** applied to group list display on group pages (however `raise_visible_to_user` will still prevent anyone from seeing a list they shouldn't have access to)
2. I re-used some of the user suggestions code for new users, for searching for potential group members. I have, however, created a new filter function to exlude non-local users, for use until groups are AP compliant.
3. I am assuming that I should *not* commit anything in the `migrations` directory, and that maintainers will run `makemigrations` where appropriate: let me know if this assumption is wrong.
4. I am not sure what to do about group members who block each other. Currently:
   1. If the `group.user` has blocked a user, the second user cannot access the group page but can access any lists assigned to the group if they are owned by someone else, and can add & delete items on those lists;
   2. If a member of a group who is not `group.user` creates a list, other members who are blocked by that user cannot see the list (but *can* see it listed on the group page)
   3. Users who are blocked cannot invite the blocker to join a group because they won't appear in the user-inviting search results.

   The second case could be fixed if `privacy_filter` worked for groups and group lists. The first case may not matter much because a group owner can simply remove any member from the group, but it is a bit weird and not really ideal.
5. I get a `Object of type LocalStream is not JSON serializable` error when attempting to unblock a user in my development environment. I'm not sure whether this is an actual bug in Bookwyrm, a problem with my setup, or if I somehow broke something in the codebase. This may need to be checked!
6. I'm guessing you probably want some tests - I created some new classes and filters. What are the expectations/requirements around this?

## Groups

1. any user may create a `Group`.
2. this `group.user` may send a `GroupMemberInvitation` to any other user as long as they are not blocked either way
3. the invitee must accept the `GroupMemberInvitation` to create a `GroupMembership`.
4. members (users with a `GroupMembership` connecting them to a `Group`) can:
   1. create a `List` with a `curation` value of `group` and assign the List to this group
   2. See all `List`s associated with the group, on the group home page
   3. see the list page for any `List` associated with the group, regardless of its privacy value (unless the `list.user` has blocked them)
   4. add books to and remove books from any List associated with the group (unless the `list.user` has blocked them)
5. Groups have their own page visible to all users, at `/group/{id}`

## Lists

1. Lists now have a new `curation` value of `group`
2. Group-curated lists must be associated with one (and only one) `Group`
3. Any member of the group can add and remove books to/from the list, unless blocked by the `list.user`.
4. Lists can have their curation value changed at any point: either by changing it to a non-group curation (closed, curated, open), or by changing the group it is associated with.
5. If a list's curation value is changed, group members no longer have access (unless curation is "public")
6. If a user is removed from the group, their group-curated lists will be changed to `closed` curation with a `group` of `None`.

## Notifications

Group members are notified when:

* a new member joins
* a member leaves
* a member is removed
* a book is added to a group-curated list

`group.user`s (group owners) are notified when:

* a user accepts their invitation to join

Users are notified when:

* a `group.user` invites them to join a group

## Abuse and privacy

I have attempted to think about the potential for abuse and mitigate it, however I've probably missed something, and as noted above there are some outstanding problems to resolve.

1. Users must be invited to join a group: you cannot simply make someone a member of a group without them opting in
2. Group curated lists are removed from a group when the `list.user` leaves the group and given `curation=private`
3. Users can leave voluntarily or be removed by the `group.owner`
4. The same privacy levels apply to groups and group-curated lists as for any other object. This means if I group-curate a list with privacy level `direct`, only group members will be able to access it.
5. `privacy_filter` needs to be adjusted, or an alternative created for groups, so that the ability to see the existence of a list is also determined by privacy levels (e.g. `direct` lists don't appear on a group's list of Lists)

## Design

As much as possible all changes are based on existing code and attempt to integrate seamlessly. I'm open to presenting any of the new options differently.

group_list_creation

user_groups

group_homepage

add_book_to_shared_list

notifications


mouse-reeve:
The automated checks aren't as bad as they look -- if you run `./bw-black` that will fix the Python formatting check and a lot of the things Pylint is upset about, and the tests aren't passing because there isn't a migration file yet for the new models (`./bw-dev makemigrations`)

mouse-reeve:
To address some of your questions (I'm going have to think more about the privacy controls stuff):

> * I re-used some of the user suggestions code for new users, for searching for potential group members. I have, however, created a new filter function to exlude non-local users, for use until groups are AP compliant.

Ah this is a great idea! I think instead of creating another function, you could modify `get_suggestions` to take a `**filters` kwargs param and then pass in `local=True` as a kwarg when you call it. Something like:
``` python
 def get_suggestions(self, user, **filters):
    ...
    user = models.User.objects.get(
         id=user_id, is_active=True, bookwyrm_user=True, **filters
    )
    ...
```

> * I am assuming that I should _not_ commit anything in the `migrations` directory, and that maintainers will run `makemigrations` where appropriate: let me know if this assumption is wrong.

Go ahead and commit the migration. Thanks for checking <3


> * I get a `Object of type LocalStream is not JSON serializable` error when attempting to unblock a user in my development environment. I'm not sure whether this is an actual bug in Bookwyrm, a problem with my setup, or if I somehow broke something in the codebase. This may need to be checked!

That sounds like a bug in BookWyrm, I'll take a look

> * I'm guessing you probably want some tests - I created some new classes and filters. What are the expectations/requirements around this?

I don't have a specific set of expectations around tests, but in broad strokes, for models, I like to test special logic in overwritten functions (like custom `save`), and for views, the minimum would be a test that checks if the view renders with valid html (for example: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/tests/views/preferences/test_block.py#L42 -- this test just loads the view and renders the template).

mouse-reeve:
> * I can't work out how to adjust `privacy_filter` so that Group members can see groups they wouldn't otherwise see (i.e. groups with privacy levels of `direct`, or of `followers` when the viewer isn't a follower), whilst ensuring that the privacy filter nevertheless works in all other cases (e.g. when a user is blocked). Currently therefore `privacy_filter` is **NOT** applied to group list display on group pages (however `raise_visible_to_user` will still prevent anyone from seeing a list they shouldn't have access to)

It sounds like the correct behavior here is that the privacy of a group is the privacy settings OR group membership - so for example, a followers-only group is visible to you if you follow the creator of the group or you are a member of the group. This is simple enough to implement with Q in the queryset, but not simple to incorporate into the general-purpose function that handles lots of different models. I think the best approach here would be to make `privacy_filter` a class function on models, and subclass it for groups, like how `raise_visible_to_user` works. I can do this refactoring bit if you'd like!

> * I am not sure what to do about group members who block each other. Currently:
>   
>   1. If the `group.user` has blocked a user, the second user cannot access the group page but can access any lists assigned to the group if they are owned by someone else, and can add & delete items on those lists;

I'm trying to think about groups more holistically, since there are a lot of ways they could be used in the future in addition to controlling list curation. For example, posting to a group (like for people in a book club). I think given that, it would make sense for a user to be automatically removed from a group if the group owner blocks them? Otherwise the privacy controls just seem very complex. The blocked user would still be able to see lists owned by other users in the group, but not edit them. I guess I think of blocking as a pretty severe step, so I err on the side of limiting access more rather than less.

hughrun:
> It sounds like the correct behavior here is that the privacy of a group is the privacy settings OR group membership - so for example, a followers-only group is visible to you if you follow the creator of the group or you are a member of the group. This is simple enough to implement with Q in the queryset, but not simple to incorporate into the general-purpose function that handles lots of different models. I think the best approach here would be to make `privacy_filter` a class function on models, and subclass it for groups, like how `raise_visible_to_user` works. I can do this refactoring bit if you'd like!

Yes that's exactly it! If you are able to make this change that would be great - I'm new to Django and the Q stuff still confuses me.

> I'm trying to think about groups more holistically, since there are a lot of ways they could be used in the future in addition to controlling list curation. For example, posting to a group (like for people in a book club). I think given that, it would make sense for a user to be automatically removed from a group if the group owner blocks them? Otherwise the privacy controls just seem very complex. The blocked user would still be able to see lists owned by other users in the group, but not edit them. I guess I think of blocking as a pretty severe step, so I err on the side of limiting access more rather than less.

This seems right. I suppose the complexity I was trying to think through is what happens if group members who are not the group owner block each other. But really that is a social problem, not a technical problem - it's probably not cleanly solvable with a technical rule. If the group owner blocks someone then I agree it's a reasonable assumption that they also want them expelled from their group. I was also thinking ahead to the possibility of uh "collectively owned" groups, but perhaps that's something to worry about later, if that's ever implemented. I'm happy to look at adding something to remove a user from relevant groups if the owner blocks them - I believe it should be impossible to add a blocked user, or for a blocked user to _see_ the group, already.

Oh and apologies for all the linting commits gunking this PR up. I forgot that when I push to my fork it updates the PR. I also finally remembered I can run pytest in the Docker container so now I can see why my tests are all failing locally.

mouse-reeve:
Thank you so much for all your work on this! It's a big task. Would it be more helpful to you to get feedback/testing/code review comments as you go, or would you prefer to get to a stopping place and then request feedback then?

hughrun:
> Thank you so much for all your work on this! It's a big task. Would it be more helpful to you to get feedback/testing/code review comments as you go, or would you prefer to get to a stopping place and then request feedback then?

I am ready for any feedback now! Thanks for your patience. Happy to make any changes to make this work more sensibly if necessary.

I have added some tests, which work locally but I've obviously done something wrong because the redis connection is having issues here in the automated runs and most of them are failing. Any advice would be great regarding that.

Possibly related or possibly irrelevant, when I run pytest locally the `test_remote_id` test in `test_shelf_model` fails due to a mismatch on the url protocols (i.e. https vs http). As far as I know this has nothing to do with my additions so I'm not sure what that's about, and it doesn't appear to be failing here.

I believe other than that, everything should be working as discussed in the original issue and in the thread here. Specifically:

* user is removed from group and any group-curated lists they own are changed to closed curation with no group
* `privacy_filter` now works correctly via over-riding `followers_filter` and `direct_filter` in both `List` and `Group`. That is, group members can always see group-curated lists and the group itself
* I finally understood what eslint was telling me to do!

hughrun:
Thanks @mouse-reeve for reviewing this, your comments all make sense, with the exception of the `suggested_users` file: as per my comment I don't understand your suggestion, though I can see a different way to make it more efficient.

I should be able to clean all this up in the next couple of days, please let me know if you spot any other problems.

mouse-reeve:
I'm going through and using the groups feature -- first of all, this rules so much. Very excited about it, thank you, I really appreciate how thorough and diligent your work has been on this!

Some usability things that occurred to me:

## Lists
1. The first thing I did was go to create a list and select the group curation type. Since I hadn't created any groups, I couldn't add one, and it wasn't clear how to. I think either the group options should be disabled and include a link on how to create a group, or there should be an option to create a group on the fly (the latter might be a good piece of future follow up work though, since it's sounds complicated).
Screen Shot 2021-10-13 at 11 08 07 AM

## Add user
2. The edit group form is missing in the add-users state:
Screen Shot 2021-10-13 at 11 14 56 AM
3. I think either the suggested users should be in the right-hand column, or the search panel should be in the main region and not use the columns in the add user flow? Using both makes the search element and the results element feel disjointed to me:
Screen Shot 2021-10-13 at 11 21 14 AM
4. It would be nice to have a confirmation modal before removing a user (as someone who is terrible at clicking the right things) (also something that doesn't need to be part of this PR if it would add a lot of work, imo)
5. On removal, I think it would make more sense for there to either be some kind of message (like, "Removed"), or have the remove form submit synchronously so that it's clear that a user is no longer a member. The state with no button and no text felt unclear to me:
Screen Shot 2021-10-13 at 11 25 02 AM
6. Maybe just "Leave group" instead of "Remove self from group" for the sake of being concise?

## User group page
7. Using the inline_form component would make this a little more consistent with the other forms:
Screen Shot 2021-10-13 at 11 36 37 AM
 
But fwiw these are all last-bit-of-polish changes and I hope you're patting yourself on the back.

hughrun:
@mouse-reeve LOL so clearly I need some help with tests, I've managed to break ALL of them.

Other than that:

- all suggestions from your code review have been resolved
- all suggestions [in your comment above](https://github.com/bookwyrm-social/bookwyrm/pull/1490#issuecomment-942609222) have been resolved
- users can now delete their groups (whoops, we both missed that one) and any group lists are changed to closed curation with the group removed

Looking forward to seeing this in production once we've sorted out the tests.

marek-lach:
Some fediverse applications implement the URL for a group as "https://tavern.town/`groups`/sneakerheads". Bookwyrm could probably accept _both_ `/group`, and also `/groups` as a valid group URL, to be compatible in this regard with as many fediverse software as possible?

Awesome feature tho,

PR #1492, opened 2021-10-03
fix for #1491 new shelf creation not displaying name field

### LINKED ISSUE #1491: Creating a new shelf when a default shelf selected does not provide name option
**Describe the bug**
The `Create Shelf` form does not display a `name` field if the currently selected shelf is one of of the default shelves (`All books`, `To Read`, `Currently Reading`, `Read`). When you save the new shelf it is created, but with a duplicate name of whatever you had selected at the time of creation.

**To Reproduce**
1. Go to 'user/{yourname}/books/'
2. Click on 'Create Shelf'
3. See only the 'Description' field
4. fill in description and save shelf
5. Notice that you have created a new shelf with the duplicate name 'All books`

**Expected behavior**
The `Name` field should appear when creating a new shelf, regardless of which shelf is currently selected.

**Screenshots**
Create shelf with default shelf selected:
bad

Create shelf with non-default existing shelf selected:

good

**Instance**
bookwyrm.social

**Additional context**
I haven't previously noticed there is a description field for shelves, maybe this is new?

---

    - OS: MacOS
    - Browser Firefox
    - Version 92.0

hughrun:
When selecting "Create shelf" whilst a default shelf is currently selected, the option to create a new name is suppressed and the new list will take on the same name as the selected default list. This PR changes the behaviour inherited from the "Edit shelf" logic such that `editable` is always true when _creating_ shelves.

fixes #1491


mouse-reeve:
Perfect! Thank you :)

Issue #1345, opened 2021-08-31
About the German Translation of "Finished Reading"


oragegu:
**Describe the bug**
I have noticed that for "Finished Reading", it is translated in German as "lesen abschliessen". 

It sounds a bit strange to my ears. 

What do you people think of "zu Ende gelesen" instead?

**Code Reference**
https://github.com/bookwyrm-social/bookwyrm/blob/8d23f1d35613dad6e605e1e41725c79f3f30b7be/locale/de_DE/LC_MESSAGES/django.po#L3093

mouse-reeve:
Thank you! Tagging @tofuwabohu

tofuwabohu:
I agree that this sounds a bit weird. "zu Ende gelesen" implies that the book has actually been finished (read through), while "Lesen abschließen" was my idea of leaving it open whether the book was just put aside. Since the original "Finished reading" also implies that the book was read through, it should probably be changed.  
I also think I remember there was some talk about stopping reading a book without finishing it which is currently not supported, before I change the translation, is there somethin that's going to change (soon)?

mouse-reeve:
I see! It makes more sense for the button to imply that you read the whole book. I should prioritize adding an "abandoned" feature - I added a button to remove books from their current shelf, which to me seems equivalent, but I think it's both not clear in the UI and not doing quite what people want

PR #1500, opened 2021-10-04
Changing the "Lesen abschließen" as discussed in issues

oragegu:
I tried `./bw-dev compilemessages -l de`, but it seemed not to do anything to the German .mo file.

Link of discussion: https://github.com/bookwyrm-social/bookwyrm/issues/1345


oragegu:
Requested along with the french translation

PR #1511, opened 2021-10-06
Oulipo localizations

mouse-reeve:
(cc @reesporte )


mouse-reeve:
trying this through crowdin instead

PR #1529, opened 2021-10-13
Add sorting to shelf. Use table-sort-header

### LINKED ISSUE #1519: Book sorting in library (your books)
I have more than 300 books and the way they're sorted by default (guess it's by rating) doesn't make any sense to me personally.
For me sorting by date added would make more sense but it might well be by author for others or other criteria.

scientia-ac-labore:
# Description
This PR add sorting to shelves. I used template from announcements. Sorting works for title, author, shelved, started, finished, rating fields. Default shelved. 
Close #1519
P.S pylint and black without problem on my side now


scientia-ac-labore:
I have made all changes. And I have noticed one problem if I annotate authors using F() in line 77, if book has two or more authors there was books occur again. So rewrite this using Subquery.

PR #1575, opened 2021-10-26
Move actor model attrs into their own abstract class

mouse-reeve:


mouse-reeve:
I'm closing this because it's too old; if this task is worth revisiting, it would be easier to start fresh.

PR #1589, opened 2021-11-05
match page title to active shelf

### LINKED ISSUE #1586: Make Page Title Match Shelf Name
Rather minor suggestion, but it came to my attention when I was managing three of my shelves (To Read, Currently Reading, Audiobooks) across multiple tabs. In this screenshot I have each of the three open in separate tabs but they're indistinguishable. 

image

All of the page titles say "Your Books" when I think it'd make more sense to have it be the shelf title so people can identify tabs easier.

hughrun:
Resolves #1586

### DIFF
--- a/bookwyrm/templates/shelf/shelf.html
+++ b/bookwyrm/templates/shelf/shelf.html
@@ -5,7 +5,7 @@
 {% load i18n %}
 
 {% block title %}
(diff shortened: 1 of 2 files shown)


mouse-reeve:
thank you!!

PR #1591, opened 2021-11-06
Misc/small template fixes

nycterent:
fixes some mistypes and removes quotes from smtp hostname - solves connection problem.


mouse-reeve:
ah sorry about this merge weirdness! It looks like you made this change to the configuration in the production branch, so it's getting confused about the branch differences. I think the files you want to change are available in the `main` branch, so if you're down to make the fixes to the files there, that would work as expected.

I really appreciate the fixes

PR #1605, opened 2021-11-14
Don't delete non-form data when editing authors 

### LINKED ISSUE #1584: Editing an author deletes author data not in form
**Describe the bug**
Editing an author saves the form data as an update. `isni` and `viaf_id` are not part of the form, so these values are deleted when the author details are edited.

**To Reproduce**
1. Go to 'author page'
2. Click on 'Edit'
3. Make a change
4. Save
5. Most users will not see any change because these values are not displayed, however they have been removed.

**Expected behavior**
Only values that have changed should change. ISNI and VIAF data should be retained on edits.

**Instance**
Development site but it presumably would be replicated on any production server.

**Additional context**
I only noticed this because I am working on a PR that will display a link to the ISNI record, and it disappears whenever I edit an author.

hughrun:
Fixes #1584

This is a temporary fix. As @mouse-reeve has suggested, ultimately it would be good to re-import data from one or more of the linked data sources if there is anything missing, each time the author form is saved.

### DIFF
--- a/bookwyrm/forms.py
+++ b/bookwyrm/forms.py
@@ -210,6 +210,7 @@ class AuthorForm(CustomForm):
             "born",
             "died",
             "openlibrary_key",
+            "inventaire_id",
             "librarything_key",
             "goodreads_key",
         ]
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you!

PR #1608, opened 2021-11-15
refactor shelf activity on book page

### LINKED ISSUE #1582: date unpopulated when starting reading a book
**Describe the bug**
> A clear and concise description of what the bug is.

moving a book to the default/auto-generated "currently reading" shelf doesn't populate a start date.

**To Reproduce**
1. Go to any book in your "to read" shelf
2. mark the book as currently reading, somehow[a]; eg:
    - 2.1) Click on "move book" drop-down under the book's description, choose "currently reading"
    - 2.2) _OR_ click on "start reading" button under the books cover art
3. Come back another day to your "currently reading" shelf, to use "update progress"
4. **Bug/unexpected**: notice the book seems to have undefined data for the start date

[a]: this choose-your-own-journey option for this step might be key to the bug being reproducible; I'm not sure which approach (or if both) causes the bug. I just know I keep seeing the end result of the bug (step 4). Screenshot attached of the two options for this step.

Screenshot showing book on to-read list and the UI for marking it as currently reading

**Expected behavior**
> A clear and concise description of what you expected to happen.

Step 4 should have revealed an already populated date value of the day that step 2 occurred.

**Screenshots**
> If applicable, add screenshots to help explain your problem.

**Instance**
> On which BookWyrm instance did you encounter this problem.

bookwyrm.social

**Additional context**
> Add any other context about the problem here.

I'm filing this as a bug because I could swear this used to work as I'm describing expected. But I could be wrong, in which case I'd say this maybe is a legitimate feature request.

---

    - Device: Google pixel 5
    - OS: Android 12
    - Browser: Google Chrome Beta
    - Version: `96.0.4664.27` _(that's browsers version.. if this prompt was meant to ask for the web version of Bookwyrm being served to me, then I'm not sure - I'm not seeing a tag anywhere in the footer of the UI or anything)_

hughrun:
- disallow moving from custom shelf to a reading status shelf with `shelf_selector`
- always use `shelve_button` for moving books from a reading status shelf
- redesign shelf information as a list of boxes

Fixes #1582 

If the book is on a shelf, it will now display as shown below. Note that 

1. the heading only appears if the book has been shelved, and
2. it is not possible to move directly from a custom shelf to a reading status shelf, as seen in the image.

I haven't touched the reading status button so you can still add to a custom shelf using that. I find this a little confusing because it adds it rather than moving it, but that's not in scope for this fix so it's a conversation we can have, if necessary, some other time.

Screen Shot 2021-11-15 at 8 53 01 pm

### DIFF
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -90,6 +90,11 @@ let StatusCache = new class {
             trigger.removeAttribute('disabled');
         })
         .then(response => {
+            if (response.headers.get("forceReload")) {
+                BookWyrm.addRemoveClass(form, 'is-processing', true);
+                trigger.setAttribute('disabled', null);
+                return location.reload();
+            }
             if (!response.ok) {
                 throw new Error();
             }
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -93,8 +93,10 @@ let StatusCache = new class {
             if (response.headers.get("forceReload")) {
                 BookWyrm.addRemoveClass(form, 'is-processing', true);
                 trigger.setAttribute('disabled', null);
+
                 return location.reload();
             }
+
             if (!response.ok) {
                 throw new Error();
             }
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -194,7 +194,7 @@ let StatusCache = new class {
             .forEach(item => BookWyrm.addRemoveClass(item, "is-hidden", false));
 
         // Remove existing disabled states
-        // BUG: this affects all shelves, not just shelving status shelves
+
         button.querySelectorAll("[data-shelf-dropdown-identifier] button")
             .forEach(item => item.disabled = false);
 
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -192,6 +192,7 @@ let StatusCache = new class {
             .forEach(item => BookWyrm.addRemoveClass(item, "is-hidden", false));
 
         // Remove existing disabled states
+        // BUG: this affects all shelves, not just shelving status shelves
         button.querySelectorAll("[data-shelf-dropdown-identifier] button")
             .forEach(item => item.disabled = false);
 
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -74,7 +74,7 @@ let StatusCache = new class {
 
         // This allows the form to submit in the old fashioned way if there's a problem
 
-        if (!trigger || !form || response.headers.get("forceReload")) {
+        if (!trigger || !form) {
             return;
         }
 
@@ -90,7 +90,6 @@ let StatusCache = new class {
             trigger.removeAttribute('disabled');
         })
         .then(response => {
-
             if (!response.ok) {
                 throw new Error();
             }
--- a/bookwyrm/static/js/status_cache.js
+++ b/bookwyrm/static/js/status_cache.js
@@ -74,7 +74,7 @@ let StatusCache = new class {
 
         // This allows the form to submit in the old fashioned way if there's a problem
 
-        if (!trigger || !form) {
+        if (!trigger || !form || response.headers.get("forceReload")) {
             return;
         }
 
@@ -90,12 +90,6 @@ let StatusCache = new class {
             trigger.removeAttribute('disabled');
         })
         .then(response => {
-            if (response.headers.get("forceReload")) {
-                BookWyrm.addRemoveClass(form, 'is-processing', true);
-                trigger.setAttribute('disabled', null);
-
-                return location.reload();
-            }
 
             if (!response.ok) {
                 throw new Error();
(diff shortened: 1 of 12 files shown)


bookwyrm/templatetags/bookwyrm_tags.py
```diff
@@ -41,6 +41,17 @@ def get_book_description(book):
     return book.description or book.parent_work.description
 
 
+@register.filter(name="is_shelf_type")
+def is_shelf_type(current_shelf, shelf_type):
+    """is this shelf a readthrough shelf?"""
+    readthrough = current_shelf in ["to-read", "reading", "read"]
```
mouse-reeve, line 47: A more future-proof way to check this would be to check the boolean field `shelf.editable`

mouse-reeve (changes requested):
Looks really nice, what a relief to finally have that button a normal size

bookwyrm/templatetags/bookwyrm_tags.py
hughrun: Ohh yeah that is a lot simpler lol.

hughrun:
I just noticed there's a problem with the shelve button if the first `active_shelf` is editable. I'll take another look at this tonight.

hughrun:
After considering how the "move" button works in the shelves view (as opposed to the book view) I changed my approach to this. 

1. "Move" button now triggers modals for read status shelves just like `shelve_button` does
2. "Move" button always removes the shelf you are moving from. `shelve_button`, on the contrary, adds shelves
3. Fixes bug where `shelve_button`offered to remove the book from the last shelve it was added to. Now `active_shelf` is always a read status shelf
4. Fixes bug in Firefox where form state persists across page reloads
5. Improves styles in Book view

I noticed #1626 whilst working on this however I can't work out how to fix it so I've logged it as a separate issue.

bookwyrm/static/js/status_cache.js
```diff
@@ -90,6 +90,13 @@ let StatusCache = new class {
             trigger.removeAttribute('disabled');
         })
         .then(response => {
+            if (response.headers.get("forceReload")) {
```
mouse-reeve, line 93: I wonder if it's possible to toggle whether the button uses ajax or a regular submit, rather than doing the ajax and then forcing a reload? It checks if the `submit-status` class is present on the form to attach the javascript listener. If that class is absent, it's just a regular old form submission.

mouse-reeve (changes requested):
I extremely appreciate you digging through this surprisingly complicated bit of code!

I'm running into an issue with the "Remove" button when used from the "all books" view that aggregates all the shelves and the book is on a custom shelf:
```
ValueError at /unshelve/
Field 'id' expected a number but got ''.
...
app/bookwyrm/views/shelf/shelf_actions.py, line 99, in unshelve
```

bookwyrm/static/js/status_cache.js
hughrun: That would certainly feel less like a hacky workaround!

hughrun:
> I extremely appreciate you digging through this surprisingly complicated bit of code!
> 
> I'm running into an issue with the "Remove" button when used from the "all books" view that aggregates all the shelves and the book is on a custom shelf:
> 
> ```
> ValueError at /unshelve/
> Field 'id' expected a number but got ''.
> ...
> app/bookwyrm/views/shelf/shelf_actions.py, line 99, in unshelve
> ```

hmm that error looks familiar, I must have missed this. I think that's referring to the shelf code, which makes sense given "All books" isn't a real shelf. "All books" is a bit of an issue because books appear multiple times: once for each shelf. Will take another look.

mouse-reeve:
Yeah, it's a weird view! I think not having a remove button in that view would also be fine, if that's an easier fix

bookwyrm/static/js/status_cache.js
hughrun: We can't just remove `submit-status` because it's coming from the modals and we don't (I mean, unless we do...) want to force a refresh on the `shelve_button`. I've adjusted this to require less code.

hughrun:
Ok, this _should_ be working now.

- removes `submit-status` class when using `shelf_selector` button (to override ajax and submit ye olde way instead)
- removes a bunch of (my) pointless code as a result of the above
- when in the "All books" shelf view, you can remove a book from any shelf it is on

There's a couple of things I would ideally have like to do in addition, but they're arguably out of scope and require further consultation:

1. each book in "All books" is listed once for every shelf it is on. I played around with the query but couldn't work out how to make it so each book appears once regardless of how many shelves it is on. If we can dedupe by `Shelfbook.book.id` or something like that, it should work but I couldn't work out how to do that without presumably breaking anything else using the generic query.
2. Buttons only show spinners if they are submitting an ajax request. Button classes seem to be a bit inconsistent so I wasn't game to just throw in a JS listener like `button[type="submit"]` but if there's a safe choice that would be nice to indicate something is, indeed, happening.

mouse-reeve:
Thank you! I think this is a bug improvement. There's still some weird behavior when a book is on multiple shelves, but that behavior isn't caused by this PR, so I think it's future work and I'll open an issue for it.

Screen Shot 2021-12-02 at 11 47 40 AM

Issue #1465, opened 2021-09-28
Consistency between how views and templates are organized


mouse-reeve:
The views, templates, and tests should be organized into directories in a way that reflects how they are used. Right now it's a bit of a mishmash. While we're at it, let's improve naming and markup conventions.

- [x] User preferences (#1464)
- [x] Site admin (#1467)
- [x] Directory (#1468)
- [x] User profile (#1469 and #1471)
- [x] User shelves (#1469)
- [x] Create book (#1477)
- [x] Book and editions (#1477)
- [x] Author (#1538)
- [ ] Feeds
- [ ] Discover
- [ ] Lists
- [ ] Get started
- [x] Import (#1600)
- [x] Notifications (#1485)
- [x] Logged out views (landing, login, register, lost password) (#1544)

mouse-reeve:
Closing this because it's overall in a better state

PR #1613, opened 2021-11-15
Use a redis cache to store lists for user

### LINKED ISSUE #1319: Bad Gateway errors when trying to use lists
**Describe the bug**
I am having the same issue as #885  - every time I try to view or otherwise interact with lists, I get a 502 Bad Gateway error. I have tried multiple browsers and instances and this problem recurs. It is very frustrating.

**To Reproduce**
1. Go to https://bookwyrm.social/list/130
2. See `502 Bad Gateway`

**Expected behavior**
A list of books

**Screenshots**
See console details below (idk what these mean, I am only a vaguely technical person)
image

**Instance**
I had this problem on both bookwyrm.social and cutebook.club.

---

    - OS: macOS 10.14.5 Mojave
    - Browser: Chrome 91.0.4472.164 *and* Firefox 91.0

mouse-reeve:
Tasks:
 - [x] Create stream class for lists
 - [x] Consider group membership when computing audiences
 - [x] Look into cache max length
 - [x] Tests for stream class
 - [x] Use stream to load lists view
 - [x] Management command for populating lists cache

Fixes #1319 
Works on #1465


bookwyrm/tests/models/test_list.py
```diff
@@ -27,7 +27,7 @@ def test_remote_id(self, _):
             book_list = models.List.objects.create(
                 name="Test List", user=self.local_user
             )
-        expected_id = "https://%s/list/%d" % (settings.DOMAIN, book_list.id)
+        expected_id = f"https://{settings.DOMAIN}/list/{book_list.id}"
```
cincodenada, line 30: :tada: This in particular is so much nicer!

bookwyrm/tests/models/test_relationship_models.py
```diff
@@ -24,14 +28,39 @@ def setUp(self):
...
+        with patch("bookwyrm.models.activitypub_mixin.broadcast_task.apply_async"):
+            models.UserBlocks.objects.create(
+                user_subject=self.local_user, user_object=self.remote_user
+            )
+
+        with self.assertRaises(IntegrityError):
+            models.UserFollows.objects.create(
+                user_subject=self.local_user, user_object=self.remote_user
+            )
+
```
cincodenada, line 62: Yay!

bookwyrm/views/list.py
```diff
@@ -29,18 +30,7 @@ class Lists(View):
...
-                request.user, privacy_levels=["public", "followers"]
-            )
-            .annotate(item_count=Count("listitem", filter=Q(listitem__approved=True)))
-            .filter(item_count__gt=0)
-            .select_related("user")
-            .prefetch_related("listitem_set")
-            .order_by("-updated_date")
-            .distinct()
-        )
+        lists = ListsStream().get_activity_stream(request.user)
```
cincodenada, line 33: :partying_face:

bookwyrm/management/commands/populate_lists_streams.py
```diff
@@ -0,0 +1,34 @@
...
+    print("This may take a long time! Please be patient.")
+    for user in users:
+        print(".", end="")
+        lists_stream.populate_lists_task.delay(user.id)
+
+
+class Command(BaseCommand):
+    """start all over with user streams"""
+
+    help = "Populate streams for all users"
```
cincodenada, line 22: This help copy should be updated to reflect that it's the list streams

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+
+class ListsStream(RedisStore):
+    """all the lists you can see"""
+
+    def stream_id(self, user):  # pylint: disable=no-self-use
+        """the redis key for this user's instance of this stream"""
+        return f"{user.id}-lists"
+
+    def get_rank(self, obj):  # pylint: disable=no-self-use
+        """lists are sorted by date published"""
```
cincodenada, line 19: published -> updated

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+        return obj.updated_date.timestamp()
+
+    def add_list(self, book_list):
+        """add a list to users' feeds"""
+        # the pipeline contains all the add-to-stream activities
+        self.add_object_to_related_stores(book_list)
+
+    def add_user_lists(self, viewer, user):
+        """add a user's listes to another user's feed"""
+        # only add the listes that the viewer should be able to see
```
cincodenada, line 29: listes -> lists on lines 28 and 29 (altho I do think "listes" is v. cute, along the lines of "fingies and toesies")

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+        lists = models.List.filter(user=user).privacy_filter(viewer)
+        self.bulk_add_objects_to_store(lists, self.stream_id(viewer))
+
+    def remove_user_lists(self, viewer, user):
+        """remove a user's list from another user's feed"""
+        # remove all so that followers only lists are removed
+        lists = user.list_set.all()
+        self.bulk_remove_objects_from_store(lists, self.stream_id(viewer))
+
+    def get_activity_stream(self, user):
```
cincodenada, line 39: `get_list_stream` maybe? Since the rest of this is from `ActivityStream` -> `ListStream`

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def add_list(self, book_list):
+        """add a list to users' feeds"""
+        # the pipeline contains all the add-to-stream activities
+        self.add_object_to_related_stores(book_list)
+
+    def add_user_lists(self, viewer, user):
+        """add a user's listes to another user's feed"""
+        # only add the listes that the viewer should be able to see
+        lists = models.List.filter(user=user).privacy_filter(viewer)
```
cincodenada, line 30: I'm only starting to get a hang of Django, but should this be `List.objects.filter`?

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+            .prefetch_related("listitem_set")
+            .order_by("-updated_date")
+            .distinct()
+        )
+
+    def populate_streams(self, user):
+        """go from zero to a timeline"""
+        self.populate_store(self.stream_id(user))
+
+    def get_audience(self, book_list):  # pylint: disable=no-self-use
```
cincodenada, line 57: I wonder if this (and its companion in activitystream) should be a method on the List object instead, sort of a beefed up version of `privacy_filter`. That would probably be a separate PR tho anyway, since it's also in activitystream.

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+                    | Q(memberships__group__id=book_list.group.id)
+                )
+            else:
+                audience = audience.filter(
+                    Q(id=book_list.user.id)  # if the user is the list's owner
+                    | Q(following=book_list.user)  # if the user is following the pwmer
+                )
+        return audience.distinct()
+
+    def get_stores_for_object(self, obj):
```
cincodenada, line 95: I realize this is just copying from activitystream but is there a reason this shouldn't be `get_stores_for_list`? Or maybe even `get_stores_with_list`/`get_stores_containing_list`, if I'm understanding correctly.

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+        return [self.stream_id(u) for u in self.get_audience(obj)]
+
+    def get_lists_for_user(self, user):  # pylint: disable=no-self-use
+        """given a user, what lists should they see on this stream"""
+        return models.List.privacy_filter(
+            user,
+            privacy_levels=["public", "followers"],
+        )
+
+    def get_objects_for_store(self, store):
```
cincodenada, line 105: Same as above, I think this would be more clear as `get_lists_for_store`

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+        user = models.User.objects.get(id=store.split("-")[0])
+        return self.get_lists_for_user(user)
```
cincodenada, line 107: I think pylint is right and these last four methods and `stream_id` should all be class methods. They don't have anything to do with a given instance, and some of them even return instances, which is very much a class method's job.

bookwyrm/management/commands/populate_lists_streams.py
```diff
@@ -0,0 +1,34 @@
+""" Re-create user streams """
```
cincodenada, line 1: -> list streams

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
+""" access the activity streams stored in redis """
```
cincodenada, line 1: -> list streams

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+        return self.get_lists_for_user(user)
+
+
+@receiver(signals.post_save, sender=models.List)
+# pylint: disable=unused-argument
+def add_list_on_create(sender, instance, created, *args, **kwargs):
+    """add newly created lists to activity feeds"""
+    if not created:
+        return
+    # when creating new things, gotta wait on the transaction
```
cincodenada, line 116: phew, this is a tricky one!

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+    transaction.on_commit(lambda: add_list_on_create_command(instance))
```
cincodenada, line 117: I'm not super familiar with Python lambdas, but if this is just one line, could this be `lambda: add_list_task.delay(instance.id)`?

bookwyrm/tests/lists_stream/test_signals.py
```diff
@@ -0,0 +1,111 @@
...
+                outbox="https://example.com/users/rat/outbox",
+            )
+
+    def test_add_list_on_create(self, _):
+        """a new lists has entered"""
+        book_list = models.List.objects.create(
+            user=self.remote_user, name="hi", privacy="public"
+        )
+        with patch("bookwyrm.lists_stream.add_list_task.delay") as mock:
+            lists_stream.add_list_on_create_command(book_list)
```
cincodenada, line 39: Hmm. So if I'm reading correctly, this doesn't really test that the post_save signal is doing anything, since we're just manually calling what the signal handler _would_ call here. Do we need to mock something so that `on_transaction` fires instead?

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+
+
+@receiver(signals.pre_delete, sender=models.List)
```
cincodenada, line 120: Shouldn't this be a post_delete? Otherwise if something keeps us from actually deleting the list, we'll remove it from streams prematurely? I checked and we do still get an instance with post_delete, it's just not in the database any more.

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+# pylint: disable=unused-argument
+def remove_list_on_delete(sender, instance, *args, **kwargs):
+    """add newly created lists to activity feeds"""
```
cincodenada, line 123: fix docstring (remove deleted lists from...)

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+# pylint: disable=unused-argument
+def remove_lists_on_block(sender, instance, *args, **kwargs):
+    """remove lists from all feeds on block"""
+    # blocks apply ot all feeds
+    if instance.user_subject.local:
+        remove_user_lists_task.delay(instance.user_subject.id, instance.user_object.id)
+
+    # and in both directions
+    if instance.user_object.local:
+        remove_user_lists_task.delay(instance.user_object.id, instance.user_subject.id)
```
cincodenada, line 160: Caching sure is a lot of work!! Well done

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+            instance.user_subject.id,
+            instance.user_object.id,
+        )
+
+    # add lists back to streams with lists from anyone
+    if instance.user_object.local:
+        add_user_lists_task.delay(
+            instance.user_object.id,
+            instance.user_subject.id,
+        )
```
cincodenada, line 186: Do we need to wait on the transaction here as well? Otherwise we'll be forbidden from adding these lists if the block hasn't actually been deleted, yeah? I'd think it wouldn't be an issue since it's a delayed task, but that's the case above where we waited for the transaction too, so maybe it is!

bookwyrm/lists_stream.py
```diff
@@ -0,0 +1,239 @@
...
+    if not created or not instance.local:
+        return
+
+    populate_lists_task.delay(instance.id)
+
+
+# ---- TASKS
+@app.task(queue=MEDIUM)
+def populate_lists_task(user_id):
+    """background task for populating an empty activitystream"""
```
cincodenada, line 202: -> empty list stream

cincodenada (commented):
I haven't looked carefully through the tests yet, but I figured I'd submit what I've got here. Sorry if you were midstream here, I just saw this and thought it looked interesting to give a look over :)

bookwyrm/tests/activitypub/test_base_activity.py
```diff
@@ -31,7 +31,7 @@ def setUp(self):
         """we're probably going to re-use this so why copy/paste"""
         with patch("bookwyrm.suggested_users.rerank_suggestions_task.delay"), patch(
             "bookwyrm.activitystreams.populate_stream_task.delay"
-        ):
+        ), patch("bookwyrm.lists_stream.populate_lists_task.delay"):
```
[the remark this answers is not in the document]
mouse-reeve: yeah these patch lines are gnarly in more ways than one

bookwyrm/lists_stream.py
mouse-reeve: changing all instances to listies

bookwyrm/lists_stream.py
mouse-reeve: yes, I must be missing a test for this function

bookwyrm/lists_stream.py
mouse-reeve: yeah, I think you're right on both fronts

bookwyrm/lists_stream.py
mouse-reeve: Those would be more sensible names, but it's overriding a function in `bookwyrm/redis_store.py` which is called in that parent class

bookwyrm/lists_stream.py
mouse-reeve: yeah it took me ages to debug this

bookwyrm/lists_stream.py
mouse-reeve: I think I did it like this because the task has to take an id (constraint in celery), and the re-load the object to figure out where it's being stored, so the object has to be queryable at the time the task runs. Which, given how tasks work, may not really be the case anyway 🤔 . Hmm. I guess it could be bulk removed from every redis stream, which would mean only the id was needed.....

bookwyrm/tests/lists_stream/test_signals.py
mouse-reeve: Yeah, it isn't testing the function it says it is. I'm not sure how to mock (or just complete) the transaction, though

bookwyrm/lists_stream.py
mouse-reeve: the block existing doesn't do anything to prevent the lists from being added to redis, so the transaction won't stop it from working

mouse-reeve:
@cincodenada thank you _so_ much for this, it's incredibly helpful and I appreciate it immensely

bookwyrm/lists_stream.py
cincodenada: Ah okay, yeah that makes a ton of sense, I missed that these are overrides. Now I'm sad ABC doesn't have an override decorator, but Stack Overflow tells me there are reasons...looks like someone [made a library](https://stackoverflow.com/questions/1167617/in-python-how-do-i-indicate-im-overriding-a-method) that provides an @override decorator, but that's for another day.

bookwyrm/lists_stream.py
cincodenada: I'm curious about the celery constraints here - I'm not familiar enough with it yet to know the details there. You have a good point about delayed tasks making it possible the object isn't there anyway.

I think this is worth following up on (since it could result in deleted lists showing up in a stream) but it doesn't have to be a blocker for merge imo.

cincodenada (approved):
I gave this a quick once-over - if helpful I can take a little closer look this evening, but if you want to merge go ahead, I agree that it makes sense to circle back on the remaining pieces.

mouse-reeve:
You were right to remind me about that part -- it was not working. I refactored it so that it doesn't use the object of the deleted list, but just removes the object from every store

mouse-reeve:
@bookwyrm-social/instance-admins Once this is merged into production, you'll need to run `./bw-dev populate_lists_streams` to get lists working again

PR #1627, opened 2021-11-20
Don't change the preview image URL on update

joachimesque:


mouse-reeve:
Thank you for catching and fixing this so quickly!

PR #1629, opened 2021-11-21
Add reviews stream with only reviews and comments

joachimesque:
I wanted a way to display only reviews and comments (mostly to get rid of "noise" aka generated notes about shelves updates). This might not be the best way to do it but I think it works.

An alternative would be to filter each activity stream in `bookwyrm/views/feed.py` and have per-feed status type preference, but that's a bit more complicated (dealing with forms and stuff).


joachimesque:
Closed because of #1630

PR #1630, opened 2021-11-21
Add feed filters

joachimesque:
As an alternative to  #1629

image


bookwyrm/views/feed.py
```diff
@@ -230,3 +248,16 @@ def get_suggested_books(user, max_books=5):
...
+
+def filter_stream_by_status_type(activities, allowed_types=[]):
+    if "review" not in allowed_types:
+        activities = activities.filter(Q(review__isnull=True))
+    if "comment" not in allowed_types:
+        activities = activities.filter(Q(comment__isnull=True))
+    if "quotation" not in allowed_types:
+        activities = activities.filter(Q(quotation__isnull=True))
+    if "everything" not in allowed_types:
+        activities = activities.filter(Q(generatednote__isnull=True))
```
joachimesque, line 261: This query will have to incorporate more types of notes.

bookwyrm/views/feed.py
```diff
@@ -230,3 +248,16 @@ def get_suggested_books(user, max_books=5):
         suggested_books.append(shelf_preview)
         book_count += len(shelf_preview["books"])
     return suggested_books
+
+
+def filter_stream_by_status_type(activities, allowed_types=[]):
+    if "review" not in allowed_types:
+        activities = activities.filter(Q(review__isnull=True))
```
joachimesque, line 255: I don't know if it's the best way to filter activities, I don't think I have access to the class name of each object to filter out.

bookwyrm/models/user.py
```diff
@@ -128,6 +128,13 @@ class User(OrderedCollectionPageMixin, AbstractUser):
     show_suggested_users = models.BooleanField(default=True)
     discoverable = fields.BooleanField(default=False)
 
+    # feed options
+    feed_status_types = ArrayField(
+        models.CharField(max_length=10, blank=False),
+        size=8,
+        default=list(["review", "comment", "quotation", "everything"]),
```
joachimesque, line 135: On app start I get the following warning, concerning this list:

```
bookwyrm-web-1             | WARNINGS:
bookwyrm-web-1             | bookwyrm.User.feed_status_types: (fields.E010) ArrayField default should be a callable instead of an instance so that it's not shared between all field instances.
bookwyrm-web-1             | 	HINT: Use a callable instead, e.g., use `list` instead of `[]`.
```

mouse-reeve: You can use choices to simply a few things here. First, you can define a choice block that has the translations in it like you have in the multi select choices:
``` python3
FeedFilterChoices = [
    "review", _("Reviews"),
    ...
]
```

And then you can define `choices` in the `ArrayField`, which will enforce the expected values:

```python3
ArrayField(
    models.CharField(max_length=10, blank=False),
    choices=FeedFilterChoices,
    ...
)
```

And finally, define a function that returns the default list so that it isn't grumpy about that:
```python3
def get_feed_filter_choices():
    return [f[0] for f in FeedFilterChoices]

...

ArrayField(
    models.CharField(max_length=10, blank=False),
    choices=FeedFilterChoices,
    default=get_feed_filter_choices,
    ...
)
```

I think if the model field is declared this way, you won't have to pass `choices` to `widgets.CheckboxSelectMultiple`.

bookwyrm/views/feed.py
mouse-reeve: I can't think of a better way to do this

mouse-reeve: I just realized boosts are getting filtered as an "everything else" type, rather than under the status type of the boosted status. You can access the type boosted status with, for example: `boost__boosted_status__review__isnull=True`

mouse-reeve (changes requested):
Thank you so much for making this change, it's going to make the feed so much nicer to use!

bookwyrm/templates/feed/feed.html
```diff
@@ -16,10 +16,45 @@ <h1 class="title">
...
+            </button>
+        </div>
+    </form>
+</details>
+
 {# announcements and system messages #}
 {% if not activities.number > 1 %}
 <a href="{{ request.path }}" class="transition-y is-hidden notification is-primary is-block" data-poll-wrapper>
     {% blocktrans with tab_key=tab.key %}load <span data-poll="stream/{{ tab_key }}">0</span> unread status(es){% endblocktrans %}
+    {{ allowed_status_types|json_script:"unread-notifications-wrapper" }}
```
joachimesque, line 57: Django doesn't have a cleaner way of inserting JSON values into the code. At least it's safely inserted.

bookwyrm/static/js/bookwyrm.js
```diff
@@ -113,9 +113,44 @@ let BookWyrm = new class {
...
     updateCountElement(counter, data) {
+        let count = data.count;
+        const count_by_type = data.count_by_type;
         const currentCount = counter.innerText;
-        const count = data.count;
         const hasMentions = data.has_mentions;
+        const allowedStatusTypesEl = document.getElementById('unread-notifications-wrapper');
+
+        // If we're on the right counter element
+        if (counter.closest('[data-poll-wrapper]').contains(allowedStatusTypesEl)) {
```
joachimesque, line 123: I'm not completely happy with this way of discriminating between poll sites, but this is dependent on Django's ways of inserting JSON data into the code.

mouse-reeve:
Looks great!

PR #1633, opened 2021-11-27
Front-end: New modal component + header alignment

joachimesque:
## Scope of this PR

This PR adds a new modal component. The handling of the modal via JS can be used to open any modal.

I also added an accessibility enhancement, so that keyboard surfers can see focus states for all elements (Bulma is very bad on that point).

image
This captures shows a previous use of the modal, out of the scope of this PR.

## How to use the modal?

Invocation button:

```html
<button class="button is-primary" type="button" data-modal-open="CUSTOM_MODAL_NAME" aria-haspopup="dialog">Click to open the modal</button>
```

Modal code:

```html
{# Login modal #}
<div class="modal" id="CUSTOM_MODAL_NAME">
    <div class="modal-background" data-modal-close></div><!-- modal background -->
    <div class="modal-card" role="dialog" aria-modal="true" aria-labelledby="CUSTOM_MODAL_HEADER" tabindex="-1">
        <header class="modal-card-head">
            <h1 class="modal-card-title" id="CUSTOM_MODAL_HEADER">
                Modal title
            </h1>
            <button type="button" class="delete" data-modal-close aria-label="{% trans 'Close' %}"></button>
        </header>
        <section class="modal-card-body">
            Modal content
        </section>
        <footer class="modal-card-foot">
            Modal footer
        </footer>
    </div>
    <button type="button" data-modal-close class="modal-close is-large" aria-label="{% trans 'Close' %}"></button>
</div>
```

The optional class `is-thin` can be added to the `div.modal-card` if you need a smaller modal.


mouse-reeve:
Thank you! I didn't notice that the login form was higher until you pointed it out and now it will bug me until this gets merged.

Your new modal javascript also looks really good - having the semi-transparent background be clickable is a big improvement. It seemed to work smoothly for me with VoiceOver as well. The existing modals in the app use the same toggle javascript that inline show/hide elements use, so for the sake of consistency, I think it makes sense to switch them over to your new code, but that's a to-do for me, not work that should be part of this pull request.

I think it makes sense to use the modal template component for the login modal (`bookwyrm/templates/components/modal.html`), but it will need to be updates to work with the new javascript, so that can be considered future work as well unless you're excited about retrofitting template code :)

As an aside, I think I'm going to try and get something equivalent to Black, probably https://prettier.io/, for javascript to reduce the amount of squinting at pylint required.

bookwyrm/static/js/bookwyrm.js
```diff
@@ -368,4 +374,84 @@ let BookWyrm = new class {
...
+        
+        if (!modal) {
+            return;
+        }
+
+        // Helper functions
+        function handleModalOpen(modalElement) {
+            htmlElement.classList.add('is-clipped');
+            modalElement.classList.add('is-active');
+            modalElement.getElementsByClassName('modal-card')[0].focus();
```
mouse-reeve, line 392: Will this work for situations where there are multiple modals on a page?

bookwyrm/static/js/bookwyrm.js
```diff
@@ -368,4 +374,84 @@ let BookWyrm = new class {
...
+        const htmlElement = document.querySelector('html');
+        const modal = document.getElementById(targetModalId);
+        
+        if (!modal) {
+            return;
+        }
+
+        // Helper functions
+        function handleModalOpen(modalElement) {
+            htmlElement.classList.add('is-clipped');
```
mouse-reeve, line 390: Is this to prevent the background from scrolling?

bookwyrm/static/js/bookwyrm.js
```diff
@@ -368,4 +374,84 @@ let BookWyrm = new class {
...
+
+        function handleFocusTrap(event) {
+            const focusableEls = event.currentTarget.querySelectorAll(
+                [
+                    'a[href]:not([disabled])',
+                    'button:not([disabled])',
+                    'textarea:not([disabled])',
+                    'input[type="text"]:not([disabled])',
+                    'input[type="radio"]:not([disabled])',
+                    'input[type="checkbox"]:not([disabled])',
```
mouse-reeve, line 430: I don't know querySelector super well, but it seems like is going to miss inputs of types `date`, `password`, and `number`. I wonder if it's possible to select all inputs that are _not_ of type `hidden`?

bookwyrm/static/js/bookwyrm.js
```diff
@@ -368,4 +374,84 @@ let BookWyrm = new class {
...
+                    'select:not([disabled])'
+                ].join(',')
+            );
+            const firstFocusableEl = focusableEls[0];  
+            const lastFocusableEl = focusableEls[focusableEls.length - 1];
+
+            if (event.key !== 'Tab') { 
+                return;
```
mouse-reeve, line 438: I think it would be more efficient to check this at the top of the function, before doing the query selection.

bookwyrm/static/js/bookwyrm.js
joachimesque: I think so, but I haven't tested. I tried to get `handleModalOpen` to work in the context of each specific modal. For this line, we focus the first element with that class that is inside the `modalElement` that has been passed, and that element is the one that's targeted by the value of `data-modal-open`

bookwyrm/static/js/bookwyrm.js
joachimesque: yes, it's recommended by the Bulma doc

> To activate the modal, just add the `is-active` modifier on the `.modal` container. You may also want to add `is-clipped` modifier to a containing element (usually html) to stop scroll overflow.

mouse-reeve:
I've been dithering on this, but I think, because the login form is likely to be the one and only thing someone is interested in on the landing page, it is better to have it readily available, rather than behind an additional click. But your modal code absolutely rules, so I'd like to merge that. And I think it's possible to solve the top bar misalignment issue while keeping the form in there, even though I haven't figured out the perfect fix yet

mouse-reeve:
I know there are approximately one million modals that use the existing system, so I'm happy to merge the javascript and css changes you've already and then work on retrofitting.

joachimesque:
Ok, I'll handle all that :)

joachimesque:
Here's what the header looks like:

Screenshot 2021-12-17 at 17-29-16 BookWyrm
Screenshot 2021-12-17 at 17-28-57 BookWyrm
Screenshot 2021-12-17 at 17-27-34 BookWyrm

I decided for using a `button` instead of a `div role="button"`, for accessibility matters. Hence the button reset styles.

mouse-reeve:
I'm stoked to start updating modals :)

Issue #1139, opened 2021-05-24
Remote Follow on Profile Pages


TomatDividedBy0:
Remote follow is one of the advantages of the Fediverse, and a button to remote follow a profile for both logged-in and logged-out profiles would make it easier for people to realize they can follow an account from Mastodon/Pleroma/wherever without necessarily needing to create a Bookwyrm account.

hughrun:
I'm working on a PR for this. Stay tuned.

hughrun:
Okey dokey, **seeking feedback on #1635.** I still need to do a little bit of cleanup and testing but it's close to complete and proven to work, at least with Mastodon. This should work with any other ActivityPub service that uses the ostatus 'subscribe' protocol. That definitely includes Mastodon and Pleroma, not sure if any others do the same.

**Follow a bookwyrm user from remote server**

1. Click new button on user profile page, regardless of whether or not you are logged in. Happy to adjust this button including the wording.

![follow_button](https://user-images.githubusercontent.com/17669239/143765372-cf40ad38-60db-48de-b2e3-c06419c937d7.png)

2. a new window pops up (non-javascript fallback will own in new window/tab - happy to receive advice on the most accessible way to indicate that this happens in a new window). Enter the handle you wish to follow from.

![follow_form](https://user-images.githubusercontent.com/17669239/143765385-e67573c6-20f1-4f97-a1e4-dda6c80188e4.png)
 
3. User is redirected to their "home" server to complete the follow activity using their own ostatus subscribe template

![redirect_mastodon](https://user-images.githubusercontent.com/17669239/143765464-6b9cb226-07af-42fa-a291-171be95773db.png)

4. Follow completed

![proof](https://user-images.githubusercontent.com/17669239/143765562-85e2d062-9a3d-48cf-932f-255227b127a6.png)

 **Follow remote user from bookwyrm account**

1. Click on follow button on Mastodon whilst logged out, window pops up asking for remote user handle.

![follow_masto_from_bookwyrm](https://user-images.githubusercontent.com/17669239/143765517-fb9db0fd-5ce1-4d3c-8654-fcf400d9edd3.png)

2. User is redirected to Bookwyrm ostatus subscribe template to complete follow

![redirect_to_bookwyrm](https://user-images.githubusercontent.com/17669239/143765544-ebfe60c8-0402-458f-943d-af77b1212f49.png)

[...]

PR #1635, opened 2021-11-28
Remote follow

hughrun:
~~Work in Progress seeking feedback. More testing needed.~~ See #1139 for discussion.

This PR uses the same strategy as Mastodon: the `http://ostatus.org/schema/1.0/subscribe` protocol [indicates the endpoint to use for remote following](https://socialhub.activitypub.rocks/t/represent-endpoint-for-remote-interaction/480).

- enables remote ActivityPub actors to follow Bookwyrm users by clicking a button on Bookwyrm user profile
- enables Bookwyrm users to follow remote ActivityPub actors by clicking a button on remote user profile

This has been tested as working on Mastodon, and appears very likely to be working with Pleroma.


bookwyrm/templates/ostatus/remote_follow_button.html
```diff
@@ -0,0 +1,12 @@
+{% load i18n %}
+{% if request.user == user %}
+{% else %}
+
+<div class="field mb-0">
+    <div class="control">
+        <a class="button is-small is-link" href="{% url 'remote-follow-page' %}?user={{ user.username }}" target="_blank" rel="noopener noreferrer" onclick="BookWyrm.displayPopUp(`{% url 'remote-follow-page' %}?user={{ user.username }}`, `remoteFollow`); return false">
```
joachimesque, line 7: I saw in the issue that you weren't sure as to how to make the pop-up window accessible: I suggest adding an ARIA annotation, saying that the link opens in a pop-up window. From that point on, the APIs from the OS take over and describe the new (focused) window.

Why `aria-description`?

- It is [pretty well supported](https://a11ysupport.io/tech/aria/aria-description_attribute), except on iOS)
- `aria-label` override the text content of the link, which is not good in this situation
- `title` has horrible screen reader support
- `is-sr-only` could be used in the button's text content, but I think it would insert whitespace and mess up the button's dimensions

```suggestion
        <a class="button is-small is-link" href="{% url 'remote-follow-page' %}?user={{ user.username }}" target="_blank" rel="noopener noreferrer" onclick="BookWyrm.displayPopUp(`{% url 'remote-follow-page' %}?user={{ user.username }}`, `remoteFollow`); return false" aria-description="{% trans "This link opens in a pop-up window" %}">
```

Or, we could add the notice directly in the button text, without hiding it. The button would be bigger, the interface could appear messier, but it could be better for some users.

An alternative would be to have the text visible, outside the button but visually linked to it (close enough that we know it's that link that opens in a new window), with an `aria-describedby` on the link pointing to that text.

mouse-reeve: Thank you! Is this a best practice for any link with a `target="_blank"`, or just for popups specifically?

hughrun: Thanks @joachimesque!

Generally speaking I'm not in favour of pop-ups, this is one of those rare exceptions where I think it is genuinely the smoothest option for most users. Still, it's confusing if you are blocking pop-ups. I think I actually prefer the last option. What do you think about this, @joachimesque @mouse-reeve ? 

pop-up

If we go with it, is it best to use only `aria-describedby` or is it helpful to include both `aria-describedby` _and_ `aria-description`? Or is that actually annoying/confusing for people using screen-readers?

joachimesque: The last option is IMO the best scenario: the information is available to all users. `aria-description` won't be  useful in this case, you can only use `aria-describedby` :)

mouse-reeve: I like that option as well

hughrun:
This should now be ready. A couple of things to note:

1. This only works for fediverse sites that use all of ActivityPub, webfinger, and the ostatus subscribe protocol. As best I can tell this means Mastodon, Pleroma, and (once this feature is rolled out) Bookwyrm. Pixelfed _might_, I can't work it out. PeerTube doesn't use webfinger and Lemmy doesn't use ostatus subscribe.
2. The full username is required for webfinger lookups to work, but actually the follow will be by whichever account is logged in. That is, you could enter `mouse@bookwyrm.social` as the remote address and it will all work, but if you are logged in to `bookwyrm.social` as the user `hugh`, then the remote account will be followed by `hugh`, not `mouse`. This is also how Mastodon remote follow works and it will only cause confusion in a tiny number of users with multiple accounts on a single instance who aren't paying attention.

bookwyrm/templates/ostatus/remote_follow_button.html
```diff
@@ -0,0 +1,15 @@
...
+
+<div class="field mb-0">
+    <div class="control">
+        <a class="button is-small is-link" href="{% url 'remote-follow-page' %}?user={{ user.username }}" target="_blank" rel="noopener noreferrer" onclick="BookWyrm.displayPopUp(`{% url 'remote-follow-page' %}?user={{ user.username }}`, `remoteFollow`); return false;" aria-describedby="remote_follow_warning">
+            {% blocktrans with username=user.localname %}Follow on Fediverse{% endblocktrans %}
+        </a>
+    </div>
+    <p id="remote_follow_warning" class="mt-1 is-size-7 has-text-weight-light">
+        {% trans 'This link opens in a pop-up window' %}
+    </p>
```
joachimesque, line 13: 🎉

bookwyrm/templates/ostatus/template.html
```diff
@@ -0,0 +1,60 @@
+{% load layout %}
+{% load i18n %}
+{% load static %}
+{% load utilities %}
+{% load markdown %}
+
+<!DOCTYPE html>
+<html lang="{% get_lang %}">
+<head>
+    <title>{% block title %}{% endblock %}{{ site.name }}</title>
```
joachimesque, line 10: Accessibility-wise, it would be great to have the title set with the current step of the follow

bookwyrm/templates/ostatus/template.html
```diff
@@ -0,0 +1,60 @@
...
+    {% endif %}
+    <meta name="twitter:title" content="{% if title %}{{ title }} - {% endif %}{{ site.name }}">
+    <meta name="og:title" content="{% if title %}{{ title }} - {% endif %}{{ site.name }}">
+    <meta name="twitter:description" content="{{ site.instance_tagline }}">
+    <meta name="og:description" content="{{ site.instance_tagline }}">
+
+    {% block opengraph_images %}
+        {% include 'snippets/opengraph_images.html' %}
+    {% endblock %}
+    <meta name="twitter:image:alt" content="BookWyrm Logo">
```
joachimesque, line 33: IMO, the OpenGraph properties are not useful for the pop-up window

```suggestion
```

bookwyrm/templates/ostatus/template.html
```diff
@@ -0,0 +1,60 @@
...
+        function closeWindow() {
+            window.close();
+        }
+    </script>
+</head>
+<body>
+<nav class="navbar" aria-label="main navigation">
+    <div class="container">
+        <div class="navbar-brand">
+            <a class="navbar-item" href="/">
```
joachimesque, line 44: If we're in the pop-up window I think that all links could be disactivated, except for those necessary to the follow transaction

bookwyrm/templates/ostatus/template.html
```diff
@@ -0,0 +1,60 @@
...
+        </div>
+    </div>
+</nav>
+<div class="section is-flex-grow-1 columns is-centered">
+    <div class="block column is-one-third">
+        {% block content%}{% endblock %}
+    </div>
+</div>
+
+{% include 'components/footer.html' %}
```
joachimesque, line 57: Same thing with the footer, I guess?

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
+{% load i18n %}
+
+{% block content %}
+<div class="block">
+    {% if not request.user.is_authenticated and not error == 'remote_subscribe' %}
+    <div class="navbar-item">
```
joachimesque, line 6: From what I gather we're not in the main navbar, so I think we can do without this `div`

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+                        <div class="column is-narrow">
+                            <button class="button is-primary" type="submit">{% trans "Log in" %}</button>
+                        </div>
+                    </div>
+                </form>
+            </div>
+        </div>
+    </div>
+    {% elif error == 'ostatus_subscribe' %}
+    <div class="notification is-warning has-text-centered">
```
joachimesque, line 30:
```suggestion
    <div class="notification is-warning has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}Something went wrong trying to follow <strong>{{ account }}</strong>{% endblocktrans %}</p>
+        <p>{% trans 'Check you have the correct username before trying again.' %}</p>
+    </div>
+    {% elif error == 'remote_subscribe' %}
+    <div class="notification is-warning has-text-centered">
```
joachimesque, line 35:
```suggestion
    <div class="notification is-warning has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}Something went wrong trying to follow from <strong>{{ account }}</strong>{% endblocktrans %}</p>
+        <p>{% trans 'Check you have the correct username before trying again.' %}</p>
+    </div>
+    {% elif error == 'is_blocked' %}
+    <div class="notification is-danger has-text-centered">
```
joachimesque, line 40:
```suggestion
    <div class="notification is-danger has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}You have blocked <strong>{{ account }}</strong>{% endblocktrans %}</p>
+    </div>
+    {% elif error == 'has_blocked' %}
+    <div class="notification is-danger has-text-centered">
```
joachimesque, line 44:
```suggestion
    <div class="notification is-danger has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}<strong>{{ account }}</strong> has blocked you{% endblocktrans %}</p>
+    </div>
+    {% elif error == 'already_following' %}
+    <div class="notification is-success has-text-centered">
```
joachimesque, line 48:
```suggestion
    <div class="notification is-success has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}You are already following <strong>{{ account }}</strong>{% endblocktrans %}</p>
+    </div>
+    {% elif error == 'already_requested' %}
+    <div class="notification is-success has-text-centered">
```
joachimesque, line 52:
```suggestion
    <div class="notification is-success has-text-centered" role="status">
```

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,60 @@
...
+        <p>{% blocktrans %}You have already requested to follow <strong>{{ account }}</strong>{% endblocktrans %}</p>
+    </div>
+    {% endif %}
+</div>
+<div class="block is-pulled-right">
+    <a href="" onclick="closeWindow()">Close window</a>
```
joachimesque, line 58: Accessibility-wise, I'd suggest a button instead of the anchor.

```suggestion
    <button type="button" onclick="closeWindow()">Close window</button>
```

joachimesque (commented):
I checked out the accessibility of your implementation and I couldn't resist leaving a few comments not related to a11Y stuff, I hope it's ok.

The `role="status"` suggestions will give screen readers some info as to how to announce the notification. Basically,  `role="status"` tells them to read it when idle, it's like `aria-live="polite"` with more semantics. Alternatively, you could use `role="alert"`, which is way more urgent (equivalent to `aria-live="assertive"`), and is read whenever it's displayed on the page, so at page load in this situation.

I left some remarks about code not being necessary (header links, footer…), but I haven't tested the feature locally, so it's really more questions than affirmations.

All in all it's a great feature, I can't wait to use it!

hughrun:
Thankyou @joachimesque! These all look like good suggestions, I was a bit sloppy. Always happy for a11y advice, it's important but I have a lot to learn.

On reflection you're right about not only the header meta but also the footer so I'll also undo the move into its own include file.

bookwyrm/templates/ostatus/error.html
mouse-reeve: I have a button to commit these suggestions into the PR and I assume you do too @hughrun, but if for some permissions reason you don't, let me know

hughrun: I do! Does that commit to `hughrun:remote-follow`? I don't want to mess up the git branches if it commits directly into `bookwyrm:remote-follow` or something.

mouse-reeve: I have no idea! I'd guess it commits it to the main repo and you'd have to pull to your local branch from upstream?

joachimesque:
I noticed something that could complicate somewhat the translations, some languages use different word orders so for example this code would be harder to translate: `{% trans 'You are now following ' %}{{ user.display_name }}`. I think that using `blocktrans` would be better in many cases

hughrun:
> I noticed something that could complicate somewhat the translations, some languages use different word orders so for example this code would be harder to translate: `{% trans 'You are now following ' %}{{ user.display_name }}`. I think that using `blocktrans` would be better in many cases

I agree, the problem is that I couldn't get `{% blocktrans %} to work in some cases (usually when including a template), and i couldn't work out why so I though maybe I'd misunderstood something. Tips appreciated, I'll take another look at this.

mouse-reeve:
> I agree, the problem is that I couldn't get `{% blocktrans %}` to work in some cases (usually when including a template), and i couldn't work out why so I though maybe I'd misunderstood something. Tips appreciated, I'll take another look at this.

I don't know if this is what you ran into, but the problem I kept having was that you can't pass an object into blocktrans. So to see the display name, you have to use `{% blocktrans with username=user.display_name %}You are now following {{ username }}{% endblocktrans %}`

hughrun:
> > I agree, the problem is that I couldn't get `{% blocktrans %}` to work in some cases (usually when including a template), and i couldn't work out why so I though maybe I'd misunderstood something. Tips appreciated, I'll take another look at this.
> 
> I don't know if this is what you ran into, but the problem I kept having was that you can't pass an object into blocktrans. So to see the display name, you have to use `{% blocktrans with username=user.display_name %}You are now following {{ username }}{% endblocktrans %}`

Ahh yes that's it. I guess I though all that "with x=y" stuff was just to be able to use simpler variables.

bookwyrm/templates/ostatus/error.html
hughrun: Let's find out haha

bookwyrm/views/follow.py
```diff
@@ -84,3 +94,68 @@ def delete_follow_request(request):
...
+    )
+    account = f"{username_parts[2]}@{username_parts[1]}"
+    user = handle_remote_webfinger(account)
+    error = None
+
+    if user is None or user == "":
+        error = "ostatus_subscribe"
+
+    # don't do these checks for AnonymousUser before they sign in
+    if request.user.id:
```
mouse-reeve, line 113: You can check `request.user.is_authenticated:` -- it'll be functionally the same, but it's a little clearer what's being checked.

bookwyrm/templates/ostatus/remote_follow.html
```diff
@@ -0,0 +1,46 @@
...
+<div class="block">
+    <p>{% blocktrans with username=user.display_name %}Follow {{ username }} from another Fediverse account like Bookwyrm, Mastodon, or Pleroma.{% endblocktrans %}</p>
+</div>
+<div class="card">
+    <section class="card-content content">
+        <form name="remote-follow" method="post" action="{% url 'remote-follow' %}">
+            {% csrf_token %}
+            <input type="hidden" name="user" value="{{ user.id }}">
+            <label class="label" for="remote_user">{% trans 'User handle to follow from:' %}</label>
+            <input class="input" type="text" name="remote_user" id="remote_user" placeholder="user@example.social">
```
mouse-reeve, line 41: This `input` should be `required`

bookwyrm/views/follow.py
```diff
@@ -84,3 +94,68 @@ def delete_follow_request(request):
...
+    return TemplateResponse(request, "ostatus/remote_follow.html", data)
+
+
+@require_POST
+def remote_follow(request):
+    """direct user to follow from remote account using ostatus subscribe protocol"""
+    remote_user = request.POST.get("remote_user")
+    template = subscribe_remote_webfinger(remote_user)
+    if template is None:
+        data = {"account": remote_user, "user": None, "error": "remote_subscribe"}
```
mouse-reeve, line 157: Would it be possible to create separate errors for when a user isn't found on the server vs when the server isn't equipped for remote follow?

mouse-reeve (changes requested):
This is so cool! All my feedback is pretty minor. At some point it needs tests that render the new templates and check html validity, but I can add that later if you're tested-out for now :)

bookwyrm/views/follow.py
hughrun: Oh yeah ghat would be much clearer for the BookWyrm user. May need to add some additional code to differentiate no server / user not found / user found but service doesn't use ostatus subscribe, though we'd use friendlier descriptions

hughrun:
@mouse-reeve I am trying to set up page rendering tests but I keep getting an error:

`bookwyrm.models.site.SiteSettings.DoesNotExist: SiteSettings matching query does not exist.`

Here's an example of what I've tried for my tests:

```python
def test_remote_follow_page(self, _):
    """make sure remote follow page loads"""
    request = self.factory.get("", {"acct": "mouse@local.com"})
    request.user = self.remote_user
    result = views.remote_follow_page(request)
    self.assertIsInstance(result, TemplateResponse)
    validate_html(result.render())
    self.assertEqual(result.status_code, 200)
```

It seems to be `validate_html` that pytest doesn't like, but I'm not sure why. Any tips?

mouse-reeve:
yeah, there's middleware the loads the site name and logo, so when you're rendering html, the `setUp` function has to have `models.SiteSettings.objects.create()` in it to initialize the default values

bookwyrm/templates/ostatus/error.html
```diff
@@ -0,0 +1,76 @@
+{% load i18n %}
+
+{% block content %}
+<div class="block">
+    {% if error == 'invalid_username' %}
+    <div class="notification is-warning has-text-centered" role="status">
+        {% blocktrans %}
```
mouse-reeve, line 7: The whitespace on these trans blocks is going to cause formatting issues in the locale file. I think it makes more sense to translate each sentence separately, with as little html in each as possible

hughrun: I've pulled them into separate strings but we need to keep _some_ html inside blocktrans otherwise it will be impossible to translate properly into languages that don't share the same grammar and word order as English.

mouse-reeve: yeah for sure - your change looks perfect, thank you

mouse-reeve:
🎉

PR #1638, opened 2021-11-28
Add `aria-describedby` to field help/error messages

joachimesque:


bookwyrm/templates/preferences/edit_user.html
```diff
@@ -34,29 +34,29 @@ <h2 class="title is-4">{% trans "Profile" %}</h2>
...
-                <p class="help is-danger">{{ error | escape }}</p>
+                <p class="help is-danger" id="desc_summary">{{ error | escape }}</p>
                 {% endfor %}
             </div>
             <div class="field">
                 <label class="label" for="id_email">{% trans "Email address:" %}</label>
                 {{ form.email }}
                 {% for error in form.email.errors %}
-                <p class="help is-danger">{{ error | escape }}</p>
+                <p class="help is-danger" id="desc_email">{{ error | escape }}</p>
```
mouse-reeve, line 59: I think these ids should be in a parent element, so that when there are multiple errors, they don't have duplicate ids

joachimesque: I hadn't thought about that, thank you!

mouse-reeve:
This reads so nicely for me with VoiceOver! Thank you

PR #1639, opened 2021-12-01
add translation strings

### LINKED ISSUE #1637: Some translation strings missing/not applied
**Describe the bug**
Some translations are not being applied.

**To Reproduce**

change Bookwyrm language to other than English (tested with Lithuanian)

https://ziurkes.group.lt/user/athinkingmeat/books/read
https://ziurkes.group.lt/user/athinkingmeat/books/reading
https://ziurkes.group.lt/user/athinkingmeat/books/to-read

**Expected behavior**
All these links should have "read", "currently reading" and "to read" strings translated, but they are shown in English

**Screenshots**

image

image

image

**Instance**

https://ziurkes.group.lt/

**Additional context**
Probably is a problem with other languages as well

hughrun:
- add logic when default shelves used
- add trans strings in a few spots they were missing
- increase timeout on oclc isni api 

Fixes #1637


bookwyrm/templates/user/user.html
```diff
@@ -49,7 +54,7 @@ <h3>{{ shelf.name }}
 
 {% if goal %}
 <div class="block">
-    <h2 class="title">{% now 'Y' %} Reading Goal</h2>
+    <h2 class="title">{% now 'Y' %} {% trans 'Reading Goal' %}</h2>
```
mouse-reeve, line 57: this one should be a blocktrans I think -- you can save the result of the `now` tag by doing `{% now 'Y' as current_year %}` to get a variable you can use in blocktrans

hughrun: I don't want to be a pain, but now you've pointed this out I just noticed a potential (niche) problem with the template.

`now` for the goal comes from the server. `now` for the template comes from the client. So when I'm excited about my new year resolution on the morning of 1 January in Australia and log on to bookwyrm.social, hosted in the USA, it will tell me I have already read a bunch of things towards my "2022" reading goal, but I haven't set one yet, and I can't until the afternoon.

We probably should pass `current_year` through as a string from the server with the reading goal number?

hughrun: Ah, perfect. I saw it was locale aware and made an assumption. Maybe I should set a reminder and find out for sure in a month haha.

PR #1640, opened 2021-12-01
Comment out trailing block ending in production nginx config template

NonSparkly:
For people installing an instance with only the reverse proxy server, the hidden trailing `}` at the end of the second server block is quite hard to catch and it took me a good while to figure it out. Having the entire server commented out makes the whole process more understandable in my opinion.

### DIFF
--- a/nginx/production
+++ b/nginx/production
@@ -17,7 +17,8 @@ server {
 
 #     # redirect http to https
 #     return 301 https://your-domain.com$request_uri;
-# }
+}
+
 #
 # server {
 #     listen [::]:443 ssl http2;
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you! It's always a struggle to balance readability for different use cases :)

NonSparkly:
I realize now that the reason I got caught out by this twice is because I did the installation via SSH / CLI / Nano, which means I had absolutely no IDE-help which should have caught something like this. Even so, I did the exact same thing twice and even though it just took me a couple of minutes, it could have taken me hours instead.

Thanks for being my first foray into Open Source, this project is the first one that I've added anything to! :D

PR #1647, opened 2021-12-04
List embed

### LINKED ISSUE #1646: Give users the option to embed lists in other websites
Similar to how spotify playlists can be embedded in websites, it would be cool for users to be able to embed a list in their own blog or website.

I think this is pretty straightforward, and would entail:
1. Creating an alternate list view that uses the same Django view with a simplified template designed to look nice in an embed
2. Generating an iframe html that can be copied and pasted

(suggested by Erik Kemp on matrix)

joachimesque:
Closes  #1646

A new List view is created with the following properties:

- Can be called as an `iframe` from unsafe origins (the `X-Frame-Options` default header is disabled)
- The new unique `embed_key` List field allows us to bypass the privacy filters in order for any list to be displayed.\
  That key is only displayed in the embed code, which is displayed on the list page. In the case of a private list, only the list author will have access to the embed code (as they will be the only user allowed on the list page).\
  More checks can be added, but I felt that this method allows for some use cases (private lists displayed elsewhere by their owner), while any lists safety is protected. (I may have missed some use cases though 🤔 )

Capture d’écran 2021-12-04 à 15 55 21


bookwyrm/templates/layout.html
```diff
@@ -34,7 +34,7 @@
     <div class="container">
         <div class="navbar-brand">
             <a class="navbar-item" href="/">
-                <img class="image logo" src="{% if site.logo_small %}{% get_media_prefix %}{{ site.logo_small }}{% else %}{% static "images/logo-small.png" %}{% endif %}" alt="Home page">
+                <img class="image logo" src="{% if site.logo_small %}{% get_media_prefix %}{{ site.logo_small }}{% else %}{% static "images/logo-small.png" %}{% endif %}" alt="{% blocktrans with site_name=site.name %}{{ site_name }} home page{% endblocktrans %}">
```
mouse-reeve, line 37: good catch!

bookwyrm/models/list.py
```diff
@@ -43,6 +45,7 @@ class List(OrderedCollectionMixin, BookWyrmModel):
         through="ListItem",
         through_fields=("book_list", "book"),
     )
+    embed_key = models.UUIDField(unique=True, null=True, editable=False)
```
mouse-reeve, line 48: I think you can set `default=uuid4` here, and you won't have to compute it manually or add it to the migration

bookwyrm/static/js/bookwyrm.js
```diff
@@ -430,4 +433,38 @@ let BookWyrm = new class {
...
+
+        copyButtonEl.textContent = textareaEl.dataset.copytextLabel;
+        copyButtonEl.classList.add(
+            "mt-2",
+            "button",
+            "is-small",
+            "is-fullwidth",
+            "is-primary",
+            "is-light"
+        );
```
mouse-reeve, line 459: why generate this element in javascript as opposed to putting it in the template?

bookwyrm/views/list.py
```diff
@@ -200,6 +210,63 @@ def post(self, request, list_id):
         return redirect(book_list.local_path)
 
 
+class EmbedList(View):
+    """embeded book list page"""
+
+    def get(self, request, list_id, list_key):
+        """display a book list"""
+        book_list = get_object_or_404(models.List, id=list_id)
```
mouse-reeve, line 218: I think it makes sense to check `book_list.raise_visible_to_user(request.user)` to ensure that the viewer is authorized to see the list

bookwyrm/views/list.py
```diff
@@ -200,6 +210,63 @@ def post(self, request, list_id):
...
+
+        embed_key = str(book_list.embed_key.hex)
+
+        if list_key != embed_key:
+            raise Http404()
+
+        query = request.GET.get("q")
```
mouse-reeve, line 225: is this used for anything? normally I think it would be searching to add books to a list, which you wouldn't do in this context. I think you can omit it

mouse-reeve (commented):
This looks so good!

bookwyrm/models/list.py
joachimesque: From what I tested, the `default` value is not unique. I may have overlooked something, but the addition in the migration was the best way I saw to handle the uniqueness requirement.

bookwyrm/static/js/bookwyrm.js
joachimesque: If the client doesn't use JS, the button in the template will do nothing. In that case I prefer showing only a textarea in which the user can select and copy manually and no expectation of interactivity with a button. If the client uses JS, it can add the button with interactivity.

bookwyrm/views/list.py
joachimesque: I chose not to do it, based on a use case that might happen:

I'm a BookWyrm user who has a password-protected private blog. On Bookwyrm I make a private list, that I want to embed on my blog, so that the readers of my blog—non Bookwyrm users—will be able to see it. Checking `book_list.raise_visible_to_user(request.user)` for the display of the embed will guarantee that my friends won't be able to see the contents of this private list.

On the other hand, the use of an UUID key that's unique to that list will keep brute-force attempts at displaying private lists (by guessing the list id). That key is only displayed to the list creator if the list is private. There could be further limits to the visibility of the embed code for non-public lists. An alternative (that would complicate things) would be to generate a new key for each share (I think that's what Nextcloud or Google Documents do)—so that the key can be easily revoked.

On the other hand, if that use case is judged too risky for privacy, we'll have to restrict the embedding to Public lists only.

bookwyrm/views/list.py
joachimesque: Ooh yes you're right, I thought it was part of the List view

bookwyrm/views/list.py
mouse-reeve: Gotcha! I think what you've done makes sense, carry on :)

bookwyrm/models/list.py
mouse-reeve: Weird! I would never have suspected it wouldn't be unique -- thanks for looking into it

PR #1660, opened 2021-12-05
Button to manually load remote data for editions and authors

### LINKED ISSUE #591: re-fetch book info from source
If additional info is added to the source lib, e.g. Open Library, it won't be accesible in bookwyrm. 

Have a possiblity to update the book data manually, e.g. by a button "Re-fetch data" next  to the "View on Open Library"-link.

Another possibility would be to automatically check for new data periodically but it'd be a lot of request of which most wouldn't do anything.

**Additional context**
Examples:

1. There's a book on bookwyrm that doesn't have a cover, but it does have a cover now on Open Library.
2. There's a book on bookwyrm where the author is missing. It's been corrected on Open Library.

mouse-reeve:
Looks up the remote data for a book or author, and copies over any fields that are found in the remote copy but not in the local copy. No data should be overwritten locally.

Fixes #591 

Remaining work:
- [x] Tidy up UI and add explanatory text
- [x] Screen reader review
- [x] Tests
    - [x] ISNI loaded from OL
    - [x] Connector update functions
    - [x] Connector url generator functions
    - [x] views

Currently this happens synchronously, which I think is fine?

Screen Shot 2021-12-07 at 1 42 44 PM
Screen Shot 2021-12-07 at 1 42 52 PM


mouse-reeve:
@hughrun I think I can expand this to isni as well, but it will probably involve moving a bunch of code around. It also turns out that isni wasn't being fetched from openlibrary 😓, so that's fixed here too

hughrun:
> @hughrun I think I can expand this to isni as well, but it will probably involve moving a bunch of code around. It also turns out that isni wasn't being fetched from openlibrary 😓, so that's fixed here too

If I remember correctly, I set up authors so that any time an author with an ISNI is edited, it silently checks and enhances the Bookwyrm record. Making that an explicit thing requiring a button click is probably sensible, instead of it just mysteriously happening.

PR #1672, opened 2021-12-10
Removes unused groups view

mouse-reeve:
And more tests for the view

### DIFF
--- a/bookwyrm/models/group.py
+++ b/bookwyrm/models/group.py
@@ -73,7 +73,7 @@ class GroupMember(models.Model):
             )
         ).exists():
             raise IntegrityError()
-        # accepts and requests are handled by the GroupInvitation model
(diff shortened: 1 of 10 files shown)


mouse-reeve:
I've been going through the tests for #1613 and noticed that this doesn't appear to be used anywhere, looks like it duplicated UserGroup -- unless I'm missing something, @hughrun ?

hughrun:
> I've been going through the tests for #1613 and noticed that this doesn't appear to be used anywhere, looks like it duplicated UserGroup -- unless I'm missing something, @hughrun ?

Whoops, I think you're right. I must have changed the way this worked after I wrote the tests and neglected to clean this one up.

hughrun:
thanks for cleaning up my mess 😀

PR #1677, opened 2021-12-12
removing quotes as they are interpreted as part of a value

nycterent:
quotes in env file are treated as a part of variable, removed.

fixes bookwyrm.preview_images.generate_edition_preview_image_task[6784c797-f91f-4d56-9cf3-3ffae5587e3f] raised unexpected: ValueError('unknown color specifier: '"#363636"'')


mouse-reeve:
thank you for troubleshooting this!

PR #1686, opened 2021-12-17
Adds sync_media_to_s3 command

cincodenada:
This just uses the AWS CLI `sync` command instead of `copy`. Sync checks against existing files in the bucket before uploading, so it's a good bit slower for an initial upload (for which `copy` should be used instead), but way faster for syncing a few updated files to an already-populated bucket. There's a companion PR over at https://github.com/bookwyrm-social/documentation/pull/50 that documents its use.

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -155,13 +155,13 @@ case "$CMD" in
             --acl public-read" "$@"
         ;;
     set_cors_to_s3)
-	set +x
+        set +x
         config_file=$1
         if [ -z "$config_file" ]; then
             echo "This command requires a JSON file containing a CORS configuration as an argument"
             exit 1
         fi
-	set -x
+        set -x
         awscommand "$(pwd):/bw"\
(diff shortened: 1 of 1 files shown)


bw-dev
```diff
@@ -140,20 +140,33 @@ case "$CMD" in
...
+        awscommand "bookwyrm_media_volume:/images"\
+            "s3 sync /images s3://${AWS_STORAGE_BUCKET_NAME}/images\
+            --endpoint-url ${AWS_S3_ENDPOINT_URL}\
+            --acl public-read" "$@"
         ;;
     set_cors_to_s3)
+        set +x
+        config_file=$1
+        if [ -z "$config_file" ]; then
+            echo "This command requires a JSON file containing a CORS configuration as an argument"
```
joachimesque, line 161: Great addition!

mouse-reeve:
thank you!

PR #1690, opened 2021-12-19
add "Create list" form to group template

### LINKED ISSUE #1666: button to add group lists from group page
At the moment if you want to add a group-curated list, you have to do it from the Lists page, rather than the home page of the Group. This makes sense if you know how the backend functionality works but is not very intuitive.

  Add a `Create list` button (like the one in the lists view) to the Group page.

Leaving the current, obscure, functionality.

**Additional context**
I intended to add this once the initial group functionality was completed, but I forgot about it.
This Issue is to remind me to put in a PR to add this.

hughrun:
Fixes #1666


hughrun:
😭 this is causing tests to fail claiming improper use of `<label>` but I can't work out why, given the same form works fine elsewhere.
```python
{% if request.user.is_authenticated %}
<div class="block content">
    {% include 'lists/create_form.html' with controls_text="create_list" curation_group=group %}
</div>
{% endif %}
```

mouse-reeve:
I think #1321 will fix it -- the reason the test was passing elsewhere that that markup is that it wasn't being checked

mouse-reeve:
whew I think I didn't totally jank up that merge -- the remaining test failure is because both the create and edit forms use the same IDs, so I think the input/textarea elements have to be hard coded with a differentiating value in the id (like group id) to make the tests happy.

hughrun:
> whew I think I didn't totally jank up that merge -- the remaining test failure is because both the create and edit forms use the same IDs, so I think the input/textarea elements have to be hard coded with a differentiating value in the id (like group id) to make the tests happy.

Yeah that makes sense. I noticed this myself eventually but it didn't seem to make any difference. Will clean it up and see what happens! Thanks for #1321

hughrun:
Right. Fixed the ID problem, plus a couple of other things I had neglected to check around visibility of things in the group user suggestion view.

mouse-reeve:
This works beautifully! What do you think about putting the create list button next to the lists header, rather than at the top?

hughrun:
> This works beautifully! What do you think about putting the create list button next to the lists header, rather than at the top?

Hmm that probably makes more sense. Stay tuned.

hughrun:
hmm I thought maybe I'd not pulled in something but it appears there's something new that's making tests fail. Fairly sure it's nothing I did! But ...maybe it is.

mouse-reeve:
Oh yeah that is definitely not your doing! That's not a well written test - I'll fix it

mouse-reeve:
This should fix it: https://github.com/bookwyrm-social/bookwyrm/pull/1760

mouse-reeve:
Woo! Looks great! Thank you

PR #1691, opened 2021-12-20
Feature: Create annual summary page

joachimesque:
## What?

A page with a little summary of your books this year

## How does it look?

image

image


bookwyrm/views/annual_summary.py
```diff
@@ -0,0 +1,79 @@
...
+            "pages"
+        )
+        pages_list = [p.pages for p in read_books_with_pages]
+
+        pages_total = 0
+        pages_average = 0
+        book_pages_lowest = 0
+        book_pages_highest = 0
+        if len(pages_list) > 0:
+            pages_total = sum(pages_list)
```
mouse-reeve, line 42: You can get these page stats using the queryset, rather than using a list:
```python3
from django.db.models import Max, Min, Avg, Sum
stats = read_books_with_pages.aggregate(Max("pages"), Sum("pages"), Min("pages"), Avg("pages"))
```

And you'll get an object with the stats in this format:
```
{'pages__max': 3990, 'pages__min': 1, 'pages__sum': 129623, 'pages__avg': 451.64808362369337}
```

bookwyrm/views/annual_summary.py
```diff
@@ -0,0 +1,79 @@
...
+            .exclude(deleted=True)
+            .exclude(rating=None)
+            .filter(book_id__in=read_book_ids_in_year)
+        )
+        best_ratings_books_ids = [review.book.id for review in ratings.filter(rating=5)]
+
+        rating_average = 0
+        if len(ratings) > 0:
+            ratings_list = [review.rating for review in ratings]
+            rating_average = round(sum(ratings_list) / len(ratings_list), 2)
```
mouse-reeve, line 62: You can use `ratings.aggregate(Avg("rating"))` here as well

mouse-reeve (changes requested):
This is going to be so cool! I added a couple django queryset things you can use to avoid making calculations manually.

mouse-reeve:
I'm going to try to get this live today, so excited for it!

PR #1699, opened 2021-12-27
Autocomplete script

mouse-reeve:
Quick proof-of-concept for text box autocomplete 

Remaining work:
 - [x] Decouple the file types trie from the general script
 - [x] Include the script files in the appropriate places


joachimesque:
I think that you could reduce the complexity of this feature by using the `datalist` element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist

PR #1706, opened 2021-12-28
Fixes: Annual summary book list

joachimesque:
- Fixed image sizing and quality
- Fixed book list order


mouse-reeve:
thanks for the quick fixes -- there's nothing quite like seeing things with real data in them

PR #1711, opened 2021-12-28
Fixes the year for the All Books heading on the Annual Summary page.

Flameborn:


bookwyrm/templates/annual_summary/layout.html
```diff
@@ -40,7 +40,7 @@
 
     <h1 class="title is-1 is-serif has-text-centered">
         📚✨
-        {% blocktrans %}{{ year }} <em>in the books</em>{% endblocktrans %}
```
mouse-reeve, line 43: I think the original text was correct here?

bookwyrm/templates/annual_summary/layout.html
```diff
@@ -224,7 +224,7 @@ <h2 class="title is-3 is-serif">
     <div class="columns">
         <div class="column has-text-centered">
             <h2 class="title is-3 is-serif">
-            {% blocktrans %}All the books {{ display_name }} read in 2021{% endblocktrans %}
```
mouse-reeve, line 227: thank you! good catch

mouse-reeve: ah @joachimesque beat you to the punk on this one with #1710 -- thank you both for catching this and fixing it so quickly

PR #1714, opened 2021-12-28
Redirect /@<username> to /user/<username>

### LINKED ISSUE #1685: deal with requests for /@username
### Problem
Mastodon puts public accounts at `example.org/@username`. People might expect bookwyrm to do the same.

### Solution
Redirect `/@mouse` to `/user/mouse`. With regards to code, this sounds pretty easy. Keep in mind that this uses up any URL path starting with "@".

### Other Stuff
Come to think of it, why _is_ the user profile at `/user/mouse`? The at-symbol ("@")  is rarely used in URL paths otherwise. It's a nifty way of shortening the URL while keeping other URL paths safe.

joachimesque:
Closes #1685

### DIFF
--- a/bookwyrm/urls.py
+++ b/bookwyrm/urls.py
@@ -276,6 +276,7 @@ urlpatterns = [
     # users
     re_path(rf"{USER_PATH}\.json$", views.User.as_view()),
     re_path(rf"{USER_PATH}/?$", views.User.as_view(), name="user-feed"),
+    re_path(rf"@(?P<username>{regex.USERNAME})$", views.user_redirect),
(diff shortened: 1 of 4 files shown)


bookwyrm/urls.py
```diff
@@ -276,6 +276,7 @@
     # users
     re_path(rf"{USER_PATH}\.json$", views.User.as_view()),
     re_path(rf"{USER_PATH}/?$", views.User.as_view(), name="user-feed"),
+    re_path(rf"@(?P<username>{regex.USERNAME})$", views.user_redirect),
```
mouse-reeve, line 279: This just needs a `^` at the beginning of the string, otherwise it will match any url that ends in `@<username>`

PR #1717, opened 2021-12-28
Remove last lingering traces of fedireads name

cincodenada:
The defaults should be irrelevant if there's an env file in place, but the backup script actually will break backups, I think.

I also was blown away at what `apt-get install cron` was bringing with it, after some digging I added the flag to not install "recommended" dependencies which gets us a bunch of stuff we definitely don't need.

### DIFF
--- a/bookwyrm/management/commands/populate_streams.py
+++ b/bookwyrm/management/commands/populate_streams.py
@@ -6,7 +6,7 @@ from bookwyrm import activitystreams, models
 def populate_streams(stream=None):
     """build all the streams for all the users"""
     streams = [stream] if stream else activitystreams.streams.keys()
-    print("Populations streams", streams)
+    print("Populating streams", streams)
     users = models.User.objects.filter(
         local=True,
         is_active=True,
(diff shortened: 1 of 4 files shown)


mouse-reeve:
Can you open a duplicate of this PR relative to `main`? Some of these files are canonical to that branch and copied form there to `production` (everything except the `postgres-docker` stuff I think)

cincodenada:
Done - but for my understanding, what do you mean by "canonical to that branch"? I'm accustomed to `main` having everything that `production` has in it, plus new stuff that's not ready for production yet, so I'm confused by the existence of things in `production` that _aren't_ in `main`.

mouse-reeve:
There are a few ways that production is configured differently (like using the `postgres-docker` configuration), plus some thing available in main aren't available in production (like `resetdb`, and docker installs fewer libraries) -- it would be clearer to have everything in main, and it's gotten closer to that (with the .env.prod vs .env.dev, for example), but as you've found, there is still divergence.

cincodenada:
Cool, thanks for the context there! Do you think it's worthwhile to continue thinking about ways to converge those further? I could see it being nice to have everything switch based on e.g. `DEBUG=true/false` or something, but I'm not the one steering the ship here and I don't wanna aim somewhere that's not your vibe.

mouse-reeve:
Yeah, I think having them converge further would be great! Ops is... not my passion, so a little help steering is welcome :)

PR #1725, opened 2021-12-29
Front-end: On book page, click on cover to show/add image

### LINKED ISSUE #1009: When clicking on a cover, show it enlarged or allow to upload one if it’s missing.
> do you mean the one to add the cover? I've been thinking for a while that it might be nice to have an "enlarge cover" button on the cover, which could be the add cover button if the cover is absent

_Originally posted by @mouse-reeve in https://github.com/bookwyrm-social/bookwyrm/pull/994#discussion_r619850052_

joachimesque:
Closes #1009

- Cover show modal uses new modal JS
- Cover add modal uses old modal for now


mouse-reeve:
This is really nice -- how would you feel about including an overlay on hover that indicates what will happen when you click? Since the behavior is different than elsewhere in the application, I think that would be clarifying

joachimesque:
I kept it simple: 

image

mouse-reeve:
Simple is great! Could you add an equivalent one for the state when you add a cover?

The only issue I'm seeing now is that when I click to enlarge a cover, and then click on the background to close the modal, the "Click to enlarge" overlay stays on the image until I click somewhere else on the page. It's super low-stakes, so if it's not something that's easily fixed I think it's fine to just not worry about it

joachimesque:
> when I click to enlarge a cover, and then click on the background to close the modal, the "Click to enlarge" overlay stays on the image

Indeed, that was because the overlay reacted to the `:focus` state. When the modal closes, the focus returns to the button, but now I changed the reaction to `:focus-visible`, which works only on keyboard focus.

PR #1728, opened 2021-12-29
Front-end: Update Feed summary card style

joachimesque:
Now the header's background is contained by the border-radius. I took this approach because it's not possible to pass variables up to an extended template


mouse-reeve:
those smoothed corners are so soothing

PR #1736, opened 2021-12-30
Switch old modals to new modals

mouse-reeve:
Stoked about this! The markup is so much simpler and clear this way -- thank you @joachimesque 

Once all the modals are converted, the `new_modal.html` can be the only modal, but it'll be less of a drag to do this incrementally in the meantime.

grep says these are left:
- [x] `bookwyrm/templates/groups/delete_group_modal.html`
- [x] `bookwyrm/templates/snippets/delete_readthrough_modal.html`
- [x] ~`bookwyrm/templates/snippets/reading_modals/layout.html`~ (separate PR)
- [x] `bookwyrm/templates/snippets/report_modal.html`

Remaining work:
 - [ ] In the no script state for reading modals, the close buttons don't work


mouse-reeve:
I'm going to merge this despite the close buttons, because those views should be used very rarely, and a PR like this that touches a lot of files could become a nightmare to maintain in term of merge conflicts.

PR #1738, opened 2021-12-31
Refactors reading modals

mouse-reeve:
There are a couple tweaks here related to particular behavior in these modals:
- There's a landing page as a javascript fallback that displays the modals in an always-open state, so the modal component has an option to include the `is-active` class at page load
- Because the buttons submit as forms making GET requests, I added `event.preventDefault()` to the javascript that handles opening the modals, so that the forms aren't submitted when javascript is available

Remaining work:
- [ ] Now the close buttons don't work in the new modal in the no script state


mouse-reeve:
I'm merging this into #1736

PR #1739, opened 2021-12-31
Don't do fancy javascript clicks on the notification page

### LINKED ISSUE #1733: Notifications blocks accessibility and usability
I noticed that the notifications have a problem since PR #1485

The clickable blocks are not accessible:

- a block cannot be accessed with the keyboard:
    - there's no `tabindex`
    - there's no reaction to keyboard Enter and Space key
- the ARIA role isn't set
- it isn't read by screenreaders as a link
- as it's not a link, the url isn't shown (in the browser status bar) on hover
- the block doesn't react on hover
- cmd-click or ctrl-click don't open the link in a new tab

Plus, when clicking "Follow" or unfolding the dropdown, the block click is fired, even though it shouldn't be.

The problem is interactive element nesting. Here's a write up on the subject, by Adrian Roselli: https://adrianroselli.com/2016/12/be-wary-of-nesting-roles.html

The best way forward would be to revert the changes of #1485 relating block click.

mouse-reeve:
Fixes #1733

Screen Shot 2021-12-30 at 7 47 47 PM


joachimesque (approved):
👍

PR #1747, opened 2022-01-01
update awkward translation

reesporte:
idk if this is how you want these to be done now, since you're using crowdin, but i just figured i'd try both lol :) 

lmk which way works best for you and i can just do that in the future


mouse-reeve:
Thanks for being so thorough! Crowdin is the place to make the change though -- it should get merged into this file soon

PR #1750, opened 2022-01-01
Default to hidden state for local storage visibility toggle

mouse-reeve:
This will prevent the card for showing briefly before javascript loads


cincodenada (approved):
Simple enough!

PR #1751, opened 2022-01-01
Adds OpenTelemetry exporter for use with various monitoring tools

cincodenada:
I wanted to give [Honeycomb.io](https://ui.honeycomb.io/) a try, and their setup guide involved OpenTelemetry, which is apparently a standard that you can implement to send traces and telemetry to a bunch of different monitoring tools - DataDog and NewRelic will import it too. Neat!

There were already Django and Celery extensions avaialble, so I just got those integrated with the exporter, and now I have request data dumping into my monitoring app. In theory this should allow folks to send request data to any client that ingests OTLP, which seems like a lot of them!

charts

I wasn't sure where to put "run at startup" code, I didn't see anywhere we had anything quite like that already. So I consulted StackOverflow and Django things and it seems like the `ready()` callback in `apps.py` is the place for such things. 

We reliably load `settings.py` at startup and I looked into stuffing it in there but decided against that because it's settings, not initialization, we currently don't do any initialization logic beyond setting variables there, and we _do_ do stuff like load Bookwyrm's settings into Celerywyrm that might do weird stuff if we did init things in settings.

I also didn't want to put it in `wsgi.py` because then it won't run if you use `runserver` I think, and because Celerywyrm has `wsgi` and `asgi` in addition to the manual run, so none of that seemed like the right place.

If there's somewhere better that I missed do let me know! I think I set up `apps.py` correctly though - it looks like if Django doesn't find an `apps.py` it just generates its own app name and things, and since nothing broke horribly I'm assuming I set it up right :sweat_smile:


bookwyrm/settings.py
```diff
@@ -233,3 +233,7 @@
     MEDIA_FULL_URL = f"{PROTOCOL}://{DOMAIN}{MEDIA_URL}"
     STATIC_FULL_URL = f"{PROTOCOL}://{DOMAIN}{STATIC_URL}"
     MEDIA_ROOT = os.path.join(BASE_DIR, env("MEDIA_ROOT", "images"))
+
+OTEL_EXPORTER_OTLP_ENDPOINT = env("OTEL_EXPORTER_OTLP_ENDPOINT")
```
mouse-reeve, line 237: Can you add default values (presumably of `None`) to these? That way there won't be an error in `bookwyrm/apps.py` for installations where the variables are absent from the env file -- otherwise admins will have to manually copy over the lines from `.env.prod.example` to .`env`

mouse-reeve (changes requested):
Thank you for adding this!

PR #1753, opened 2022-01-02
increasing rsa key size

nycterent:
should address this issue:

Creation of an RSA key uses  bits, which is below 2048 and considered breakable.

Not sure if that breaks anything - have to be careful..

### DIFF
--- a/bookwyrm/signatures.py
+++ b/bookwyrm/signatures.py
@@ -15,7 +15,7 @@ MAX_SIGNATURE_AGE = 300
 def create_key_pair():
     """a new public/private key pair, used for creating new users"""
     random_generator = Random.new().read
-    key = RSA.generate(1024, random_generator)
+    key = RSA.generate(2048, random_generator)
     private_key = key.export_key().decode("utf8")
     public_key = key.public_key().export_key().decode("utf8")
 


bookwyrm/signatures.py
```diff
@@ -15,7 +15,7 @@
 def create_key_pair():
     """a new public/private key pair, used for creating new users"""
     random_generator = Random.new().read
-    key = RSA.generate(1024, random_generator)
+    key = RSA.generate(2048, random_generator)
     private_key = key.export_key().decode("utf8")
     public_key = key.publickey().export_key().decode("utf8")
```
jaschaurbach, line 20: With version bum of pycryptodome to 3.9.4 key.publickey needs to change to key.public_key.

jaschaurbach: I think merge of main branch will do this.

mouse-reeve:
@jaschaurbach thank you so much for the review!
@chdorner do you have the capacity to do a final check and merge this?

chdorner:
Sounds good, I will have a look. Should be able to test this with my local federation setup 👍 

Is there something we should / can do for existing users? I'm assuming upgrading / effectively replacing the keypair for existing users is probably not as easy, do you maybe know @jaschaurbach?

chdorner (approved):
Thank you so much! 🙌 

Just a note to myself: I tested this locally with two federated instances, two users still had an old 1024-bit key, and one user had the new 2048-bit key. Those users were able to follow each other, get statuses federated into their timelines, etc.

If there is something we can and should do about existing users with an old 1024-bit key, then we can do that in a follow-up PR, but I suspect that won't be a straightforward task 🙂

jaschaurbach:
@chdorner in theory the keys can be rotated. The moment the keys are rotated all messages will be signed with the new key. 

Important: The key generated for the instance will be used to sign the keys generated per user. This one needs to be changed first. The key generated for the instance will make use of the secret in .env. This is why it needs to be unique. 

I have no two instance setup running so I can not check this atm.

PR #1768, opened 2022-01-05
Cache queries in feed view

mouse-reeve:
This about halves the number of database queries required to load the feed page.


mouse-reeve:
Tagging @cincodenada and @hughrun for a reality check on this. I'm not sure what the limitations are for using the default cache at prod scale? But it's nice to see such a substantial reduction in queries.

hughrun:
@mouse-reeve Apologies, I got a bit distracted from looking at caching. The docs specifically warn against using the default in-memory cache for production, which is what we have here. I did a bit of experimenting with Memcached which was definitely fast, but it occurs to me that given we are already using Redis, that's the obvious choice. 

I was going to see what we need to install to get Redis working for caching and it looks like the answer might be "several python packages", but it appears that if we can upgrade Django to 4.0 without breaking anything else, that allows us to use Redis for caching out of the box.

Alas, when I try that, I get an error:
`_chain() got an unexpected keyword argument 'subclasses'`

cincodenada:
> it appears that if we can upgrade Django to 4.0 without breaking anything else, that allows us to use Redis for caching out of the box.

Unfortunately, as you discovered, that's not really practical - it may be worth looking into just to keep up with things, but much of the purpose of a major version change (3.x -> 4.x) is specifically to allow for breaking changes, so doing a major version upgrade of your framework is almost always going to be a bit of an undertaking, since various APIs and ways of doing things are gonna change out from under you.

cincodenada:
Generally though, I agree with the idea of using Redis since we have it. It looks like [https://github.com/jazzband/django-redis](django-redis) is an actively-maintained Redis caching backend for Django - if I understand correctly, switching the default cache to that shouldn't be much more than adding `django-redis` to `requirements.txt` and updating the cache configuration according to their docs, but I haven't actually tried that yet.

mouse-reeve:
Thank you both for your thoughts on cache background! I like using redis too, always happy not to add another dependency :)

I would like to upgrade to 4.0 but based on their list of breaking changes, it will be a bit of a task. But switching from a library redis cache backend to the native one should be relatively straightforward, since it's a matter of switching out the backend rather than making code changes.

I went a little beyond the scope and moved the dev setup for redis to be the same as the prod one, with passwords set. I figure that'll be easier to maintain and get prod and main a little closer. And I'm not sure if using the existing redis instance is the most correct approach.

mouse-reeve:
This should reduce the number of queries on the feed page after the initial page load by 85%. It does appear now that clearing the cache also clears users's feeds (which makes total sense but didn't occur to me at first), so that would be a surprising side effect for an admin who wanted to clear the cache for whatever reason.

The one thing I'm unsure about and would love to get feedback on is caching the don't have timeouts -- as written, a number of queries are cached forever until they are invalidated in a model or view. I'm not sure if it's better to have a very long timeout for those so that the app isn't caching data on users who, for example stopped, using the service. On the other hand, I think of bookwyrm as a very slow-paced application, where it's reasonable to expect data to only update every few months if you're reading _Middlemarch_ or something.

hughrun:
I think you're right about it being fine to be a low frequency site. However I think we don't want to create a situation where the cache slowly fills up for large sites. A very long cache expiry time could be a reasonable compromise, like 3 or 6 months or something.

PR #1769, opened 2022-01-05
Change thumbnail generation strategy

joachimesque:
I've looked at the thumbnail generation settings. Right now it's working with the default config:

```
IMAGEKIT_DEFAULT_CACHEFILE_BACKEND = 'imagekit.cachefiles.backends.Simple'
IMAGEKIT_DEFAULT_CACHEFILE_STRATEGY = 	'imagekit.cachefiles.strategies.JustInTime'
```

(from https://django-imagekit.readthedocs.io/en/latest/configuration.html)
- The `JustInTime` strategy generates the thumbnail file when it's existence is needed (if not already generated)
- The `Simple` backend uses the main process to generate the thumbnail files

So basically, if you have a page with loads of books with a thumbnail size that haven't been generated yet, the server will generate them on the fly—hence, the timeouts I experienced on my Annual Summary page

~This PR switches to a `Celery` backend, and an `Optimistic` strategy, where the thumbnails will be generated asynchronously (through Celery) when the cover image is uploaded/the book is imported~

~Because the `Optimistic` strategy disables the check to see if the image has been generated, if we change the strategy from `JustInTime` we'll have to generate all the images with `./bw-dev generate_thumbnails` (I think)~

This PR switches the default strategy to a custom strategy that will generate thumbnails on source creation (when the cover image is uploaded/book is imported), as well as when the thumb is required but didn't exist

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -191,7 +191,7 @@ USER_AGENT = f"{agent} (BookWyrm/{VERSION}; +https://{DOMAIN}/)"
 ENABLE_THUMBNAIL_GENERATION = env.bool("ENABLE_THUMBNAIL_GENERATION", False)
 IMAGEKIT_CACHEFILE_DIR = "thumbnails"
 #IMAGEKIT_DEFAULT_CACHEFILE_BACKEND = "imagekit.cachefiles.backends.Celery"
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Thank you for looking into this!

PR #1779, opened 2022-01-07
Fix follow recommendation modal failing on signup

hughrun:
You can't concatenate a list with a queryset, nor two querysets, apparently.


mouse-reeve:
Oops! #1767 definitely caused this one

PR #1788, opened 2022-01-07
Moves to a single example .env file

mouse-reeve:
The prod and dev ones hae gotten so close that it seems unnecessary to
have two different ones.

Documentation needs to be updated in parallel with this change.

This was the diff between the two files:
``` zsh
% diff .env.prod.example .env.dev.example 
5,6c5,6
< DEBUG=false
< USE_HTTPS=true
---
> DEBUG=true
> USE_HTTPS=false
9c9
< EMAIL=your@email.here
---
> #EMAIL=your@email.here
```


mouse-reeve:
Likely relevant to your interests @cincodenada

PR #1792, opened 2022-01-08
save language preferences with cookies that last

### LINKED ISSUE #1756: language settings don't persist past browser sessions
**Describe the bug**
Language settings don't persist past browser sessions. If I change my language settings, and close my browser without logging out, when I open the browser again, the language has reverted to English. However, if I logout and log back in, bookwyrm is in the language I specified. 

**To Reproduce**
( assuming your bookwyrm is already in English)
1. Go to https://bookwyrm.social/preferences/profile
2. Select a language other than English under the Language setting and Save
3. Close the browser window and reopen bookwyrm
4. See that bookwyrm is in English and not the language you selected

**Expected behavior**
I would expect that my language settings are persistent across browser sessions. 

**Instance**
bookwyrm.social

**Additional Context**
I have been able to repro this bug on Safari and Brave on macOS BigSur

reesporte:
(as long as login ones do)

fixes #1756

### DIFF
--- a/bookwyrm/views/helpers.py
+++ b/bookwyrm/views/helpers.py
@@ -1,12 +1,13 @@
 """ helper functions used in various views """
 import re
-from datetime import datetime
+from datetime import datetime, timedelta
 import dateutil.parser
 import dateutil.tz
 from dateutil.parser import ParserError
 
 from requests import HTTPError
 from django.db.models import Q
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Perfect! Thank you for fixing this

Issue #1787, opened 2022-01-07
I want to be able to set the log level independently of prod/dev


cincodenada:
**Is your feature request related to a problem? Please describe.**
Sometimes I want to be able to turn on more verbose logging in production to track down an issue. I'd like to be able to do that via an environment variable

**Describe the solution you'd like**
I'd like a `LOG_LEVEL` environment variable or similar in `.env` that adjusts what types of messages are logged.

**Describe alternatives you've considered**
I think the main alternative is the current setup, where there are (I think?) different logging levels in dev/prod, which makes it hard to track down production issues

**Additional context**
[Matrix conversation](https://matrix.to/#/!zoxBMxLlvIyeEKkHuB:matrix.org/$yBr16_gvZ6pHHhFDUsDArypuxjR7E5qOuV1OgROLQS4?via=matrix.org&via=tchncs.de&via=privacytools.io)

(this is a note-to-self, I literally have an implementation of this sitting around uncommitted on my server, so _definitely_ ping me if I don't get it up as a PR this weekend)

mouse-reeve:
Fixed by #1802

PR #1802, opened 2022-01-10
Make log level configurable, override default logging config

### LINKED ISSUE #1789: We should be logging 500 errors in prod
By default Django doesn't log 500 errors in prod, we should figure out how to make that happen

I would like information about 500 errors logged to the docker log

This is probably accomplishable more generally by making the log level more verbose (see #1787) but it would be nice to have 500s logged by default, regardless.

(this is mostly a note-to-self)

cincodenada:
This should fix #1789 and #1787 - I have yet to actually test this, so don't merge right away, I couldn't figure out how to make it a draft.

The reason we aren't logging 500s in production is becuase Django [by default doesn't log _anything_ in production](https://docs.djangoproject.com/en/3.2/topics/logging/#default-logging-configuration), it just emails. This changes the default to log INFO messages and above regardless of the value of DEBUG - we could change that if we wanted, and, like, only log WARNING/ERROR (which is what it maps 400s and 500s to, respectively) by default in production, by changing what the default log level is depending on the value of DEBUG.

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -124,6 +124,10 @@ LOGGING = {
             'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
             'propagate': False,
         },
+        'bookwyrm': {
+            'handlers': ['console'],
+            'level': os.getenv('LOG_LEVEL', 'DEBUG' if DEBUG else 'INFO').upper(),
+        }
     },
 }
 
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -106,31 +106,31 @@ TEMPLATES = [
     },
 ]
 
-LOG_LEVEL = env('LOG_LEVEL', 'INFO').upper()
+LOG_LEVEL = env("LOG_LEVEL", "INFO").upper()
 # Override aspects of the default handler to our taste
 # See https://docs.djangoproject.com/en/3.2/topics/logging/#default-logging-configuration
 # for a reference to the defaults we're overriding
 LOGGING = {
-    'version': 1,
-    'disable_existing_loggers': False,
-    'handlers': {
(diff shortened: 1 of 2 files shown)


mouse-reeve:
I think the simplest way to fix the tests is to check the same variable that disables the cache, `USE_DUMMY_CACHE` (maybe not with that name) to wrap the block of code that sets the cache. That env variable is set in the test environment, and lets you skip certain configuration pieces that don't make sense in a test environment.

News to me that pylint wants % style notation for logging strings, when it can't abide it anywhere else!

cincodenada:
> I think the simplest way to fix the tests is to check the same variable that disables the cache, `USE_DUMMY_CACHE` (maybe not with that name) to wrap the block of code that sets the cache. That env variable is set in the test environment, and lets you skip certain configuration pieces that don't make sense in a test environment.

Ah, okay, thanks - I'll take a look at that

> News to me that pylint wants % style notation for logging strings, when it can't abide it anywhere else!

Oh yeah, I actually know why this is, I've run into it (or similar) before - the goal of logging stuff is to have as little performance impact as possible when logging is turned off, so that you can add whatever logging you want without bogging things down.

Format strings defeat this, because if you use f-strings Python is gonna build them to pass in to `log.debug()` even if the function doesn't do anything. But with a %-style strings and the parameters as arguments, there's no work to do up front, and `log.debug()` can just not build the string if it's disabled.

cincodenada:
The logging config was actually breaking in non-tests too - see the commit note, turns out I misunderstood how Django's log overriding worked. I thought I gave this a test run, so I don't know what went on there, but I think it should be good now.

mouse-reeve:
Looks good! This will save a lot of headaches debugging people's prod installs

Issue #1785, opened 2022-01-07
`./bw-dev` should always use `run --rm` for running one-off commands instead of `exec`


cincodenada:
**Describe the bug**
`exec` runs things in a currently-running container (failing if there isn't one), whereas `run --rm` runs things in a temporary instance. Currently `bw-dev` uses `exec` for `initdb` and `update`, both of which cause problems.

In `initdb`, if you're not already running an instance - which, if you're running `initdb`, you're often not - `exec` will fail because there's no instance. `run --rm` will work fine as long as the image is built. In particular,  trying to run the setup instructions through `./bw-dev` will not work (see https://github.com/bookwyrm-social/documentation/pull/51/files)

In `upgrade`, things will usually work, unless you've installed a new requirement that is loaded by `manage.py`. In that case, the old docker image will try to run the new code (since it's mounted into the container), and will crash out with a missing requirement.

**To Reproduce**
Run `./bw-dev initdb` without an instance running already, or run `./bw-dev` upgrade after adding a requirement that's required in the Django startup.

**Expected behavior**
Bookwyrm reinits or upgrades successfully

**Instance**
https://kirja.casa (thanks Stoori!)

**Tasks**

- [x] Use `run --rm` in `update` (#1803)
- [ ] Update `initdb` to `run --rm` (#1804)
- [ ] Update tests to not depend on a running container, probably with `--no-deps`

(This is a note-to-self, ping me if I don't put up a PR this weekend)

mouse-reeve:
The only one this will cause any issue for is the tests. In most cases it's easier and much more intuitive to run the tests with `run --rm`, but  because `run` spins up related images, the tests will always have redis running, so tests that need mocks will pass locally but fail in CI. My process for running tests that need mocks is to use `docker-compose stop` to shut down redis and celery, which lets me `exec` the tests on the web image. 

This is janky and complicated! There is probably a better way to isolate the tests? At any rate, I'm 99% completely all for this change.

nycterent:
also up -d should be used when issuing update command

cincodenada:
> because `run` spins up related images, the tests will always have redis running, so tests that need mocks will pass locally but fail in CI. My process for running tests that need mocks is to use `docker-compose stop` to shut down redis and celery, which lets me `exec` the tests on the web image.

[...]

PR #1803, opened 2022-01-10
Use run --rm in `update` command

cincodenada:
Partial fix for #1785 - preivously we would run the migrate and collect
commands with exec, which would run them in the running (and thus old)
contianers, but with the new code. This caused issues when, for example,
new dependencies were introduced, which weren't built into the old
containers.

Instead, use run --rm to spin up temporary instances of the new
containers to do the commands.

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -81,9 +81,9 @@ case "$CMD" in
     update)
         git pull
         docker-compose build --no-cache
-        docker-compose exec web python manage.py migrate
-        docker-compose exec web python manage.py collectstatic --no-input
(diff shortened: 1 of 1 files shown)


cincodenada:
One other thing here - what is the `--no-cache` doing here? I haven't double-checked but I think that may make things much slower, since the three main containers all use the same image, and I don't think it should be necessary if we're doing our images right.

mouse-reeve:
Thank you for updating this! I'm going to add this command to the main branch bw-dev as well, and try to get them a little closer.

PR #1804, opened 2022-01-10
Use run --rm instead of exec for initdb

cincodenada:
This way we don't depend on the containers already being up and running. Another part of #1785


cincodenada:
I think we should also deal with the tests as part of this PR, I just haven't gotten to that bit yet.

mouse-reeve:
@cincodenada I think this is merge-able as is, do you have strong feelings about looking into testing first?

cincodenada:
Oh, thanks for the ping here! Yeah I think you're right, it's fine to merge this and deal with tests separately. I'll file an issue for looking into `--no-deps` for the tests on their own.

PR #1805, opened 2022-01-10
Urlencode the redis password in the cache URI

### LINKED ISSUE #1790: Server Errors if the redis activity password has non-URL-friendly characters in it
**Describe the bug**
If `REDIS_ACTIVITY_PASSWORD` has a `?` (and presumably other similar characters), Redis will fail to start up. This is because the new Redis-backed cache uses a Redis URI for a connection configuration, and doesn't escape the parameters when building said URL.

**To Reproduce**
1. Set REDIS_PASSWORD to something "abc?123"
2. `./bw-dev up -d`
3. Try to load a page
4. Should get "Server Error"

**Expected behavior**
Passwords with special  characters should be just fine

**Instance**
https://kirja.casa (thanks again Stoori!)

(another note-to-self, but like if you're feeling excited about it, go for it 😆)

cincodenada:
This should fix #1790 where we had issues if passwords have non-URL-safe characters in them


cincodenada:
Again, still need to actually test this one fully

mouse-reeve:
I tested this out with a password with `?` in it, and on `main` I get a weird error about the port as a result of the url encoding issue, which was fixed by this code change. But I still got a `WRONGPASS` error on this branch; I manually authenticated in `redis-cli` with the password successfully so I think there may be an encoding issue with using the password, even though the `urlencode` fixes the `LOCATION` field syntax.

mouse-reeve:
I tested this out with a password with `?` in it, and on `main` I get a weird error about the port as a result of the url encoding issue, which was fixed by this code change. But I still got a `WRONGPASS` error on this branch; I manually authenticated in `redis-cli` with the password successfully so I think there may be an encoding issue with using the password, even though the `urlencode` fixes the `LOCATION` field syntax.

mouse-reeve:
It has just been _so_ long, I'm going to close this. I'd be more than pleased to revisit but there are so many pull requests 😅

PR #1818, opened 2022-01-13
add support for different redis db

iamruinous:
I'm trying to support a smaller memory footprint by sharing a redis instance. Added settings to support specifying a redis database other than `0`. 

You can now set `REDIS_ACTIVITY_DB=x` and `REDIS_BROKER_DB=y` to keep them separated.


mouse-reeve:
This looks great! Can you run `./bw-dev black` to make the python formatter happy? Otherwise perfect :)

cincodenada:
Agreed, this is great! And exactly what Redis' logical databases are designed for, so I think it makes a ton of sense. Thanks!

I had a quick thought on naming, reading over this - I think it's unfortunate that Redis decided to name their logical partitions "databases" when the whole thing is also commonly referred to as a "database" (I heard you like databases...), which makes keys like `REDIS_ACTIVITY_DB` ambiguous - it could easily be mistaken for a key that takes a Redis connection string, for instance, which could lead to confusion.

I don't have a great suggestion because again, "database" is the correct term, we're just inheriting the ambiguity, but maybe `REDIS_ACTIVITY_DB_INDEX`? That matches the language the Redis docs use for them (the other option would be `LOGICAL_DB` but I think the meaning of "index" makes it more apparent we're looking for a number).

iamruinous:
Thats a great suggestion, will update the PR shortly using `_DB_INDEX`. I've been testing this out on a small redis instance shared between many different services with great results so far.

mouse-reeve:
Thank you both!

PR #1819, opened 2022-01-13
use correct environment variable for EMAIL_SENDER_DOMAIN

iamruinous:
settings incorrectly uses the `EMAIL_SENDER_NAME` for the `EMAIL_SENDER_DOMAIN`


mouse-reeve:
Oops! My bad, thanks for catching this

PR #1847, opened 2022-01-19
Very minor style updates adressing issues caused by flex-shrink: 0

OhScee:
Long header titles will not wrap in an open modal, and will have the remainder hidden in the overflow.
Form buttons will over extend past their parent.

Updates are minor, but allows containers to properly auto adjust height when wrapping content


mouse-reeve:
Thank you doing this! Could you post some screenshots of what you mean? That'll help me understand which components are causing the problem.

Rather than editing the vendor file from bulma, it would be preferable to override the problematic css in `/bookwyrm/static/css/bookwyrm.css` -- that way the changes won't be lost if we upgrade the library or switch to rendering it from SASS.

mouse-reeve:
I'm going through and closing pull requests that are stalled or out of date -- if you'd like to get this PR back in action, I'm super happy to re-open and work on getting it in sync with the current state of main ❤️

PR #1848, opened 2022-01-19
Use Source Han Sans for preview images generation

joachimesque:


cincodenada:
In addition to tests, assuming we get a font repo going, we'll need to adjust this PR to pull the font into the Dockerfile instead, and rewrite the branch to drop the font from the commit history. If you're not familiar with that kind of git wrangling, let me know and I can go into more detail!

joachimesque:
The tests are now passing. If you want to continue this PR, I suggest you also delete all files in `bookwyrm/public/fonts/public_sans`, as they won't be used anymore

PR #1855, opened 2022-01-21
Makes default language configurable

### LINKED ISSUE #1791: Let admins set default interface language
Right now, BookWyrm's interface language is determined by, in order of preference:
1. The language selected in the user's settings
2. The language requested by the browser, if it's available in the app
3. English

Admin should be able to set the default display language of the interface for logged out viewers and users that don't have a preference saved.

mouse-reeve:
Fixes #1791 - if the browser sends a request for a language that is available, that will still be preferred to the provided `LANGUAGE_CODE`, but if no available language is requested, whatever is set in the env will be used.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,8 @@ USE_HTTPS=true
 DOMAIN=your.domain.here
 EMAIL=your@email.here
 
+# Instance defualt language (see options at bookwyrm/settings.py "LANGUAGES"
+LANGUAGE_CODE="en-us"
 # Used for deciding which editions to prefer
 DEFAULT_LANGUAGE="English"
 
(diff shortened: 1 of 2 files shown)


mouse-reeve:
I'd love to get your feedback on this @lisbonjoker -- what I wrote will decide the language by looking at:

1. The logged in user's saved lnaguage
1. The browser's requested language
1. The instance default language

So when I, with my English-requesting browser, look at an instance that uses a different language, I will still see it in English. I'd only see it in the desired language if my browser language was not sent, or wasn't available in bookwyrm

If an instance is meant to be particular to a language, I could imagine that the admins would prefer I see it in that language, regardless of my browser settings

lisbonjoker:
I will try it as soon as I can, at work at the moment!

mouse-reeve:
No rush! I'm happy to change how this works any time

Issue #1828, opened 2022-01-16
Preview Card doesn't display non-latin characters properly


oragegu:
**Describe the bug**
Preview Card doesn't display non-latin characters properly

**To Reproduce**
1. Follow with a mastodon account the bookwyrm account that is about to publish a review the name of the book non-latin
2. Publish review
3. See the preview card with not-properly-displayed-non-latin-chars

**Expected behavior**
Display preview cards with non-latin book names

**Screenshots**
![image](https://user-images.githubusercontent.com/78921858/149646051-4eac4608-9ca3-49f3-8937-965871bab002.png)

**Instance**
Dans Mon Catalogue: book.dansmonorage.blue

mouse-reeve:
Thank you! Can you link the book? It could be the font, or the character encoding.

oragegu:
link of “wants to read”: https://book.dansmonorage.blue/user/pikapika/generatednote/4180#anchor-4180


Link to the book:
https://book.dansmonorage.blue/book/19137

oragegu:
> Thank you! Can you link the book? It could be the font, or the character encoding.

Just tested with a French book, the accents are properly rendered. Guess in this case the encoding is fine 

![C1F17A0B-09A3-4A60-AF39-095F397F074D](https://user-images.githubusercontent.com/78921858/149658277-309e7cbe-bd0c-4c79-8712-781cf548e5ce.jpeg)

link: https://book.dansmonorage.blue/user/pikapika/generatednote/4230

mouse-reeve:
Thank you! It seems like Open Sans just doesn't support a wide enough range of unicode characters. Based on this person who had a similar issue, we could check is character, or use a different font: https://stackoverflow.com/questions/11411746/drawing-multilingual-text-using-pil/11458418. Either seems doable although the latter is easier code-wise.

@joachimesque do you have an alternate font that you think would look as nice for this that has unicode support? I trust your aesthetic judgment.

nycterent:
i have stumbled on the same issue too: 

https://akmenys.group.lt/file/uodegos/images/previews/covers/15518-2cddb83b-f8eb-4353-888a-a9b1c0160e50.jpg

the book - https://ziurkes.group.lt/book/15518

joachimesque:
[...]

PR #1860, opened 2022-01-21
 Use Source Han Sans for preview images generation (bis)

cincodenada:
This is a partial fix for #1828 - these are the same changes as #1848, I've just rewritten the history to not include the font file in the repo itself, and added a line to download it into the Docker file at build time instead.

I ran a test clone to make sure I got this branch properly scrubbed, and it looks good - I pulled this branch down and the size of my `.git` dir stayed the same as it was before (21MB), and then pulled down the original branch and (after much downloading) my `.git` dir was 42MB, so we should be good!

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -35,6 +35,9 @@ LOCALE_PATHS = [
 ]
 LANGUAGE_COOKIE_NAME = env.str("LANGUAGE_COOKIE_NAME", "django_language")
 
+STATIC_ROOT = os.path.join(BASE_DIR, env("STATIC_ROOT", "static"))
+MEDIA_ROOT = os.path.join(BASE_DIR, env("MEDIA_ROOT", "images"))
+
 DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
 
 # Preview image
@@ -44,6 +47,16 @@ PREVIEW_TEXT_COLOR = env.str("PREVIEW_TEXT_COLOR", "#363636")
 PREVIEW_IMG_WIDTH = env.int("PREVIEW_IMG_WIDTH", 1200)
 PREVIEW_IMG_HEIGHT = env.int("PREVIEW_IMG_HEIGHT", 630)
 PREVIEW_DEFAULT_COVER_COLOR = env.str("PREVIEW_DEFAULT_COVER_COLOR", "#002549")
+PREVIEW_DEFAULT_FONT = env.str("PREVIEW_DEFAULT_FONT", "Source Han Sans")
+
+FONTS = {
+    "Source Han Sans": {
+        "directory": "source_han_sans",
+        "filename": "SourceHanSans-VF.ttf.ttc",
+        "url": "https://github.com/adobe-fonts/source-han-sans/raw/release/Variable/OTC/SourceHanSans-VF.ttf.ttc",
+    }
+}
+FONT_DIR = os.path.join(STATIC_ROOT, "fonts")
 
 # Quick-start development settings - unsuitable for production
 # See https://docs.djangoproject.com/en/3.2/howto/deployment/checklist/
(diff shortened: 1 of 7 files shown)


cincodenada:
I'm still working on confirming this works, my dev machine is being annoying tonight so I'll probably have to dig back into this tomorrow. But I think this should do the trick.

joachimesque:
Thank you! I'll close #1848

cincodenada:
Okay, this ended up being trickier than I had anticipated - because we're mounting the app as a volume, everything under `/app` is replaced, which means we can't stash things there.

We could load the font under some other directory in the docker image (say, `/resources`), but then it's not in the `/app/static/fonts` directory, which is kinda messy. We could copy it into the `/app` directory at startup or somewhere, but at that point I figured we might as well just download it at that point instead.

So, I ended up here, where we download the font the first time we need it. Upside here is we only download it if we actually need it - so folks without the environment variable set don't need to download it. Gotchas here are race conditions - it's possible we could try to download the file twice if two parts of the code ask for the font at the same time, but I'm not sure how likely that is.

It may make more sense to just download the font at app startup instead - maybe `apps.py` would be a reasonable place to do that, as we did with celerywyrm in #1751?

So, in short: this works (altho has some tests/linting to fix, it looks like) but I'm not convinced it's the best solution here.

bookwyrm/settings.py
```diff
@@ -310,13 +327,8 @@
     MEDIA_FULL_URL = MEDIA_URL
     STATIC_FULL_URL = STATIC_URL
     DEFAULT_FILE_STORAGE = "bookwyrm.storage_backends.ImagesStorage"
-    # I don't know if it's used, but the site crashes without it
```
mouse-reeve, line 313: thank you for this

cincodenada: It was handy, I was like "dang, I want these higher up in the file" and then realized that they were the same whether we were using S3 or not, so they could be! 😁

mouse-reeve:
I'm trying to test this, but my images that use web are all crashing becuase they can't find the font file. The web image itself give me the message `Just a sec, downloading Source Han Sans`, but still crashes with the file not found error.

cincodenada:
Hmm! Thanks for testing, I'll try spinning up a new instance myself when I have a minute to see if I can reproduce. Can you paste the error trace you're getting? If it can't find the file it should use the fallback font rather than crashing, so that's extra strange!

mouse-reeve:
This is the error I'm seeing:
```
web_1             | Django version 3.2.10, using settings 'bookwyrm.settings'
web_1             | Starting development server at http://0.0.0.0:8000/
web_1             | Quit the server with CONTROL-C.
web_1             | /app/bookwyrm/settings.py changed, reloading.
web_1             | Just a sec, downloading Source Han Sans
web_1             | Watching for file changes with StatReloader
web_1             | Exception in thread django-main-thread:
web_1             | Traceback (most recent call last):
web_1             |   File "/usr/local/lib/python3.9/threading.py", line 950, in _bootstrap_inner
web_1             |     self.run()
web_1             |   File "/usr/local/lib/python3.9/threading.py", line 888, in run
web_1             |     self._target(*self._args, **self._kwargs)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
web_1             |     fn(*args, **kwargs)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/runserver.py", line 110, in inner_run
web_1             |     autoreload.raise_last_exception()
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/utils/autoreload.py", line 87, in raise_last_exception
web_1             |     raise _exception[1]
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/core/management/__init__.py", line 375, in execute
web_1             |     autoreload.check_errors(django.setup)()
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/utils/autoreload.py", line 64, in wrapper
web_1             |     fn(*args, **kwargs)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/__init__.py", line 24, in setup
web_1             |     apps.populate(settings.INSTALLED_APPS)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/apps/registry.py", line 122, in populate
web_1             |     app_config.ready()
web_1             |   File "/app/bookwyrm/apps.py", line 40, in ready
web_1             |     download_file(config["url"], font_path)
web_1             |   File "/app/bookwyrm/apps.py", line 17, in download_file
web_1             |     with open(destination, "b+w") as outfile:
web_1             | FileNotFoundError: [Errno 2] No such file or directory: '/app/static/fonts/source_han_sans/SourceHanSans-VF.ttf.ttc'
```

cincodenada:
Okay, great, that makes sense. It's a confusing error: the actual error here is that the directory `/app/static/fonts/source_han_sans` doesn't exist, so it can't create the fontfile to write to it.

Which...after looking closer at how we build up the docker things, I see the problem, I think - I forgot (again) that the static directories aren't mounted like the rest of the app, so creating the directory in the repo doesn't actually create it in the image.

It _might_ work if you run `collectstatic` and then try again, but that's just a workaround - we should be creating the directory here (and catching file errors) anyway. I'll update the PR today to do that. Thanks!

cincodenada:
Okay, just pushed up a commit that should create the directory, and also not crash if we can't, I tested both of those by meddling around with my local dockers, and they both worked as expected. Thanks for testing!

PR #1866, opened 2022-01-24
Fix typo from models refactor

cincodenada:
Looks like this got caught up when moving to importing `models`
instead of the individual models, and was throwing an error on
`./bw-dev initdb`

With this fix I was able to run a full `resetdb`/`initdb` 🥳


mouse-reeve:
Thank you!

Issue #1839, opened 2022-01-18
Logo in onboarding dialog is not being size-limited


cincodenada:
**Describe the bug**
There appears to be no size-limiting on the onboarding dialog, as a large logo image is being rendered apparently full-size:

**To Reproduce**
1. Set the logo image to a high-resolution/large image
2. Go through the onboarding process
2. Look at the header of the dialog

**Expected behavior**
The logo is limited to a couple line's worth of text or so in height

**Screenshots**
![biglogo](https://user-images.githubusercontent.com/479715/149884096-182fb903-f8c7-40c8-8fc3-a26a4d6cc475.png)

**Instance**
https://ziurkes.group.lt

---

**Desktop (please complete the following information):**
    - OS: Xubuntu, I think, 20.04
    - Browser Firefox
    - Version 94

officialkzhang:
Hello! I'm a student in an Intro to SWE class, and I was looking for an issue for a "first contribution" assignment. Can I attempt this issue, and would you say it's beginner-friendly?

mouse-reeve:
Yes! Please do -- it should be a matter of adding the right class the HTML element in the template. Let me know if I can help point you to the right places

officialkzhang:
Hello! So I've found a working class for the image HTML element that will allow the image to be a reasonable size, however, I can't seem to locate where to add this class? Is there a JavaScript file that I could locate when exactly the onboarding logo would be produced, and therefore add the class? Or does the usage of Bulma prohibit directly adding classes to a CSS element?

mouse-reeve:
Ah, you've run afoul of the arcane templating system! It's all rendered in Python using Django templates, which live in `bookwyrm/templates/` -- this is the spot you're looking for: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/get_started/layout.html#L15

mouse-reeve:
Fixed by #1872

PR #1872, opened 2022-01-27
Added a class for fixing onboarding logo's size

officialkzhang:
This addresses #1839 by adding a CSS class to the onboarding logo's element and fixing the logo's size to a reasonable dimension of about 3rem. I was unable to test this on my local machine due to some bugs with running an instance, so further confirmations of this fix are appreciated.

### DIFF
--- a/bookwyrm/templates/get_started/layout.html
+++ b/bookwyrm/templates/get_started/layout.html
@@ -11,7 +11,7 @@
     <div class="modal-card is-fullwidth">
         <header class="modal-card-head">
             <img
-                class="image logo mr-2"
+                class="image logo mr-2 icon is-large"
                 src="{% if site.logo_small %}{% get_media_prefix %}{{ site.logo_small }}{% else %}{% static "images/logo-small.png" %}{% endif %}"
                 aria-hidden="true"
                 alt="{{ site.name }}"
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you! Your fix does solve the image size issue, but the `icon` class is meant to control how characters from the icon font display, rather than for controlling images. As a general principal, it's preferable to use classes as intended, just so that it's easier to follow. In this case, if a logo is rectangular rather than square, it will get stretched to a square and display oddly.

I wonder if it would work to use the same pattern that the main menu uses, since that uses the same logo file. In that instance, the parent div has the class navbar`. What do you think about using the `navbar` class on the `<header>` element that wraps the logo?

officialkzhang:
Looked into adding the navbar to the parent header element of the onboarding logo, seems to replicate the movement of the main menu as intended. The image is now size-limited (as we wished before), but it thankfully doesn't have to deal with the inconsistency and unintended ramifications of the icon class I previously added. However, now since the header has the navbar class, it will in fact move with the user as they scroll through the onboarding page (and won't be static). However, I feel as that isn't much of a hindrance, especially since I couldn't locate an image class to append that worked similarly to the icon one I had found previously.

mouse-reeve:
Perfect! Thank you so much

PR #1874, opened 2022-01-27
This fixes #1871 and fixes #1873

### LINKED ISSUE #1871: Improve bot management with robots.txt
Looking at traffic to bookwyrm.social, there is a huge amount of traffic coming in from various bots. Adding politeness settings for them could help reduce load.

Top bots:
1. AhrefsBot/7.0; +http://ahrefs.com/robot/
2. SemrushBot/7~bl; +http://www.semrush.com/bot.html
3. DotBot/1.2; +https://opensiteexplorer.org/dotbot
4. BLEXBot/1.0; +http://webmeup-crawler.com/
5. MJ12bot/v1.4.8; http://mj12bot.com/
6. PetalBot;+https://webmaster.petalsearch.com/site/petalbot

### LINKED ISSUE #1873: Test creates file.json,  but does not delete it.
**Describe the bug**
A pytest creates a file.json but does not delete it at the end. Which leads to a change in the brunch. Also, the file.json is not entered in .gitignor.

**To Reproduce**
1. Setting up the developer environment
2. enter `docker-compose run --rm web pytest`

**Expected behavior**
pytest would run without errors and does not produce a change in brunch.

**Additional context**
I would change the test a little bit to remove file.json at the end.

willhoh:
It fixes #1871 huge amount of traffic from bots
and closes #1873 that leaves a file after testing.


willhoh:
I'm new at contributing and also checks in github. How do I solve the lint errors the right way?

willhoh:
It was much easier than I thought. Have made a new commit to my main.

mouse-reeve:
Huh, it does seem like your change should have fixed the issue with the line breaks. I wonder if your editor uses a different type of line ending? Luckily, to fix the Python file, you can run `./bw-dev black`, which is a command that runs the [Black](https://github.com/psf/black) code formatter. That should get you set to go and resolve both checks.

Let me know if you're having trouble with these, it can be so frustrating to trial-and-error automated checks and I don't want you to get stuck trying to wrangle it.

willhoh:
I'm using vs-code and have LF as line ending. Should it be CRLF? Or am I completely wrong?

mouse-reeve:
It looks like you have it correct, thank you for troubleshooting!

mouse-reeve:
This is so thorough, I appreciate that you went through the different semrush bots and figured out which ones to block. Thank you!

PR #1901, opened 2022-02-02
Fixes name of Swedish locale

mouse-reeve:


cincodenada:
Ha, so fast! Fixes #1902 now 😁 

We should also update the translation files and switch them around, I'm not sure what the process is there though.

mouse-reeve:
ahaha you beat me to it

mouse-reeve:
> We should also update the translation files since it's easy enough to switch them around without having to know the language - I'm not sure what the process is there though.

Yep, that should be a straightforward fix in https://translate.joinbookwyrm.com/

cincodenada:
Great, yeah I realized afterwards that would be a separate thing - how/when do the updated translation files get pushed to Crowdin so that we can update things?

mouse-reeve:
It's manual -- the files are updated using `./bw-dev update_locales`, and once the changes are in `main`, Crowdin detects the updated `en_US` locale file.

Issue #1906, opened 2022-02-03
Exception resolving a bunch of template variables


cincodenada:
**Describe the bug**

After upgrading, I get a ton of errors similar to the following. Things seem to work, but they're very slow:

```
Exception while resolving variable 'title' in template 'book/editions/editions.html'.
web_1             | Traceback (most recent call last):
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 829, in _resolve_lookup
web_1             |     current = current[bit]
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/template/context.py", line 83, in __getitem__
web_1             |     raise KeyError(key)
web_1             | KeyError: 'title'
web_1             | 
web_1             | During handling of the above exception, another exception occurred:
web_1             | 
web_1             | Traceback (most recent call last):
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 835, in _resolve_lookup
web_1             |     if isinstance(current, BaseContext) and getattr(type(current), bit):
web_1             | AttributeError: type object 'RequestContext' has no attribute 'title'
web_1             | 
web_1             | During handling of the above exception, another exception occurred:
web_1             | 
web_1             | Traceback (most recent call last):
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 843, in _resolve_lookup
web_1             |     current = current[int(bit)]
web_1             | ValueError: invalid literal for int() with base 10: 'title'
web_1             | 
web_1             | During handling of the above exception, another exception occurred:
web_1             | 
web_1             | Traceback (most recent call last):
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/template/base.py", line 848, in _resolve_lookup
web_1             |     raise VariableDoesNotExist("Failed lookup for key "
[...]

PR #1907, opened 2022-02-03
Revert logging changes that seem to be breaking things

cincodenada:
I really don't understand how, but after bisecting, this logging configuration seems to be what is causing #1906 - if I remove it, the resolution errors go away, if I reintroduce it, they show up again.

Very strange, but since configurable logging (and 500 errors) are only nice-to-have, I think we should remove the logging configuration

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -106,59 +106,6 @@ TEMPLATES = [
     },
 ]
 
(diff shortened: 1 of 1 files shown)


mouse-reeve:
I'm closing this because I think your logging configation is great and the over-reporting in `DEBUG` mode is not _because_ of this code (and also, while not great, livable since it's just in one mode that shouldn't be used in production anyway)

PR #1925, opened 2022-02-06
Check HTTP status codes

### LINKED ISSUE #1831: Check HTTP status codes on file links
When an edition has a link out to a copy of the book, it would be great to automatically check that the link isn't returning a 404 or error type status code. I think this can be a `HEAD` request. I'm not sure when the best time to do it is.

willhoh:
This will check that http status code of the file link url is between 200 and 399. It should fix #1831

### DIFF
--- a/bookwyrm/forms.py
+++ b/bookwyrm/forms.py
@@ -2,6 +2,7 @@
 import datetime
 from collections import defaultdict
 from urllib.parse import urlparse
+import requests
 
 from django import forms
 from django.forms import ModelForm, PasswordInput, widgets, ChoiceField
@@ -241,7 +242,8 @@ class FileLinkForm(CustomForm):
                 self.add_error(
                     "url",
                     _(
-                        "This domain is blocked. Please contact your administrator if you think this is an error."
+                        """This domain is blocked. Please contact your administrator 
+                        if you think this is an error."""
                     ),
                 )
             elif models.FileLink.objects.filter(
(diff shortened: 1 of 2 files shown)


willhoh:
Can someone explain to me what the error means that was found during the checks?

I also typed `./bw-dev pytest bootwyrm` and only get
```
+ case "$CMD" in
+ execweb pytest --no-cov-on-fail bookwyrm
+ docker-compose exec web pytest --no-cov-on-fail bookwyrm
```
what am i doing wrong?

mouse-reeve:
If you merge main into your branch, it should get the tests working again. Sorry for the confusion!

mouse-reeve:
Sorry to leave you hanging for so long on this, things have been hectic for me, but I haven't forgotten about you! This PR is raising a security alert because it queries a user-provided url. I don't think it's a big issue because it's a `head` request and nothing is being done with the results. But I will probably pass it through the function that verifies urls for other types of requests, just for good measure -- that's something I can do separately once this is ready to merge.

I think if you run black, it should resolve the lingering whitespace issues, it looks like those are just trailing whitespace issues.

bookwyrm/forms.py
```diff
@@ -242,14 +243,39 @@ def clean(self):
         filetype = cleaned_data.get("filetype")
         book = cleaned_data.get("book")
         domain = urlparse(url).netloc
+
+        try:
+            response = requests.head(url, timeout=3)
+
+            if response.status_code in range(300, 308):
```
mouse-reeve, line 250: I believe you can pass `allow_redirects=True` to the original request instead of checking for redirect status codes

bookwyrm/forms.py
```diff
@@ -242,14 +243,39 @@ def clean(self):
...
+
         if models.LinkDomain.objects.filter(domain=domain).exists():
             status = models.LinkDomain.objects.get(domain=domain).status
             if status == "blocked":
                 # pylint: disable=line-too-long
                 self.add_error(
                     "url",
                     _(
-                        "This domain is blocked. Please contact your administrator if you think this is an error."
+                        """This domain is blocked. Please contact your administrator
```
mouse-reeve, line 277: Did this get changed automatically? It ought to stay on one line, so that the translation string doesn't have unexpected whitespace

mouse-reeve (changes requested):
Thank you so much for doing this work, and being patient with my much-slower-than-usual turnaround time on review

mouse-reeve:
I will deal with these merge conflicts! They're entirely on me

bookwyrm/forms.py
```diff
@@ -242,14 +243,39 @@ def clean(self):
...
+            response = requests.head(url, timeout=3)
+
+            if response.status_code in range(300, 308):
+                headers = response.headers
+                url_redirect = headers["Location"]
+                print(url_redirect)
+
+                response = requests.head(url_redirect, timeout=3)
+            if int(response.status_code) < 200 or int(response.status_code) > 299:
+                # status code between 200 and 399 should be fine
```
cincodenada, line 257: This comment is inconsistent with the condition above it, one of them should be updated (are 3xx codes okay or not?)

mouse-reeve:

PR #1935, opened 2022-02-12
[Accessibility] Fixes

joachimesque:
I'm using both [Deque axe](https://www.deque.com/axe/) and [WebAIM Wave](https://wave.webaim.org/) browser extentions to automate the finding of accessibility errors. One error is fixed per commit, more context is available in the commit message.

## The following errors won't be fixed yet

- [Text elements must have sufficient color contrast against the background](https://dequeuniversity.com/rules/axe/4.3/color-contrast)
  This issue is linked to Bulma. A good solution would be a server-side compilation of custom Bulma styles, with accessible colors.
- [Heading levels should only increase by one](https://dequeuniversity.com/rules/axe/4.3/heading-order)
  Right now this issue is linked to the homepage's design and structure.

### DIFF
--- a/bookwyrm/static/css/bookwyrm.css
+++ b/bookwyrm/static/css/bookwyrm.css
@@ -301,6 +301,159 @@ details.dropdown .dropdown-menu a:focus-visible {
     }
 }
 
+/** Bookwyrm Tabs
+ ******************************************************************************/
+
+.bw-tabs {
+    -webkit-overflow-scrolling:touch;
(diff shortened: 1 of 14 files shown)


mouse-reeve:
Seeing the linting errors for the css, it occurs to me that I think Prettier can auto-format css as well. I just would never ever want to have to fix those manually. https://github.com/bookwyrm-social/bookwyrm/pull/1936

Issue #1932, opened 2022-02-11
Covers in lists aren't properly sized on mobile


mouse-reeve:
**Describe the bug**
On mobile, the covers in the list view aren't being scaled down to fit on the screen.

**Screenshots**
![Screenshot_20220211-135440](https://user-images.githubusercontent.com/1807695/153675913-3a15232b-c428-4ebe-9e93-629b0c21991b.png)

**Instance**
Bookwyrm.social

mouse-reeve:
Fixed by #1938

PR #1938, opened 2022-02-12
Change cover class

willhoh:
The cover class "is-h-m-tablet" hasn't show and resize the image properly. With "is-h-m-mobile" it does it very well. This should fix issue #1932.


mouse-reeve:
Upon testing, I think this needs both `is-h-m-mobile` and `is-h-m-tablet` -- when it just has the mobile class, the full-sized view seems to have the wrong image size.

mouse-reeve:
Thank you so much for fixing this!

Issue #1931, opened 2022-02-11
Searching for ISBN with dashes (as seen on some books) does not show the same results as without dashes


viviicat:
**Describe the bug**
When using the search, if you type the ISBN in a format as seen on some books, (978-1-78330-350-2) you will get results from Inventaire that contain some of the sub-numbers (e.g. "350 Arbres et arbustes" is the first result in this case). If you expand OpenLibrary you can see the correct book but it is less clear.

This is inconsistent with typing the ISBN as an un-dashed list of numbers, in which case you clearly see a single book matching the description.

**To Reproduce**
1. Type an ISBN containing dashes, e.g. 978-1-78330-350-2 into the search bar.
2. Hit Search.
3. Note many results from Inventaire, all of which do not contain the target book.
4. Expand OpenLibrary results and note the correct book is present there.

**Expected behavior**
When a number matching an ISBN dash-included format (seems there are a few variations of this?) is entered, it should be parsed as an ISBN and sent as a flat number to the search engines, so that only the correct book is shown. Users may not realize that removing the dashes from their search would help them find their book.

**Screenshots**
![image](https://user-images.githubusercontent.com/287339/153527587-418810b5-2cb8-4915-ba6b-f2088a5a2b3b.png)
![image](https://user-images.githubusercontent.com/287339/153527620-c5049ba8-77bf-4a2b-a027-e811aeed7a52.png)

**Instance**
My instance https://books.solarpunk.moe

---

**Desktop (please complete the following information):**
    - OS: Windows
    - Browser Firefox
    - Version 97

willhoh:
Hi, I would love to work out a solution for this. I have already done some research and found 2 algorithms to check if they are ISBN numbers. Unfortunately I'm stuck on the realization of the search and where/how to include such helper functions. Would it be possible to give me some tips for this?

I have already added a few tests to check if ISBNs are really recognized.

PR #1942, opened 2022-02-13
Isbn check befor search

willhoh:
The system checks whether an ISBN has been entered. If so, hyphens are removed and passed on to the search. This should fix issue #1931.

### DIFF
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -12,9 +12,6 @@ from bookwyrm.settings import MEDIA_FULL_URL
 
 # pylint: disable=arguments-differ
 def search(query, min_confidence=0, filters=None, return_first=False):
-    # check if query is isbn
-    if isbn_check(query):
-        query = query.replace('-','')
     """search your local database"""
     filters = filters or []
     if not query:
@@ -134,32 +131,6 @@ def search_title_author(query, min_confidence, *filters, return_first=False):
             return result
         list_results.append(result)
     return list_results
-
-
-def isbn_check(query):
-    n = query.replace('-','')
-    if len(n) == 13:
-        # Multiply every other digit by  3
-        # Add these numbers and the other digits
-        product = (sum(int(ch) for ch in n[::2]) 
-                + sum(int(ch) * 3 for ch in n[1::2]))
-        if product % 10 == 0:
-            return True
-    elif len(n) == 10:
-        if n[0:8].isdigit() and (n[9].isdigit() or n[9].lower() == "x"):
-            product = 0
-            # Iterate through code_string
-            for i in range(9):
-                # for each character, multiply by a different decreasing number: 10 - x
-                product = product + int(n[i]) * (10 - i)
-            # Handle last character
-            if n[9].lower() == "x":
-                product += 10
-            else:
-                product += int(n[9])
-            if product % 11 == 0:
-                return True
-    return False
 
 
 @dataclass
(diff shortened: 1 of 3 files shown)


bookwyrm/book_search.py
```diff
@@ -133,6 +136,32 @@ def search_title_author(query, min_confidence, *filters, return_first=False):
     return list_results
 
 
+def isbn_check(query):
+    n = query.replace('-','')
```
viviicat, line 140: Does this need a check to make sure all the characters are numeric after replacing them? Will it throw for 13-character strings that are non-numeric due to use of `int()`?

kopischke:
Apologies for the armchair commenting, but may I suggest two things?

1. Handle separators other than hyphens. In Germany, long numeric codes like ISBNs are often space separated; I am pretty sure other examples exist elsewhere, so it might be useful to make this generic.
2. Deduplicate the transform by having the ISBN check function  return the (transformed, if it is an ISBN, else untransformed) query instead of a Boolean.

bookwyrm/book_search.py
willhoh: Yes and No :-)
I'm not checking for numeric values here, but whether the 13-digit/10-digit value is an ISBN. I need int() to vivisect this (see [ISBN](https://en.wikipedia.org/wiki/International_Standard_Book_Number)). The passed value is always a string. I check for ISBN because someone should also be able to enter a title like 1950-2000.

willhoh:
> Apologies for the armchair commenting, but may I suggest two things?
> 
> 1. Handle separators other than hyphens. In Germany, long numeric codes like ISBNs are often space separated; I am pretty sure other examples exist elsewhere, so it might be useful to make this generic.
> 2. Deduplicate the transform by having the ISBN check function  return the (transformed, if it is an ISBN, else untransformed) query instead of a Boolean.

Of course you can interfere and I even wish that, otherwise how should I know if I'm doing something wrong?

Now I have tried to incorporate your improvements, but I had to realize that I have only corrected the search for the data of bookwyrm. For "inventaire" or "openlibrary" the input is still passed as it is entered.

Therefore I believe that the upper commits will not solve the issue #1931. The choice of my test data was bad.

willhoh:
I think I found it and was able to solve it.

bookwyrm/book_search.py
viviicat: I just tested and indeed, if you search for `fffffffffffff` you get

```
ValueError at /search

invalid literal for int() with base 10: 'f'
```

after line 140, you should add

```
if not n.isnumeric():
    return False
```

or similar.

bookwyrm/views/search.py
```diff
@@ -123,3 +125,35 @@ def list_search(query, viewer, *_):
         )
         .order_by("-similarity")
     ), None
+
+
+def isbn_check(query):
+    """isbn10 or isbn13 check, if so remove separators"""
+    if query:
```
mouse-reeve, line 132: This will be just a bit easier to read if you return when there is no query right at the top, so the body of the function doesn't have to be embedded in the if statement.

Something like:
``` python
if not query:
    None

su_num = re....
```

bookwyrm/views/search.py
```diff
@@ -123,3 +125,35 @@ def list_search(query, viewer, *_):
...
+        su_num = re.sub(r"(?<=\d)\D(?=\d|[xX])", "", query)
```
mouse-reeve, line 133: I'm not sure what `su_num` means - is there a more verbose way to name it so it's more clear? Alternatively, a comment on the line aboveexplaining it would help me out

mouse-reeve (changes requested):
Very small changes requested! Thank you for your work on this ❤️

PR #1152, opened 2021-05-30
Front-end styles compilation

joachimesque:
## Description

The goal of this PR is to rationalize the handling of styles, and allow an instance owner to painlessly add custom styles.

This feature uses [Django Sass Processor](https://github.com/jrief/django-sass-processor), which doesn't force the use of NPM or other package managers. The CSS compilation is done when rendering a template, and is cached for later use. See above link for doc.

## How to set up a custom stylesheet:

- copy `.css-config-sample/_instance-settings.scss` to `bookwyrm/static/css/_instance-settings.scss`
- edit `bookwyrm/static/css/_instance-settings.scss` to add Bulma vars definitions or any CSS/SCSS/SASS style you like
- ✨ it works ✨

## Development

The PR proposes a new architecture for Bookwyrm's app styles:

- `bookwyrm/static/css/bookwyrm/components` for Bookwyrm components
- `bookwyrm/static/css/bookwyrm/overrides` for vendor libraries overrides and extensions
- `bookwyrm/static/css/bookwyrm/utilities` for utilities

A new `_bookwyrm-settings.scss` file is added, to define settings for Bulma (or any vendor lib). These will be superceded by the optional `_instance-settings.scss`.

## TODO

- [ ] `eclint`: Disable linting on **/vendors/**
  This should be possible per [`eclint`'s doc](https://gitlab.com/greut/eclint#usage), but I don't know how to pass the argument `-exclude` to the GitHub action in `.github/workflows/lint-global.yaml`

PR #1956, opened 2022-02-18
Compiles css framework from sass

mouse-reeve:
This is based on @joachimesque's work in #1152 -- it adds npm-sass to the dependencies and sets up the tooling for compiling the sass into css. Converting the existing css into sass is a separate PR.


.github/workflows/lint-frontend.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       # See .stylelintignore for files that are not linted.
       - name: Run stylelint
         run: >
-          npx stylelint bookwyrm/static/css/*.css \
+          npx stylelint bookwyrm/sass/*.scss \
```
joachimesque, line 30: Low importance: instead of naming the directory after the file format (`bookwyrm/sass`), I think it might be more relevant to name it after its function (`bookwyrm/style` or `bookwyrm/stylesheets`). That way if we change the compilation process (e.g. if Sass isn't supported anymore), the dir name won't have to change.

joachimesque (approved):
Looks great to me :)

If I understand correctly, the styles will have to be compiled and minified in the PR, and won't be compiled at docker image build time? In that case we should perhaps ensure that the minified version is always the latest version. Or perhaps a compilation could be launched when a new release is prepared?

dev-tools/package.json
```diff
@@ -1,18 +1,22 @@
...
-    "watch:static": "yarn watch \"./bw-dev collectstatic\" bookwyrm/static/**"
+    "css-build": "node-sass --omit-source-map-url ../bookwyrm/sass/bookwyrm.scss ../bookwyrm/static/css/bookwyrm.min.css --output-style=compressed",
+    "css-watch": "npm run css-build -- --watch",
+    "watch:static": "watch \"cd .. && ./bw-dev collectstatic\" ../bookwyrm/static/**"
   },
   "devDependencies": {
+    "bulma": "^0.9.3",
     "eslint": "^8.9.0",
+    "merge": "2.1.1",
+    "node-sass": "^7.0.1",
```
joachimesque, line 11: We could perhaps try with `sass` instead of the deprecated `node-sass`, what do you think?

https://www.npmjs.com/package/sass

mouse-reeve:
I took a second look at the Django process you set up for sass and I think it was a better choice than using npm, so I copied your config over from the original PR. Thank you for the feedback!

PR #1975, opened 2022-02-26
Themes

### LINKED ISSUE #815: Make BookWyrm theme-able
It would be great to let admins create a default css theme for their instances, and let users select a theme at their preference. Dark mode!

I think this involves a number of steps, which can be handled atomically:
 - [x] Switching to a compiled-from-sass version of the bulma styles and custom css
 - [ ] Update the css colors used in the templates to be semantic rather than color-names
 - [x] Create a theme file
 - [ ] Give admins a way to set available and default themes
 - [ ] Give users a way to select their own theme

mouse-reeve:
Fixes #815 

Remaining work:
 - [x] Fix `is-light` buttons and notifications (resolved by #1977)
 - [x] Finish dark mode (resolved by #1977)
 - [x] Upload themes
 - [x] Link from themes page to settings
 - [x] Add theme to edit user form
 - [x] Add instructions
 - [x] Fix icons


bookwyrm/models/user.py
```diff
@@ -173,6 +174,17 @@ class User(OrderedCollectionPageMixin, AbstractUser):
     property_fields = [("following_link", "following")]
     field_tracker = FieldTracker(fields=["name", "avatar"])
 
+    @property
+    def get_theme(self):
```
viviicat, line 178: I noticed in testing that this breaks when you're logged out -- the template resolves to empty string.

I assume that this is because the user object is `None` in that case.

mouse-reeve: Thank you! The user model is definitely not the right place for this function

joachimesque:
Oooh!! I'll be sure to update :)

I'll have to open a PR for accessibility color fixes, it should be easier now!

PR #1976, opened 2022-02-27
Add barcode scanning support

### LINKED ISSUE #318: let users add books to their shelves by scanning ISBN / barcodes
In Goodreads mobile app, you're able to add books to your shelves by scanning the barcode / ISBN from the back of a book. At this time, you can also assign shelves. In the past, I've used this as a way to track books I'm interested in reading and also my personal library. Would be cool to have this option in Bookwyrm, too!

viviicat:
Use [Quagga2](https://github.com/ericblade/quagga2) to find valid EAN barcodes from the user's webcam/phone camera, and perform a search for the found ISBN. This is a JS-based procedure which sends no image data to the server.

Resolves #318 -- it does not immediately add the items to the shelf but users can follow the normal flow of adding a book from a search. If the book is present in their library they can select it and add it to a shelf as they would otherwise.

Note: If a bulk import UI is ever added it would be great to adapt this for when you have a pile of books to add.


viviicat:
When I started I think there may have been less reliance on npm? Does it make sense to just install quagga with that now? Or is including the min.js the right way to do it?

mouse-reeve:
> When I started I think there may have been less reliance on npm? Does it make sense to just install quagga with that now? Or is including the min.js the right way to do it?

At the moment there is no still use of npm in production. I think including min.js is fine!

mouse-reeve:
Wow!! this is so cool!

PR #1985, opened 2022-02-28
Update README.md

Strubbl:
fix link to instances


mouse-reeve:
Good catch, thank you!

PR #1987, opened 2022-03-01
Fix for TypeError/no focus when clicking on a reply

viviicat:
Remove uuid for status reply panel
- The focus target did not include this uuid, so was throwing `Uncaught TypeError: node is null` when clicking on Reply button.
- I wasn't able to figure out how to share the uuid between blocks but it doesn't seem like the uuid is really needed -- the same block has other ids that do not have the uuid in them. So if the uuid is needed for uniqueness the other ones should be changed too.


mouse-reeve:
Thank you!

PR #1994, opened 2022-03-04
Adds custom compile management command

### LINKED ISSUE #1991: Theme selection not working in production
On a closer scan, it's expected that the `compilescss` management command doesn't work with a variable for the scss file path, but the processor still ought to be able to produce the stylesheet on the fly. This is working locally (with or without s3 enabled) but not in production.

mouse-reeve:
Fixes #1991

### DIFF
--- a/bookwyrm/management/commands/compilethemes.py
+++ b/bookwyrm/management/commands/compilethemes.py
@@ -19,7 +19,7 @@ class Command(BaseCommand):
         """compile themes"""
         storage = StaticFilesStorage()
         theme_files = list(get_files(storage, location="css/themes"))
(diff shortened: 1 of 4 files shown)


jalada:
@mouse-reeve am I right in saying this step (`compile_themes`) is missing from your [dockerless production setup docs](https://docs.joinbookwyrm.com/install-prod-dockerless.html)? I'm just following them now (in combo with using S3) and the sync to S3 is just copying a bunch of scss files rather than the compiled css files. (I guess as an aside it probably doesn't need to copy the uncompiled files, either).

mouse-reeve:
Yes, the docs need to be update; sorry about that!

mouse-reeve:
Fixed, that was my bad

PR #1998, opened 2022-03-07
Custom question option and field for spammed bookwyrm instances

oragegu:
image
image
image


bookwyrm/templates/landing/layout.html
```diff
@@ -70,6 +70,14 @@ <h2 class="title">
...
                             {% include 'snippets/form_errors.html' with errors_list=request_form.email.errors id="desc_request_email" %}
                         </div>
+						{% if site.invite_request_question %}
+							<div class="block">
+								<label for="id_answer_register" class="label">{{ site.invite_question_text }}</label>
+								<input type="answer" name="answer" maxlength="50" class="input" required="true" id="id_answer_register" aria-describedby="desc_answer_register">
+								{% include 'snippets/form_errors.html' with errors_list=request_form.answer.errors id="desc_answer_register" %}
+							</div>
+						{% endif %}
+
```
cincodenada, line 80: It looks like this block has tabs, which is why it looks misaligned here but probably not in your editor - can you convert these to spaces to be consistent?

bookwyrm/views/admin/invite.py
```diff
@@ -169,10 +172,11 @@ def post(self, request):
         if form.is_valid():
             received = True
             form.save()
+		#answer =
```
cincodenada, line 175: I think this can just be removed

bookwyrm/models/site.py
```diff
@@ -48,8 +48,14 @@ class SiteSettings(models.Model):
     # registration
     allow_registration = models.BooleanField(default=False)
     allow_invite_requests = models.BooleanField(default=True)
+    invite_request_question = models.BooleanField(default=False)
     require_confirm_email = models.BooleanField(default=True)
 
+    invite_question_text = models.CharField(
+		max_length=255,
+		null=not(invite_request_question),
+        default="What is your favourite book?"
```
cincodenada, line 57: I think the `invite_request_question` field could be removed, and this could just be a field with `null=true` (`null` here means "can be set to null"), and if the field is null, we don't show the prompt.

corentin-feys: Hello,
Do you mean remove the tick- box entirely and assume if people leave the field empty, it means they don't want a question?

bookwyrm/templates/landing/layout.html
corentin-feys: Yes you are correct. I have replaced them.

bookwyrm/models/site.py
mouse-reeve: I think there might be a benefit to having both the boolean and the charfield, so the question doesn't have to be deleted if an admin doesn't want to use it for a period and then does again in the future. It's a more explicit pattern, and I think the downside is that someone might miss that they need to use both the checkbox and the question field.

mouse-reeve:
Thank you so much for this PR! I'm running slower than usual on review, so I appreciate your patience with the response time, because I know it's an iterative process. When I get a chance, I'll take a close look at the code and try it out locally -- I'll let you know if I find anything that needs tweaking, since there can be weird edge cases that it takes a second or third pair of eyes to find 😄 . Once we get the automated checks passing and I'm able to test it, I'll go ahead and merge.

It looks like the tests are failing because you haven't checked in the migration file -- that should be a quick fix. If you don't have the file, you can generate it by running `./bw-dev makemigrations`. Just `git add` that file and push it to this branch, and the tests will run. The linting errors are probably all things you can resolve just by running `./bw-black`, which will automatically re-formated your code to it's extremely specific standards.

Let me know if you have any trouble with the CI or anything else and we'll get you sorted.

bookwyrm/models/site.py
cincodenada: To me, the database representation is a separate question from the UI - I think it makes sense to have the checkbox and text field for UI, but that doesn't mean you have to have it that way in the database. That said, this would allow you to have a question and check/uncheck it without clearing out the question, so maybe it's reasonable to have both.

In any case, I don't think `null=not(invite_request_question)` is what you want - unless I'm missing something, this is evaluated at startup, and is evaluated against the value of the `invite_request_question` variable in Python, not anything int he database, so it'll always be `false` (since `models.CharField()` is truthy).

And `null` is just an attribute of the field specifying [whether it _can_ be null or not](https://docs.djangoproject.com/en/4.0/ref/models/fields/#null) - I think you just want this to be `null=false` (so sites will default to "What is your favourite book" with the checkbox disabled). That's the default, so you can leave `null` out if you want, I think.

cincodenada:
On migrations, I see you had migrations and then removed them - it looks like there were several smaller migrations, so as @mouse-reeve said, you can just run `./bw-dev makemigrations` to generate one tidier migration and add that. And I think they mistyped, you want to run `bw-dev black` to do the nitpicky formatting :)

mouse-reeve:
omg yes thank you @cincodenada I only have a half functional brain today

bookwyrm/models/site.py
corentin-feys: Thank you so much for the feedback! I agree with cinodenada that the null parameter can just be left as false. I was afraid that you wouldn't be able to save the settings if you unticked the box and left the field blank. I'll test it more thoroughly and get back to you.

cincodenada: I see, and yeah that makes sense! It's very confusing that the property is called `null`, really it should be called `nullable` (or `canBeNull` or whatever), but for historical reasons it's just `null` :shrug: 

You should be able to set the field to an empty string (`""`) even with null(able)=false, so I think you'll be okay, but yeah definitely check on that and make sure Django does set it to empty string instead of null!

corentin-feys: I ran into the issue I mentioned in my previous comment. If the field is allowed to be left blank, then one might tick the box and have no question, which we don't want. If the field is not allowed to be empty, one may be forced to set a question even though they don't plan to use one. 
To get around it, I made it so that when the question field is left blank, it is reset to the default question. Let me know what you think.

mouse-reeve: That seems like a sensible behavior to me

mouse-reeve:
This works great! I'm doing some merge conflict resolution stuff hence the base change. Really appreciate yalls work on this <3

mouse-reeve:
I realize I may have merged this preemptively, I've been trying to get the pull request situation under control. If any of you feel like there are additional changes this needs I'd be happy to see a new PR for that. But also when I tested this it did what it said on the box 😄

oragegu:
> I realize I may have merged this preemptively, I've been trying to get the pull request situation under control. If any of you feel like there are additional changes this needs I'd be happy to see a new PR for that. But also when I tested this it did what it said on the box 😄

Yaaaaay it’s merged!

Everything works (I think). It’s just one of the checks keeps saying “string too long” or something, and wouldn’t let the code pass. And we didn’t know what to do 😢 

But apart from that check everything works!

PR #2002, opened 2022-03-10
Refactors content warning field

### LINKED ISSUE #1420: Clarify UI around content warnings
A few problems with content warnings:
 - [ ] The button is icon-only and hard to find for visual users in the compose UI (I think it's fine for screen reader users):
Screen Shot 2021-09-18 at 11 11 48 AM

 - [x] ~When content warnings are active in the compose box, they aren't clearly labeled (again, for visual users only):~ (fixed by #1422)
Screen Shot 2021-09-18 at 11 12 44 AM

- [x] #1430
Screen Shot 2021-09-18 at 11 13 09 AM

mouse-reeve:
Fixes #1420

This uses the details panel rather than a javascript toggle to show/hide the content warning field in the compose box.

Closed:
Screen Shot 2022-03-10 at 9 16 11 AM

Opened:
Screen Shot 2022-03-10 at 9 16 16 AM


viviicat:
Feels like an improvement to me!

Tak:
I guess I missed my chance on this one, but: the toggle is now almost as big as the expanded field - why not just have it always visible?

mouse-reeve:
You haven't missed your chance, I'm happy to revisit after merging! I suppose I think the show/hide makes it more clear that it's an optional field, especially since it has a large impact on how the resulting post is displayed. In the previous setup, hiding the content warning in the compose meant that it wouldn't be included in the status. As it stands, that isn't still the case, but I wonder if it should be 🤔

PR #2003, opened 2022-03-10
Consistent positioning of success buttons

mouse-reeve:
Currently, the "success" button for anything from posting a status to following a link can be either on the right or left, after or before the close or cancel options. This PR just puts the success button on the right and the cancel (or whatever) button the left.

For example, before:
Screen Shot 2022-03-10 at 9 34 53 AM

And after:
Screen Shot 2022-03-10 at 9 34 38 AM


oragegu:
I think the uniformity is very neat.

PR #2004, opened 2022-03-10
Uses details instead of javascript for inline forms

mouse-reeve:
I'd like to get rid of the javascript "toggle button" component entirely and use noscript-friendly modals and details instead. This replaces the "Create"/"Edit" buttons next to headers with details panels. The main downside is that the button next to the header was more compact and I liked how it looked, but I think it's a worthwhile tradeoff.

Before
-------
Closed:
Screen Shot 2022-03-10 at 11 14 51 AM

Open:
Screen Shot 2022-03-10 at 11 14 55 AM

After
-----

Closed:
Screen Shot 2022-03-10 at 11 06 12 AM

Open:
Screen Shot 2022-03-10 at 11 06 21 AM


mouse-reeve:
I may make this a modal rather than an inline form 🤔

PR #2007, opened 2022-03-12
Add names of books/lists/authors/etc as slugs, redirect to slugified version of the page

### LINKED ISSUE #1947: Add book title or author name to url
It would be great if urls were more human readable, by including the book title in book urls and author name in author urls. This field should be optional and non-determing, so if you enter a url without the name, or with the wrong name, it should resolve to the correct name. When the title or name changes, the url should reflect that change.

For example:
`https://bookwyrm.social/book/304370` or `https://bookwyrm.social/book/304370/sdkjhfjksdf`

Would redirect to:
`https://bookwyrm.social/book/304370/oldest-cuisine-in-the-world`

viviicat:
Closes #1947 

TODO: I would like to write unit tests for the slugs, that's the main thing keeping this WIP. Not aware of any issues remaining and my local run of existing unit tests is green.


bookwyrm/views/author.py
```diff
@@ -12,20 +12,23 @@
...
 
-    def get(self, request, author_id):
+    def get(self, request, author_id, slug=None):
         """landing page for an author"""
         author = get_object_or_404(models.Author, id=author_id)
 
         if is_api_request(request):
             return ActivitypubResponse(author.to_activity())
 
+        if r := maybe_redirect_local_path(request, author):
```
mouse-reeve, line 29: I didn't know you could do this! Very cool

bookwyrm/views/books/books.py
```diff
@@ -15,22 +15,26 @@
...
-from bookwyrm.views.helpers import is_api_request
+from bookwyrm.views.helpers import is_api_request, maybe_redirect_local_path
 
 
 # pylint: disable=no-self-use
 class Book(View):
     """a book! this is the stuff"""
 
-    def get(self, request, book_id, user_statuses=False):
+    def get(self, request, book_id, **kwargs):
```
mouse-reeve, line 25: what the advantage of using kwargs here? is that because of the changes to urls.py?

bookwyrm/views/author.py
```diff
@@ -12,20 +12,23 @@
...
-from bookwyrm.views.helpers import is_api_request
+from bookwyrm.views.helpers import is_api_request, maybe_redirect_local_path
 
 
 # pylint: disable= no-self-use
 class Author(View):
     """this person wrote a book"""
 
-    def get(self, request, author_id):
+    def get(self, request, author_id, slug=None):
```
viviicat, line 22: So this is failing the lint because of the unused arg. Is this a case where we just do a linter ignore, or is there some other way we can use `{% url %}` to construct a URL and then discard a param?

bookwyrm/views/books/books.py
viviicat: So I believe it's only relevant for API requests at the moment (and came up in unit tests of `views/books/test_book.py`) but before this change the view supported `view(request, self.book.id, "review")` (or `comment`/`quotation`) to filter user statuses -- but since `user_statuses` was an optional argument and `slug` is also an optional argument, there's no way to differentiate between those args without specifying  both. So I'm using kwargs instead so that they're named.

An alternative would be to force an ordering. If we did that, then if we specified user_statuses first we'd have to pass `False` around everywhere we wanted to specify the slug, and (afaik) it would need to be in the URL. So that doesn't seem right.

bookwyrm/views/author.py
mouse-reeve: If I understand correctly, it needs to be there because the slug is a url parameter and so the view has to receive the parameter for the url to work. Can the slug be not a parameter? Like if the regex was just `r"/s/[-_a-z0-9]+"` instead of `r"/s/(?P<slug>[-_a-z0-9]+)"`. If that works, then the parameter wouldn't be necessary in the view.

But if that doesn't work I'd definitely just do a pylint disable on it.

mouse-reeve: now that I think about it, that idea probably won't work because, as you pointed out right there, it's needed for the `{% url %}` tag

bookwyrm/views/books/books.py
mouse-reeve: I see! Thanks for explaining -- I think the way you have it works :)

bookwyrm/views/author.py
viviicat: Do you think it's reasonable to append the slug manually everywhere we call `{% url %}`, like this? 
```
href="{% url 'book' book.id %}/{{ book.name | slugify}}"
```
Does that feel too hacky?

mouse-reeve: Hmm, I think that seems fine to me? I feel like either solution is kind of odd in its own way, and this one at least keeps the oddness in slightly fewer places?

bookwyrm/views/books/books.py
```diff
@@ -46,6 +50,9 @@ def get(self, request, book_id, user_statuses=False):
         if not book or not book.parent_work:
             raise Http404()
 
+        if r := maybe_redirect_local_path(request, book):
```
viviicat, line 53: Investigated more and this is falsely redirecting if the user statuses are set. Need to only redirect if user_statuses is false.

bookwyrm/views/books/books.py
viviicat: This is used outside the API--for the tabs at the bottom of the book page.

bookwyrm/views/author.py
viviicat: I think I'll go with the linting annotations because I was running into complexities in spots where we have `{% url 'book' book.id slug as book_url %}` and it feels like it would be messy to have to compose the URL manually.

mouse-reeve: Sounds good! Thanks for thinking it through with me

mouse-reeve (changes requested):
I've found that there is an empty slug generated when the title of the book/list/author doesn't have any latin characters, which leads to a 404 error. For example, when I load a copy of https://bookwyrm.social/book/356220 in my test instance, it resolves to a 404.

viviicat:
> I've found that there is an empty slug generated when the title of the book/list/author doesn't have any latin characters, which leads to a 404 error. For example, when I load a copy of https://bookwyrm.social/book/356220 in my test instance, it resolves to a 404.

I've tweaked the routing so that empty slugs are allowed to support pages like that. This of course keeps a stray /s/ at the end of the URL, but I think that's probably fine? Stripping that seemed like it would be messy as I would have to wrap the existing `slugify` code.

mouse-reeve:
Thanks for your patience while I took approximately 100 years to get around to this! Stoked to have it live soon :)

PR #2008, opened 2022-03-12
Add autocompletions for fish shell

viviicat:
These are nice autocompletions--for example if you do `./bw-dev pytest <tab>` it'll give you a list of all the test paths only. `./bw-dev up <tab>` shows you the list of containers to bring up. etc.

Can be used by copying to `~/.config/fish/completions/bw-dev.fish`

Also added the `setup` command to the help output of bw-dev.


mouse-reeve:
Thank you!

PR #2009, opened 2022-03-13
add SASS_PROCESSOR_ENABLED = True

viviicat:


viviicat:
> During development, **or when SASS_PROCESSOR_ENABLED = True**, the compiled file is placed into the folder referenced by SASS_PROCESSOR_ROOT (if unset, this setting defaults to STATIC_ROOT). Having a location outside of the working directory prevents to pollute your local static/css/... directories with auto-generated files. Therefore assure, that this directory is writable by the Django runserver.

PR #2020, opened 2022-03-14
Array input field for forms

mouse-reeve:
Screen Shot 2022-03-14 at 10 42 21 AM


mouse-reeve:
@Tak I realized after I merged that subjects with commas in them were going to cause issues in a comma separated field, so I switched it to this

Tak:
This looks a lot better anyway. I was initially looking at attempting something like this, but then I noticed we already had some "just separate with commas" fields, so I thought I'd PR that and you'd say something if you thought it should be more sophisticated. :slightly_smiling_face:

PR #2032, opened 2022-03-17
fix for npm dev commands (stylelint and prettier)

viviicat:
Fixing a number of overlapping issues that prevented stylelint from working on a fresh install (at least for me). I was getting a "permission denied" error. I believe if your install has node_modules and/or package.json *outside* the container, you may be able to run the commands ok. But maybe try with a fresh install dir--I believe it's reproducible.

The dev-tools image was being created okay, but then when the `/app` directory was mapped, the npm files/folders were being overlaid and discarded. With my change, npm now installs to the `dev-tools/` directory, and that folder is mapped to the container's directory.

I then altered the commands in `./bw-dev` to run with a prefix of `dev-tools/`.

Then I modified the stylelint config as it had a number of issues:
* Use scss config to mute the warning.
* Lint all files, but exclude fonts.
* Then discovered a linter error due to the import at the bottom of the themes -- I believe it makes sense to exclude it?

And finally, cleaned up the existing linting issues.


viviicat:
Looks like the workflows may need to be changed for this to work, though.

mouse-reeve:
Thank you for updating the dev-tools image, I'll try and take a look next week. Getting these checks working right is such a hassle 😵

dev-tools/package.json
```diff
@@ -11,6 +11,7 @@
     "stylelint-config-recommended": "^7.0.0",
     "stylelint-config-standard": "^25.0.0",
```
cincodenada, line 12: These two other `stylelint-config` packages should be able to be removed now, `stylelint-config-standard-scss` includes the ones that it needs.

docker-compose.yml
```diff
@@ -102,6 +102,7 @@ services:
     build: dev-tools
     env_file: .env
     volumes:
+      - /app/dev-tools/
```
cincodenada, line 105: This confused me for a bit, I think a comment here would be helpful for future me's :slightly_smiling_face: 

Maybe something like this, if I'm understanding correctly:
```
# Create an empty volume for `/app/dev-tools` so that node_modules is isolated to the container
```

cincodenada:
I'd already merged in main and re-ran the stylelint, so I opened a PR in @viviicat's repo to save y'all the trouble: https://github.com/viviicat/bookwyrm/pull/1 - I think if @viviicat merges that PR, it should just show up as updated here.

mouse-reeve:
Is there something I can do to get this PR ready? I feel bad for neglecting it for so long but i'm not sure what needs to happen at this point 😅

viviicat:
Hmm, not sure either! I'll have to take a look at this. Probably tonight.

alopix:
Any progress on this or workaround until this PR is merged? `stylelint` unfortunately does not work for me without changing the docker files

docker-compose.yml
dato: I thought this line was going to "mask" _dev-tools_ from the working directory, and show the _dev-tools_ from the image (i.e., containing the dependencies). But if /app/dev-tools/node_modules is going to be empty, it seems to me it defeats the purpose of creating the image in the first place.

(I read Docker docs for a bit: it would be great if "dev-tools" subdirectory could be skipped when mounting /app, but I didn't see a way

dato (commented):
> Is there any chance we could revisit this one and bring it to an end?

There some outstanding conflicts against main, which I fixed and pushed into my repo, in case it's useful.

It seems to work for me (there is a warning about duplicate selectors that were introduced since the last merge to main).



~I think an alternative way could be to install these JS tools globally inside the image (as black is). But I couldn't get the new way to work either. Because of my comment below, it always prompts to install things~

docker-compose.yml
dato: > I thought this line was going to "mask" dev-tools from the working directory, and show the dev-tools from the image

It seems it does that!, and I mistyped something; sorry for the noise.

PR #2103, opened 2022-05-06
Replace user|username with request.user.localname

### LINKED ISSUE #2096: Profile link on user menu incorrect if currently on another user’s profile
https://github.com/bookwyrm-social/bookwyrm/blob/e0fa0b859aa16a15ee82c496f938c92daee3141c/bookwyrm/templates/user_menu.html#L24

The profile link works as expected unless I’m on someone else’s profile and then use the user menu to try to navigate back to my own. At a glance, I suspect we need `request.user.localname` here as in the link for My Books.

Does that seem right?

denmch:
Closes #2096

### DIFF
--- a/bookwyrm/templates/user_menu.html
+++ b/bookwyrm/templates/user_menu.html
@@ -21,7 +21,7 @@
             role="menu"
         >
             <li role="menuitem">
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thanks for catching and fixing this!

PR #2140, opened 2022-05-30
Ignore VariableDoesNotExist errors in debug logging

cincodenada:
They're so noisy as to make debug logging useless otherwise, and we can add a logging filter to ignore them, turns out!


mouse-reeve:
Oh this is great! Thank you. It's irritating that so many are generated at all.

cincodenada:
Oh, I forgot I had filed a bug about this too - this should close #1906, also I ran `black`, it looks like pylint is failing on unrelated things.

cincodenada:
Okay, I finally circled back to this and fixed up the linting - I was helping another admin debug things the other day and was reminded that this was still outstanding, because the debug output was very noisy again :smile:

PR #2144, opened 2022-05-31
Don't add imported statuses to timeline

mouse-reeve:
Imported items are creating a huge number of add_status_task tasks, and
really they don't need to go in to timelines at all.


mouse-reeve:
This doesn't work because the `linked_review` field on the import item isn't set until after the review is created, which is also after this signal is triggered. An alternative would be to add a database field on review to flag it as imported.

PR #2149, opened 2022-06-05
Update preview image generation to only query ids

cincodenada:
I just turned preview images back on and tried to regenerate preview images on my instance which apparently now has nearly 50,000 books, and the preview generation failed to generate the books. It turns out the process was getting OOM killed because it was querying the whole book object and trying to build a list of a fully-hydrated object for every book in my database in memory, which was more than my meager VPS could handle.

Fortunately, since we're just dispatching jobs here, we only need to query for the id of each book, which is much less memory-intensive (and faster, to boot). With this change in place, my server is happily on its way regenerating all the preview images once again 🎉


bookwyrm/management/commands/generate_preview_images.py
```diff
@@ -56,12 +56,17 @@ def handle(self, *args, **options):
             self.stdout.write(" OK 🖼")
 
             # Books
-            books = models.Book.objects.select_subclasses().filter()
+            book_ids = [
+                values["id"]
+                for values in models.Book.objects.select_subclasses()
+                .filter()
+                .values("id")
+            ]
```
mouse-reeve, line 64: I think you can use `.values_list("id", flat=True)` instead of a list comprehension here

mouse-reeve (changes requested):
Minor tweak requested, but thank you so much for noticing and fixing :)

bookwyrm/management/commands/generate_preview_images.py
cincodenada: Oooh, I looked briefly for something like that but skimmed right over the `flat` option that's exactly what I wanted. Thanks!

cincodenada:
Real-world testing update: since adding this patch and starting the process yesterday my server is now about halfway through generating all the book thumbnails 😅 Slowly but surely!

PR #2150, opened 2022-06-05
Fix pylint config for pylint 2.14.0

cincodenada:
Pylint 2.14.0 was just released with new warnings and a deprecated one that broke the pipeline, since it uses the latest pylint.

I added a `.pylintrc` and comands to `bw-dev`, and added `pylint` to `requirements.txt`, so we can run pylint locally.

I also rearranged the dockerfile for optimal caching (assuming that system packages will change less often than the python requirements)

I also updated the GitHub workflow, with pylint in `requirements.txt` it doesn't need to install it any more (and now we're pinned to a version so it won't randomly break in the future), and it can use the `.pylintrc` instead of having the options inline.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -21,8 +21,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install pylint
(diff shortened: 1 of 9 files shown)


mouse-reeve:
Hurray!

Issue #2052, opened 2022-03-22
Create a walkthrough for new users


mouse-reeve:
It would be great to give new users who want it a "guided tour" of the application.

hughrun:
I'm not promising to do it right now but I would like to volunteer to look into this. It's difficult to get your head around all the concepts at first.

hughrun:
I'm working on this now, using Shepherd.

hughrun:
whoops @mouse-reeve I must have forgotten to trigger an auto-close for this.

PR #2168, opened 2022-07-03
Add guided tour / walkthrough

hughrun:
This is intended to address #2052 

**I expect to have to make changes in response to feedback for this, so please provide any you have**. I've tried to cover all the basics without overwhelming new users. I may not have gotten that balance right. 

Uses [Shepherd](https://shepherdjs.dev/docs/index.html) (MIT license). My understanding is that this is pretty good with accessibility, hopefully that's correct.

I have set this up so that it _should_ pick up styles from the active theme, whatever it is. This requires some slightly hacky work on the default Shepherd CSS, so I'm happy to receive any feedback on that if I messed it up.

All elements used to anchor guided tour modals have an id - if I added the id (which is the case for nearly all of them) I've used a `tour-` prefix to make it harder to accidentally mess up the tour and not fix when inevitably things get moved around in future. All tour wording should be translatable, but it would be great if someone can double check that.

Every user will now have a `show_guided_tour` setting, set to `True` by default. This is easy to turn off on the first pop up modal on every page the tour appears on. Users can trigger the tour (i.e. turn their setting back to `True`) by clicking on a link in the footer. If JavaScript is disabled a message will appear next to this link noting that JS is required for the tour.


bookwyrm/templates/guided_tour/book.html
```diff
@@ -0,0 +1,160 @@
...
+        {
+            text: "{% trans 'You can also manually add reading dates here. Unlike changing the reading status using the previous method, adding read-through dates manually will not automatically add them to your ' %}\
+            <code>\
+            {% trans 'Read' %}\
+            </code>\
+            {% trans ' or '%}\
+            <code>\
+            {% trans 'Reading' %}\
+            </code>\
+            {% trans ' shelves. Got a favourite you re-read every year? We\'ve got you covered - you can add multiple read dates for the same book 😀' %}",
```
mouse-reeve, line 62: Breaking up the translation string will make this hard to translate. It would work better to include it in one long block or break it up by sentence. It's better to include the html in the string rather than end the translation block to avoid it.

If you want to keep the whitespace for readability, you can use `{% blocktrans trimmed %}...{% endblocktrans %}` instead of `{% trans "..." %}`.

bookwyrm/migrations/0151_user_show_guided_tour.py
```diff
@@ -0,0 +1,19 @@
...
+    dependencies = [
+        ("bookwyrm", "0150_readthrough_stopped_date"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="user",
+            name="show_guided_tour",
+            field=bookwyrm.models.fields.BooleanField(default=True),
+        ),
```
mouse-reeve, line 18: I think for existing users, we don't want to show the tour (right?) -- the default value is correct here, but there could be a `RunPython` step that sets the boolean to false for existing users.

bookwyrm/templates/guided_tour/search.html
```diff
@@ -0,0 +1,171 @@
...
+                        return this.next();
+                    },
+                    text: "{% trans 'Next' %}",
+                },
+            ],
+        });
+    } else if (localResult) {
+    tour.addStep(
+        {
+            text: "{% trans 'If the book you are looking for is already on this server, you can click on the title to go to the book\'s page.' %}",
```
mouse-reeve, line 48: From an internal consistency perspective, `instance` is more correct than `server`. But I wonder if there's a way to phrase this that doesn't use it at all? I'm trying to avoid making people have to learn jargon....

bookwyrm/templates/guided_tour/user_books.html
```diff
@@ -0,0 +1,139 @@
...
+            <code>{% trans 'Read' %}</code>,\
+            {% trans ' and ' %}\
+            <code>{% trans 'Stopped Reading' %}</code> \
+            {% trans 'are default shelves used to indicate the current reading status of books. A book can only be assigned to one of these shelves at any given moment e.g. a  book is not allowed to be on both the ' %}\
+            <code>{% trans 'Currently Reading' %}</code>, \
+            {% trans ' and the ' %}\
+            <code>{% trans 'To Read' %}</code>, \
+            {% trans ' shelves. When you change the reading status of a book it will automatically be moved to the matching shelf.' %}\
+            ",
+            title: "{% trans 'Creating a group' %}",
```
mouse-reeve, line 42: I think this is the wrong header :)

bookwyrm/templates/guided_tour/user_books.html
```diff
@@ -0,0 +1,139 @@
...
+            text: "<code>{% trans 'To Read' %}</code>, \
+            <code>{% trans 'Currently Reading' %}</code>, \
+            <code>{% trans 'Read' %}</code>,\
+            {% trans ' and ' %}\
+            <code>{% trans 'Stopped Reading' %}</code> \
+            {% trans 'are default shelves used to indicate the current reading status of books. A book can only be assigned to one of these shelves at any given moment e.g. a  book is not allowed to be on both the ' %}\
+            <code>{% trans 'Currently Reading' %}</code>, \
+            {% trans ' and the ' %}\
+            <code>{% trans 'To Read' %}</code>, \
+            {% trans ' shelves. When you change the reading status of a book it will automatically be moved to the matching shelf.' %}\
```
mouse-reeve, line 40: I'm trying to think of a more concise way to phrase this but haven't yet 🤔

bookwyrm/templates/guided_tour/user_books.html
```diff
@@ -0,0 +1,139 @@
...
+                {
+                    action() {
+                        return this.next();
+                    },
+                    text: "{% trans 'Next' %}",
+                },
+            ],
+        },
+        {
+            text: "{% trans 'If you have an export file from another service like Goodreads or LibraryThing, you can import your books and shelves here.' %}",
```
mouse-reeve, line 87: This doesn't actually let you import shelves, just books

bookwyrm/templates/guided_tour/lists.html
```diff
@@ -0,0 +1,214 @@
...
+                {
+                    action() {
+                        return this.next();
+                    },
+                    text: "{% trans 'Next' %}",
+                },
+            ],
+        },
+        {
+            text: "{% trans 'You can also decide how your list is to be curated - only by you, by anyone, or by a group.' %}\
```
mouse-reeve, line 122: I think this part of the tour would be easier to follow if it steps you through a simpler list create process, rather than nesting learning about groups in learning about creating lists. Maybe instead it could do groups first, and then lists, so the group is already available?

mouse-reeve (changes requested):
This is so cool! I'm going to think more about the content, but some initial thoughts:

The tour boxes blend into the background for me, and I think it would help to have a sense of depth or highlight on the box. Maybe a more prominent `box-shadow`? Probably not this exactly, but just thinking about visual distinction:
Screen Shot 2022-07-03 at 7 51 54 AM
vs.
Screen Shot 2022-07-03 at 7 51 23 AM

I think it would help to visually pull out and emphasize when the tour wants you to do something specific. It's hard at a glance to tell the difference between "if you wanted to do x, this is where you'd do it" and "please do x now to continue the tour." Maybe having the action in a box with an icon or something like that.

Something seems to have caused the user icon to end up on a different line than the username in the top bar profile menu:
Screen Shot 2022-07-03 at 7 55 41 AM

The main thing I think needs adding is a guide to the timelines and how to post.

bookwyrm/templates/guided_tour/search.html
hughrun: Agreed. In this particular case it does matter (I think?). Will have a think.

bookwyrm/migrations/0151_user_show_guided_tour.py
hughrun: That is probably sensible. I'm bit sure how to do that - can you point me to the relevant script/file?

bookwyrm/templates/guided_tour/book.html
```diff
@@ -0,0 +1,160 @@
...
+                    action() {
+                        return this.next();
+                    },
+                    text: "{% trans 'Next' %}",
+                },
+            ],
+        },
+        {
+            text: "{% trans 'You can also manually add reading dates here. Unlike changing the reading status using the previous method, adding read-through dates manually will not automatically add them to your ' %}\
+            <code>\
```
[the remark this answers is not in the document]
hughrun: I think I agree, I initially was just playing around and then ...just kept using <code> 🙃

bookwyrm/templates/book/book.html
```diff
@@ -405,4 +405,7 @@ <h2 class="title is-5">{% trans "Lists" %}</h2>
 {% block scripts %}
 <script src="{% static "js/tabs.js" %}?v={{ js_cache }}"></script>
 <script src="{% static "js/autocomplete.js" %}?v={{ js_cache }}"></script>
+{% if user.show_guided_tour %}
```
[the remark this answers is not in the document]
hughrun: it's only needed if the user wants the guided tour?

bookwyrm/templates/guided_tour/user_books.html
hughrun: LOL whoops

bookwyrm/templates/guided_tour/lists.html
hughrun: Originally I did it that way! But I felt that learning shelves and adding a book is likely the first thing people want to know, and lists naturally follows from that, so doing groups first seemed a bit obtuse. Will think a bit more about this, the right order is tricky.

bookwyrm/templates/book/book.html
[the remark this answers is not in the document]
mouse-reeve: am I misreading this? My impression is that it's checking if the logged in user has the guided tour enabled, in which case `request.user` represents the logged in user, rather than `user`, which may or may not be set to anything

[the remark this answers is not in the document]
hughrun: oh you're right, I misunderstood what you meant! Nice catch.

bookwyrm/migrations/0151_user_show_guided_tour.py
mouse-reeve: There are a bunch in the migrations, this being one: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/migrations/0142_auto_20220227_1752.py#L7. But the docs are actually probably clearer than the examples: https://docs.djangoproject.com/en/3.2/ref/migration-operations/#django.db.migrations.operations.RunPython

bookwyrm/templates/guided_tour/search.html
mouse-reeve: Yeah, it's hard to explain books being available but not currently on your instance without talking about that explicitly...

bookwyrm/migrations/0151_user_show_guided_tour.py
hughrun: I'm having a bit of trouble understanding the docs. Does `RunPython` run immediately after the migration? The way it's written sound to me like it runs immediately before, which doesn't seem all that helpful.

Would I add a function to `bulk_update` here to overwrite the default for existing users?

mouse-reeve: The migration runs a series of steps that are in the `operations` array, and since they're in an array, they run sequentially. That means you can add your own step before or after any of the steps in the migrations. If you had a bunch of different fields changed, each `AddField`, `RemoveField`, `AlterField`, or whatever, would be its own step in the `operations`, and you could add `RunPython` scripts anywhere you'd like.

 So you have:
``` python
    operations = [
        migrations.AddField(
            model_name="user",
            name="show_guided_tour",
            field=bookwyrm.models.fields.BooleanField(default=True),
        ),
    ]
```

And you can add a script step after `AddField`:
``` python
    operations = [
        migrations.AddField(
            model_name="user",
            name="show_guided_tour",
            field=bookwyrm.models.fields.BooleanField(default=True),
        ),
        migrations.RunPython(forward_function, migrations.RunPython.noop),
    ]
```

Then you just have to write a function (which here I called `foreward_function` to be generic) that does what you want. The second parameter (I used `migrations.RunPython.noop`, which does nothing) runs if you undo the migration. 

I don't know why, but in a migration function you have to call models using `schema_editor` and `db_alias`, so what would normally be:
``` Python
from bookwyrm import models
def forward_func():
    models.User.objects.filter(...)
```

Becomes
``` Python
def forward_func(apps, schema_editor):
    db_alias = schema_editor.connection.alias
    user_model = apps.get_model("bookwyrm", "User")
    user_model.objects.using(db_alias).filter(...)
```

I think you can use `.update()` on the queryset to bulk update? Something like:
``` Python
user_model.objects.using(db_alias).filter(local=True).update(show_guided_tour=False)
```

But there might be a more efficient way to do that.

hughrun:
> Something seems to have caused the user icon to end up on a different line than the username in the top bar profile menu

I've tracked down what is making this happen, but not yet _why_. For some reason classes are being applied twice, which causes `.img` (which has `display: block`) to override `.avatar` instead of the other way around.

hughrun:
> I've tracked down what is making this happen, but not yet _why_. For some reason classes are being applied twice, which causes `.img` (which has `display: block`) to override `.avatar` instead of the other way around.

This happened because I was using `@import` instead of `@use` in the scss file.

hughrun:
There are still some tests failing. Seems I may have broken something to do with views, not sure what.

mouse-reeve:
It looks like the tests are all html validation errors caused by elements with id `tour-spoiler-alert` appearing multiple times on the same page

bookwyrm/templates/guided_tour/home.html
```diff
@@ -0,0 +1,216 @@
...
+            {
+                action() {
+                    return this.next();
+                },
+                text: "{% trans 'Next' %}",
+            },
+        ],
+    },
+    {
+        text: "{% trans 'Updates from people you are following will appear on your timeline here.<br><br>The <strong>Books</strong> tab shows just the updates about reading activity.' %}",
```
mouse-reeve, line 144: I would change the description of the books tab here, because as it is I think it makes it sound like you see a subset of your home timeline, rather than updates from anyone posting about your books.

Maybe something like "The **Books** tab shows you updates from anyone about your books"?

bookwyrm/templates/guided_tour/user_profile.html
```diff
@@ -0,0 +1,148 @@
...
+            {
+                action() {
+                    return this.next();
+                },
+                text: "{% trans 'Next' %}",
+            },
+        ],
+    },
+    {
+        text: "{% trans 'The Books tab shows your book shelves. We\'ll explore this next.' %}",
```
mouse-reeve, line 100: I got this text, but then the next step was to search for a book, which seems like a different next thing. Maybe "We'll explore this later."?

bookwyrm/templates/guided_tour/direct_messages.html
```diff
@@ -0,0 +1,81 @@
+{% load i18n %}
+
+<script>
+    const tour = new Shepherd.Tour({
+        exitOnEsc: true,
+    });
+
+    tour.addSteps([
+        {
+            text: "{% trans 'You can send a Message to other Bookwyrm users. To send a message, use the <strong>@</strong> symbol with their username(s), followed by your message.<br><br>For example: <strong>@mouse Hi Mouse, I love reading books too!</strong>' %}",
```
mouse-reeve, line 10: I think it's better to skip the example, since it won't make a lot of sense to people who don't know who I am, and if there's a poor unsuspecting user on another instance with that username, they might get some odd DMs

bookwyrm/templates/guided_tour/direct_messages.html
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    action() {
+                        return this.next();
+                    },
+                    text: "{% trans 'Next' %}",
+                },
+            ],
+        },
+        {
+            text: "{% trans 'If your message is about a book you\'re reading, click here to put a spoiler alert on your message. This will hide your message behind the warning until the recipient wants to see it.' %}",
```
mouse-reeve, line 36: This step isn't working for me -- after the info about DMs, I click next and the tour disappears

bookwyrm/templates/guided_tour/direct_messages.html
mouse-reeve: Overall, I wonder if it would be better to do the post status box on the book page rather than DMs? DMs are less commonly, and the book-related statuses are more platform-specific

mouse-reeve (changes requested):
This looks great and works great, just some quibbling about wording and steps :)

bookwyrm/templates/guided_tour/home.html
hughrun: To be honest I was struggling to understand what the Books tab is for - I didn't realise it was for "your books", though not you've pointed that out it's rather obvious. Thanks for clarifying!

bookwyrm/templates/guided_tour/user_profile.html
hughrun: Ah yes. In my mind it is "next", but we have to search for a book and add it before we can look at our book shelves, which is extremely not "next" to someone brand new to the app.

bookwyrm/templates/guided_tour/direct_messages.html
hughrun: Yep makes sense, will adjust that. It's useful to have made this step though because I found a bug lol.

bookwyrm/templates/guided_tour/direct_messages.html
hughrun: Per your comment above, I think you're right that this is pretty obscure for most new users, and it's best to skip this whole section of the tour.

hughrun:
thanks for the review, I'll make some adjustments in the next day or two.

bookwyrm/templates/guided_tour/home.html
mouse-reeve: The tour is doing it's work already :)

hughrun:
Ok I think I addressed all concerns now. I also fixed an issue with responsive menu elements not being picked up in the tours when they became hidden.

Assuming @mouse-reeve is happy with this, there's a couple of others things to do to cover this off:

1. Instance admins will need to run `./bw-dev collectstatic` in addition to the migrations to make the tour work, because it amends the JS and CSS files.
2. We need to decide and document somewhere what the process will be when interface or functional changes are made that require the guided tour to be amended, and _also_ when a new feature is added and how we decide whether/where/how to add that to the tour.

mouse-reeve:
`collectstatic` is run as part of the instance update command, so that shouldn't be an issue (if people aren't using that command, they'll have run into problems long before now).

It should be possible to get unit tests to verify that the tour elements on the page match up to elements on the rendered page, which would be a basic way to at least see if there's been a code change that disconnects part of the tour. I think this could be added to the helper function that checks html validity in the unit could do this. #2201 will be a good test case of changing something in the UI that's part of the tour.

bookwyrm/templates/guided_tour/home.html
```diff
@@ -0,0 +1,220 @@
...
+            {
+                action() {
+                    return this.next();
+                },
+                text: "{% trans 'Next' %}",
+            },
+        ],
+    },
+    {
+        text: "{% trans 'Your profile, books, direct messages, and settings can be accessed by clicking on your name here.<p class=\'notification is-warning is-light mt-3\'>Try selecting <strong>Profile</strong> from the drop down menu to continue the tour.</p>' %}",
```
mouse-reeve, line 195: For some reason this step isn't showing for me?

mouse-reeve (changes requested):
I hate requesting changes _again_ because this is rad and I'd like to get it merged, I just can't figure out why that one step isn't showing up. Am I missing something?

But pending that I think it's great and I'm very appreciative of your patience while I haven't been the most responsive

hughrun:
> I hate requesting changes _again_ because this is rad and I'd like to get it merged, I just can't figure out why that one step isn't showing up. Am I missing something?
> 
> But pending that I think it's great and I'm very appreciative of your patience while I haven't been the most responsive

No problem - I realised about halfway through coding this all up that I had created a never-ending backlog of updates, given the app will change haha

Lemme take a look at this, I suspect it may be to do with the responsive design.

bookwyrm/templates/guided_tour/home.html
hughrun: Ok I am pretty sure I have fixed this now - thanks for catching it. It was a dodgy element selector to anchor the step - see the commit message for more info.

hughrun:
> It should be possible to get unit tests to verify that the tour elements on the page match up to elements on the rendered page, which would be a basic way to at least see if there's been a code change that disconnects part of the tour. I think this could be added to the helper function that checks html validity in the unit could do this. #2201 will be a good test case of changing something in the UI that's part of the tour.

Let me know if you'd like me to work on this also - I may need some initial pointers re how it works at the moment.

mouse-reeve:
I think that should be a follow-up PR, and I'm happy to do it, it will be fussy and require knowing specifics about how I set up the tests that aren't that interesting

mouse-reeve (approved):
Looks good! Thank you :)

PR #2196, opened 2022-07-07
Block some bw-dev commands from being run in prod

mouse-reeve:
Right now, commands that should not be run in production are removed
from the bw-dev file in the `production` branch. Since eventually I'd
like to get rid of that branch, this change would use the `DEBUG`
environment variable to determine if a command should be disabled,
rather than depending on the file available in the branch.

Works on #1858

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -3,6 +3,17 @@
 # exit on errors
 set -e
 
+# check if we're in DEBUG mode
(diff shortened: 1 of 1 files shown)


mouse-reeve:
cc @cincodenada I can't figure out how to request a review through the UI for some reason but I think this would be of interest to you

PR #2201, opened 2022-07-08
Adds "Your Books" to the header and removed "Feed"

mouse-reeve:
The instance icon is already a link to the feed, and I think it's weird not to have a quick link to your books, since books are important.

Before:
Screen Shot 2022-07-08 at 10 54 32 AM

After:
Screen Shot 2022-07-08 at 10 51 11 AM


kopischke:
I’m wondering if the loss of discoverability (the icon is not obviously clickable), never mind accessibility (loss of a text label), is worth this. Wouldn’t it be preferable to simply have the currently static “Your Books” heading in the sidebar link to the books?

mouse-reeve:
> I’m wondering if the loss of discoverability (the icon is not obviously clickable), never mind accessibility (loss of a text label), is worth this. Wouldn’t it be preferable to simply have the currently static “Your Books” heading in the sidebar link to the books?

Accessibility is a good point! If we go with this design, the home icon can have a screen reader label that clarifies that it directs to the feed.

But I agree that making the feed hard to find is my main concern. I'm not totally sure what you mean by side bar link? An alternative option would be to keep all four, but my concern is that the top bar risks becoming cluttered:
Screen Shot 2022-07-08 at 11 36 11 AM

joelchrono12:
Perhaps a more drastic change is needed.

Most search bars in most websites, including those dedicated to reading, are in the right side of the screen. Moving it there alongside the profile pic and notifications would allow the feed, lists and rest of links together to the left.

kopischke:
> I'm not totally sure what you mean by side bar link?

I was thinking of the left sidebar area of the feed.

image

As I write this, I realise this is only available from the feed, which tells me two things:

1. I’ve never felt the need to jump to my books from anywhere but the feed, which works as a “Home” view for me (something the fact that the BW icon links to it implicitly reinforces);
2. The items in the top menu might not actually be equal. Maybe a more general redesign of the top bar menu is in order?

joelchrono12:
Oooh, yeah, that sounds like a nice way to do it too. Maybe something like this (really badly done using the inspector xD) may work too
image

hughrun:
At the risk of over-complicating this, what if we made what appears in this menu configurable by the user? If space is a problem, we could limit it to three menu items you can "pin", but each user can choose from a list of options in the user settings. Of course, that doesn't resolve the question of what the default options would be, though I think from an onboarding point of view the current three (Feed, Lists, Discover) is probably correct.

mouse-reeve:
I re-read this thread and i think the general consensus is to merge this and also that in the future we could re-think the title bar in a bigger picture sense? Please comment if you strongly object to me merging this -- I'm trying to clean up old PRs, and this one keeps irking me in the UI.

dato:
I haven't been using BookWyrm for long, and after reading all comments my humble suggestion would be: _do keep the Feed header, remove the Discover one, make the icon a link to it instead_.

Rationale: when you're logged out, the home page shows "discovery entries". If, say, I'm logged out and visit https://lectura.social/about, then the top-left instance icon takes me back to a "discovery-like view".

_Perhaps having the icon do the same while logged in would make sense._ Also the icon represents the whole instance, which seems a good fit for discovery content.

Just a thought, sorry if it just adds noise to an almost-settled topic.

hughrun:
@mouse-reeve I think you should merge this per your comment on August 7, and additional finessing can be discussed further if desired.

PR #2212, opened 2022-07-09
Specify metadata about an author's relationship to a book

### LINKED ISSUE #1120: Multiple Types of Authors
A lot of books have editors, contributors, or translators. Being able to specify author types would be useful, as opposed to the current system where everyone is treated as an author.

### LINKED ISSUE #1944: Add illustrator field for books
Occasionally I run into a book on bookwyrm that has an illustrator. In order to credit the illustrator and allow people to search for an illustrator they like, I'll often include the illustrator as a secondary author. However, this isn't ideal and could be confusing. I would love to see an optional "illustrator" field alongside the "author" field.

This could either work exactly as the author field does but with an "Illustrator" label, or it could be implemented as a drop-down label selection for new authors/contributors. This second option might also be helpful for editors, essayists, or people who have contributed a foreword. 

An example of a book with an illustrator listed as an author is [Plain Bad Heroines](https://bookwyrm.social/book/95045), where [Emily M. Danforth](https://bookwyrm.social/author/12944) is the author and [Sara Lautman](https://bookwyrm.social/author/12945) is the illustrator.

mouse-reeve:
This PR creates an explicit join table for the many-to-many relationship between Book and Author, which contains a field for the type of author (ie, author, illustrator, translator, etc). 

The current list of author types is:
```
    ("author", _("Author")),
    ("editor", _("Editor")),
    ("translator", _("Translator")),
    ("contributor", _("Contributor")),
    ("preface", _("Author of Preface")),
    ("illustrator", _("Illustrator")),
    ("letterer", _("Letterer")),
    ("narrator", _("Narrator")),
    ("other", _("Other")),
```

Work:
 - [x] Create new join table
 - [x] Write migration
 - [x] Display author type in the UI
 - [x] Set author type when adding author to book
 - [ ] Make author type editable
 - [ ] Specify author type in ActivityPub

Fixes #1120
Fixes #1944


bookwyrm/migrations/0154_bookauthor.py
```diff
@@ -0,0 +1,74 @@
...
+                    "remote_id",
+                    bookwyrm.models.fields.RemoteIdField(
+                        max_length=255,
+                        null=True,
+                        validators=[bookwyrm.models.fields.validate_remote_id],
+                    ),
+                ),
+                (
+                    "author_type",
+                    models.CharField(
```
jaschaurbach, line 39: Would love to have "Narrator" for Audiobooks. I know, there is no schema.org property for this - which is a problem from schema.org as they never went with time. 

For Graphic Novels it would be great if there would be "Lettering" as well and in general: "Cover Artwork" as this does not need to be a by the person which did the illustrations.

mouse-reeve: Thank you! Agreed

jaschaurbach: And if we really want to show appreciation to authors: Author of Preface (which has no schema.org representation, only in ONYX)

Blechpirat: I stumbled about the lack of an editor (in scientific books there often is one ore more editors who compiled the book by adding pieces from different authors). At the moment the first author is often the only one mentioned, but it should be the editor.

mouse-reeve:
If I finish enough pull requests that were started in 2022, will I start to age backwards? Anyway, this is close to working but I am not sure how to proceed including author type in activitypub. Right now, it looks like:
```
{
  "id": "https://<url>.app/book/2",
  "type": "Edition",
...
 "authors": [
    "https://<url>.app/author/1",
    "https://<ulr>.app/author/39"
  ],
...
}
  ```

I think it could be a nested dict? Or should it reference the join table somehow? I'll have to look into how things like this are parsed 🤔 


I used a simple select field when you're adding a new author like so:
Screenshot 2026-06-17 at 4 34 29 PM

I think the best course of action is to copy how series work when editing the author type on existing authors, so I'm going to look at how that was implemented to finish that aspect of this RP

@bookwyrm-social/code-review

**Comment by mouse-reeve** [comment]
I'm also not sure how to get the [search vector](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/book.py#L444) to include author after this change

**Comment by hughrun** [comment]
If you're looking for more options for contributions types, see my comments on Contributor at https://github.com/bookwyrm-social/bookwyrm/issues/3681 Specifically, check [ONIX values](https://ns.editeur.org/onix/en/17) which is generally the standard used by publishers.

I had been thinking of doing some subclassing gymnastics to make `Author` a subclass of `Contributor`, for similar effect.

Something to think about here is connectors. Most won't give us more than a generic "author" but Inventaire provides much more specific contribution types and potentially SRU connectors will as well.

@mouse-reeve can you explain more about the problem with search-vector?

**Comment by mouse-reeve** [comment]
Ah I missed that, amazing. I will take a look and see whether this makes sense in all that context. I've only really determined that search vector is no longer adding the author after this change, but I don't know why

**Comment by jaschaurbach** [comment]
> I think it could be a nested dict?

The first thing that came to my mind was https://codeberg.org/fediverse/fep/src/branch/main/fep/e232/fep-e232.md (Object Links) - so the authors get linked to the object AND some software can understand this already.

**Comment by mouse-reeve** [comment]
Okay, I'm glad you thought this through more than I did in 2022/yesterday. Tell me if this all seems like I'm understanding correctly.

It sounds like `Contribution` is the same functionally as `AuthorBook` here, but more sensibly named. Using ONIX types is great and I'm 100% in. It acts as a join table between the contributor and the book, and stores additional data about the type of contribution. I'm not seeing why you would need to subclass it, though. It seems like you get the same information from the [enum of contributor types](https://github.com/bookwyrm-social/bookwyrm/pull/2212/changes#diff-ffdad197341d69624eed89fcfa140fd47de514c7b217b4ee6202e1e7cf1baf71R950) - that encompasses the fields you suggest for `contribution_type` and `contribution_desc`. (I'm also not sure why you need both `onix_code` and `contribution_type`? And I think it would make more sense to align wikidata keys with onix codes in the inventaire connector, but that's just implementation details).

Authors already linked to `Book` instead of `Edition` or `Work`, but `Book` is used as an abstract class (even though it isn't), so what that means is that both `Edition` and `Work` can have authors. Because `Work`s are kind of hidden, this isn't obvious, but it's generally designed to have `Edition`s and `Work`s have their own, complete, independent set of author links -- so a work would have an author, and an edition would have the same author (separate entry in the join table), and also its translator, which is not linked to the work. A more philosophically perfect schema might be for the Editions to inherit the Work's author, but I think this would be very challenging for implementation. (Or at least a separate piece from creating the `Contributor` table)

I'm struggling to figure out how to keep it backwards compatible. Maybe by rending the existing `Author` activity json merged with the new `Contributor` activity?

**Comment by jaschaurbach** [comment]
> Authors already linked to `Book` instead of `Edition` or `Work`, but `Book` is used as an abstract class (even though it isn't), so what that means is that both `Edition` and `Work` can have authors. Because `Work`s are kind of hidden, this isn't obvious, but it's generally designed to have `Edition`s and `Work`s have their own, complete, independent set of author links -- so a work would have an author, and an edition would have the same author (separate entry in the join table), and also its translator, which is not linked to the work. A more philosophically perfect schema might be for the Editions to inherit the Work's author, but I think this would be very challenging for implementation. (Or at least a separate piece from creating the `Contributor` table)

Works has one or several authors
Every book has at least one work in it.
The book makes the edition which has the author(s) from the work plus more roles: Cover illustration, interior illustration, maybe translation, editor(s), can have one or several narrators when it is an audiobook. 

So: Work and edition ALWAYS need two have the same author(s) - as an edition can have several authors (e.g an anthology - I know, we so not have that at the moment but maybe one day - we should stay prepared) the edition needs to inherit the authir(s) from the work. They can not be independent from each other. so I think a first step after an update to the version with this feature needs to be to change the database entries and let reflect that edition inherits the author(s) from the work. 

> I'm struggling to figure out how to keep it backwards compatible. Maybe by rending the existing `Author` activity json merged with the new `Contributor` activity?

This is a good idea. If there is a new contributor json bookworm needs to ignore the author json - an older version will only work with the author and disregard the part it does not understand. But: I think it is important to think about WHEN we will deprecate author - because otherwise we will have this part of code unmaintained but forever in the codebase.

BTW: I am back thinking about this beautiful piece of software :)

**Comment by todrobbins** [comment]
> If you're looking for more options for contributions types, see my comments on Contributor at #3681 Specifically, check [ONIX values](https://ns.editeur.org/onix/en/17) which is generally the standard used by publishers.
> 
> I had been thinking of doing some subclassing gymnastics to make `Author` a subclass of `Contributor`, for similar effect.
> 
> Something to think about here is connectors. Most won't give us more than a generic "author" but Inventaire provides much more specific contribution types and potentially SRU connectors will as well.
> 
> @mouse-reeve can you explain more about the problem with search-vector?

One more option to consider for contributor is [BIBFRAME relators/roles](https://id.loc.gov/vocabulary/relators.html).

**Comment by hughrun** [comment]
> It sounds like Contribution is the same functionally as AuthorBook here. ...It acts as a join table between the contributor and the book, and stores additional data about the type of contribution.

Yes!

> It seems like you get the same information from the [enum of contributor types](https://github.com/bookwyrm-social/bookwyrm/pull/2212/changes#diff-ffdad197341d69624eed89fcfa140fd47de514c7b217b4ee6202e1e7cf1baf71R950) - that encompasses the fields you suggest for contribution_type and contribution_desc. (I'm also not sure why you need both onix_code and contribution_type?

This was mostly just in my head rather than any functioning code so this may well be all correct! It's pretty common for me to overengineer models or views on the first run and then pare things back. 

In principle though, what I think we need is a standard set of identifiers (ideally not yet another standard we invent) to define what kind of contribution it is, and then everything flows from there. So an enum might be the right approach, but I kind of want something that not only provides a translatable name but also a translatable description, and can be found via a common identifier. I'm suggesting ONIX because up until we added the Finna connector all of our data was coming from sources based more in the publishing world than the library world. @todrobbins is right to point to BIBFRAME because that's the future/present of library description, and as soon as we have an SRU connector (which is probably my next job since you're already on the case for contribution types) we will increasingly be bringing in data from library systems. So we ideally want something that can easily be mapped ("cross-walked") between ONIX, Wikidata IDs, BIBFRAME relators, and any other relevant standard identifiers. 

I agree this is an implementation detail in one sense, but in another sense we need to have a pretty clear idea about how we expect things to be implemented when defining the `Contributor`/`AuthorBook` model in the first place. It probably makes sense as you suggest to map them in the Connector code (wikidata IDs for Inventaire, BIBFRAME and MARC values for SRU), but we still need something solid and capacious enough for them to latch onto. I do think ONIX is still our best bet, but hopefully this all explains what I was trying to think through.

> I'm not seeing why you would need to subclass it, though.
> I'm struggling to figure out how to keep it backwards compatible.

I hadn't completely thought this through but my idea was to essentially leave `Author` in place but pull everything out of it into a `Contributor` model that it inherits from. The reason to do this would be that then when a "latest version" BookWyrm server gets an activity from an "old version" BookWyrm server, we still have `Author` and everything is happy. Likewise if we continue to make authors an `Author` and everyone else some other kind of `Contributor`, then activities from "latest version" BookWyrm still send `Author` objects. They would also send `Contributor` objects but old BookWyrm doesn't know what to do with that, so it will just dump it.

That was my theory, but as I'm typing it out now I'm not completely convinced because I guess new-style Books would still have a new Contributor attribute, which old-style BookWyrm won't expect and will probably choke on. I think I might have been mixing up activities with objects in my head 🤔.

A final note - I was also considering whether it makes sense to allow contributions directly to a Series (i.e. not via a contribution to a book). Primarily my mental model for this is an editor of a non-fiction series, who might not always be listed on each book but plays a key role in commissioning and bringing together the series. But this is probably a level of detail that would just make things confusing. We already can map authors to series via Works and their Editions, so we probably can just continue to do that and ensure a series editor is connected to at least one book in the series.

**Comment by mouse-reeve** [comment]
Oh I see! So the activitypub could serialize as:
```
{
  "id": "https://<url>.app/book/2",
  "type": "Edition",
...
 "authors": [
    "https://<url>.app/author/1",
  ].
  "translators": [
    "https://<ulr>.app/author/39"
  ],
...
}
```

And then older version would just ignore the non-author fields. I wonder how feasible it is to do this for the activitypub schema without actually subclassing the model 🤔

mouse-reeve:
I think the ONIX codes and how they map onto Wikidata and BIBFrame should not be stored in the database since they're a fixed set of things we recognize, rather than something that changes with user input. I will look into how I can change the enum to hold context for them (mappings and text descriptions), which I think will be very doable.

mouse-reeve:
> > I think it could be a nested dict?
> 
> The first thing that came to my mind was https://codeberg.org/fediverse/fep/src/branch/main/fep/e232/fep-e232.md (Object Links) - so the authors get linked to the object AND some software can understand this already.

This does seem like a sensible approach, but it wouldn't be backwards compatible. I guess an alternative approach to the one I just posted would be to use a new schema (like this one) under a new name (`contributor`) and keep the `author` field as is for backwards compatibility for a certain amount of time

PR #2246, opened 2022-07-29
Moves annual goal to its own file

mouse-reeve:
The user model file is just,, very long

### DIFF
--- a/bookwyrm/migrations/0122_alter_annualgoal_year.py
+++ b/bookwyrm/migrations/0122_alter_annualgoal_year.py
@@ -14,6 +14,8 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name="annualgoal",
             name="year",
-            field=models.IntegerField(default=bookwyrm.models.user.get_current_year),
+            field=models.IntegerField(
+                default=bookwyrm.models.annual_goal.get_current_year
+            ),
         ),
     ]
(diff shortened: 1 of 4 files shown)


franbarra:
I have 2 quick questions:

1. What does "gte" and "lt" stand for? As in "finish_date__year__gte" and "finish_date__year__lt".

2. Why is `finish_date__year__gte` equal to `year` but `finish_date__year__lt` equals `year + 1`? Let's imagine we take the year 2022; if we set an annual reading goal for 2022 does it mean we take into account books finished by 2023 as well?

Thanks in advance

mouse-reeve:
`gte` means "greater than or equal to" and `lt` means "less than" -- here's the [django documentation](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#gt) on it. So `finish_date__year__gte=year` means the year in the database row is greater than or equal to the value in the `year` variable. 

Looking at this block again:
```python
            self.user.readthrough_set.filter(
                finish_date__year__gte=self.year,
                finish_date__year__lt=self.year + 1,
            )
```

I think it could be written equivalently (and more clearly) as:
```python
            self.user.readthrough_set.filter(
                finish_date__year=self.year,
            )
```

I suspect that I was originally comparing the entire date to another date, which is why it's written like that, and then changed it to just year but didn't re-think it fully.

franbarra:
Thanks! Now it's clearer.

PR #2255, opened 2022-08-05
Add notifications count on mobile header

joachimesque:
### Normal state
Capture d’écran 2022-08-05 à 11 03 13

### Notifications
Capture d’écran 2022-08-05 à 11 14 23

### Mentions
Capture d’écran 2022-08-05 à 11 03 45


mouse-reeve:
Thank you, this works perfectly!

PR #2275, opened 2022-08-13
strip spaces from user search

### LINKED ISSUE #2205: Adding a space after typing a username will not show that user in the search results
**Describe the bug**
When I search for "@user@example.com" it won't work if the input has a space at the end ("@user@example.com ")
It may work if the user has already been searched for before though. 

**To Reproduce**
1. search a user you don't currently follow
2. add a space at the end
3. the user won't be found

**Expected behavior**
spaces should be ignored when looking for usernames

**Instance**

**Additional context**
Bookrastinating.com

---

**Desktop**
    - OS: Fedora
    - Browser Firefox
    - Version 102

hughrun:
Strips leading and trailing spaces from user search to prevent errors when doing webfinger lookup.

Fixes #2205

### DIFF
--- a/bookwyrm/views/search.py
+++ b/bookwyrm/views/search.py
@@ -90,6 +90,7 @@ def user_search(request):
     """cool kids members only user search"""
     viewer = request.user
     query = request.GET.get("q")
+    query = query.strip()
     data = {"type": "user", "query": query}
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Perfect! Thank you :)

Issue #1870, opened 2022-01-25
Followers-only replies to followers-only posts should not appear in the home feed if you do not follow the person the post was a reply to


bcj:
**Describe the bug**
This one is skirting the line between bug report and feature request. The most direct bug is: If someone you don't follow makes a followers-only post, and someone you do follow replies, the reply will appear in your feed but the 'status' (or other post type) link will take you to a 404 page.

Personally, I would consider it a bug if a reply to a followers-only post appears on my timeline if I don't follow everyone involved because it seems like it would be a faux pas to interact with that post but that is definitely more subjective and posts having different visibility makes that complicated. If pressed I would argue:

| original post | reply | visible | notes |
| ------------ | ----- | ------ | ------ |
| public           | public | If I follow replier or reply is boosted | also would be fine with "if I follow both" but that's a bigger change and not in line with mastodon |
| public     | followers-only | If I follow replier | see above |
| followers-only | public | If I follow replier or reply is boosted | kind of think this shouldn't be allowed but that's a fediverse complaint |
| followers-only | followers-only | If I follow both | |

The above is definitely subjective, so mostly putting out feelers on whether it makes sense as a feature request to be made on how the home feed displays, a feature request for filters (e.g., Add a 'Hide replies to accounts you don't follow' Checkbox to Filters) or just undesired behavior

**To Reproduce**
1. Create a followers-only status with account A, which account C doesn't follow
2. Reply to that status with account B, which account C does follow
3. Find the reply in account C's feed and click on the 'status' link in the header 'Account B replied to Account A's status'
4. View 404 page

**Expected behavior**

One of:
• reply not in feed in the first place
• 'status' not presented as a link
• 403 instead of 404 (I guess that would imply that any link of the form `SITE/user/USER@SITE/POSTTYPE/ID` should be a 403 if you don't follow user/aren't logged in. Hmm)

**Screenshots**
N/A

**Instance**
bookwyrm.social

**Additional context**

Sorry, this is a ramble-y/subjective bug

---

**Desktop (please complete the following information):**
    - OS: macOS
    - Browser Firefox
    - Version 97.0b7

quaayd:
This kind of bothers me too. if i click on a comment that is a response to someone i don't follow, I get a "Not Found, the page you requested doesn't seem to exist", which feels like gaslighting, tbh. the page exists, i'm just not able to access it.

mouse-reeve:
[...]

PR #2277, opened 2022-08-14
hide replies to posts user cannot see

hughrun:
This is a possible solution to #1870

Users should not see links to posts they are not allowed to see in their feed. The main question is how to stop that happening.

This commit hides all replies to posts if the original post was "followers only" and the user is not a follower of the original poster. The privacy of the reply is not considered relevant (except "direct").

I believe this is the cleanest way to deal with the problem, as it avoids orphaned replies and confusing 404s, and a reply without access to the context of the original post is not particularly useful to anyone. This also feels like it respects the wishes of the original poster more accurately, as it does not draw attention from non-followers to the original followers-only post.

A less draconian approach might be to remove the link to the original status in the feed interface, however that simply leads to confusion of another kind since it will make the interface inconsistent.

Alternatively this could be moved down to only apply to "followers only" replies to "followers only" statuses, but that doesn't make much sense to me since a public reply to a followers-only post is just as problematic.

Small side-note: if the original author does not follower the replier, they will get a **notification** (assuming they are `@mentioned` in the reply) but the reply will not appear in their feed.

This commit does not change any ActivityPub behaviour - it only affects the Bookwyrm user feeds. This means orphaned posts may be sent to external apps like Mastodon.

### DIFF
--- a/bookwyrm/activitystreams.py
+++ b/bookwyrm/activitystreams.py
@@ -111,8 +111,15 @@ class ActivityStream(RedisStore):
             Q(id__in=status.user.blocks.all()) | Q(blocks=status.user)  # not blocked
         )
 
+        # only visible to the poster and mentioned users
+        if status.privacy == "direct":
+            audience = audience.filter(
+                Q(id=status.user.id)  # if the user is the post's author
+                | Q(id__in=status.mention_users.all())  # if the user is mentioned
+            )
+
         # don't show replies to statuses the user can't see
-        if status.reply_parent and status.reply_parent.privacy == "followers":
+        elif status.reply_parent and status.reply_parent.privacy == "followers":
             audience = audience.filter(
                 Q(id=status.user.id)  # if the user is the post's author
                 | Q(id=status.reply_parent.user.id)  # if the user is the OG author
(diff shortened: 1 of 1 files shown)


bookwyrm/activitystreams.py
```diff
@@ -111,8 +111,18 @@ def get_audience(self, status):  # pylint: disable=no-self-use
...
+                Q(id=status.user.id)  # if the user is the post's author
+                | Q(id=status.reply_parent.user.id)  # if the user is the OG author
+                | (
+                    Q(following=status.user) & Q(following=status.reply_parent.user)
+                )  # if the user is following both authors
+            ).distinct()
+
         # only visible to the poster and mentioned users
-        if status.privacy == "direct":
+        elif status.privacy == "direct":
```
mouse-reeve, line 125: If the parent status is a followers-only status, and then a reply is sent as a DM, will this mean that the DM is visible to followers, rather than only visible to tagged users?

hughrun: Hmm good question.

I *think* I tested this exact scenario and it wasn't a problem, but it might have been the other way around (a followers-only response to a DM) so I'll double check.

`HomeStream` excludes DMs so I think that prevents it, but for a belt and braces approach we could probably do:

```python
if status.reply_parent.privacy == "followers" and status.privacy not "direct" :
```

hughrun: I've checked this and it all seems to be ok. However to make it clearer, I changed the order so that DMs are at top, `elif` the parent post is followers-only... etc.

I did discover a privacy problem whilst checking this, but it wasn't created in this PR, and I'm a bit unsure how to fix it. I will add another issue for that.

mouse-reeve:
Thank you! I should be able to review this tonight

mouse-reeve:

PR #2281, opened 2022-08-27
fix resetdb

### LINKED ISSUE #2276: resetdb doesn't work
**Describe the bug**
`./bw-dev resetdb` fails with the following error message:

```
dropdb: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
	Is the server running locally and accepting connections on that socket?
```

**To Reproduce**
1. run command
2. see error
3. database is not dropped

**Expected behavior**
Dev database is wiped and docker build process is restarted

**Screenshots**

**Instance**
n/a

**Additional context**

---

    - OS: MacOS
    - Browser [n/a
    - Version `main` 

hughrun:
Instead of just whining I thought maybe I should fix the problem.

This replaces a manual reset of the database with deletion of the bookwyrm volumes using docker.

fixes #2276

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -84,16 +84,14 @@ case "$CMD" in
         ;;
     resetdb)
         prod_error
-        clean
-        # Start just the DB so no one else is using it
-        docker-compose up --build -d db
-        rundb dropdb -U ${POSTGRES_USER} ${POSTGRES_DB}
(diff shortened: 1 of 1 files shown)


mouse-reeve:
This mercifully killed my current dev setup in which I'd forgotten all the passwords I'd set for all my test accounts. Thank you!

PR #2282, opened 2022-08-28
Normalise ISBNs for searching

### LINKED ISSUE #2269: Ignore missing leading zeros when searching ISBN
Many older books have an ISBN (ISBN-10 to be precise), but omit any leading zeroes.

When searching ISBNs, permit leading zeroes to be omitted. As it stands, matches are precise, so without the leading zeroes, nothing is found.

### LINKED ISSUE #2274: Perform case-insensitive search on ISBNs
This issue interacts with #2270

The last digit of ISBNs is a checksum, which may take on a value that is not a number, i.e. X. When typing this in the search bar in lower case, books may not be found that are registered with the X in upper case.

Do case insensitive matching instead. That's probably the best default for all searchable fields.

hughrun:
- make ISBN search strings upper case for any 'X' check digits
- prepend missing zero on ancient ISBN-10s
- trim surrounding whitespace from ISBN search strings

fixes #2269  
fixes #2274

### DIFF
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -30,7 +30,9 @@ def isbn_search(query):
     """search your local database"""
     if not query:
         return []
-
+    # Up-case the ISBN string to ensure any 'X' check-digit is correct
+    # If the ISBN has only 9 characters, prepend missing zero
+    query = query.upper().rjust(10, '0')
     filters = [{f: query} for f in ["isbn_10", "isbn_13"]]
     results = models.Edition.objects.filter(
         reduce(operator.or_, (Q(**f) for f in filters))
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -32,7 +32,7 @@ def isbn_search(query):
         return []
     # Up-case the ISBN string to ensure any 'X' check-digit is correct
     # If the ISBN has only 9 characters, prepend missing zero
-    query = query.upper().rjust(10, '0')
+    query = query.strip().upper().rjust(10, '0')
     filters = [{f: query} for f in ["isbn_10", "isbn_13"]]
     results = models.Edition.objects.filter(
         reduce(operator.or_, (Q(**f) for f in filters))
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -32,7 +32,7 @@ def isbn_search(query):
         return []
     # Up-case the ISBN string to ensure any 'X' check-digit is correct
     # If the ISBN has only 9 characters, prepend missing zero
-    query = query.strip().upper().rjust(10, '0')
+    query = query.strip().upper().rjust(10, "0")
     filters = [{f: query} for f in ["isbn_10", "isbn_13"]]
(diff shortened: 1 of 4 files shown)


mouse-reeve:
This is so counterintuitive, but the `isbn_search` function only handles the case where a user navigates to an isbn using the `<domain>/isbn/00000000X` url. And you fixed it perfectly there. But for searches executed through the search box in the UI, the query goes through [`search_identifiers`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/book_search.py#L73). I think it might actually make a lot of sense to use `maybe_isbn` to redirect isbn-like queries to `isbn_search` rather than trying to modify `search_identifiers`, since that tries to do an exact match on a bunch of different database fields in aggregate, so it would be a hassle to modify the query for just the isbn field

hughrun:
I don't want to argue with the person who made this thing, but I tested this by using the search box in the UI! When using the UI searchbox it runs through `get_search_url` in  `abstract_connector.py` if `maybe_isbn` returns `True`. The original search string is displayed in the UI but it actually constructs the search URL with the normalised version.

I didn't really test `isbn_search` because I wasn't even sure when it would be used - is that when you scan a barcode with your camera? 

It did occur to me that I probably should update/add tests but then I ...didn't. This is probably also where I admit I had a feeling `search_identifiers` might be the place I needed to be amending, but I was struggling to work out how one might identify whether the deduplication field is an isbn vs something else. Admittedly I did find all the logic on this pretty confusing so I'm not surprised I missed something. Are you saying this will miss anything that is already stored in the local database, or ..something else?

>  I think it might actually make a lot of sense to use maybe_isbn to redirect isbn-like queries to isbn_search rather than trying to modify search_identifiers, since that tries to do an exact match on a bunch of different database fields in aggregate, so it would be a hassle to modify the query for just the isbn field

This seems sound, would we just do an `if` like in `get_search_url`?

mouse-reeve:
The case where I'm seeing it not work is only when you're searching for a book that is already imported into the local database. The remote search is working perfectly after your fix, so nothing else there needs to change. But try searching for an ISBN with a lowercase x (I've been using `080720563x`, then importing the book, and then searching again. You should see that the remote result shows up, but there are no local results, even though that book does exist in the database. That's because the local results are using `search_identifiers`.

`isbn_search` is literally only called if you type in the `/isbn/<isbn?>` url -- there's no instance where the app uses it implicitly. I added it at someone's request a long while back and kind of forget what the rationale was. Your fix there is correct and necessary, it's just one of the many labyrinthine endpoints in search 😵‍💫

I think the best fix would be to modify `search_identifiers` to start out at the very top by checking if something is ISBN-y, and then returning the results of `isbn_search` if so.

hughrun:
Awesome thanks for explaining! Will take a look tonight probably.

hughrun:
@mouse-reeve ok I think I fixed it. There was additionally a problem with lowercase 'x', so I'm not sure how I didn't notice that earlier.

mouse-reeve:
Perfect! Thank you

PR #2283, opened 2022-08-29
fix visibility of stars when editing a review

### LINKED ISSUE #2213: Editing a review with a partial star rounds the star value up
Editing a review with a partial star rounds the star value up

1. Post a review with a partial star
2. Click `Edit` in the post's context menu

**Expected behavior**
The star value is unchanged

**Actual behavior**
The star value is rounded up

**Instance**
reading.taks.garden

**Additional context**
Probably shares a similar root cause as https://github.com/bookwyrm-social/bookwyrm/issues/1505

hughrun:
Previously the star rating appeared to be five stars when editing a review, regardless of what value was actually stored.
Now it will show the actual rating, including half stars.

Fixes #2213

### DIFF
--- a/bookwyrm/templates/snippets/form_rate_stars.html
+++ b/bookwyrm/templates/snippets/form_rate_stars.html
@@ -37,7 +37,7 @@
             type="radio"
             name="rating"
             value="{{ forloop.counter0 }}.5"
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you for wading into this one!

PR #2288, opened 2022-09-01
conditional display of notifs button

### LINKED ISSUE #2264: Delete Notifications button should not appear if there are no notifications to delete
**Describe the bug**
The `X Delete notifications` button always appears on the page at `/notifications`.

**To Reproduce**
1. Go to `/notifications`
2. Click on `X Delete notifications`
3. Notice that button remains on screen even if there are not any notifications to delete

**Expected behavior**
It seems more logical that this button would only be visible if there are notifications to be deleted.

**Screenshots**
Screen Shot 2022-08-09 at 8 47 09 pm
Screen Shot 2022-08-09 at 8 46 46 pm

**Instance**
bookwyrm.social

**Additional context**
I'm guessing it would be easier to simply not show the button on page load if there are not any notifications, whereas "disappearing" it after clearing notifications is likely going to require JS.

---

    - OS: MacOS
    - Browser Firefox
    - Version 103.0.1

    - Device: iPhone

mooseyboots:
we make the display of the "delete notifications" button conditional on the variable `notifications` being non-nil.

from exploratory testing, it looks like nothing needs to be done regarding redisplay.

fixes #2264

### DIFF
--- a/bookwyrm/templates/notifications/notifications_page.html
+++ b/bookwyrm/templates/notifications/notifications_page.html
@@ -10,6 +10,7 @@
         <h1 class="title">{% trans "Notifications" %}</h1>
     </div>
 
+    {% if notifications %}
     <form name="clear" action="/notifications" method="POST" class="column is-narrow">
         {% csrf_token %}
         {% spaceless %}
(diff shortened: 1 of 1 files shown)


mooseyboots:
[there's no develop branch, so i guess you PR into main]

mouse-reeve:
Works like a charm, thank you!

PR #2293, opened 2022-09-09
Fix star rating default and minimum

### LINKED ISSUE #2291: unrated books show a rating of half a star
**Describe the bug**
When viewing a book you haven't yet rated, instead of 5 empty stars there is half a star already filled.

**To Reproduce**
1. Go to a book's home page
2. Notice half a star is already filled

**Expected behavior**
Stars should all be empty.

**Screenshots**

**Instance**
bookwyrm.social

**Additional context**
Pretty sure this is my fault, and was introduced in #2283 

---

hughrun:
#2283 accidentally surfaced a formerly-hidden issue with star rating defaults.

This PR:

* ensures the default rating is zero (previously the default was 0.5 star)
* allows 0.5 star ratings (previously the minimum was 1 star)

Fixes #2291


mouse-reeve:
I think this is missing a migration. Otherwise looks perfect!

hughrun:
@mouse-reeve oh yeah, I forgot I'd need to do that.

It wants to also `Alter field preferred_timezone on user` which I assume comes from something else since I didn't touch the `User` model - is that going to cause problems?

mouse-reeve:
I don't think it will cause problems, I suspect it's a side effect of updating Django that I didn't notice

hughrun:
> I don't think it will cause problems, I suspect it's a side effect of updating Django that I didn't notice

Are we on 4.1 now?

mouse-reeve:
No, just a minor version change. I can't think why else the timezones would have changed.

Issue #2250, opened 2022-08-01
2FA Support


SrEstegosaurio:
Having suppart for 2FA would be an amazing security improvement for everyone. That's it. :D

Btw: thanks for working on this amazing project. <3

SrEstegosaurio:
Something important that sometimes I see neglected on other projects are the **recovery codes**. Those need to be provided at 2FA activation time to act as a backup in case of not having acces to the codes. 

I say this because I lost a few accounts already due to them not providing recovery codes. :/

Anyways, a lot of thanks for reading my previous issue. (:

hughrun:
This would be cool.

We could probably use [pyOTP](https://pyauth.github.io/pyotp/) and [qrcode.js](http://davidshimjs.github.io/qrcodejs/)

TOTP for the standard 2FA
HOTP for backup codes

hughrun:
I'm working this now.

mouse-reeve:
We have thisnow ! Thank you Hugh!!

PR #2294, opened 2022-09-11
Enable optional 2FA

hughrun:
This addresses #2250

Users with an authentication app like Google Authenticator, MS Authenticator, Authy etc will be able to secure their account with 2FA.

This is still WIP, but I'd be grateful for any suggestions or advice.

Implemented:

- user can enable 2FA from user profile settings
- user can disable 2FA
- prevent login until TOTP code verified if 2FA enabled
- alert users without 2FA setting that it exists (once off)

TODO:

- [x] backup code generation (HOTP)
- [x] templates tidy up
- [x] display errors in templates where useful
- [x] test with MS Authenticator
- [x] write tests
- [x] more manual testing

**Questions** 
1. should admins be able to disable 2FA for users via the admin menu? I can see how this would be useful if you lose your phone or something, but also that some people may find it disconcerting.
~~2. I'm passing the username through as a hidden form input on the `two_factor_login.html` template because `request.user` doesn't work (since they're not logged in yet, which is the point). Is there a better way to do this?~~

Oh also sorry about the weird star ratings stuff confusing things - I pulled that in before it had gone into `main` because I knew there was a migration that would screw things up if it wasn't already there.


bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,123 @@
...
+        return img.to_string()
+
+
+class Confirm2FA(View):
+    """confirm user's 2FA settings"""
+
+    def post(self, request):
+        """confirm the 2FA works before requiring it"""
+        form = forms.Confirm2FAForm(request.POST, instance=request.user)
+        # TODO: show an error here
```
hughrun, line 68: remember to add error

bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,123 @@
...
+                "2fa_user": user,
+                "error": "Code does not match, try again",
+            }
+            return TemplateResponse(
+                request, "two_factor_auth/two_factor_login.html", data
+            )
+
+        # log the user in - we are bypassing standard login
+        login(request, user)
+        user.update_active_date()
```
hughrun, line 114: this is causing tests to break - take a look at this.

hughrun:
> I don't know a better way to do this, just take into account a hypothetical situation where the username is changed by the user (if they're trying to be sneaky)

I *think* this should be ok because they still need the OTP code.

hughrun:
@mouse-reeve I'm a bit confused about why these tests are failing. The login should only redirect to `/login-2FA-prompt` if `two_factor_auth` is `True`, which it shouldn't be for your test user because the value is never set, and I deliberately set the default to `None`. I really don't want to break anyone's access with this PR so if you can give me any tips for making sure the tests are accurate but also pass, that would be appreciated 😀

hughrun:
> @mouse-reeve I'm a bit confused about why these tests are failing. 

I had a sudden thought this morning and now it's _obvious_ why tests are failing.

I'll update the tests and do another check today, then hopefully this is ready for a final review.

mouse-reeve:
Oh I'm glad, I somehow completely missed that you'd tagged me in this!

hughrun:
Ok @mouse-reeve (or anyone else) this is ready for checking.

I added some tests but I had difficulty testing the `post` function on `login_with_2FA`. I think because it checks the DB for the username? Currently that test is commented out - it can be deleted if you're comfortable with that, or I will need some advice on how to get it to work (it's throwing a Redis error).

mouse-reeve:
I'm stoked to review this, and should be able to next week -- I'm having a surgery on monday though so it might be a minute before I'm coding

hughrun:
Oh my goodness! Good luck with the surgery and no rush at all! 😀

bookwyrm/templates/preferences/2fa.html
```diff
@@ -0,0 +1,78 @@
...
+    <div class="block">
+        <p>{% trans "You can generate backup codes to use in case you do not have access to your authentication app. If you generate new codes, any backup codes previously generated will no longer work." %}</p>
+        <a class="button" href="{% url 'generate-2fa-backup-codes' %}">{% trans "Generate backup codes" %}</a>
+    </div>
+    {% elif password_confirmed %}
+    <form name="confirm-2fa" action="{% url 'conf-2fa' %}" method="post" enctype="multipart/form-data">
+    {% csrf_token %}
+        <p>Scan the QR code with your authentication app and then enter the code from  your app below to confirm your app is set up.</p>
+        <div>
+            <svg width="200" height="200" style="background: white;" xmlns="http://www.w3.org/2000/svg">
```
mouse-reeve, line 47: This ended up being too small a window, which clipped the svg (I'm using Chrome on a Mac)
Screen Shot 2022-10-10 at 3 19 26 PM

bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,175 @@
...
+    """Check 2FA code matches before allowing login"""
+
+    def get(self, request):
+        """Display 2FA form"""
+        data = {"form": forms.Confirm2FAForm()}
+        return TemplateResponse(request, "two_factor_auth/two_factor_login.html", data)
+
+    def post(self, request):
+        """Check 2FA code and allow/disallow login"""
+        user = models.User.objects.get(username=request.session["2fa_user"])
```
mouse-reeve, line 111: I got a user does not exist error here when logging in with what I thought were correct credentials. I'll need to restart my janky computer before I debug further and get better replication steps, but this should handle invalid users without throwing a DoesNotExist

mouse-reeve (changes requested):
I haven't fully tested this because of my computer is too tired to move pixels and I need to restart it, but here's some initial thoughts! Overall super happy with the flow here :)

bookwyrm/templates/two_factor_auth/two_factor_prompt.html
```diff
@@ -0,0 +1,92 @@
...
+            <div class="column is-one-third m-4">
+                <header class="block">
+                    {% block header %}
+                    <h1 class="title">
+                        {% trans "2FA is available" %}
+                    </h1>
+                    {% endblock %}
+                </header>
+                <div class="is-centered p-2">
+                    <p class="block">You can secure your account by setting up two factor authentication in your user preferences. This will require a one-time code from your phone in addition to your password each time you log in.</p>
```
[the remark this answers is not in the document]
hughrun: I knew I would miss some of those...

bookwyrm/views/landing/login.py
```diff
@@ -51,11 +54,26 @@ def post(self, request):
         # perform authentication
         user = authenticate(request, username=username, password=password)
         if user is not None:
-            # successful login
+            # if 2fa is set, don't log them in until they enter the right code
+            if user.two_factor_auth is True:
```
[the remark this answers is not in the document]
hughrun: Probably!

bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,175 @@
...
+        elapsed_time = datetime.now() - datetime.fromtimestamp(
+            int(request.session["2fa_auth_time"])
+        )
+        form = forms.Confirm2FAForm(request.POST, instance=user)
+        # don't allow the login credentials to last too long before completing login
+        if elapsed_time > timedelta(seconds=60):
```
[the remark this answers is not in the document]
hughrun: I can't think of any reason it should vary on an instance level, but I don't have any particular resistance to making it a setting.

bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,175 @@
...
+            request.session["2fa_user"] = None
+            request.session["2fa_auth_time"] = 0
+            return redirect("/")
+        if not form.is_valid():
+            # make life harder for bots
+            # humans are unlikely to get it wrong more than twice
+            if "2fa_attempts" not in request.session:
+                request.session["2fa_attempts"] = 0
+            request.session["2fa_attempts"] = request.session["2fa_attempts"] + 1
+            time.sleep(2 ** request.session["2fa_attempts"])
```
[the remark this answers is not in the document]
hughrun: Honestly I hadn't thought of that! 

If I'm understanding correctly, nginx will actually take care of this [via `server_config`](https://github.com/bookwyrm-social/bookwyrm/blob/7bbe42fb30a79a26115524d18b697d895563c92f/nginx/server_config#L2) which if admins don't change anything is set to 1 request per second. I wasn't aware of that config, and it seems unnecessarily confusing to add an extra bunch of code here aimed at the same thing. Seems best if I just take this out?

bookwyrm/templates/preferences/2fa.html
hughrun: Will take a look at this, might be more sensible to use vw/vh or a percentage.

bookwyrm/views/preferences/two_factor_auth.py
hughrun: Sounds like a `session` issue but regardless yes, it should gracefully handle that, will take a look at it.

bookwyrm/views/preferences/two_factor_auth.py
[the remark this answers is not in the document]
mouse-reeve: Yes, I think take this out, and add the names of the views that should be rate limited to the nginx file, so something like
`location ~ ^/(login|password-reset|resend-link|login-2fa-check) {`

bookwyrm/templates/two_factor_auth/two_factor_prompt.html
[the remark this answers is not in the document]
mouse-reeve: I always do too, why are they like this 😅

mouse-reeve:
I'm running into a rate limiting issue -- it may have not come up for you unless you manually updated the `nginx/nginx.conf` file in your development environment after rate limiting was added (I should see if there's a better way to manage nginx configurations, but that won't happen today). When I log in, I immediately end up with a 503 service temporarily unavailable message, which goes away if I refresh.

I'm looking into why this is happening (maybe the login view is being automatically accessed multiple times which is triggering the rate limiting?)

mouse-reeve:
Oh oh, I figured it out, it's because `location ~ ^/(login|password-reset|resend-link)` matches anything that starts with `login`, so it rate limits `/login` and `/login-2fa-check` as if they are the same path. Changing the regex to `location ~ ^/(login[^-]|password-reset|resend-link)` fixes it I think

bookwyrm/views/preferences/two_factor_auth.py
```diff
@@ -0,0 +1,175 @@
...
+    def get(self, request):
+        """Display 2FA form"""
+        data = {"form": forms.Confirm2FAForm()}
+        return TemplateResponse(request, "two_factor_auth/two_factor_login.html", data)
+
+    def post(self, request):
+        """Check 2FA code and allow/disallow login"""
+        user = models.User.objects.get(username=request.session["2fa_user"])
+        elapsed_time = datetime.now() - datetime.fromtimestamp(
+            int(request.session["2fa_auth_time"])
```
mouse-reeve, line 113: It's probably safer to use `request.session.get("2fa_auth_time")` here (and above) with fallback values, because if a malformed request  is sent (like via cURL) without those keys, there's no need to have a 500 error in the application.

bookwyrm/views/preferences/two_factor_auth.py
mouse-reeve, line 117: Another question about this because I don't know that much about 2fa -- is it correct for this to be the time that the 2fa page loads, rather than the time button is pressed to submit the codes? If I'm understanding it correctly (and I might not be), if I enter my credentials, get to the 2fa page, go back a cup of tea, come back, send my codes, it will have timed out even if I used the latest codes in my authenticator app.

hughrun: Yes that is the intention. Partially this is because it just feels insecure to allow a half-authorised session to last too long. But primarily it's intended to prevent replay attacks - if we store the time the initial login was requested on the server and then expire it after a short period of time, then an attacker repeatedly rolling back the client's clock and trying to guess an old token won't work.

hughrun:
> Oh oh, I figured it out, it's because `location ~ ^/(login|password-reset|resend-link)` matches anything that starts with `login`, so it rate limits `/login` and `/login-2fa-check` as if they are the same path. Changing the regex to `location ~ ^/(login[^-]|password-reset|resend-link)` fixes it I think

Would it be better to just rename `/login-2fa-check` to, say, `/check-2fa-login`?

mouse-reeve:
I think either is an equally fine fix?

bookwyrm/views/preferences/two_factor_auth.py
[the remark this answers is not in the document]
hughrun: Yes you're right. I thought the `10m` in the config was a time limit but it's actually a memory size limit.

hughrun:
Ok I think I have addressed all issues from your review @mouse-reeve! 

Two things to note:

1. There is now a new setting: `TWO_FACTOR_LOGIN_MAX_SECONDS` which will need to be added when upgrading - I'm not sure what the process is for that
2. I added a test earlier that I have commented out because I think it needs a database mock to pass, and I can't work out how to do that. Please advise whether you'd prefer it just be removed, or if you can advise on how to fix it.

mouse-reeve:
The settings.py file is checked in and used as-is so that setting change will get out without any issue. And thank you for reminding me about that test! I meant to see if I could get the mocks working and totally forgot

bookwyrm/tests/views/preferences/test_two_factor_auth.py
```diff
@@ -0,0 +1,195 @@
...
+    # request = self.factory.post("", form.data)
+    # request.user = self.local_user
+
+    # middleware = SessionMiddleware(request)
+    # middleware.process_request(request)
+    # request.session['2fa_auth_time'] = time.time()
+    # request.session['2fa_user'] = self.local_user.username
+    # request.session.save()
+
+    # with patch("bookwyrm.views.preferences.two_factor_auth.LoginWith2FA"):
```
mouse-reeve, line 132: If you also add `patch("bookwyrm.views.preferences.two_factor_auth.login")`  here, you should be good to go. The response was a 302, though, not a TemplateResponse so you'll have to change the assertions.

bookwyrm/views/preferences/two_factor_auth.py
mouse-reeve: I got this error just now but I can't get it to happen again. Maybe just for safety wrap this in a try/except for models.User.DoesNotExist and have it return bad request rather than a 500?

mouse-reeve:
Besides the DoesNotExist error and the one unit test, this works beautifully. I'm so stoked to have this in the app!

bookwyrm/views/preferences/two_factor_auth.py
hughrun: I thought I'd sorted this out but now I see what is happening. It would be unlikely to occur outside a test environment (I think you may have a cached session from a user that doesn't exist _any more_) but yes your suggestion will fix it.

mouse-reeve (approved):

mouse-reeve:
thank you again <3

Issue #1156, opened 2021-06-02
Alternative installation guide


Amolith:
## Is your feature request related to a problem? Please describe.
I run all of my applications and services inside [LXC](https://linuxcontainers.org/). At the moment, the only supported installation method for BookWyrm is through Docker. With my architecture, this would require installing the whole Docker stack in a container in order to run a single Docker container inside the first container. In my opinion, this is incredibly inefficient; it would be better to provide a "manual" or "advanced" installation guide that describes the setup process for individuals like myself with different architectural needs.

## Describe the solution you'd like
I would simply like documentation that guides potential hosts through the process of setting BookWyrm up without relying solely on Docker. I would be more than happy to write and maintain such documentation as well as support individuals who installed BookWyrm by following it. If this is a welcome contribution, I would also ask that future release notes describe architectural changes made to the Docker-related files so that "manual" hosts can see updates via RSS and easily modify their setup accordingly.

## Describe alternatives you've considered
It's certainly possible to dissect the Docker(-compose) files and manually get a functional stack set up but dissecting those files and working out the architectural differences between them for every release and then applying those changes locally would make upgrades prohibitively time-consuming and difficult.

mouse-reeve:
This would be a super welcome contribution! I know you're not alone in wanting to run the system without Docker. You're also welcome to join the matrix chat (https://app.element.io/#/room/#bookwyrm:matrix.org), which is handy for quick questions

Amolith:
\o/

Should be able to start working on it sometime tomorrow morning/afternoon. Once I have an instance set up on [nixnet.services](https://nixnet.services), I'll be adding it to the public list as well :ok_hand:

WesleyAC:
Closing, since this was completed in https://github.com/bookwyrm-social/documentation/pull/77

[Docs for Dockerless installation](https://docs.joinbookwyrm.com/install-prod-dockerless.html)

PR #2305, opened 2022-09-23
Add a dockerless install method

CactiChameleon9:
I need to do a little bit more testing and the scripts are very welcome to scrutiny... but it does work :tada: 

Hope this is helpful and laid out correctly

Solves #1156

### DIFF
--- a/dockerless-install.sh
+++ b/dockerless-install.sh
@@ -1,0 +1,76 @@
+#!/bin/bash
+
+# exit on errors
+set -e
+
+echo Installing dependencies 
+apt install postgresql redis nginx python3-venv libpq-dev
+
+echo
+echo Removing old bookwyrm install (not database, just config and files)
+sleep 2
+rm -r /opt/bookwyrm
+
+echo
+echo Creating new bookwyrm install dir in /opt/bookwyrm
+mkdir /opt/bookwyrm
+cd /opt/bookwyrm
+git clone https://github.com/bookwyrm-social/bookwyrm ./
+git checkout production
+cp .env.example .env
+
+echo
+echo Making the python virtual enviroment
+mkdir venv
+python3 -m venv /opt/bookwyrm/venv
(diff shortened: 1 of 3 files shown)


mouse-reeve:
Thank you so much for this! I wonder if these files should be checked in to the repository, or it would make more sense to have them in the documentation, and ask people to copy/paste the configurations and changes into their install. I think the downsides to that are that the files are a little harder to find, it's less obvious how to update them, and it's more work for the person doing the install to create and edit the files they need.

But the reasons I think it might be better are, since this is a much less common route, it will create less confusion for people doing a traditional install, and (perhaps more importantly), the person doing the install will likely need to change some of the details here based on their own needs. For example if someone didn't want to install in `/opt`. Another reason is that the install script has manual instruction steps, which to me seems like a better fit for a documentation page than a script.

So my proposal, and I'm curious if you agree or have qualms, would be to convert these files into a markdown document in the [documentation repo](https://github.com/bookwyrm-social/documentation/tree/main/content/running_bookwyrm), and re-format it a bit so that it's a set of instructions rather than a script with related files. And fwiw, if you think that sounds good but would prefer not to make the actual markdown files, I'd be happy to do that and make you a co-author on the commits.

CactiChameleon9:
I think I agree with you on the documentation aspect - but I wasn't sure where you wanted this kind a stuff :laughing: (especially with the production and main branch)

I tend to use my code as documentation, if that explains a few of the decisions I made there

Reformatting this into markdown shouldn't be a problem, however I don't think I can do that today. If you are feeling particularly inspired, go ahead - but I should get round to doing that soonish

In the meantime, feel free to close this PR if you want to

mouse-reeve:
Sounds good! I'll leave it to you but feel free to let me know if I can support in any way :)

CactiChameleon9:
Done :)

PR #2314, opened 2022-10-09
Closes #2194 Normalize stored ISNI

### LINKED ISSUE #2194: Normalize stored ISNI
The `ISNI` field should always be in the same format (without spaces), but right now sometimes the field is stored with spaces between the digits. There should be validation when the author is saved that cleans this value, similar to how ISBNs are validated

chambersh1129:
The Author.isni_link property has sample code to remove spaces, I just copied that up to the Author.save() function.  I left Author.isni_link as is to preserve existing functionality for isni values stored in the DB where the Author.save() function has not been run.


chambersh1129:
removed trailing whitespace which broke pylint/black linting workflows

Issue #2191, opened 2022-07-07
No check if finish date has not happened yet


joelchrono12:
**Describe the bug**
There is no check to avoid finishing a book in a date that still has not happened

**To Reproduce**
1. On a currently reading book, click on "Finish reading"
2. Set a finish date in the future
3. Click "Post"
4. There is nothing stopping you to do it

**Expected behavior**
The Post button should be disabled until a valid finish date is chosen 

**Screenshots**
As you see, the Post button is pretty easy to click here :P  
![image](https://user-images.githubusercontent.com/18448587/177686213-f904f303-575d-456e-a3d8-789a2876da54.png)

**Instance**
https://bookrastinating.com

---

**Desktop (please complete the following information):**
    - OS: Fedora
    - Browser Firefox
    - Version 102

hughrun:
@mouse-reeve we can close this issue now?

mouse-reeve:
Yes!

PR #2316, opened 2022-10-11
Make it so that finishing a book cannot happen in the future

redshiftss:
This is a commit that hopefully fixes issue  #2191, which mentions that users can stop reading a book in the future.
I just added a couple of checks in the `ReadThroughForm()` form, which I've noticed gets created whenever an user adds a book that they have read to their profile. I've also imported the datetime module in the `forms.py` file.


redshiftss:
I'd also really appreciate it if this PR could be tagged as `hacktoberfest-accepted` as well if possible

mouse-reeve:
I've been going through and adding `# pylint: disable=invalid-name` to those as I run across them, but the version in the checks doesn't seem to mind about it so it should be no issue :)

mouse-reeve:
I went ahead and fixed the whitespace complaint, I hope you don't mind! I just hate to have people stuck in an interminable cycle of petty display complaints. And this works great! Thank you for the contribution ❤️

PR #2358, opened 2022-11-14
Fix npm dev tools

cincodenada:
The npm-based devtools weren't working (for me, at least) because they need to be run out of the `dev-tools` folder. This fixes that, and does some other dev-tools cleanup while I'm at it.


cincodenada:
Actually, this seems to be a duplicate of #2032, so I'll close this in favor of that. There's some dockerfile rearranging in here, but I'll open a separate PR for that if it's still relevant.

mouse-reeve:
Oh I see that you commented on the PR -- thank you <3

PR #2373, opened 2022-11-15
Add contrib directory with systemd service files for front, worker and scheduler

codeurimpulsif:
# Goal

- Create a `contrib` directory
- Create `bookwyrm`, `bookwyrm-worker` and `bookwyrm-scheduler` systemd services

# Information

Naming a directory `contrib` for scripts and configuration files providing a helpful leg-up for integration of a project into other environments is a long established convention.

If this PR is accepted I will also propose a change in the `install-prod-dockerless` documentation to use this systemd service files instead of the current `dockerless-run.sh` script.

### DIFF
--- a/contrib/README.md
+++ b/contrib/README.md
@@ -1,0 +1,5 @@
+# Contrib
+
+This directory contain some scripts, configuration files and other useful tools around BookWyrm.
+
+These tools are not necessary for the proper functioning of BookWyrm but provide a helpful leg-up for integration with some third-party or to nicely fit BookWyrm into other environments.
--- a/contrib/systemd/bookwyrm-scheduler.service
+++ b/contrib/systemd/bookwyrm-scheduler.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=BookWyrm scheduler
-After=network.target postgresql.service
+After=network.target postgresql.service redis.service
 
 [Service]
 User=bookwyrm
--- a/contrib/systemd/bookwyrm-scheduler.service
+++ b/contrib/systemd/bookwyrm-scheduler.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=BookWyrm scheduler
+After=network.target postgresql.service
+
+[Service]
+User=bookwyrm
+Group=bookwyrm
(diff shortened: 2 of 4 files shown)


mouse-reeve:
I think I'm missing something -- where is the `dockerless-run.sh` file?

codeurimpulsif:
The `dockerless-run.sh` file is only mentionned at the end of [this documentation](https://docs.joinbookwyrm.com/install-prod-dockerless.html), I will replace this file with the systemd service files of this PR.

mouse-reeve:
Oooh gotcha, thank you!

mouse-reeve:
@marcusyoung -- I can't figure out how to get github to assign you as a reviewer so I'm just tagging you 😭

chdorner:
+1 on this, it's a much cleaner way to run it with three separate systemd services than how it is proposed [in the docs](https://docs.joinbookwyrm.com/install-prod-dockerless.html) with daemonizing the three processes in one bash script.

fwiw, the service configurations look good. I'd maybe add a `Restart=always` which is similar to the docker-compose configuration, plus maybe add Redis to the `After` (`After=network.target postgresql.service redis.service`).

codeurimpulsif:
@chdorner Thanks! I already add `redis.service` in `After` for the `bookwyrm-worker` but not the others because only the worker seem to need it after some tests, but maybe not?

chdorner:
I think all three have a dependency on Redis. The web one for caching and putting jobs on the queue, the worker for pulling jobs from the queue, and the scheduler for putting jobs I on the queue.

mouse-reeve:
@chdorner that's correct -- I think this is ready to merge once the redis related changes are made? Thank you all for this work!

codeurimpulsif:
@mouse-reeve @chdorner you're right, thanks for the clarification!
I just add it.

marcusyoung:
I know this is merged and in the latest release now but bookwyrm-worker file is incorrect as it needs to have the import queue added.

chdorner:
@marcusyoung #2622 addressed this and was merged a few hours ago

PR #2382, opened 2022-11-15
Longer invite answer field

### LINKED ISSUE #2377: Larger, extendable, or configurable character limit for invite request text response
When requesting an invite, some users complain that 50 characters is not enough to answer the question. In our case the question is "How do your interests overlap with the concept of Solarpunk?"

I would like one of the following, if possible:
- Configurable length for the invite request response in the settings form
- Longer length (100-150 would probably suffice)
- Selection of type of response (short versus long text,  this is how Funkwhale does it)

Cannot think of an alternative other than patching the code.

**Additional context**
Here is Funkwhale's solution to this problem:
image

mouse-reeve:
It was an oversight that it was only set to 50 characters I expect! Fixes #2377


viviicat:
Thanks as always! <3

PR #2385, opened 2022-11-16
Simplify title/author search logic

mouse-reeve:
This code is still gnarly but this should be an improvement

### DIFF
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -101,9 +101,11 @@ def search_title_author(query, min_confidence, *filters, return_first=False):
     # filter out multiple editions of the same work
     list_results = []
     for work_id in set(editions_of_work[:30]):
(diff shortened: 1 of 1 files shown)


bookwyrm/book_search.py
```diff
@@ -96,19 +98,16 @@ def search_title_author(query, min_confidence, *filters, return_first=False):
...
-    for work_id in set(editions_of_work):
-        editions = results.filter(parent_work=work_id)
-        default = editions.order_by("-edition_rank").first()
-        default_rank = default.rank if default else 0
-        # if mutliple books have the top rank, pick the default edition
-        if default_rank == editions.first().rank:
-            result = default
-        else:
-            result = editions.first()
+    for work_id in set(editions_of_work[:30]):
```
tBaxter, line 105: Suggest getting the slice when you get the queryset, so you're not pulling a larger queryset than you'll use.

PR #2390, opened 2022-11-16
Quote ports in docker-compose

jaschaurbach:
Per sepcification in compose-file v3 ports are in quotes.

https://docs.docker.com/compose/compose-file/compose-file-v3/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     image: nginx:latest
     restart: unless-stopped
     ports:
-      - 1333:80
+      - "1333:80"
     depends_on:
       - web
     networks:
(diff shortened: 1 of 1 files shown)


jaschaurbach:
(I came to this because PyCharme threw a warning regarding the quotes)

cincodenada (approved):
This is a good cleanup!

Fun fact: the warning is because of a quirk in YAML's attempts to guess what you mean that results in port pairs that are valid times (more specifically, base-60 numbers, with "digits" delimited by colons) being interpreted as a single number, so "22:22" will be interpreted as the number 22*60 + 22 = 1342 :roll_eyes:

Issue #2389, opened 2022-11-16
2FA signup should present the code as well as the QR image.


moopet:
2FA currently assumes a phone with a camera is going to be the 2FA device, but this isn't always the case. Presenting the TOTP authenticator key would allow a wider range of 2FA devices to be used.

hughrun:
oh lol sorry I assumed this was brand new and didn't see @jaschaurbach's PR.

@mouse-reeve I'm going to revert this re showing the code, Jascha's is better.

PR #2393, opened 2022-11-16
Show 2 fa code

jaschaurbach:
this shows the Secret below the QR Code to onfigure apps which do not read QR codes.

This solves adresses #2389


jaschaurbach:
argh. I am not so good with git...

PR #2394, opened 2022-11-16
show otp_secret when setting up 2fa

jaschaurbach:
solves #2389


mouse-reeve:
I'd like to provide a little more context in the UI around what the code is, so that people who aren't familiar with 2FA know whether they need to pay attention to the string. Since (I think) most users will be using the QR code, it might make sense to have this hidden in a `details` element. When I looked at my authentication app, it also asked me for an account name. 

What do you think of a UI like this:
Screen Shot 2022-11-17 at 11 49 23 AM

Screen Shot 2022-11-17 at 11 49 31 AM

jaschaurbach:
I can live with that. That your App asked for an account name is... well, could not replicate that. I am using Bitwarden as a client for Vaultwarden. I enter the key
Bildschirm­foto 2022-11-17 um 21 24 37
save that and get my OTP
Bildschirm­foto 2022-11-17 um 21 24 44
and this one just works. Which App do you use?

mouse-reeve:
I'm seeing it asking for username in Google Authenticator

jaschaurbach:
This is strange - I can not replicate this. I just loaded Google Authenticator for iOS and added it with the code, with QR code and deliberately entered the it time based and counter based - it just works.

mouse-reeve:
Yes, I normally use the QR code with Authenticator. I wanted to try using the code, which is an option that Authenticator has when I try to add a new site

jaschaurbach:
I know. I used it before I switched to Authy and then to Bitwarden. I believe that that happened but it should not have happened. On the other hand in the QR code is the user and the Secret encoded (`otpauth://totp/3b3e-2600-1700-bab1-3b90-e9f6-4f16-9e96-bdc2.ngrok.io:mouse?secret=ZMJNBY3WFTWAWUI2IJXSMR6WK4N6USI3&issuer=3b3e-2600-1700-bab1-3b90-e9f6-4f16-9e96-bdc2.ngrok.io`) and I have no idea how an authenticator does it without these info. Well, it is confusing,. 

Back to your question: Your mockup is very nice!

PR #2407, opened 2022-11-18
2fa qrcode fixes

### LINKED ISSUE #2386: Dark theme is incompatible with 2FA
**Describe the bug**
When using 2FA, the generated QR code does not have a white backdrop.

**To Reproduce**
1. Enable the Dark theme
2. Set up two-factor authentication until you observe a bar code to scan
3. Attempt to scan with Google Authenticator or similar, and note that it has a lot of trouble with the reduced contrast.

**Expected behavior**
2FA works for dark theme.

**Instance**
Solarpunk.moe

---

    - OS: Gentoo Linux
    - Browser: Firefox
    - Version 106.0.5

### LINKED ISSUE #2389: 2FA signup should present the code as well as the QR image.

hughrun:
- add light background to qr code when using dark theme

fixes #2386

### DIFF
--- a/bookwyrm/static/css/themes/bookwyrm-dark.scss
+++ b/bookwyrm/static/css/themes/bookwyrm-dark.scss
@@ -92,6 +92,10 @@ $family-secondary: $family-sans-serif;
     color: $grey-light !important;
 }
 
+#qrcode svg {
+    background-color: #a6a6a6;
+}
+
 @import "../bookwyrm.scss";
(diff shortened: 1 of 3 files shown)


mouse-reeve:
Thank you! looks good

PR #2412, opened 2022-11-18
check if ISNI is valid

jaschaurbach:
ISNI is always 16 characters
ISNI has a modulo 11 2 checksum
First 4 numbers always 0
if someone forgets the 0s at the beginning, fill it up and see if checksum works.

All non valid ISNI get deleted.

I have no idea how to raise a warning in the frontend :(

### DIFF
--- a/bookwyrm/models/author.py
+++ b/bookwyrm/models/author.py
@@ -7,6 +7,7 @@ from django.db import models
 
 from bookwyrm import activitypub
 from bookwyrm.settings import DOMAIN
+from stdnum.iso7064 import mod_11_2
 
 from .book import BookDataModel
 from . import fields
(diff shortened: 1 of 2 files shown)


jaschaurbach:
The build fails because in the the isni is "hi" which gets to '00000000000000hi and this can not be checked. Will update with a check if the given value is a number or has only an X at the enc (which is a valid sign at the end of an ISNI).

maybe the test should be rewritten to check for NaN, a wrong number and a valid isni. What do you think @mouse-reeve?

jaschaurbach:
OK, I fail to do it better. As far as I can get it the test values get thrown away by my change and the tests fail. 

Maybe I just took a wrong approach on this. 

I will close this one and rethink my approach.

PR #2418, opened 2022-11-19
set HTTP_X_FORWARDED_PROTO in .env

### LINKED ISSUE #2397: Running behind a proxy produces wrong embedding URLs
I was a little bit confused that the URLs for embedding were http instead of https. After looking up where that comes from 

https://github.com/bookwyrm-social/bookwyrm/blob/fdc477afdf4002ea432741121024ea75dbc68192/bookwyrm/views/list/list.py#L53

`build_absolute_uri()` will always generate an absolute URI with the same scheme the current request has - as the traffic comes via http vfrom the proxy it will always build http://URL. 

Iw as confused as my proxy is giving the correct headers in the request. I learned that Django needs to be configured to believe the headers. This is done by adding `SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')` to `settings.py`

You can read about it here: https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header

**As this can be a security issue I would not recommend to add it to `settings.py` per default but I recommend to add an option into .env that this can be set there and stays after an update.** 

I have it at the moment in my `settings.py` and I am prepared to put it back there every time - but I am waiting for the update I forget it and get confused. 

As my setup is forwarding _every_ requests to https it is no problem if someone would take the `http`-ULR by accident. I know Servers which do not forward to `https:` by default and sometimes admins do not care as port 80 its just serving nothing. 

Who knows what is coming in the future. I saw the first servers who do not have a port 80 any more - even no forwarding. I think it is just a question of time until we see more and more SSL only Setups.

Long talking short: Please take it into consideration.

jaschaurbach:
This fixes #2397.

The description in .env is whith a warning and a link to the official documentation about what this setting is doing if set to true.


bookwyrm/settings.py
```diff
@@ -364,3 +364,6 @@
 OTEL_SERVICE_NAME = env("OTEL_SERVICE_NAME", None)
 
 TWO_FACTOR_LOGIN_MAX_SECONDS = 60
+
+if HTTP_X_FORWARDED_PROTO:
```
mouse-reeve, line 368: I think you'll need to add a line creating this variable, something like:
`HTTP_X_FORWARDED_PROTO = env("HTTP_X_FORWARDED_PROTO")`

jaschaurbach: Upside. Of course I need. 

Sorry, I am just getting back to coding after _years_ of not coding :/

mouse-reeve: It's no trouble at all, glad to have your contributions :)

jaschaurbach:
The is the hardest commit in my whole life :D

Issue #2433, opened 2022-11-22
Lists completely broken - 502 error


marcusyoung:
**Describe the bug**

Lists don't appear to work at all except for the admin user. If another user clicks on them nothing happens. Sends a CPU core up to 100% and eventually goes to a 502 after about a minute. This is for public lists that I created (as admin) if that may be relevant. 

In further test I tried to create a list as a normal (editor) user. It shows the dialogue to give the list a name etc, but then after clicking to create the list it does the same thing. Freezes. Puts a CPU core at 100% and 502s.

I've put all my lists to private until it can be investigated. I have no other user public lists currently.

marcusyoung:
Comment from @jaschaurbach on Matrix
>I have this issue with new lists since 0.5.1
I checked - I had no list created after the update except one. And that fails

marcusyoung:
> Comment from @jaschaurbach on Matrix
> 
> > I have this issue with new lists since 0.5.1
> > I checked - I had no list created after the update except one. And that fails

I think I created most of mine (not the tests) just before 0.5.1, but I have changed the permissions on all of the since 0.5.1. So that may have triggered whatever the issue is.

cincodenada:
Okay yeah I was able to reproduce this too, just accessing a list that I created long before the latest release takes a while and then hits a 502.

Have we tracked down a minimum version here yet? Are we sure this is a regression in the 0.5.1 release, or could it be in 0.5.0 or earlier?

I did a quick scan of the changes and nothing obvious stood out, so I'm doing some further testing now.

marcusyoung:
@cincodenada Had you by any chance changed the permission on that list recently?

jaschaurbach:
I found this in the log

```
2022-11-23 17:13:31,805 DEBG 'bookwyrm' stderr output:
Internal Server Error: /list/2/s/alle-queerwelten-ausgaben
Traceback (most recent call last):
  File "/home/queer/bookwyrm/venv/lib64/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/queer/bookwyrm/venv/lib64/python3.9/site-packages/django/core/handlers/base.py", line 204, in _get_response
    response = response.render()
  File "/home/queer/bookwyrm/venv/lib64/python3.9/site-packages/django/template/response.py", line 105, in render
    self.content = self.rendered_content
[...]

PR #2436, opened 2022-11-23
Hotfix: Disable list suggestions

cincodenada:
I did some process of elimination, and this seems to be behind #2433, causing 502s and using up a whole CPU core. Obviously this doesn't fix the underlying issue, but since list suggestions aren't a core functionality and are already conditionally rendered, disabling them seems like a good mitigation until the underlying issue can be fixed.


cincodenada:
Closing in favor of #2437 which I believe fixes the underlying issue

PR #2445, opened 2022-11-25
Handles state in notifications where a list item was deleted

mouse-reeve:
There might be a clearer way to phrase this, but it's better than before ™️
Before:
Screen Shot 2022-11-25 at 8 57 35 AM

After:
Screen Shot 2022-11-25 at 8 56 34 AM

### DIFF
--- a/bookwyrm/templates/notifications/items/add.html
+++ b/bookwyrm/templates/notifications/items/add.html
@@ -61,7 +61,13 @@
 {% else %}
     {% with count=notification.related_list_items.count|add:"-2" %}
     {% with display_count=count|intcomma %}
-    {% if related_list.curation != "curated" %}
+    {% if count < 1 %}
+        {# This happens if the list item was deleted #}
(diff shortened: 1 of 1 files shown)


bookwyrm/templates/notifications/items/add.html
```diff
@@ -61,7 +61,13 @@
 {% else %}
     {% with count=notification.related_list_items.count|add:"-2" %}
     {% with display_count=count|intcomma %}
-    {% if related_list.curation != "curated" %}
+    {% if count < 1 %}
+        {# This happens if the list item was deleted #}
+        {% blocktrans trimmed %}
+        <a href="{{ related_user_link }}">{{ related_user }}</a>
+        added added a book to one of your lists
```
franbarra, line 68: There's an extra "added"

mouse-reeve: whoops! thank you

Issue #2451, opened 2022-11-26
Mentioning user seems to generate invalid JSON-LD


indutny:
**Describe the bug**

When sending a DM or mentioning a user on a custom ActivityPub server (not Maston, and not Bookwyrm) the generate JSON-LD which is being sent to the remote server's inbox contains Mention link that has `"id":null`, which is not valid according to JSON-LD playground:

```json
{
  "id": "https://wyrms.de/user/indutny/status/28884/activity",
  "type": "Create",
  "actor": "https://wyrms.de/user/indutny",
  "object": {
    "id": "https://wyrms.de/user/indutny/status/28884",
    "type": "Note",
    "published": "2022-11-26T03:00:40.507496+00:00",
    "attributedTo": "https://wyrms.de/user/indutny",
    "content": "<p><a href=\"https://mean.engineer/users/test\">@test@mean.engineer</a> hm.</p>",
    "to": [
      "https://mean.engineer/users/test"
    ],
    "cc": [],
    "replies": {
      "id": "https://wyrms.de/user/indutny/status/28884/replies",
      "type": "OrderedCollection",
      "totalItems": 0,
      "first": "https://wyrms.de/user/indutny/status/28884/replies?page=1",
      "last": "https://wyrms.de/user/indutny/status/28884/replies?page=1",
      "@context": "https://www.w3.org/ns/activitystreams"
    },
    "tag": [
      {
        "id": null,
        "type": "Mention",
        "href": "https://mean.engineer/users/test",
        "name": "test@mean.engineer",
        "mediaType": null,
        "attributedTo": null,
        "availability": null
      }
    ],
    "attachment": [],
    "sensitive": false,
    "@context": "https://www.w3.org/ns/activitystreams"
  },
  "to": [
    "https://mean.engineer/users/test"
  ],
  "cc": [],
  "signature": {
    "creator": "https://wyrms.de/user/indutny#main-key",
    "created": "2022-11-26T03:00:40.507496+00:00",
    "signatureValue": "g8cUYgiFpRA7JOpcz0S6ejQaM3nseUYkqv7DjYJF/GzysnFzygJyBrOnR3yPGngIbOKf/eZorDcEPulebUIUZN499uY3G9D8jkxn3KMQa220Y+h4wqQh9qY87BeL2P47NKPvC/9f+Uc02Pm6iOojlxxtyrDcAY8mE/c18lpIlqI=",
    "type": "RsaSignature2017"
  },
[...]

PR #2452, opened 2022-11-26
Fix serialization of JSON-LD subarrays

indutny:
Properties like "tag" could be lists containing multiple subclasses of ActivityObject. Make sure to serialize them recursively instead of outputting them as they are, because otherwise we could get a bunch of nulls in the resulting JSON and that wouldn't necessarily be a valid JSON-LD object.

Fix: #2451

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -194,6 +194,11 @@ class ActivityObject:
             try:
                 if issubclass(type(v), ActivityObject):
                     data[k] = v.serialize()
+                elif isinstance(v, list):
+                    data[k] = [
+                        e.serialize() if issubclass(type(e), ActivityObject) else e
+                        for e in v
+                    ]
             except TypeError:
                 pass
         data = {k: v for (k, v) in data.items() if v is not None and k not in omit}
@@ -306,7 +311,9 @@ class Link(ActivityObject):
 
     def serialize(self, **kwargs):
         """remove fields"""
-        omit = ("id", "type", "@context")
+        omit = ("id", "@context")
+        if self.type == "Link":
+            omit += ("type",)
         return super().serialize(omit=omit)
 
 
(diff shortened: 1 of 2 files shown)


indutny:
(Disclaimer: I don't write python normally so please forgive me if this is a terrible solution of the problem and is not idiomatically correct)

indutny:
My python-foo is pretty bad and I can't seem to get tests to run locally, but I think I fixed it!

bookwyrm/activitypub/base_activity.py
```diff
@@ -194,6 +194,11 @@ def serialize(self, **kwargs):
             try:
                 if issubclass(type(v), ActivityObject):
                     data[k] = v.serialize()
+                elif isinstance(v, list):
+                    data[k] = [
```
indutny, line 198: Directly nested lists are not allowed in JSON-LD, so I think we should be alright!

indutny:
Yay, looks like I figured the issue with failing tests without running them :joy:

mouse-reeve:
I'm seeing one issue with this -- the tag in a status with a mention goes from:
```json
  "tag": [
    {
      "id": null,
      "type": "Mention",
      "href": "https://friend.camp/users/tripofmice",
      "name": "tripofmice@friend.camp",
      "mediaType": null,
      "attributedTo": null,
      "availability": null
    },
    ...
```
to
```json
  "tag": [
    {
      "href": "https://friend.camp/users/tripofmice",
      "name": "tripofmice@friend.camp"
    },
    ...
```

The `type` field is missing. I'm not sure why!

indutny:
@mouse-reeve oops! Again, I didn't get any better at Python, but I think this fix should do it. Let me know if I got it wrong!

mouse-reeve:
Thank you for this! Seems to work just right :)

indutny:
Hooray! Thank you!

PR #2457, opened 2022-11-26
make use of templates instead of hidden divs which is more elegant

viviicat:
The rationale for this is that I kept seeing the barcode scanning status when refreshing the page (at least on Firefox, and because the CSS hadn't loaded yet) and it was irritating me that these are visible.

`<template>` is the appropriate way to do this and should be supported in any of the browsers that support the barcode scanner.


mouse-reeve:
Thank you for this!

PR #2460, opened 2022-11-26
Visual tweaks

viviicat:
Please let me know if you wish for any of these to be reverted, as some are certainly subjective.

* Fix the button that is inline with text, so it aligns properly.
image

* With as little visual difference as possible, change the `<details>` blocks so they do not have padding, instead their children have padding. This makes it much more clickable, at least on Firefox, because you can click anywhere in the box instead of just where the text is.
   * As a result various paddings had to be tweaked, simplifying the classes in many cases.

* Adjusted some of the styling of some of the `<details>` expanders by adding `.block` for consistency.

* The action button for changing an item's shelf has been moved to the right for better usage of screen space.
image

* Fix two-factor auth page which was too narrow for the code, so the code ended up being misformatted.
Screenshot from 2022-11-26 15-36-08

* Link domain page, use a smaller button to set the display name.
image

* When book has no description, the "Add Description" button and its form now have better spacing.
image
image

* Some other really subtle padding issues for buttons and forms.

### DIFF
--- a/bookwyrm/static/css/bookwyrm/_all.scss
+++ b/bookwyrm/static/css/bookwyrm/_all.scss
@@ -140,7 +140,7 @@ button:focus-visible .button-invisible-overlay {
     opacity: 1;
 }
 
-button .button-paragraph {
+button.button-paragraph {
     vertical-align: middle;
 }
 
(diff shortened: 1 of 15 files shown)


mouse-reeve:
I'm with you on all these!

PR #2466, opened 2022-12-03
StoryGraph import ISBN tweak

rickosborne:
Looks like SG updated their CSV headers just a little, so that the importer hadn't been picking up the ISBN.


mouse-reeve:
Good catch, thank you!

PR #2476, opened 2022-12-05
New ID: Audible ASIN

jaschaurbach:
Audible belongs to Amazon BUT they do not share the same IDs. The Audible ASIN of an audiobook is never the same as the Amazon ASIN of the same Audio Books. 

Yeah, I know, Amazon is great. The fact that the ASIN is a good distinction for different works and editions bothers me more than I will ever be willing to admint.

Prepares also to the author tables for the Audible ASIN, as the author pages have own ASINs as well.


jaschaurbach:
For the next time I create two Merges: One with the updated test (if the filed should be in the export) and one that will be merged after the test has been merged.

jaschaurbach:
Made obsolete by #2482

Issue #2487, opened 2022-12-09
Make goodreads ID for books available to edit 


jaschaurbach:
It is just missing in the frontend.

this request is a reminder for me.

jzacsh:
What is this new ID in the UI? Is it the first digits in the last page-path of the work? Eg `3899622` for the edition URL `https://www.goodreads.com/book/show/3899622-hands-of-my-father` perhaps?

jaschaurbach:
Yes, because `https://www.goodreads.com/book/show/3899622` will be forwarded to the URL from your comment. They add the name for SEO reasons.

PR #2488, opened 2022-12-10
goodreads_key for books editable

jaschaurbach:
The field goodreads_key is now editable.
ID is shown in ID section of book:

Bildschirm­foto 2022-12-10 um 10 49 07

This solves #2487


mouse-reeve:
Works beautifully, thank you!

PR #2491, opened 2022-12-10
Detect preferred timezone via JavaScript on register

### LINKED ISSUE #2024: Pre-set user's timezone
I signed up and saw you can set a timezone. It was set to UTC by default.

JavaScript has an Internationalization API that returns timezones. At attempt should be made to pre-set this for less friction.

```
const x = new Intl.DateTimeFormat()
x.resolvedOptions().timeZone
// => "Asia/Kolkata"
```

Reading:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions

chdorner:
Fixes #2024

One solution for detecting and setting the user's `preferred_timezone`. Given browser don't send that information via headers, next best thing seems to be to use JavaScript's `Intl.DateTimeFormat()` API.
In case the browser sends rubbish data, the server side defaults to UTC.

### DIFF
--- a/bookwyrm/static/js/bookwyrm.js
+++ b/bookwyrm/static/js/bookwyrm.js
@@ -48,6 +48,12 @@ let BookWyrm = new (class {
         document
             .querySelector("#barcode-scanner-modal")
             .addEventListener("open", this.openBarcodeScanner.bind(this));
+
+        document
+            .querySelectorAll('form[name="register"]')
+            .forEach((form) =>
+                form.addEventListener("submit", (e) => this.setPreferredTimezone(e, form))
+            );
     }
 
     /**
@@ -784,5 +790,17 @@ let BookWyrm = new (class {
         event.target.addEventListener("close", cleanup, { once: true });
 
         initBarcodes();
+    }
+
+    /**
+     * Set preferred timezone in register form.
+     *
+     * @param  {Event} event - `submit` event fired by the register form.
+     * @return {undefined}
+     */
+    setPreferredTimezone(event, form) {
+        const tz = Intl.DateTimeFormat().resolvedOptions().timeZone;
+
+        form.querySelector('input[name="preferred_timezone"]').value = tz;
     }
 })();
--- a/bookwyrm/templates/snippets/register_form.html
+++ b/bookwyrm/templates/snippets/register_form.html
@@ -58,6 +58,8 @@
     </div>
 </div>
 
+<input type="hidden" name="preferred_timezone" />
+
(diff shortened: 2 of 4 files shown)


bookwyrm/static/js/bookwyrm.js
```diff
@@ -48,6 +48,12 @@ let BookWyrm = new (class {
         document
             .querySelector("#barcode-scanner-modal")
             .addEventListener("open", this.openBarcodeScanner.bind(this));
+
+        document
+            .querySelectorAll('form[name="register"]')
+            .forEach((form) =>
+                form.addEventListener("submit", (e) => this.setPreferredTimezone(e, form))
+            );
```
chdorner, line 56: It wasn't immediately obvious to me how the JavaScript side of things is organised, so just tried to immitate what is already there. Happy to move this around wherever it would fit better.

mouse-reeve: It isn't immediately obvious to me either and I'm the one who organized it. This seems like a perfectly good place for it :)

mouse-reeve:
This rules! Thank you

bookwyrm/forms/landing.py
```diff
@@ -42,7 +42,10 @@ class Meta:
         model = models.User
         fields = ["localname", "email", "password"]
         help_texts = {f: None for f in fields}
-        widgets = {"password": forms.PasswordInput()}
+        widgets = {
+            "password": forms.PasswordInput(),
+            "preferred_timezone": forms.TextInput(),
```
mouse-reeve, line 47: I don't think this is necessary since it's being used as a hidden field?

mouse-reeve (changes requested):
Extremely minor quibble (although there may be something this is doing that I missed)

bookwyrm/forms/landing.py
chdorner: I'll try tomorrow 👍 iirc Django complained without this being there

mouse-reeve: thank you!

chdorner: I found some time tonight, and you were right, this isn't needed at all 🕺
I'm new to Django, so this might happen often until I get used to it after a while

mouse-reeve (approved):
:)

PR #2492, opened 2022-12-11
Manual email confirm

jaschaurbach:
This one ads `bw-dev manual_confirm $USER` and `manage.py manual-confirm $USER` to manually confirm E-Mail of a user activate the user. 

Autocompletion for shell added for this as well as for remove_2fa (we forgot that one)


mouse-reeve:
This is good to have! I think it should also be doable from the UI directly

jaschaurbach:
> This is good to have! I think it should also be doable from the UI directly

Working on that - just finding out where to put it best and how it should look like. But I need it desperately as my mailserver had some strange behavior and mails never delivered. As I am running the main-branch on this server it would be great if this can be merged - I will then add the UI-part later in another pull request :)

mouse-reeve:
Thank you! I meant that more as a general note rather than "jascha please do this" but I really appreciate that. I also wrote an admin page that lets you send a test email, I'll open a PR for it shortly.

bookwyrm/management/commands/confirm_email.py
```diff
@@ -0,0 +1,20 @@
...
+
+    help = "Manually confirm email for user"
+
+    def add_arguments(self, parser):
+        parser.add_argument("username")
+
+    def handle(self, *args, **options):
+        name = options["username"]
+        user = models.User.objects.get(localname=name)
+        user.is_active = True
```
mouse-reeve, line 18: There are actually two fields that need to be updates - `is_active`, which you have, but `deactivation_reason` should also be nulled. For a separate feature, I wrote a user method called `reactivate`, which does both those things. I think you can use that here (just call `user.reactivate()`) instead of setting the field and saving. I'm going to update the confirm email view to do that as well.

mouse-reeve (changes requested):
Besides my comment below this looks perfect

jaschaurbach:
merged 2498 to use user.reactivat() - did not work? 

I nulled deactivation_reason now.

mouse-reeve:
The change in that PR wasn't directly related to this, I just realized I should be using `reactivate()` there as well -- I think I just made the comment more confusing with that!

What I was trying to say is that rather than manually setting `is_active` and `deactivation_reason` in the management command, you can just call `user.reactivate()` and it will handle updating and saving the user.

mouse-reeve (approved):
Perfect! Thank you

PR #2489, opened 2022-12-10
Add new user tab listing only reviews and comments

chdorner:
This is my first PR, please let me know if I've missed anything obvious!

Fixes #2306 by adding a new tab to the user profile which lists the user's activities filtered by Reviews and Comments.

PR #2496, opened 2022-12-11
Trying to fix an unreliable unit test regex

mouse-reeve:


chdorner (commented):
looks sensible 🎉 

I can rebase my #2489 branch once this made it in and undo my test debugging commit 👍

PR #2497, opened 2022-12-11
Add eslint subcommand to bw-dev

chdorner:
I noticed when working on #2491 that `./bw-dev` doesn't have an easy way to run `eslint` even though there's a CI job that runs it.

This adds an `eslint` subcommand for running this locally.


mouse-reeve:
Thank you!

PR #2499, opened 2022-12-11
Updates nginx production config

mouse-reeve:
Co-authored-by: Trammell Hudson <hudson@trmm.net>

This copies over the changes Trammell added to the development file. I also realized that I think it's fine to only commend out the https redirect, rather than commenting out the entire server block for listening on port 443? If this works it makes the file a lot easier to read.


mouse-reeve:
I've been using this config on b.s and it seems to be helping 😵

PR #2510, opened 2022-12-15
delete avatar when user is deleted

jaschaurbach:
To comply with GDPR, as always.


mouse-reeve:

PR #2511, opened 2022-12-15
Added Import Limit

### LINKED ISSUE #2429: Admin determined limit on max import size
I'd like to be able determine the maximum import size a user can make. I had a user that imported 4000 books. I want to control that as you can all of a sudden have a huge increase in disk space usage, especially once the edition collation has taken place also.

Giebisch:
Fixes #2429 

It's not ready yet, as I'm still running into 2 bugs using this feature. Also tests are not available yet, but I thought it's still useful to create a draft PR so you can already give feedback on the implementation.

**Todo**
- [x] Bugfix
- [x] Add tests

The admin can set a limit and the user can see how many book imports he has left. The import button is disabled when there are no more imports remaining.

**Admin view**
import_limit

**User view**
allowed_imports


mouse-reeve:
Sorry for making your tests even more busted with that commit! I believe all you need to do is create a merge migration to be back on track (`./bw-dev makemigrations --merge`).

Giebisch:
> Sorry for making your tests even more busted with that commit! I believe all you need to do is create a merge migration to be back on track (`./bw-dev makemigrations --merge`).

Just to be clear, the commit is there so we will have consistent migrations, right ? ^^

mouse-reeve:
Creating a merge migration just tells Django how to step through the list of migrations in the right order. Does that answer your question? I'm not sure I know exactly what you're asking

Giebisch:
@mouse-reeve Ah perhaps I have not expressed myself precisely enough. What was the reason to sync the `main` branch into this one at the current state ? The way I know it is that you don't do it until you've finished developing. That's why I'm asking, maybe this has something to do with the migration files ?

mouse-reeve:
Ah I see - I merged in `main` because there was a merge conflict related to a different PR which I merged, and I wanted to resolve it so that you wouldn't have to.

Giebisch:
I believe the 2 failed tests have nothing to do with my changes, but I'm not 100% sure

Giebisch:
@mouse-reeve Okay looks like something did indeed go wrong with the migrations. Can you check again if everything was correct with your commit ? Might need your help here

mouse-reeve:
It looks like it still wants a merge migration, even though I see you made one above. Is it possible that you ran the merge before pulling the latest `main` into your branch? I think if you
 - Migrate back to the migration before your merge (`./bw-dev migrate bookwyrm 0170`)
 - Delete the 0171 merge migration
 - Merge `main` into you branch
 - Re-run the merge migration
That should fix it?

Giebisch:
@mouse-reeve Thanks, I think that solved the problem.

bookwyrm/templates/import/import.html
```diff
@@ -15,6 +15,12 @@ <h1 class="title">{% trans "Import Books" %}</h1>
     {% endif %}
 
     {% if site.imports_enabled %}
+        {% if import_size_limit and import_limit_reset %}
+            <div class="notification">
+                <p>Currently you are allowed to import {{ import_size_limit }} books every {{ import_limit_reset }} days.</p>
```
mouse-reeve, line 20: These blocks of text will need to be wrapped in `{% blocktrans %]` tags so they can be localized. For example:
```
<p>{% blocktrans %}Currently you are allowed to import {{ import_size_limit }} books every {{ import_limit_reset }} days.{% endblocktrans %}</p>
```

bookwyrm/templates/settings/imports/imports.html
```diff
@@ -57,8 +57,39 @@
...
+            id="imports-set-limit"
+            method="POST"
+            action="{% url 'settings-imports-set-limit' %}"
+        >
+            <div class="notification">
+                {% trans "Some users might try to import a large number of books, which you want to limit." %}
+                {% trans "Set the value to 0 to not enforce any limit." %}
+            </div>
+            <div class="align.to-t">
+                <label for="limit">Set import limit to</label>
```
mouse-reeve, line 78: This needs to be in a `{% trans "..." %}` tag

mouse-reeve (changes requested):
I haven't gotten a chance to properly test this, but I'm excited to have it as a feature! For now, just a couple places that need translation support ❤️

Giebisch:
Not sure why the tests are failing now, works for me locally and obviously I haven't touched anything besides the two templates. Can you have a look @mouse-reeve ? Much appreciated! :)

mouse-reeve:
The merge migrations are at it again! I guess it's because there are new migrations on main, and the tests must be merging before they run? If you merge main into your branch and create another merge migration, that should fix it.

mouse-reeve (changes requested):
This worked great for regular imports, but it misses retry imports, which allowed me to overshoot my limit by 200 books

Giebisch:
Great catch! Fixed it :)

mouse-reeve (approved):
Looks good!

PR #2512, opened 2022-12-16
Fix year of reading: phrasing when there are books without page data

### LINKED ISSUE #2508: Year of reading: phrasing when there are books without page data
The 'year of reading page' looks amazing!

I have a minor quibble which I intend to fix myself though: when there are books without page data, it shouldn't say "X books have no pages" (which makes it sound like I read empty books). Instead it should say something like "no page data was available for X books". 

image

Open to other phrasing suggestions too!

hueyy:
Resolves #2508


mouse-reeve:
Thank you so much for identifying and fixing this! It looks perfect :)

PR #2524, opened 2022-12-17
Initial hashtag support

### LINKED ISSUE #1081: Hashtag support would be nice
I take part in a book club on Mastodon that works by using a hashtag (#SFFBookClub) so we can find each other.  Since most of the traffic on that hashtag is discussions of the book we're reading each month, Bookwyrm seems like a natural home for those posts.  But I need to be able to both use hashtags in reviews and follow/search a hashtag in Bookwyrm for that to work.

1. Hashtag support in my posts here - i.e. not stripping out the # and allowing it to be indexed
2. A view in which I can see posts with a given hashtag, like https://weirder.earth/tags/SFFBookClub on Mastodon.

For the time being I think my workaround is to post reviews here but link to them from my Mastodon account.  Not terrible, but a little clunky.

chdorner:
Initial hashtag support which:
* parses out hashtags when creating a new status
* adds the hashtags to the activitypub representation of a status
* imports hashtags from activitypub statuses (rewrites the link to be point to the local instance's hashtag page)
* adds a new view for viewing statuses belonging to a given hashtag
* correctly federates status with hashtags to Mastodon (tested with 4.1.0) and takahe (tested with 0.7.0), hashtag links are rewritten to link to the instances' hashtag page

Fixes #1081 

Screenshot of the new page pretty much exclusively reusing existing snippets:
Screenshot 2022-12-18 at 19 51 32

## Specifically out-of-scope for this PR

**Search hashtags**
Probably the next PR once this finally gets merged.

**Hashtags from statuses originating from other, non-BookWyrm ActivityPub services**
When posting a status from another ActivityPub service, we only import and keep the ones that mention local BookWyrm users, in these cases we don't rewrite the hashtag links in the status, so they will end up linking to the originating instance's hashtag page. However, we do currently import the hashtag into the database.
Not sure we need this, so this might always be out-of-scope for BookWyrm's hashtag implementation.


bookwyrm/migrations/0171_hashtag_support.py
```diff
@@ -0,0 +1,53 @@
...
+                        auto_created=True,
+                        primary_key=True,
+                        serialize=False,
+                        verbose_name="ID",
+                    ),
+                ),
+                ("created_date", models.DateTimeField(auto_now_add=True)),
+                ("updated_date", models.DateTimeField(auto_now=True)),
+                (
+                    "remote_id",
```
chdorner, line 30: note that the remote id currently is just using the default, which renders this as https://your.domain.here/hashtag/1

This should probably be slugify the hashtag and use something like `/tags/MyHashtag`. I did some quick tests with Django's `slugify` and it seems that it isn't able to slugify at least Japanese characters (just returns an empty string). Maybe something like [unidecode](https://pypi.org/project/Unidecode/) would be better for this?

bookwyrm/tests/activitypub/test_base_activity.py
```diff
@@ -177,12 +177,21 @@ def test_to_model_many_to_many(self, *_):
...
+                },
             ],
         )
         update_data.to_model(model=models.Status, instance=status)
         self.assertEqual(status.mention_users.first(), self.user)
         self.assertEqual(status.mention_books.first(), book)
 
+        hashtag = models.Hashtag.objects.filter(name="#BookClub").first()
+        self.assertIsNotNone(hashtag)
+        self.assertEqual(status.mention_hashtags.first(), hashtag)
```
chdorner, line 193: I don't have a nice local setup with two separate instances to manually test federation of statuses and the new hashtags, so I built the federation imports just based on these tests. Did I maybe miss something else?

bookwyrm/views/status.py
```diff
@@ -112,6 +112,18 @@ def post(self, request, status_type, existing_status_id=None):
...
 
+        # inspect the text for hashtags
+        for (mention_text, mention_hashtag) in find_hashtags(content).items():
+            # add them to status mentions fk
+            status.mention_hashtags.add(mention_hashtag)
+
+            # turn the mention into a link
+            content = re.sub(
+                rf"{mention_text}\b(?!@)",
+                rf'<a href="{mention_hashtag.remote_id}">{mention_text}</a>',
```
chdorner, line 123: We said in #1081 that an initial hashtag support doesn't need a page for listing the statuses mentioning a given hashtag yet. I find it weird not linking a hashtag though, as it gives no visual indication that hashtags are now a thing. Maybe this PR should include a fairly simple page listing maybe just all public statuses for a hashtag?

bookwyrm/utils/regex.py
```diff
@@ -7,5 +7,6 @@
 STRICT_USERNAME = rf"(\B{STRICT_LOCALNAME}(@{DOMAIN})?\b)"
 FULL_USERNAME = rf"{LOCALNAME}@{DOMAIN}\b"
 SLUG = r"/s/(?P<slug>[-_a-z0-9]*)"
+HASHTAG = r"(#[^!@#$%^&*(),.?\":{}|<>\s]+)"
```
chdorner, line 10: Happy for better suggestions for this regex. I was looking at [Mastodon](https://github.com/mastodon/mastodon/blob/2889c686108e89a87317505f93b841f5a8a6649b/app/models/tag.rb#L29-L32)'s and [Twitter](https://github.com/twitter/twitter-text/blob/30e2430d90cff3b46393ea54caf511441983c260/rb/lib/twitter-text/regex.rb#L126-L165)'s and thought they might be slightly too complex for a first iteration.

chdorner:
@mouse-reeve I have a few open questions for this first iteration of the hashtag support. would be great if you (or others) could chime in. thank you!

bookwyrm/models/hashtag.py
```diff
@@ -0,0 +1,19 @@
...
+from bookwyrm import activitypub
+from .activitypub_mixin import ActivitypubMixin
+from .base_model import BookWyrmModel
+from .fields import CICharField
+
+
+class Hashtag(ActivitypubMixin, BookWyrmModel):
+    "a hashtag which can be used in statuses"
+
+    name = CICharField(
```
chdorner, line 11: with the current regex, this will include the hash sign (i.e. `#rad`), I'm still a bit on the fence about this, but did it because eventually we might want to support other versions of the hash sign, i.e. with the Japanese keyboard a hash results in the [fullwidth number sign](https://unicode-table.com/en/FF03/). So being able to find the hashtag again inside a status might be much harder, plus this might be important for tools that might want to parse the content for the hashtag to replace the URL with their own?

Happy to change this around though and only store the value of the hashtag, without the signifier.

bookwyrm/migrations/0171_hashtag_support.py
mouse-reeve: Another option here would be to use decorative urls, like we do with author names and book titles - the domain looks like `https://your.domain.here/hashtag/1/MyHashtag`, but the slugified part is decorative, so if it doesn't render for whatever reason, the url still works. Relevant code:
https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/urls.py#L679
https://github.com/bookwyrm-social/bookwyrm/blob/d86e6a6949f720ce1ba2eb0ab440f289e2c19cdb/bookwyrm/views/author.py#L29

bookwyrm/views/status.py
mouse-reeve: I think that's a good idea! You're right that not linking a hashtag would be a weird choice, both for the visual queue and because once the hashtag page was created, older posts wouldn't link to it.

chdorner: Just playing around with adding a new view for this as we speak, I'll see shortly how hard it'll be to add one with respecting privacy settings of activities.

bookwyrm/migrations/0171_hashtag_support.py
chdorner: ah that's a good solution, thank you!

bookwyrm/views/status.py
chdorner: Just added an initial version of such a page 👍

bookwyrm/migrations/0171_hashtag_support.py
```diff
@@ -0,0 +1,53 @@
+# Generated by Django 3.2.16 on 2022-12-17 19:28
```
chdorner, line 1: just saw that another migration got merged, will quickly rename and re-anchor this one

chdorner: should be done 👍

mouse-reeve:
This is super cool! I did some super limited testing and it looks great so far. I'll try to do more thorough review before too long, but since December is so hectic for me, I don't think I'm going to merge anything big like this until January

chdorner:
That sounds sensible 👍 let me know when you're coming back to this so that I can bring this back up to shape before reviewing and testing so that it's let's work for you

joshsimmons:
Just wanted to jump on the thread and say THANK YOU for your work on this! I had been wondering about hashtag support in BookWyrm. If/when it's useful, I have several locally hosted fedi clients (BookWyrm, Mastodon, and Pixelfed) and am happy to do some testing.

bookwyrm/views/hashtag.py
```diff
@@ -0,0 +1,54 @@
...
+        activities = (
+            models.Status.privacy_filter(
+                request.user,
+            )
+            .filter(
+                Q(mention_hashtags=hashtag),
+            )
+            .exclude(
+                privacy__in=["direct", "unlisted"],
+            )
```
mouse-reeve, line 34: The one gotcha with this that makes this view complicated is that it will end up showing public statuses by users who have blocked one another. It ends up getting really complicated! That's why the list of lists and the timelines use a redis cache system, which could be extended without too much trouble (at least for me because presumably I still understand that code) to this view. But that would probably be out of scope for this PR..

chdorner: Ah good point. I'll give it a try and see how much work it would be to handle blocked users correctly. Might take me a bit though as I'll be too busy next week

chdorner:
I'll put this back into draft mode, wanted to also manually test the federation. But I'll need a bit of time to get that up and running

bookwyrm/views/hashtag.py
mouse-reeve: There's some existing code for that which would at least be useful reference; I can dig it up tomorrow. Feel free to remind me if I don't!

chdorner: That would be great, but no hurry. I won't be able to spend any time on this all of next week.

mouse-reeve: Here we go: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/base_model.py#L155

bookwyrm/views/status.py
```diff
@@ -115,6 +115,18 @@ def post(self, request, status_type, existing_status_id=None):
         if status.reply_parent:
             status.mention_users.add(status.reply_parent.user)
 
+        # inspect the text for hashtags
+        for (mention_text, mention_hashtag) in find_hashtags(content).items():
+            # add them to status mentions fk
+            status.mention_hashtags.add(mention_hashtag)
+
+            # turn the mention into a link
+            content = re.sub(
```
chdorner, line 124: This is what I wanted to test with my local federation setup.

By parsing and linking to the hashtag on status creation, this means that when I create a status with a hashtag on instance A, when this tag is federated to instance B the hashtag will link back to instance A.

Not sure if I like this, I think I would much rather have hashtags be linked to a local hashtag page. Which would probably require some rethinking of how and when this is happening. Curious what you think about this @mouse-reeve?

mouse-reeve: I also think it would be preferable to stay on your instance when you click a hashtag. The base model has a [property that replaces the domain in a path with the local path](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/base_model.py#L54), which I think you could use by just replacing `.remote_id` with `.local_path`

chdorner: Oh, interesting! thank you for the pointer.
In that case, the best way forward seems to be to rework the hashtag URLs and calculate a unique slug to ensure that the path of the `.remote_id` stays identical across different instances since it currently includes `Hashtag.id`.

I'll get this in a mergeable shape eventually! 😄

mouse-reeve: oh yes, there's a way to do that too, by overwriting [`get_remote_id`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/base_model.py#L54) on the model. [`Shelf`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/shelf.py#L65) is a model that uses this

chdorner: I have been thinking about this a bit more. Sorry that I don't come back to this PR that often, it's quite a big one already and requires loading a lot of context into my brain every time I start.

Please let me know if this sounds way off or if you see a better way forward.

The hashtag federation will work quite differently from the other first-party objects because there's no reason to federate hashtags on their own, but only as part of a federated status. Especially in comparison with, say, Shelf or similar models.
Wouldn't this mean that for each status that comes into a user's inbox we would have to extend the importing process with:
1. parsing the status content for hashtags, or hashtag links
    * I'm not sure if I already added hashtags to the status' ActivityPub representation as tags ([similar to how Mastodon does it](https://docs.joinmastodon.org/spec/activitypub/#status))
3. find existing or create new local hashtags
4. replace the URLs in the HTML-rendered status content with the new local hashtag URL

The only other object that I've seen traces of in the code but never managed to use myself is book mentions. The federation logic for those should be identical to hashtags. But I've never managed to use them, that being said, I haven't tried _that_ hard.

And to add to all of this complexity, this only affects the federation between two BookWyrm instances, seeing how hashtags might work when federating from BookWyrm to other ActivityPub services like Mastodon, or a lesser degree even a Mastodon status to BookWyrm.

There are quite a few things that still need thinking about, not sure if it would be easier in a synchronous conversation or pairing session, if you're up for it and if we can make it work timezone-wise.

bookwyrm/tests/activitypub/test_base_activity.py
chdorner: I have a pretty nice local setup to test federation now 🙌 if anybody else is interested, I wrote about how it works [here](https://amble.blog/2023/01/29/a-local-developent-setup-for-bookwyrm-federation/).

bookwyrm/views/status.py
chdorner: alright after quite a few hours stepping through the ActivityPub code in a debugger I _think_ I have a fairly good solution to this now, it's in 84ffe3ba2247b711bc28c138bdad1a4b1f3c1053. It still needs proper test coverage though, which I'm hoping to get to in the next few days.

chdorner:
Just tested this with [takahe](https://github.com/jointakahe/takahe) to see how hashtags federate with other ActivityPub services.

If anyone else wants to test this I added a takahe setup to my local BookWyrm federation setup scripts: https://github.com/chdorner/bookwyrm-dev

It looks pretty nice:
image
takahe is replacing the hashtag link to its own site, i.e. https://takahe-domain.tld/tags/bookwyrm/

bookwyrm/activitypub/note.py
```diff
@@ -38,6 +41,35 @@ class Note(ActivityObject):
...
+    def to_model(
+        self, model=None, instance=None, allow_create=True, save=True, overwrite=True
+    ):
+        instance = super().to_model(model, instance, allow_create, save, overwrite)
+
+        # Replace links to hashtags in content with local URLs
+        changed_content = False
+        for hashtag in instance.mention_hashtags.all():
+            updated_content = re.sub(
+                rf'(<a href=")[^"]*/hashtag/[^"]*(">{hashtag.name}</a>)',
```
chdorner, line 54: This regex is really tailored to BookWyrm-style hashtags.

The status content from BookWyrm with a status looks like: `<p>Hello World <a href="https: //dev-lit.secretbearlibrary.org/hashtag/1">#BookClub</a></p>`

An example from Mastodon looks like: `<p>Hi <span class="h-card"><a href="https://dev-sci.secretbearlibrary.org/user/sciadmin" class="u-url mention">@<span>sciadmin</span></a></span> <a href="https: //secretbearsociety.org/tags/BookWyrm" class="mention hashtag" rel="tag">#<span>BookWyrm</span></a> is rad</p>`

Makes me think that maybe we should add something like `class="mention hashtag"` and `rel="tag"` as well. Writing a regex that can parse this out without relying on the order seems like a massive pain, but maybe that's overkill.

chdorner: Opted to go with a `data-mention="hashtag"` attribute on the link. Didn't feel like it justified opening up the `class` attribute in our HTML sanitizer, as it could be used to apply CSS styles to user-generated content.

bookwyrm/views/hashtag.py
chdorner: I think that's already applied, the view uses `models.Status.privacy_filter` which in turns calls `super().privacy_filter` ([here](https://github.com/chdorner/bookwyrm/blob/63f0af9c603cae7b30b6e35a608de2d8aa11e3fa/bookwyrm/models/status.py#L230)) from `BookWyrmModel` which is the one you linked, [this](https://github.com/chdorner/bookwyrm/blob/259e8ef8f6fd9bcd5580ba750d2c0755bba47964/bookwyrm/models/base_model.py#L142) one.

Did you see statuses from users who blocked each other show up in your local testing? I'll make sure to give it a test run locally before marking this PR as ready for review / merging.

chdorner: Ok, tested this out locally.

With the following users and blocking relationships:
* `admin` does not block anybody, is not blocked by anybody
* `auser` does not block anybody, is blocked by `buser`
* `buser` blocks `auser`, is not blocked by anybody

all three users have one status posted using the same hashtag. The visibility of the statuses on the hashtag page looks like this:
* `admin` sees all three statuses
* `auser` only sees their own status and the one from `admin`
* `buser` only sees their own status and the one from `admin`

Sounds like it's working correctly, is that right?

chdorner:
This is finally ready to be reviewed @mouse-reeve 🎉 I've updated the PR description at the top with additional things I've added recently.

Some extensive testing would be good. I've tested the following myself:
* Hashtag handling on the local instance
* Federation support between two BookWyrm instances
    * most importantly, a federated status' hashtags should always link to the local instance's hashtag page
* Federation support from BookWyrm to Mastodon and takahe
    * Hashtag links get rewritten to point to the Mastodon/takahe hashtag page
* Federation support from Mastodon / takahe to BookWyrm (we only import these if a local user is mentioned)
    * see out-of-scope note in the PR description, happy to re-evaluate this, it's working a bit weird now
    * Hashtag links don't get rewritten, so they will link to the Mastodon/takahe/etc. instance where the status originated from
    * Hashtags still get imported as local hashtags with a local `remote_id` (!!), but going to the BookWyrm hashtag page those statuses from Mastodon/takahe/etc. will show up as long as the privacy setting applies

bookwyrm/models/hashtag.py
mouse-reeve: What is the advantage to including the symbol? Your points against it seem compelling to me

bookwyrm/views/hashtag.py
mouse-reeve: That sounds correct!

bookwyrm/views/status.py
mouse-reeve: I don't think book mentions are used anywhere at this point? They used to be in the past. (I'm not sure if this if this is true but that's my suspicion).

I think your solution looks good! And while I didn't see your original comment in a timely manner (oops), I am super down to talk about things synchronously in general :)

mouse-reeve (commented):
My only outstanding question is about storing the # symbol in the database. This works SO well, really exciting

mouse-reeve:
Should hashtags be case insensitive? I believe they are on mastodon

bookwyrm/models/hashtag.py
chdorner: After looking at the federation piece a bit more I think it's better to keep the hash symbol as part of the tag. This way there's a canonical string that can be used to find the correct parts of the raw content. Also, other ActivityPub services include the symbol in the JSON representation where hashtags are split out as tags, i.e. here's an example from Mastodon ([this one](https://secretbearsociety.org/@chdorner/109857363838484273.json)):
image

I have the feeling that it's better to keep the hashtag from the original message rather than adding the hard-coded symbol in the JSON representation. If for whatever reason we would want to support different symbols (i.e. the Japanese keyboard using a [different sign](https://unicode-table.com/en/FF03/)) in the future, we would still be able to find the location of the tag inside the content field.

mouse-reeve: Sounds good!

chdorner:
> Should hashtags be case insensitive? I believe they are on mastodon

Ideally, hashtags should be case-sensitive for display purposes, to help screen readers make more sense when reading them out loud. But a capitalized hashtag should match an existing lowercase hashtag and vice versa, i.e. `#BookWyrm` and `#bookwyrm` should be the same hashtag.

There's still something not working when the same hashtag is posted once all lowercase and once capitalised.
I posted `#LivableCities` first, on the feed, in the ActivityPub JSON, and on the hashtag page, and everything looks great. Then I posted a second status with `#livablecities`, which uses the lowercase variant on the feed, but it seems that it actually created a new hashtag and now the status has two hashtag mentions, the `#LivableCities` and the `#livablecities` one, which is visible in the ActivityPub JSON representation.

Will have a look at this case tomorrow. I was thinking this would be handled correctly by using the `CICharField`, but apparently not!

chdorner:
@mouse-reeve I fixed the problem with finding existing hashtags that have a different case. Hoping this should be working now as intended 🎉 

Tested the following:

* Create a new status with a tag that isn't in the DB yet with capital case, i.e. #BookWyrm
* Create another status with the tag #bookwyrm
* Ensure that there's only one #BookWyrm hashtag in the table, and in the ActivityPub JSON representation

mouse-reeve:
Shoot! I messed up the formatting resolving the merge conflicts. If the tests pass I'll fix that and merge. Thank you so much for going through such a long an involved review process with this PR, i'm excited to have it live soon!

chdorner:
@mouse-reeve I just re-resolved the merge conflicts and force-pushed the branch.

Additionally, I had to add the new `allow_external_connections` argument to my new `Note.to_model` function, and I had to re-order the migration again, that tree got really messy between 0171 and 0175, but I think that's fixed now, might be worth another look from you.

varanasib:
Thanks!  A big job and important.  It will be well appreciated.

joshsimmons:
YAY! Thank you so much for your work on this, @chdorner and @mouse-reeve <3

PR #2525, opened 2022-12-18
Update forms.js

### LINKED ISSUE #2519: Adding New Books: hitting enter in book subject field submits the form
**Describe the bug**
When adding a new book, hitting enter in a subject field submits the form.

**To Reproduce**
1. Add new book
2. Type a new subject in subject field
3. hit enter
4. form submits

**Expected behavior**
Well, I wouldn't expect the form to submit! In a perfect world, hitting enter would take me to a new subject field, so that you could easily add many subjects. Baring that, enter should do nothing or possibly take you to whatever field is considered "next".

**Instance**
https://bookwyrm.social

**Additional context**
It's not directly related to https://github.com/bookwyrm-social/bookwyrm/issues/2480, but it sure compounds the problem.

---

    - OS: Windows 10
    - Browser FF
    - Version 107

jaschaurbach:
fixes #2519 

Hitting enter within the description field does not submit the form, hitting enter in any other field does.


bookwyrm/static/js/forms.js
```diff
@@ -46,4 +46,10 @@
     document
         .querySelectorAll("[data-remove]")
         .forEach((node) => node.addEventListener("click", removeInput));
+
+    document.getElementById("id_description").addEventListener("keypress", function (event) {
```
mouse-reeve, line 50: Your javascript looks perfect, but I think it should be applied to `id_add_subject` and the created fields (`id_add_subject-2`, etc) rather than the description field. As it is, this prevents adding line breads to book descriptions.

mouse-reeve (changes requested):

bookwyrm/static/js/forms.js
jaschaurbach: I updated the js - it now uses `querySelectorAll("[name='subjects']")`. 

All other fields still save by hitting enter (except description)

jaschaurbach:
I used another approach with event delegation, described in the commit message.

mouse-reeve (approved):
Works great! Thank you

PR #2527, opened 2022-12-19
Disable sass processer when not in debug mode

chdorner:
As mentioned on the Matrix chat:
> I think the way django-sass-processor is currently configured means that every time an HTML layout is rendered, the sass processor does some work which in part results in it connecting to the static storage (in my case CloudFlare's R2, an S3-compatible storage service). Judging from an application monitoring software I'm running to try and find some performance issues, this adds a good 400ms on average to each request.
> 
> Now, I can turn this off by making the SASS_PROCESSOR_ENABLED configurable and I see those external requests Python makes to R2 disappear when turning that off.
However, the CSS files still need to be compiled (collectstatic doesn't do that), and the command it comes with (python manage.py compilescss) unfortunately fails to compile any file due to errors parsing all the the template files.
> 
> Just wanted to make sure I'm not missing something else, because my instance felt super fast after turning off the scss processing that happens on each request! I had the compiled files already stored, so it loaded alright on the client side.

This adds a new managment command `compile_themes` which lists all files in `bookwyrm/static/css/themes` and compiles them. This command should be run right before `collectstatic` is run when in a production environment.

The docs changes are in bookwyrm-social/documentation#87

TODO:
* [x] Changes to the documentation repo

### DIFF
--- a/bookwyrm/management/commands/compile_themes.py
+++ b/bookwyrm/management/commands/compile_themes.py
@@ -1,0 +1,48 @@
+""" Our own command to all scss themes """
+import glob
+import os
+
+import sass
+
+from django.core.management.base import BaseCommand
+
+from sass_processor.apps import APPS_INCLUDE_DIRS
+from sass_processor.processor import SassProcessor
+from sass_processor.utils import get_custom_functions
+
+from bookwyrm import settings
+
+
+class Command(BaseCommand):
+    """command-line options"""
+
+    help = "SCSS compile all BookWyrm themes"
+
+    # pylint: disable=unused-argument
+    def handle(self, *args, **options):
+        """compile"""
+        themes_dir = os.path.join(
+            settings.BASE_DIR, "bookwyrm", "static", "css", "themes", "*.scss"
+        )
+        for theme_scss in glob.glob(themes_dir):
+            basename, _ = os.path.splitext(theme_scss)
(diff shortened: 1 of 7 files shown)


chdorner:
I ran this on my production instance, ran these commands and it correctly uploaded the built themes CSS file to my external storage provider:
* `python manage.py compile_themes`
* `python manage.py collectstatic --no-input`

chdorner:
Here are some interesting graphs from various monitoring tools to illustrate the performance savings:

From the APM (application performance monitoring):
Screenshot 2022-12-19 at 19 59 17
and the same graph only showing "Web extenral", which includes all time spent talking to external webservers in the gunicorn process:
Screenshot 2022-12-19 at 19 59 47

synthetic monitors (i.e. hitting the logged-out landing page from different locations in the world):
Screenshot 2022-12-19 at 20 02 05
Interesting is that there are much less outliers, especially noticeable from clients further away like US and Japan (the host is in Germany)

another interesting one is just "Network in" on the server itself, seems like it was fetching either the scss source files or the already compile css files from the external storage (presumably to only upload the ones that changed?)
Screenshot 2022-12-19 at 20 04 46
(the two spikes afterwards were me running my IaC scripts)

mouse-reeve:
Thank you for such a thorough PR and a great fix!

PR #2529, opened 2022-12-19
Fix titleby cache fragment clearing

chdorner:
The cache key in the snippet uses the language code: https://github.com/bookwyrm-social/bookwyrm/blob/5ea922a551e3e8173d3c91b40954d08208b7c3b4/bookwyrm/templates/snippets/book_titleby.html#L8

I'm guessing we should just clear all languages caches when updating a book or an author.

### DIFF
--- a/bookwyrm/models/author.py
+++ b/bookwyrm/models/author.py
@@ -1,8 +1,6 @@
 """ database schema for info about authors """
 import re
 from django.contrib.postgres.indexes import GinIndex
-from django.core.cache import cache
-from django.core.cache.utils import make_template_fragment_key
 from django.db import models
 
 from bookwyrm import activitypub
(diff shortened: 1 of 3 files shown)


mouse-reeve:
I wonder if it would be better to just make this cache super short (like 10 seconds), so that it only effects one load of the page or so? That would make it easier to maintain, and the cache buster bit in the model wouldn't be necessary at all.

chdorner:
What was the initial reason to cache this fragment? If a shorter cache TTL still achieves the same thing, then why not

mouse-reeve:
My original goal was just to reduce the number of sql queries when loading a page. At some point I think the cost (fragility, complex cache busting code) becomes not worth it

chdorner:
Was it on specific pages where the this was an issue with too many SQL queries? I could take a look tomorrow and see if this is still happening and if there are other ways to reduce the queries for those pages

mouse-reeve:
In this case I was trying to reduce queries on the feed page. But I don't know that that's the worst offender at the moment, performance-wise (`search`, `get-started/books`, `directory`, and `user` are up there in my analytics)

chdorner:
@mouse-reeve I went with your suggestion here and reduced the cache timeout to 10 seconds and removed the cache busting 👍

mouse-reeve:
I love a PR that removes more lines than it adds 😄

PR #2532, opened 2022-12-20
Display generated notes correctly depending on format

### LINKED ISSUE #2531: Federated generated note contains HTML content
**Describe the bug**
A generated note imported from another instance contains `<p>{content}</p>` in the database field.

When querying for a "wants to read" generated note from a local user:
```
# user_id 2 is a local user
select * from bookwyrm_status where content like '%wants to read%' and user_id = 2;
-[ RECORD 1 ]---+--------------------------------------------------------------
id              | 135
content         | wants to read
# ...
user_id         | 2
# ...
```

When querying for a "wants to read" generated note from a remote user:
```
# user_id 153 is a remote user
select * from bookwyrm_status where content like '%wants to read%' and user_id = 153;
-[ RECORD 1 ]---+---------------------------------------------------------------
id              | 803
content         | <p>wants to read</p>
...
user_id         | 153
...
```

**To Reproduce**
1. Follow a remote user
2. Wait for remote user to create a generated note
3. See the status appear in the feed rendering the HTML tag in the header

Screenshot 2022-12-20 at 00 05 15

**Expected behavior**
The header row in the status should not render the HTML tags out.

**Screenshots**

**Instance**
On my own instance running 0.5.3

**Additional context**
I'm assuming this either is related to how the statuses are rendered in activitypub on the sending side, or interpreted when parsing them on the receiving end.

---

    - OS: macOS
    - Browser: Firefox
    - Version: 108

mouse-reeve:
The related code is _such_ a mess, but I believe this fixes #2531

### DIFF
--- a/bookwyrm/templates/discover/card-header.html
+++ b/bookwyrm/templates/discover/card-header.html
@@ -4,17 +4,17 @@
 {% with user_path=status.user.local_path username=status.user.display_name book_path=book.local_path book_title=book|book_title %}
 
 {% if status.status_type == 'GeneratedNote' %}
-    {% if status.content == 'wants to read' %}
+    {% if status.content == 'wants to read' or status.content == '<p>wants to read</p>' %}
         {% blocktrans trimmed %}
         <a href="{{ user_path}}">{{ username }}</a> wants to read <a href="{{ book_path }}">{{ book_title }}</a>
         {% endblocktrans %}
     {% endif %}
-    {% if status.content == 'finished reading' %}
+    {% if finished reading or status.content == '<p>finished reading</p>' %}
         {% blocktrans trimmed %}
         <a href="{{ user_path}}">{{ username }}</a> finished reading <a href="{{ book_path }}">{{ book_title }}</a>
         {% endblocktrans %}
     {% endif %}
-    {% if status.content == 'started reading' %}
+    {% if started reading or status.content == '<p>started reading</p>' %}
         {% blocktrans trimmed %}
         <a href="{{ user_path}}">{{ username }}</a> started reading <a href="{{ book_path }}">{{ book_title }}</a>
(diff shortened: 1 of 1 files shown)


bookwyrm/templates/discover/card-header.html
```diff
@@ -4,17 +4,17 @@
 {% with user_path=status.user.local_path username=status.user.display_name book_path=book.local_path book_title=book|book_title %}
 
 {% if status.status_type == 'GeneratedNote' %}
-    {% if status.content == 'wants to read' %}
+    {% if status.content == 'wants to read' or status.content == '<p>wants to read</p>' %}
```
chdorner, line 7: This seems quite tied to a specific implementation detail that might be prone to break in the future.

What if we'd add an additional field to the activitypub representation of statuses, something along the lines of `content_md`, so that BookWyrm's parsing side (`to_model`) could look for a the `content_md` field and use that instead of `content` if it is present.
That way we get the best of both worlds, we still render these out for non-BookWyrm instances (i.e. Mastodon), but within BookWyrm handle markdown for remote and local statuses the same way.
Not knowing some the details of the federation piece, this might not be as easy as I think it is 😄

mouse-reeve: It's terrible and super prone to breaking, I agree! But I don't think the most robust fix would be to include the unformatted version of the status. Rather, I think `GeneratedNote` should be sending the reading status change as a field (like how it's done on comments that include reading status changes) with empty content. This PR is just a bandaid.

chdorner: Ah yes, that sounds even better. Which actually also makes me think about how federation of hashtags (as done in #2524) might have a bit of a design issue.

PR #2537, opened 2022-12-20
Fix permissions check for System admin navigation section

chdorner:
I recently realised that only my very first admin account was able to see the "System" navigation section in the admin interface. When I promoted another user to either Admin or Owner. Both of those groups should have the `edit_instance_settings` permission according to the `initdb` command [here](https://github.com/bookwyrm-social/bookwyrm/blob/5ea922a551e3e8173d3c91b40954d08208b7c3b4/bookwyrm/management/commands/initdb.py#L29-L33).

Looking at "Instance Settings" section I'm assuming the issue is that it was using `perms.edit_instance_settings` instead of `perms.bookwyrm.edit_instance_settings`.

This is the navigation section I meant:
image

### DIFF
--- a/bookwyrm/templates/settings/layout.html
+++ b/bookwyrm/templates/settings/layout.html
@@ -74,7 +74,7 @@
             </li>
         </ul>
         {% endif %}
-        {% if perms.edit_instance_settings %}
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Ah, thank you! I've made this mistake before

PR #2547, opened 2022-12-26
Ensure django tests pass in default Docker setting

avandeursen:
When running tests using `./bw-dev pytest` (nice script!), I encountered a few failures. This PR has some simple proposals to fix them, and make the docker tests run more similar to the `django-tests` GitHub action workflow.

First, the test `test_thumbnail_fields` will only pass if `ENABLE_THUMBNAIL_GENERATION` is set to true. However, in the `.env.example` it is set to false. Here are two fixes:

1. Set `ENABLE_THUMBNAIL_GENERATION` to true in `.env.example`
2. Skip `test_thumbnail_fields` unless `ENABLE_THUMBNAIL_GENERATION` is set to true.

Furthermore, the `docker-compose.yml` configuration uses `postgres` without version number. But some tests require postgres:13 or up. So I added this version tag (as it is in `.github/workflows/django-tests.yml`).

(Just using `postgres`  = `postgres:latest` resolved to a postgres:11 image I downloaded long ago when it actually was the latest).

Feel free to cherry-pick selected commits if you prefer not to adopt all changes.

An alternative could also be to bump to postgres 15 in both GitHub actions and Docker compose, but that is a different decision.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -60,7 +60,7 @@ SEARCH_TIMEOUT=5
 QUERY_TIMEOUT=5
 
 # Thumbnails Generation
-ENABLE_THUMBNAIL_GENERATION=false
+ENABLE_THUMBNAIL_GENERATION=true
 
 # S3 configuration
 USE_S3=false
--- a/bookwyrm/tests/models/test_book_model.py
+++ b/bookwyrm/tests/models/test_book_model.py
@@ -2,6 +2,8 @@
 from io import BytesIO
 import pathlib
 
+import pytest
+
 from dateutil.parser import parse
 from PIL import Image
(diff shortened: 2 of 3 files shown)


mouse-reeve:
This looks good to me! It should be a quick fix to get the formatting check to pass - you can run `./bw-dev black` and that will automatically resolve it.

avandeursen:
Oops  -- thanks -- fixed!

nein09:
ah, I was just looking at this and I was tempted to `patch` that setting, but it felt a bit dodgy. This is nicer.

mouse-reeve:
Thank you so much!

PR #2548, opened 2022-12-26
Support custom pg user/db names in DB backup script

### LINKED ISSUE #2546: Database backup script via cron doesn't have environment variables
Bug report from https://github.com/bookwyrm-social/bookwyrm/discussions/2541

**Describe the bug**
The [database backup script](https://github.com/bookwyrm-social/bookwyrm/blob/production/postgres-docker/backup.sh) relies on the `POSTGRES_DB` environment variable, however in the context of the cron execution this env variable is not set.

**To Reproduce**
1. Follow the production setup instructions, make sure to set the `POSTGRES_DB` env variable to something else than the default
2. Wait for the backup cron to be scheduled and see that 

or, another way to test this without waiting:
1. Add `* * * * * env > /var/log/env.log` to `postgres-docker/cronfile`
2. Follow the production setup instructions, make sure to set the `POSTGRES_DB` env variable to something else than the default
3. After about a minute check the content of that log file to see what's in the environment: `docker-compose exec db cat /var/log/env.log`
The environment is:
```
HOME=/root
LOGNAME=root
PATH=/usr/bin:/bin
SHELL=/bin/sh
PWD=/root
```
Missing the `POSTGRES_DB` variable which the backup script refers to, resulting in always empty backups because the `bookwyrm` default database doesn't exist.

**Expected behavior**
The backup script should be able to backup the database configured in the `POSTGRES_DB` env var in the `.env` file.

chdorner:
Fixes #2546

Here's a handy diff how this changes the `bookwyrm-entrypoint.sh`:

```diff
  if ! _is_sourced; then
  	if [[ $1 == cron ]]; then
+ 		echo "POSTGRES_DB=${POSTGRES_DB}" > /backups/.env
+ 		echo "POSTGRES_USER=${POSTGRES_USER}" >> /backups/.env
  		service cron start
  		shift
  	fi
```

The backup script then sources `/backups.env` before executing the rest.

### DIFF
--- a/postgres-docker/backup.sh
+++ b/postgres-docker/backup.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
+source /backups/.env
+
 if [ -z "$POSTGRES_DB" ]; then
     echo "Database not specified, defaulting to bookwyrm"
 fi
+if [ -z "$POSTGRES_USER" ]; then
+    echo "Database user not specified, defaulting to bookwyrm"
+fi
 BACKUP_DB=${POSTGRES_DB:-bookwyrm}
+BACKUP_USER=${POSTGRES_USER:-bookwyrm}
 filename=backup_${BACKUP_DB}_$(date +%F)
-pg_dump -U $BACKUP_DB > /backups/$filename.sql
+pg_dump -U $BACKUP_USER $BACKUP_DB > /backups/$filename.sql
--- a/postgres-docker/Dockerfile
+++ b/postgres-docker/Dockerfile
@@ -15,6 +15,6 @@ RUN touch /var/log/cron.log
 # We need to start the cron service the first time it runs, when it's still being run
 # as the root user. We're going to add a check that looks at the first argument and
 # if it's 'cron', starts the service and then removes that argument.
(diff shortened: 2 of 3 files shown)


postgres-docker/backup.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
```
chdorner, line 1: I think there's an issue with how this file is updated when changing it. Given it's copied to the image from the Dockerfile to `/backups`, but later when starting a container with this image the `/backups` path is put on a volume. So when updating an existing setup where `/backups` is already a volume path, the content of this file doesn't get updated.

chdorner: Should we put this script into a path that isn't on a volume to ensure that it can get updated with newer versions of the docker image?

mouse-reeve: That seems good to me!

chdorner: Moved both scripts out of the volume location and to `/usr/local/bin`, this way we can easily update them in the future.

PR #2550, opened 2022-12-26
Fix fragment caches (language codes, cache keys, etc.)

### LINKED ISSUE #2549: About page is being cached in the wrong language
The about this instance block is being cached without taking into account viewer language

chdorner:
Fixes #2549

I also went through other fragment caching instances and fixed a few small things with them.


mouse-reeve:

PR #2551, opened 2022-12-26
Audiobook play time

### LINKED ISSUE #2536: Playtime for audio books
We have pages for books but we do not have total play time for audio books. Please someone gives me total play time.

it should be used not only on the audiobook page but when listening to one it would be nice to give quotes with a timemark.

joachimesque:
Closes #2536

This PR adds an audiobook duration field whenever the "Audiobook" choice is selected as a format

- I opted for a permissive field, allowing both `hh:mm` and `minutes` formats.
- I saw no use for the seconds in the duration. There is no reason IMO to complicate the feature further.
- The localization/translation of the duration is made through `humanize`. Considering that `django.contrib.humanize` doesn't provide the support for durations, I chose to add it to the requirements, to avoid relying on a lot of localization code and translation keys.
- The JS helps toggle betwen the "pages" field and the "play time" field whenever the Format select changes.

Happy holidays!


bookwyrm/templates/book/edit/edit_book_form.html
```diff
@@ -275,14 +275,26 @@ <h2 class="title is-4">
...
+                    <label class="label" for="id_audiobook_play_time">
+                        {% trans "Audiobook play time:" %}
+                    </label>
+                    {{ form.audiobook_play_time }}
+                    <p class="help">
+                        {% trans "Can be filled as hours:minutes or as the total run in minutes" %}
+                    </p>
+
+                    {% include 'snippets/form_errors.html' with errors_list=form.audiobook_play_time.errors id="desc_audiobook_play_time" %}
+                </div>
```
cincodenada, line 297: I like the generalized toggling in `toggleOnSelect`! Because we're toggling two ways (both on render, and reactively with Javascript), this will break this portion of the form for people with Javascript disabled.

It seems there's no good way to react to a dropdown changing with just CSS, so can we fall back to showing both fields at render time, and run the javascript code to hide one of the fields on pageload instead? It's not the best experience to have two fields and only need one, but I think it's better than switching it to an audiobook and not being able to input the length.

bookwyrm/templates/book/publisher_info.html
```diff
@@ -18,7 +19,13 @@
...
 <p>
     {% if format and not pages %}
-        {{ format }}
+        {% if format_property and format_property == "AudiobookFormat" and book.audiobook_play_time %}
+            {{ format }}<br />
+            {% trans "Play time:" %}
+            {{ book.audiobook_play_time|localized_duration }}
+        {% else %}
+            {{ format }}
+        {% endif %}
```
cincodenada, line 28: We should just be checking `book.physical_format` here instead of `format_property` - the latter is `firstof book.physical_format book.physical_format_detail`, and if I understand correctly, `book.physical_format_detail` will never be `AudiobookFormat`, so there's no reason to include it.

I also think the structure of the logic here could be rearranged to match the existing flat style better:
```
{% if book.physical_format == "AudiobookFormat" and book.audiobook_play_time %}
    {{ format }}<br />
    {% trans "Play time:" %}
    {{ book.audiobook_play_time|localized_duration }}
{% elif format and not pages %}
    {{ format }}
{% elif format and pages %}
    {% blocktrans %}{{ format }}, {{ pages }} pages{% endblocktrans %}
{% elif pages %}
    {% blocktrans %}{{ pages }} pages{% endblocktrans %}
{% endif %}
```

I do like nesting to express the conditions better, I just think mixing the two styles is more confusing than sticking with either way. You could also make the whole thing nested, and personally I slightly prefer it, but that's a bigger change:
```
{% if format %}
    {% if book.physical_format == "AudiobookFormat" %}
        {{ format }}
        {% if book.audiobook_play_time %}
            <br />
            {% trans "Play time:" %}
            {{ book.audiobook_play_time|localized_duration }}
        {% endif %}
    {% else %}
      {% if pages %}
          {% blocktrans %}{{ format }}, {{ pages }} pages{% endblocktrans %}
      {% else %}
          {{ format }}
      {% endif %}
    {% endif %}
{% elif pages %}
    {% blocktrans %}{{ pages }} pages{% endblocktrans %}
{% endif %}
```
Honestly, the pages/play time switch adds enough complexity to this it might be worth pulling out as a snippet.

cincodenada (changes requested):
This looks great overall, and is an excellent addition. I left some comments, most importantly that I think we should avoid storing loosely-formatted dates directly in the database, and interpret those at input time instead. I also had a minor observation around breakage for users without Javascript. The remaining feedback is mostly style preferences and is not crucial.

Thanks for making this happen!

joachimesque:
Thank you for the review @cincodenada :)

I agree with you about storing badly formatted data in the database, and had first tried with an ArrayField, which didn’t work too well. And somehow I had overlooked the time interval field type, made exactly for this type of data 🤦 

As for the JS: you’re right, I forgot about non-JS users! What would you think of a CSS utility that’s like `.is-hidden`, but works only when there’s JS in the page? Or maybe it exists and I forgot about it? Your solution fits well, too (and saves up on somewhat unreadable Django template code)

bookwyrm/activitypub/book.py
```diff
@@ -59,6 +59,7 @@ class Edition(Book):
     isbn13: str = ""
     oclcNumber: str = ""
     pages: int = None
+    audiobookPlayTime: str = ""
```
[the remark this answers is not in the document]
joachimesque: One big question is: how will it federate? My transform (* 60 to the value at clean() time) would multiply an already correct value, coming from another server, and so on. Do you know how I can prevent that? Is it by making an activitypub-aware field with `field_to_activity` and `field_from_activity` transforms? That's a part I really don't know in the codebase.

bookwyrm/static/js/forms.js
```diff
@@ -58,4 +58,37 @@
             }
         }
     });
+
+    /**
+     * Toggle between two fields depending on select value
+     *
+     * @param {target} the select that will direct the toggle
+     */
+    function handleToggleOnSelect(target) {
```
mouse-reeve, line 67: This was seamless, love it

bookwyrm/templates/book/format_info.html
```diff
@@ -0,0 +1,39 @@
...
+    <meta itemprop="timeRequired" content="{{ book.audiobook_play_time|iso_duration }}">
+{% endif %}
+{% endif %}
+
+{% if pages %}
+<meta itemprop="numberOfPages" content="{{ pages }}">
+{% endif %}
+
+<p>
+    {% if format and not pages %}
```
mouse-reeve, line 23: This does mean that if a book is marked as an audiobook but has both playtime and pages, it will show pages. That's kind of an edge case, and it might make the boolean logic super complicated to address, though

bookwyrm/models/book.py
```diff
@@ -295,6 +295,10 @@ class Edition(Book):
         max_length=255, choices=FormatChoices, null=True, blank=True
     )
     physical_format_detail = fields.CharField(max_length=255, blank=True, null=True)
+    audiobook_play_time = models.DurationField(
+        blank=True,
+        null=True,
+    )
```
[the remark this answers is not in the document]
joachimesque: I can see if it's possible to add a check or a transform in the Form

mouse-reeve:
Does this need anything besides an update from `main` and a merge migration to be ready to go? I don't think either of my comments above need to be blockers

bookwyrm/templates/book/format_info.html
joachimesque: I agree that it would complicate things, and is rarely bound to happen.

bookwyrm/templates/book/format_info.html
```diff
@@ -0,0 +1,39 @@
...
+
+{% firstof book.physical_format_detail book.get_physical_format_display as format %}
+{% firstof book.physical_format book.physical_format_detail as format_property %}
+{% with pages=book.pages %}
+{% if format or pages %}
+
+{% if format_property %}
+<meta itemprop="bookFormat" content="{{ format_property }}">
+{% if book.physical_format == "AudiobookFormat" and book.audiobook_play_time %}
+    <meta itemprop="timeRequired" content="{{ book.audiobook_play_time|iso_duration }}">
```
mouse-reeve, line 14: When I tested this, it looks like this line was passing in a string-ified version of the duration (for the arbitrary numbers I tried, `85 days, 17:36:00`), which was throwing an error in the filter. I think that this is because I entered a big enough number that it went over hours and into days

bookwyrm/models/book.py
```diff
@@ -299,6 +299,10 @@ class Edition(Book):
         max_length=255, choices=FormatChoices, null=True, blank=True
     )
     physical_format_detail = fields.CharField(max_length=255, blank=True, null=True)
+    audiobook_play_time = models.DurationField(
```
mouse-reeve, line 302: Because this is calling a standard Django model field, rather than a custom BookWyrm activitypub model field, it's not going to be serialized in ActivityPub. To fix this, there needs to be a subclass of `DurationField` in https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/fields.py#L528 -- that code will handle parsing the string version of the play time into a DurationField. 

I know this is kind of murky depths of how the activitypub implementation works --is that something you want to do, and if so does that give you enough of an idea of how to proceed? If you'd rather not, I can go ahead and make that change.

mouse-reeve (changes requested):

mouse-reeve:
I believe the tests are just failing because this needs a merge migration

bookwyrm/models/book.py
jaschaurbach: @joachimesque Have you seen this?

mouse-reeve:

PR #2554, opened 2022-12-27
Allow to set default user auth group

### LINKED ISSUE #2388: Allow admins to set default user role
Currently user roles default to Editor. 

Either a value in settings.py or ideally something in the admin UI

see above

**Additional context**

chdorner:
Fixes #2388

This allows to set the default auth group when a new user signs up by:
- adding a new dropdown field to the Registration admin page
- queries for that group in `User.save` and adds that (instead of hardcoding the `editor` group)
- ensuring that `initdb` sets the default group to `editor` when initialising the site settings
- backfilling site settings and setting the value correctly to the `editor` group

Similarly to the user admin form, it allows to set the group to "User", which nulls out that column and prevents the user from getting a group added on `save`.

<details>
<summary>Screenshots</summary>
image
image
</details>


bookwyrm/models/user.py
```diff
@@ -356,8 +356,14 @@ def save(self, *args, **kwargs):
...
-                self.groups.add(Group.objects.get(name="editor"))
-            except Group.DoesNotExist:
+                group = (
+                    apps.get_model("bookwyrm.SiteSettings")
+                    .objects.get()
+                    .default_user_auth_group
+                )
+                if group:
+                    self.groups.add(group)
+            except ObjectDoesNotExist:
```
mouse-reeve, line 366: I think this try/except block won't be necessary now, since the value of `default_user_auth_group` can be none?

bookwyrm/migrations/0173_default_user_auth_group_setting.py
```diff
@@ -0,0 +1,37 @@
...
+import django.db.models.deletion
+
+
+def backfill_sitesettings(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    group_model = apps.get_model("auth", "Group")
+    editor_group = group_model.objects.using(db_alias).filter(name="editor").first()
+
+    sitesettings_model = apps.get_model("bookwyrm", "SiteSettings")
+    for site in sitesettings_model.objects.all():
```
mouse-reeve, line 13: totally not important, but I think you can write this as:
```python
sitesettings_model.objects.update(default_user_auth_group=editor_group)
```

and not use a loop.

mouse-reeve (changes requested):
Extremely minor tweaks! This worked really well when I tested it, thank you

bookwyrm/models/user.py
chdorner: Just tested this and the `ObjectDoesNotExist` exception is thrown in the tests given that the site settings are not (always?) initialised.

mouse-reeve: ah that makes sense! I overlooked the `objects.get()` in there.

mouse-reeve (approved):
Really appreciate your work on this

PR #2555, opened 2022-12-28
Add Website field to author

jaschaurbach:
Authors have Websites. that might be interesting for BookWyrm users. Now there is a field specify the website of an author and it is shown tu users under the wikipedia link

minor fixes:
Adds max char for ISFDB ID (6) and ISNI (19 - 16 plus three -)


mouse-reeve:
Works perfectly, thank you!

PR #2560, opened 2023-01-01
Suggestion list

joachimesque:
Books now can have a list of suggestions at the bottom of the page.

A List has:

- a one-to-one relationship with a book
- name: ""
- description: ""
- curation: open
- privacy: public
- group: None

## In this PR

- Lists are displayed at the bottom of the Book page
- You can create a list if it doesn't exist
- You can search for books to add to the list from the Book page (in the same manner as on the list page)
- You can add books from suggestion search results
- You can view the list in the List view, but some settings are disabled
- TODO : display the right info about lists in the Book page column

## Next steps

- Use `endorsement` field in `ListItem` to allow upvotes on books in lists, and ordering by upvote
- Restrict curation to admins (the author of a list is irrelevant for Book suggestions)
- Allow anyone to add comments on list items?

## Problems

- what if two users from different servers create lists for the same book? Should they merge?


mouse-reeve:
I think I successfully fixed the merge conflict logic-wise, even though I did not succeed in making Black happy. If you run black and make a merge migration things should be back in order

mouse-reeve:
I tried it out and this is so cool! There are a few UI things I ran into, but I'm guessing you're not at the point where that would be helpful yet.

> * I hesitated between "suggestion" and "recommendation". what do you all prefer?

I think "recommendation" is more standard, so I'd default to that

> * what if two users from different servers create lists for the same book? Should they merge?

Oh good question! I think having the lists merge makes the most sense as a user, but it might be weird to handle from a technical perspective 🤔

joachimesque:
> > * what if two users from different servers create lists for the same book? Should they merge?
> 
> Oh good question! I think having the lists merge makes the most sense as a user, but it might be weird to handle from a technical perspective 🤔

Right now I've set the List with a one-to-one relationship to the Book. With the problem of federation, multiple lists should be associated with the Book (is it a `ForeignKey` relationship?). What do you think of the idea of creating a new "ProxyList" list type, that's not really a list, but a merge of multiple lists? The interface displays the ProxyList, and whenever the user adds a book to the list, it gets added to the instance's own list for that book (that gets shown in the ProxyList)

There should also be a mechanism for asking other instances for their List for a book. Is there a way to have the list of all instances that have the same book? If not, do you think it would be hard to implement? (that would be on another PR, but it's vital for the feature IMO)

joachimesque:
@mouse-reeve I'm open to a re-read of this PR. There are still questions about federation and all that, it's a domain I don't know much about :)

mouse-reeve:
I know it has literally been a year (to the month, wow) since this had any movement, but I finally have time again and I think this is a huge and delightful contribution so I'd like to merge it. Since it's been so long, I'm going to go ahead and try to finish up the remaining pieces myself. Let me know if you'd rather I didn't, but I'll plow ahead unless told otherwise in #3420 :)

joachimesque:
I’m in tune with whatever you suggest! My code proposals are gifts to the project and to you so that you can do with it as you see fit, especially since I haven’t had much time to contribute to Bookwyrm this last year 😄

mouse-reeve:
I'm going through and closing pull requests that are just too out of date. This one lives on! My progress is incremental but it isn't stopped 😅

PR #2562, opened 2023-01-01
Add shelves column in shelf table

### LINKED ISSUE #2484: Add "Shelf" column in book list
When looking at your list of books, there is no indication what shelf a book is on. 

Add a column that indicates the shelf the book is on.

I know you can filter by shelf, but that's not what I am suggesting. Filtering does not tell me what shelf a book is on, except in a cumbersome way of filtering each shelf till I find the book.

**Additional context**

2022-12-09 15_37_13-Your books - bookrastinating com — Mozilla Firefox

Apologies if this feature already exists. If it does exist, then it should be more prominent.

joachimesque:
Closes #2484

### DIFF
--- a/bookwyrm/templates/shelf/shelf.html
+++ b/bookwyrm/templates/shelf/shelf.html
@@ -155,6 +155,9 @@
                         {% endif %}
                     <th>{% trans "Rating" as text %}{% include 'snippets/table-sort-header.html' with field="rating" sort=sort text=text %}</th>
                     {% endif %}
+                    {% if shelves|length > 0 %}
+                    <th>{% trans "Shelves" %}</th>
+                    {% endif %}
                     {% if shelf.user == request.user %}
                     <th aria-hidden="true"></th>
                     {% endif %}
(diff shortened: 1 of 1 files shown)


bookwyrm/templates/shelf/shelf.html
```diff
@@ -189,6 +192,19 @@ <h2 class="title is-3">
                         {% include 'snippets/stars.html' with rating=book.rating %}
                     </td>
                     {% endif %}
+                    {% if shelves|length > 0 %}
+                    <td data-title="{% trans "Shelves" %}">
+                        {% spaceless %}
+                        {% for shelf in shelves %}
+                        {% if shelf in book.shelves.all %}
```
mouse-reeve, line 199: I think `book.shelves` will produce all shelves for all users, rather than the shelf related to the current user

jaschaurbach:
@joachimesque Are you still there and have resources to take a look at the review?

joachimesque:
Sorry, I haven’t had the time (personal matters). Please don’t hesitate to deal with this PR as you see fit, I don’t know if its changes are still relevant :)

joachimesque:
@mouse-reeve @jaschaurbach The PR is open for another look :)

bookwyrm/templates/shelf/shelf.html
mouse-reeve: Oh, we're actually both sort of right -- `book.shelves.all` _does_ produce all shelves for all users, but because it's checking if a `Shelf` object is present, and a Shelf is tied to a user, it still displays correctly, so this does work exactly how you intended. 

My new qualm is that on a large instance, that's going to be a costly lookup, since there are a lot of shelves. It may be that assigning `book.shelves.all` to a variable using `{% set ... %}`, so that the query is only called once rather than once per shelf, will be enough mitigation?

joachimesque:
Ah, I see what you mean, I hadn’t thought of it! I’ll try that out. What would be the best way to test the cost in terms of performance?

mouse-reeve:

Issue #2558, opened 2022-12-31
Please make it so we can add the site to our iOS Home Screen - progressive web app 


shaunrobinson:
In the absence of any apps for Bookwyrm, I’d like to add it to my iOS home screen and for it to launch on its own (not in safari). 

This can be done for e.g Mastodon, but cannot currently be done for bookwyrm. Please add this feature so I don’t have to keep using safari.

cincodenada:
Yes! Having Bookwyrm as a PWA has long been on my wishlist, thanks for adding this as an issue.

For just the "not have to go through Safari" bit, does [Add to Home Screen](https://www.howtogeek.com/667910/how-to-add-a-website-to-your-iphone-or-ipad-home-screen/) in the Share menu satisfy your needs? It looks similar to the functionality I'm familiar with in Android. It's still using your browser under the hood, but that's the case with a full-blown PWA like Mastodon as well.

Enabling a proper PWA is a much larger endeavor that involves adding a layer to support offline use that syncs back up to the website when you're back online, which is definitely something I'd like to see, but there's not a quick fix for.

shaunrobinson:
No, when you add a home screen icon all it does is launch the bookmark in Safari when you tap it - not ideal. 

I don’t think it’s that complicated to make it stay in its own “app”, you don’t need to make a full blown PWA to do that (it’s just a few lines of HTML, I’ve done it before myself with sites)

https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/ConfiguringWebApplications/ConfiguringWebApplications.html

bahuma20:
The requirements to be an installable PWA are actually quite low.

A manifest file and a service worker with no logic is enough: https://web.dev/install-criteria/

pabloscloud:
> For just the "not have to go through Safari" bit, does [Add to Home Screen](https://www.howtogeek.com/667910/how-to-add-a-website-to-your-iphone-or-ipad-home-screen/) in the Share menu satisfy your needs?

On iOS when adding something to the homescreen it still opens it in safari. Not so with pwa's.

gutierri:
To be an installable PWA just a manisfest.json is enough. I made a small sketch of what a PR proposal could look like:

``` json
{
  "name": "BookWyrm",
  "short_name": "BookWyrm",
  "description": "Social network for tracking your reading",
  "start_url": "/",
  "display": "fullscreen",
  "dir": "auto",
  "orientation": "portrait",
[...]

PR #2564, opened 2023-01-01
Add Metadata for iOS stand alone Website without Browser Bar

jaschaurbach:
Quickfix to solve #2558 

Will dodge it after successful implementation of manifest.json

### DIFF
--- a/bookwyrm/templates/layout.html
+++ b/bookwyrm/templates/layout.html
@@ -14,7 +14,8 @@
 
     <link rel="shortcut icon" type="image/x-icon" href="{% if site.favicon %}{% get_media_prefix %}{{ site.favicon }}{% else %}{% static "images/favicon.ico" %}{% endif %}">
     <link rel="apple-touch-icon" href="{% if site.logo %}{{ media_full_url }}{{ site.logo }}{% else %}{% static "images/logo.png" %}{% endif %}">
-
+    <meta name="apple-mobile-web-app-title" content="BookWyrm">
+    <meta name="apple-mobile-web-app-capable" content="yes">
     {% if preview_images_enabled is True %}
     <meta name="twitter:card" content="summary_large_image">
     {% else %}
--- a/bookwyrm/templates/layout.html
+++ b/bookwyrm/templates/layout.html
@@ -13,7 +13,7 @@
     <link rel="search" type="application/opensearchdescription+xml" href="{% url 'opensearch' %}" title="{% blocktrans with site_name=site.name %}{{ site_name }} search{% endblocktrans %}" />
 
     <link rel="shortcut icon" type="image/x-icon" href="{% if site.favicon %}{% get_media_prefix %}{{ site.favicon }}{% else %}{% static "images/favicon.ico" %}{% endif %}">
-    <link rel="apple-touch-icon" href="{% if site.logo %}{{ media_full_url }}{{ site.logo }}{% else %}{% static "images/logo.png" %}{% endif %}">
+    <link rel="apple-touch-icon" href="{% static "images/logo.png" %}">
     <meta name="apple-mobile-web-app-title" content="BookWyrm">
     <meta name="apple-mobile-web-app-capable" content="yes">
     {% if preview_images_enabled is True %}


jaschaurbach:
Forgot to show how it looks like. This is on my iPhone SE. The safari bar is gone :) 

IMG_3309

jaschaurbach:
Forgot the sentence: This fixes it.

bookwyrm/templates/layout.html
```diff
@@ -13,8 +13,9 @@
     <link rel="search" type="application/opensearchdescription+xml" href="{% url 'opensearch' %}" title="{% blocktrans with site_name=site.name %}{{ site_name }} search{% endblocktrans %}" />
 
     <link rel="shortcut icon" type="image/x-icon" href="{% if site.favicon %}{% get_media_prefix %}{{ site.favicon }}{% else %}{% static "images/favicon.ico" %}{% endif %}">
-    <link rel="apple-touch-icon" href="{% if site.logo %}{{ media_full_url }}{{ site.logo }}{% else %}{% static "images/logo.png" %}{% endif %}">
-
+    <link rel="apple-touch-icon" href="{% static "images/logo.png" %}">
```
chdorner, line 16: I think this should still keep the `site.logo` check for when an instance has configured a custom logo, no?

jaschaurbach: If it is a .jpg or anything else than png the touch.icon will not work. Apple Touch Icon only works with png-files.

chdorner: Ah interesting, didn't know. I'm not sure it's a great user experience to choose the default one regardless of the customization on the instance. There might be a different solution to this

jaschaurbach: Yeah, if we convert every uploaded logo to a png. Or create a separate smaller files as png. BTW: If the logo is not square the touch icon will make not to a square - no matter how it looks like. THAT is a bad user experience.

chdorner: You're right, but admin's can fix that by uploading a square logo. Only referencing the default image doesn't give admins any workaround to get their own logo into this.

Just tested it with a JPG and it does work, even though Apple's docs say PNG. So technically not up to spec, but does work in practice.

jaschaurbach: Did not work on my iPhone or iPad while it work on another iPhone which was a minor fix more up to date...

I think BTW it is a bad idea to use the manual uploaded logo as the Touch Touch Logo should be uploaded separatly - we do not use the Logo as Favicon as well and we should not do that but define clear fields for that.

cincodenada: I agree we should allow admins to customize this logo, many instances have custom logos and it would be confusing to have the default bookwyrm.social logo show up on their homescreen.

Perhaps initially we can add a separate configuration item for the app icon logo to use here, so that admins can create the proper logo if they want to. It could default to the main logo, or to the default Bookwyrm logo -  I'm not sure whether it would be better to have an incorrect logo, or a squashed/potentially logo.

A further enhancement may be to automatically generate a PNG of the correct size (similar to generating thumbnails) if the admin doesn't provide a mobile-specific logo.

cincodenada:
If we add this, we should also add the non-apple `mobile-web-app-capable` as well. The main idea behind both of these tags seems to be that the website works "as an app", which at bare minimum means it doesn't require back/forward buttons to operate, which Bookwyrm probably meets fine. As a user, I would be a little disappointed that there weren't any other PWA-type features, so I don't think this solves #2558 completely.

cincodenada:
Alternatively, it looks like [A2HS/Web app manifests](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Add_to_home_screen) may be a cross-platform way to specify this?

jaschaurbach:
Not sure if I want to do it at the moment as I am kind of frustrated. This was introduced as _quick fix_. I wrote clearly that there will be a manifest.json (like what @cincodenada just wrote about). It is nearly finished - the missing part is to auto-generate the different sizes of the logo which should be provided as png. 

In fact it solves the need of #2558 as shaunrobinson did not want a full blown pwa. For a full blown pwa there is need for some .js to get the notifications to the home-screen and the badge to the app-logo. 

I just close this PR now - it leads to nowhere.

rokejulianlockhart:
Actually, is it possible to retain the browser bar even as an installed PWA? https://play.google.com/store/apps/details?id=com.senecalearning.app does this, and it's a wonderful improvement, since the user gains the speed benefits of a PWA *with* the versatility of things like a consistent refresh button.

Issue #2571, opened 2023-01-04
When validating the form for adding book manually, some fields are emptied if a field is not valid


iGormilhit:
**Describe the bug**
When adding manually an item (book), if a field is not valid, after hitting enter some fields are emptied (values removed), such as the author, the subject, the cover, etc.

1. Add manually a book.
2. Fill the form with a publication date having only the year.
3. Add an author, subjects and a cover.
4. Save the form.
5. See the emptied fields.

**Expected behavior**
All filled in fields should keep their value after validating the form.

**Instance**
https://bw.diaspodon.fr

**Additional context**

---

**Desktop (please complete the following information):**
    - OS: Archlinux
    - Browser: Firefox
    - Version: 108

chdorner:
For the cover file select field this is related to #2421

chdorner:
Did a bit more manual testing on this:

When form submit returns with a validation error (i.e. only selecting a publishing year, not month/day):
* Subjects field
  * Adding two subjects in the initial create form
  * on the form with validation errors: Only one subject field visible and with an empty value 
* Author field
  * Adding two authors in the initial create form
  * on the form with validation errors: Only one author field visible and with an empty value
* Load cover from url field
  * Paste image url in this field on the initial create form
  * on the form with validation errors: value is empty

When form submit ends up on the "Confirmation" screen:
* Subjects field
  * Adding two subjects in the initial create form
  * On the confirmation screen there is only one subject field with an empty value
  * When going back from the confirmation screen to the create form only the first subject appears again
  * When confirming, no subjects are stored

Issue #2421, opened 2022-11-19
Book cover image not saved when book manually added


marcusyoung:
**Describe the bug**
If you add a new book manually and complete the details and add the cover, when you save the book the cover is not saved and you have to add it again.

jaschaurbach:
I just had the same experience just at this moment on 0.5.1

chdorner:
I can reproduce this by:

1. Open Create new Book page (same should apply to editing an existing book)
2. Select a file (the bug only appears when uploading a local file, not when pasting a file URL)
3. Add an author's name
4. Submit the form, but ensure that the form is shown again to confirm for example the author (the "Is "Jane Doe" one of these authors?")
5. Notice that on this step the "Upload cover" form is empty
6. Confirm to create the book
7. Notice no cover uploaded

If I'm not mistaken, this is a browser behaviour which can't be worked around. So the solution would either be a fairly involved temporary storage of this information on the backend before it's confirmed and applied, or a design change that prompts the user to select the cover file again (which also is not ideal).

pi3r8:
can confirm I am also having this issue as of 30/11/23. Adding cover manually a second time does work.

hughrun:
Fixed in [#3083](https://github.com/bookwyrm-social/bookwyrm/pull/3083)

PR #2572, opened 2023-01-04
Ensure transient `cover_url` field value gets persisted across form renders

chdorner:
Given this field doesn't map to an `Edition` model field it lost its values when re-rendering the form. It worked only when the form was valid and rendered as part of the confirmation screen, which is due to the context data value being set in `add_authors` which was only getting called after the form validation.

I've opted to pull it out into a separate new function that gets called before form validation.

This addresses parts of:
* #2421
  * the same problem exists on the cover file selection field, but afaik there's no easy fix for file selection
* #2571
  * similar behaviour exists for the authors and subjects fields. but those fixes are slightly more involved, so will separate them into different PRs

### DIFF
--- a/bookwyrm/tests/views/books/test_edit_book.py
+++ b/bookwyrm/tests/views/books/test_edit_book.py
@@ -82,6 +82,7 @@ class EditBookViews(TestCase):
         form = forms.EditionForm(instance=self.book)
         form.data["title"] = ""
         form.data["last_edited_by"] = self.local_user.id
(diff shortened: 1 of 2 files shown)


mouse-reeve:
What steps did you do to replicate the issue related to the url field specifically? I'm trying to test this but not sure what I'm looking for

chdorner:
Adding a cover url and then having the form validation fail, which is currently easiest done by adding a published date with only the year leaving month/day blank.

mouse-reeve:
Thank you! Works like a charm :)

PR #2576, opened 2023-01-06
Use the .url method on the avatar object

tebriel:
The current method of constructing an avatar by URL does not work in my CDN (Azure BlobStorage) setup. Asking Django to build the URL instead of contstructing it with string building seems like the correct way anyways, hopefully this is backwards compatible.


tebriel:
I didn't properly clone the botos3 storage classes with their root paths, I think this is still probably a best practice but is no longer a bug and is just a different way of doing something.

PR #2586, opened 2023-01-10
{list} Fix duplicate suggestions in Add Books section

### LINKED ISSUE #2584: (List page) Book showing up twice in Add Books suggestions
**Describe the bug**

When I set up a dev instance with one book, it showed up twice in the _Add Books_ suggestions on a list page:

**To Reproduce**
Steps to reproduce the behaviour:
1. Set up a fresh dev instance & create an admin user.
2. Add a book
3. Create a new list

The _Add Books_ suggestion list shows the same book twice. These are the exact same book (not multiple editions).

**Expected behaviour**

Should only show each book once.

**Screenshots**

Screen Shot 2023-01-09 at 21 35 50 PM

**Instance**

Dev

**Additional context**

It's this code in **views/list/list.py** (_get_list_suggestions_):

```python
    # just suggest whatever books are nearby
    suggestions = user.shelfbook_set.filter(~Q(book__in=book_list.books.all()))
    suggestions = [s.book for s in suggestions[:5]]
    if len(suggestions) < 5:
        suggestions += [
            s.default_edition
            for s in models.Work.objects.filter(
                ~Q(editions__in=book_list.books.all()),
            ).order_by("-updated_date")[: 5 - len(suggestions)]
        ]
```

I tried replacing `suggestions` (`suggestions = [`) instead of appending to it (which works for this case), but I don't know if that's correct because I don't understand what this is supposed to be doing.

Is it trying to query other editions of the same book? (If so, why?)

I'm happy to submit a PR for a fix once it's clear.

asmaloney:
(Also fix a spelling mistake in a comment in book_search.py)

Fixes #2584

Note that I'm not a Python person, so please tell me if any of this is non-Python-y.

- made the number of suggestions a parameter instead of a magic number
- I'm not sure if the `distinct()`s on queries are strictly necessary? But it does make it clearer.
- instead of appending to the suggestion list directly I create a second list
- I combine them to make them unique and limit the number of items to `num_suggestions`

**Edit:** I'm sure there's a fancy way to do this all in one query using a join, but handling it in Python seems fine given the size of the things we're dealing with.

### DIFF
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -20,7 +20,7 @@ def search(query, min_confidence=0, filters=None, return_first=False):
     query = query.strip()
 
     results = None
-    # first, try searching unqiue identifiers
+    # first, try searching unique identifiers
     # unique identifiers never have spaces, title/author usually do
     if not " " in query:
         results = search_identifiers(query, *filters, return_first=return_first)
--- a/bookwyrm/views/list/list.py
+++ b/bookwyrm/views/list/list.py
@@ -94,7 +94,7 @@ class List(View):
         return redirect(book_list.local_path)
 
 
-def get_list_suggestions(book_list, user, query=None):
+def get_list_suggestions(book_list, user, query=None, num_suggestions=5):
     """What books might a user want to add to a list"""
     if query:
         # search for books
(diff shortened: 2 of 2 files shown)


mouse-reeve:
Looks like python to me 😄

PR #2589, opened 2023-01-10
Fix two "aria-hidden" tags

asmaloney:


mouse-reeve:
Omg 🤦 -- thanks so much for fixing these!

PR #2590, opened 2023-01-10
Fix spelling in blocktrans arg

asmaloney:
I couldn't track down exactly how this arg is used, but I'm guessing it wasn't intentionally misspelled?


mouse-reeve:
This is definitely a bug, good catch -- it's used in the admin views and probably was causing the string to always display using the plural

PR #2598, opened 2023-01-12
Revert "Add Website field to author"

mouse-reeve:
Reverts bookwyrm-social/bookwyrm#2555


mouse-reeve:
@jaschaurbach I should have thought of this beforehand, but the constraints on the database table you added mean that if, for example, an entry exists with an ISNI longer than 19 characters, the migration will not run

jaschaurbach:
> ISNI longer than 19 characters, the migration will not run

I see the point. There should not be an ISNI longer than 19 characters (including space). 

But wouldn’t it be better to just unchange the character limit for this field instead of reverting the whole pull?

mouse-reeve:
I reverted it because I needed to remove the migration from the chain, rather than just update it with a followup migration, because then it would still try to apply the limit and then un-apply it (I think! It's possible I was wrong). Would you be down to re-open the PR minus the field length limits? I think those would be good to apply as form validation for now.

jaschaurbach:
Will do when I find my yubikey to sign the commit -.-

PR #2600, opened 2023-01-12
Fix generated note content checks in card header template

chdorner:
#2532 unfortunately introduced a bug with invalid Python in the card-header template 😬


mouse-reeve:
omg this is why I need to start having people code review _all_ my PRs 🤦

PR #2603, opened 2023-01-14
Redis: add Unix socket support

alopix:
Currently, Redis is only supported using a dedicated hostname, but some hosters (e.g. uberspace.de) only document/support how to run Redis using a Unix socket.

I hope this simple PR adding support for the `unix://` schema is a good first contribution.


alopix:
I was contemplating if I should introduce a new env var but wanted to reuse what was there to keep it simple. But definitely happy to close this in favour of the other PR 😄

PR #2604, opened 2023-01-14
Add support to connecting to redis via unix socket

rrgeorge:
This PR will add support for specifying the redis url, which will allow for connect to redis via unix socket with `unix://` urls


rrgeorge:
Fixed it so it was on the main branch

rrgeorge:
I think ive corrected all the linting problems

celerywyrm/settings.py
```diff
@@ -8,9 +8,13 @@
...
 REDIS_BROKER_PORT = env("REDIS_BROKER_PORT", 6379)
 REDIS_BROKER_DB_INDEX = env("REDIS_BROKER_DB_INDEX", 0)
+REDIS_BROKER_URL = env(
+    "REDIS_BROKER_URL",
+    f"redis://:{REDIS_BROKER_PASSWORD}@{REDIS_BROKER_HOST}:{REDIS_BROKER_PORT}/{REDIS_BROKER_DB_INDEX}",
+)
 
-CELERY_BROKER_URL = f"redis://:{REDIS_BROKER_PASSWORD}@{REDIS_BROKER_HOST}:{REDIS_BROKER_PORT}/{REDIS_BROKER_DB_INDEX}"
-CELERY_RESULT_BACKEND = f"redis://:{REDIS_BROKER_PASSWORD}@{REDIS_BROKER_HOST}:{REDIS_BROKER_PORT}/{REDIS_BROKER_DB_INDEX}"
+CELERY_BROKER_URL = REDIS_BROKER_URL.replace("unix:", "redis+socket:")
```
mouse-reeve, line 16: I think it would be clearer to re-build the url than the use replace here, if possible

bookwyrm/redis_store.py
```diff
@@ -4,12 +4,7 @@
 
 from bookwyrm import settings
 
-r = redis.Redis(
-    host=settings.REDIS_ACTIVITY_HOST,
-    port=settings.REDIS_ACTIVITY_PORT,
-    password=settings.REDIS_ACTIVITY_PASSWORD,
-    db=settings.REDIS_ACTIVITY_DB_INDEX,
-)
+r = redis.from_url(settings.REDIS_ACTIVITY_URL)
```
mouse-reeve, line 7: will redis connect correctly to the broker in [bookwyrm/tasks.py](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/tasks.py#L10) using the `CELERY_BROKER_URL` given the way you're using the `REDIS_` url here?

bookwyrm/settings.py
```diff
@@ -207,7 +207,10 @@
 REDIS_ACTIVITY_PORT = env("REDIS_ACTIVITY_PORT", 6379)
 REDIS_ACTIVITY_PASSWORD = env("REDIS_ACTIVITY_PASSWORD", None)
 REDIS_ACTIVITY_DB_INDEX = env("REDIS_ACTIVITY_DB_INDEX", 0)
-
+REDIS_ACTIVITY_URL = env(
+    "REDIS_ACTIVITY_URL",
+    f"redis://:{REDIS_ACTIVITY_PASSWORD}@{REDIS_ACTIVITY_HOST}:{REDIS_ACTIVITY_PORT}/{REDIS_ACTIVITY_DB_INDEX}",
+)
```
mouse-reeve, line 213: Forgive my ignorance here - would it be possible to specify the protocol as its own .env variable, defaulting to `redis://` which could be changed to `unix://`, rather than making the entire url an env variable?

mouse-reeve (commented):
I'm not super familiar with the way you're running things here, so I appreciate you bearing with me!

celerywyrm/settings.py
rrgeorge: This is because redis-py (used everywhere else) requires `unix://` for socket urls, but celery requires `redis+socket://` for socket urls

bookwyrm/redis_store.py
rrgeorge: Yes. The idea is that you'd use `unix://` in your env for socket urls, since that format is used everywhere else, and the url is changed `redis+socket://` for the variables used by celery

bookwyrm/settings.py
rrgeorge: I thought about that, but since these urls are constructed out of the other variables, it would be much more complex to assemble the urls for sockets as file paths. Also, this method theoretically would allow for redis over ssl (`rediss://`).
It basically allows for you to alternatively specify a full redis url instead of select parts.

alopix:
What a coincidence that we started work on this 1/2h apart. 🤣

Any indication when this will make it into main, @mouse-reeve?

mouse-reeve:
This week! I will try to test tonight or tomorrow morning

PR #2608, opened 2023-01-17
Enable RSS autodiscovery on user profile

kvibber:
This is a simple change that adds a <link> tag to the header of the user profile so that browsers and feed readers can auto-detect the feed location, making it easier to follow a user by RSS.

### DIFF
--- a/bookwyrm/templates/user/user.html
+++ b/bookwyrm/templates/user/user.html
@@ -5,7 +5,7 @@
 {% block title %}{{ user.display_name }}{% endblock %}
 
 {% block head_links %}
-    <link rel="alternate" type="application/rss+xml" href="{{ user.local_path }}/rss" title="{{ user.display_name }}" />
+    <link rel="alternate" type="application/rss+xml" href="{{ user.local_path }}/rss" title="{{ user.display_name }} - {{ site.name }}" />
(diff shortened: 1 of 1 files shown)


bookwyrm/templates/user/user.html
```diff
@@ -4,6 +4,10 @@
 
 {% block title %}{{ user.display_name }}{% endblock %}
 
+{% block head_links %}
+    <link rel="alternate" type="application/rss+xml" href="{{ user.local_path }}/rss" title="{{ user.display_name }}" />
```
hughrun, line 8: The bare user display name could be a little terse for subscribers. Consider including the instance name. To make this internationalised, you can do something like:

```python
{% translate "on" as on %}
...
{% block head_links %}
    <link rel="alternate" type="application/rss+xml" href="{{ user.local_path }}/rss" title="{{ user.display_name }} {{ on }} {{ site.name }}" />


```

hughrun (commented):
This looks really great! Just a small suggestion to add some context for subscribers when they see the feed in their reader.

hughrun:
@mouse-reeve this has been manually tested in dev and looks good to me. Needs workflow approval and assuming it passes, can be pulled.

Issue #2556, opened 2022-12-29
RSS Feed only for reviews


jaschaurbach:
I would love to get an RSS-Feed with only the _reviews_ of a user.

Giebisch:
I like this idea and am willing to work on it. How do you envision this being displayed on the profile, a second link to an RSS feed right next to the other?

chdorner:
Might be interesting to make this more generic and have different rss feeds for the different content types. I'm sure someone else will want a quotes feed or a comments feed fairly soon 😊

jaschaurbach:
> I like this idea and am willing to work on it. How do you envision this being displayed on the profile, a second link to an RSS feed right next to the other?

I can imagine the RSS logo in a dropdown. if you just click the logo you get the full feed, in the dropdown you chose which "subfeed" you want (incorporates the remark of @chdorner).

But in the end I am no UI-person and do not care as in "I have no fracking idea"

Giebisch:
> if you just click the logo you get the full feed, in the dropdown you chose which "subfeed" you want

Yes, I have already thought of that as well. I'll play around with it for a bit, let's see what I can come up with. If there is a UI/UX professional, I am of course open for help.

Giebisch:
Yea I think a dropdown works quite well. What do you think ? :)

![rss-feeds](https://user-images.githubusercontent.com/39874595/211503693-51320169-0698-441c-9cbd-59da7933b2ea.PNG)

mouse-reeve:
I like the drop-down a lot!

Giebisch:
I'm glad you like it. Copying to the clipboard would imo be the neatest solution. But I understand that BookWyrm generally tries to use as little JavaScript as possible, so I would be fine with regular links. Maybe in the future we can make a case distinction for users with and without JS.

mouse-reeve:
If the HTML element that the javascript is applied to is a regular link, both use cases will be supported, because if javascript is unavailable, the link will work as written. So it shouldn't be in conflict at all, just a note on implementation :)

hughrun:
Merged, thanks @Giebisch!

PR #2610, opened 2023-01-17
Add RSS feed for reviews, quotes, comments only

Giebisch:
#2556 

Added dropdown to choose rss feed and implemented them

rss-feeds

ToDo:

- [x] Add tests


bookwyrm/views/rss_feed.py
```diff
@@ -42,3 +42,114 @@ def items(self, obj):
...
+        """link to the user's profile"""
+        return obj.local_path
+
+    def title(self, obj):
+        """title of the rss feed entry"""
+        return _(f"Reviews from {obj.display_name}")
+
+    def items(self, obj):
+        """the user's activity feed"""
+        return obj.status_set.select_subclasses("review").filter(
```
hughrun, line 75: `select_subclasses("review")` doesn't do what one might expect here. Instead of filtering out everything that isn't a review, it returns all reviews as `Review` objects and everything else as the parent class (`Status`). This means we still get all other public items in the feed, but with only the username as the title, and the URL, without a description.

To filter this to only reviews (or quotations, etc) query the subclass directly instead of using a reverse lookup:

```python
return Review.objects.filter(
    user=obj,
    privacy__in=["public", "unlisted"],
)[:10]
```

hughrun (changes requested):
This looks great, there's just a problem with the queries as noted. I can't work out a way to filter properly with a reverse lookup but you should be able to do it directly on the relevant subclass, especially given we know what subclass we want to query.

bookwyrm/views/rss_feed.py
Giebisch: I just skimmed through the documentation, but yea it looks like I actually made a mistake. Strange how that one works. Well spotted, thanks! I will update it tomorrow

hughrun (approved):
Looks great!

PR #2613, opened 2023-01-20
Enable communication with "authorized_fetch" Mastodon servers

### LINKED ISSUE #1179: Cannot interact with mastodon instances that have activated 'secure mode'
**Describe the bug**
Accounts from instances activated secure mode cannot subscribe to any bookwyrm account. They can interact (comment and favourite) a bookwyrm post if that post is forwarded to them from other mastodon instances, but the bookwyrm account will not receive any notifications about these intereactions.

**To Reproduce**
1. Go to a mastodon instance with secure mode on
2. Subscribe to a bookwyrm account
3. Publish something with the bookwyrm account
4. The publication does not appear on the mastodon side
5. The subscription does not appear on the bookwyrm side
6. Boost a bookwyrm post with another mastodon account that is followed by the "secure mode account"
7. Comment and boost the bookwyrm post with the "secure mode account"
8. Nothing is shown on the bookwyrm stream and no notification

**Expected behavior**
Stream and notifications should be updated on both sides when intereacting.

**Remark**
I have tested this on two bookwyrm instances. Both have this problem.

hughrun:
This builds on the initial work of @renatolond in #427 and resolves a number of bugs preventing that PR moving forward.

In short, this PR should allow communication between Bookwyrm instances and Mastodon instances that use "secure mode" aka `AUTHORIZED_FETCH`.

I have run tests and also tested this manually against accounts running on both "secure mode" and non "secure mode" Mastodon instances, as well as between 2 Bookwyrm instances, and it appears to work, but I definitely want at least one other person to test it independently in case I missed something.

## New settings

There is a new value in `settings.py`:

* `INSTANCE_ACTOR_USERNAME` - defaults to ~~the instance domain name~~ `bookwyrm.instance.actor`, so that the instance actor (required to sign requests to "secure mode" Mastodon servers) will have a username like `@bookwyrm.instance.actor@example.net`
~~* `INSTANCE_ACTOR_EMAIL` - defaults to `representative@DOMAIN` so that instance actors will have a registered email like `representative@example.net`~~

The new instance actor will have an email address registered as `bookwyrm@localhost`.

resolves #1179 
closes #427


renatolond:
Thanks for picking this up 🙏🏼

bookwyrm/settings.py
```diff
@@ -370,3 +370,8 @@
 HTTP_X_FORWARDED_PROTO = env.bool("SECURE_PROXY_SSL_HEADER", False)
 if HTTP_X_FORWARDED_PROTO:
     SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
+# AUTHORIZED_FETCH Instance Actor
+# WARNING this must both be unique - not used by any other user
+INSTANCE_ACTOR_USERNAME = DOMAIN
+INSTANCE_ACTOR_EMAIL = f"representative@{DOMAIN}"
```
jaschaurbach, line 377: I would suggest a total random string instead of `representative`.

My thoughts behind that suggestion:
An admin updates the production version with `git pull` and there is already a user with the name `representative` the name is not unique any more. I know, one should read the update notes before updating but in this case the admin would need toward around this. This breaks the communicated update process and is a pitfall for those who are not reading patch notes. 

If BookWyrm introduces a new way to update an instance there should be a communication within the instance for admins and a documentation on how to perform the update.

renatolond: I think this used to be `DOMAIN@DOMAIN` in mastodon and now is `mastodon.internal`: https://github.com/mastodon/mastodon/blob/main/app/models/concerns/account_finder_concern.rb#L20
I think something like `bookwyrm.internal` should be reserved in any case, but indeed representative might be a common-enough name

hughrun: As currently written the username is DOMAIN@DOMAIN - it's the email address (which presumably will never be used) that is representative@DOMAIN.

I'm not really fussed what we use, other than agreeing it needs to be unlikely to be currently in use. the suggestion of `bookwyrm.internal@DOMAIN` seems good.

It also occurred to me when working on this that we should give admins the ability to reserve a bunch of usernames they don't want people to be able to register  (e.g. "admin" or "support" or whatever they want). I'll make a separate Issue for that.

jaschaurbach (approved):
This looks very good to me.

I recommend to add `INSTANCE_ACTOR_USERNAME ` to the documentation that admins know what this does, that they _could_ change it but there is most time no need to change it.

hughrun:
@mouse-reeve if you're happy with @jaschaurbach's review, then this is ready to go. The only change I made since that review was to add a more verbose note in the `settings.py` file.

mouse-reeve:
Seems to work to me!

Issue #644, opened 2021-02-24
Group books by series


tastytea:
**Is your feature request related to a problem? Please describe.**
When I see a book I might like and notice it has a series after the title, I want to find out what the other books of the series are.

**Describe the solution you'd like**
I'd like to be able to click on the series name and get an ordered list of the complete series.

**Describe alternatives you've considered**
Looking the book up on GoodReads, but that's tedious.

mouse-reeve:
yes! right now series is stored a list of strings in the database on each book, but I think it should be converted into it's own database table, so that it can be a proper relationship between books and can be edited by users

PR #2616, opened 2023-01-24
Book Series v1: a very simple version to link to book series by name

alopix:
This is a first pass at listing books in a series. In order not to touch the DB (yet) and have a book series link, this just matches the series name.
A future version would do what was suggested in #644.

Screenshots of what the change looks like:
image
image

The URLs would be `/series/by/AUTHOR_ID?series_name=Url%20Encoded%20Series%20Name` in order to easily distinguish them from a future version.

This is obviously a quick win solution.


bookwyrm/views/books/series.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if is_api_request(request):
+            pass
+
+        author = get_object_or_404(models.Author, id=author_id)
+
+        results = models.Edition.objects.filter(authors=author, series=series_name)
+
+        # when there are multiple editions of the same work, pick the closest
+        editions_of_work = results.values_list("parent_work__id", flat=True).distinct()
```
hughrun, line 26: If there is an edition with a series number, but a "closer" one without a series number, we should ideally prefer to use the one with the number.

bookwyrm/views/books/series.py
```diff
@@ -0,0 +1,60 @@
...
+            )
+            if result.series_number:
+                numbered_books.append(result)
+            elif result.first_published_date or result.published_date:
+                dated_books.append(result)
+            else:
+                unsortable_books.append(result)
+
+        list_results = (
+            sorted(numbered_books, key=lambda book: book.series_number)
```
hughrun, line 44: This is sorting by series number _alphabetically_, so the tenth item in the series will be listed prior to the second. Coercing it to an `int` should fix it:
```python
sorted(numbered_books, key=lambda book: int(book.series_number))
```
You may need to do something similar with `book.first_published_date` below.

hughrun (changes requested):
This is cool!

Just a couple of things I've noted that aren't quite right, but otherwise it looks great. I haven't looked into how you'd prefer and edition with a series number but if you're not sure how to do it I'm happy to help investigate.

bookwyrm/views/books/series.py
alopix: Shouldn’t the published dates be okey since they are stored as `DateTimeField` so I would expect comparison to automatically work for dates?!

hughrun: I think you're right, but I didn't check it thoroughly.

joachimesque: I haven't checked the code, just jumping into the conversation to say that series numbers are not always `int`s. A `float` type might be more relevant. E.g.: https://lire.boitam.eu/book/30636/s/as-yet-unsent

Also, sometimes series are not published in the right order (Star Wars, anyone?), and you get works numbered as 0.1, 0.2, etc. E.g.:  or https://www.goodreads.com/series/229503-the-locked-tomb where 1 was published in 2019 and 0.5 in 2020. Sorting by published dates won't necessarily be the same as sorting by series order.

(Apart from that it looks good, thank you for the PR :))

hughrun: Oh, good thinking re floats. Sorting by date is only a backup best-guess for when a book has a series listed but not a series number: a listed series number will take precedence.

alopix: I actually read lots of series with `.5` ratings so no idea why I thought int would be ok haha

As hughrun said, the published date is just a backup (suggested on element) if the series number was missing.

bookwyrm/views/books/series.py
alopix: The latest change should sort the editions with a series number first, then the top ranking edition

bookwyrm/views/books/series.py
```diff
@@ -0,0 +1,66 @@
...
+            )
+            if result.series_number:
+                numbered_books.append(result)
+            elif result.first_published_date or result.published_date:
+                dated_books.append(result)
+            else:
+                unsortable_books.append(result)
+
+        list_results = (
+            sorted(numbered_books, key=lambda book: float(book.series_number))
```
hughrun, line 47: I really want to approve this PR, but we have a bit of a problem here because `book.series_number` is a `CharField`, when it arguably should be an `IntegerField` or maybe a `DecimalField`. This means the form input will take any string, and if anyone makes the series number something like "Three" or "Book 6" with this sort it will throw a `ValueError` because it can't be coerced into a float.

This can be solved, but you'll have to give up on using a `lambda` and use named function instead so we can handle the exception. e.g. you could do something like this:

```python
class BookSeriesBy(View):

...

     list_results = (
            sorted(numbered_books, key=self.sort_by_series)
            + sorted(
                dated_book
...

    def sort_by_series(self, book):
        try:
            return float(book.series_number)
        except ValueError:
            return 999999999

```

hughrun (changes requested):
This is **so close**!

It just needs a fix to handle `ValueError`s when the book number can't be turned into a float. The rest looks pretty good to me.

bookwyrm/views/books/series.py
alopix: Would using `sys.float_info.max` make more sense instead of a magic number?

hughrun: Yes! 😄

alopix: That should be fixed now as well 😊

hughrun (approved):
Looks good!

Issue #2614, opened 2023-01-20
Feed not updating with any content besides on-instance follows


gqqjhffr:
**Describe the bug**
No federated content is appearing in my feed (most of my follows being on Mastodon).

**To Reproduce**
1. Go to Bookwyrm homepage or "feed" page while logged in and following off-instance accounts.
2. Scroll down to feed.
4. Be disappointed at the lack of a flood of federated goodness.

**Expected behavior**
I expect to see content from the 100+ accounts I've followed across the Fediverse.

**Instance**
bookwyrm.social

**Additional context**
Federated posts stopped appearing in my feed after Jan 3.

---

**Desktop (please complete the following information):**
    - OS: Windows 10
    - Browser Firefox
    - Version 109.0

hughrun:
Can you explain more? When you say "most of my follows are on Mastodon" do you mean you are following Mastodon accounts? If so there are likely 2 explanations for what you're experiencing:

1. `bookwyrm.social`'s Celery task manager fell over for a little while and there is a large backlog of tasks chugging through (e.g. I just saw an update that happened 2 weeks ago - it's getting there but will take a while to catch up)
2. The only content from non-Bookwyrm fediverse applications you will see on Bookwyrm is interactions with your posts – e.g. likes, boost, replies, and DMs. Bookwyrm is not a general-purpose social media application - the feed includes the aforementioned interactions, plus ActivityPub activities that are only currently used by Bookwyrm: `Readthrough` activity, `Review`s etc. It doesn't include `Notes` (e.g. Mastodon toots), `Articles` etc. There's nothing stopping Bookwyrm's activity types being used from other software, but they need to support those types.

solarbirdy:
Is this also why my attempt to follow my wife's new bookwyrm.social account from my mastodon account keeps turning into a "Pending approval"?

Also, she turned that off so it shouldn't require approval but is it just really lagged and that's why?

hughrun:
That might be to do with "secure mode" on the Mastodon server, we may need to some more testing to be sure.

arjanboltjes:
[...]

Issue #2545, opened 2022-12-26
Cannot follow my new account from Mastodon


xaviershay:
I'm seeing the exact same issue as #1928 using https://mas.to (of which I am just a user, not an admin), but I suspect might be a different cause.

**Describe the bug**

Following my account with my mas.to account leads to a pending approval request. My account is not set to moderate follow requests, nor do I see any request in my account profile.

**To Reproduce**

* Try to follow @xavier@bookwyrm.social from [mas.to](https://mas.to/)

**Expected behavior**

Account is followed with no pending approval.

**Instance**

bookwork.social

**Additional context**

* Following [@followTest@wyrms.de](http://wyrms.de/user/followTest) works as expected.
* My eyes don't detect any significant differences between [my JSON](https://bookwyrm.social/user/xavier.json) and followTest.
  
Sorry I'm too new to this whole thing to know how debug more effectively, but am happy to chase down leads. I don't _really_ need this feature, I was mostly curious whether it would be a good option for publishing a feed to mastodon.

Thanks for all your work 🙏

weijia-cheng:
I am also having this problem trying to follow `@weijia_sophia_cheng@bookwyrm.social` from `@weijia_sophia_cheng@deacon.social`. Same issue with the pending approval request.

weijia-cheng:
What I see right after I click "follow" from `deacon.social`:

![Screenshot from 2022-12-26 09-20-13](https://user-images.githubusercontent.com/1240463/209571264-6a31e3df-5c76-4eb5-9656-a22e81b4dd2c.png)

What happens shortly after I refresh the page:

![Screenshot from 2022-12-26 09-20-23](https://user-images.githubusercontent.com/1240463/209571302-c7907882-0aa9-4726-913f-57d520c9bab1.png)

xaviershay:
Confirming that those screenshots match my experience, and I get the same behaviour attempting to follow `@weijia_sophia_cheng@bookwyrm.social`.

marcusyoung:
In my experience this will resolve itself in a day or two. It appears to be caused by a backlog of tasks on the bookwyrm.social instance.

weijia-cheng:
@marcusyoung You're right, it did resolve itself a little afterwards.

fasiha:
Hmm, the issue hasn’t resolved for me for several weeks, but will try to toggle “withdraw” and “follow” again now and check in a couple of days.

quisquous:
[...]

PR #2617, opened 2023-01-24
Make follow activities a high priority

mouse-reeve:
This should go a long way towards fixing the problems with follows not going through to remote servers. All it does is move relationship related activities from the medium priority queue, which gets backlogged easily, to the high priority queue, which is less backlogged.

The risk here is that the high priority queue could end up getting backlogged, so this isn't the last word on fixing this, but I think the volume of activities that this will add to it will be manageable.

Works on #2614 #2545


hughrun (approved):
Looks good to me, I've run this in dev and can't see any problems.

PR #2623, opened 2023-01-26
Fix passwordless redis connection string

chdorner:
#2604 has introduced an issue where the `REDIS_URL` will have `None` set as the password if the `REDIS_ACTIVITY_PASSWORD` does not exist. The solution is to use an empty string for the default password value instead of `None`

Additionally, I thought it might be worth doing the same `requests.util.quote`ing for both the activity and broker redis passwords.


mouse-reeve:
Thanks so much for fixing this promptly. Tangentially, would you be interested in doing code review for PRs like that one?

chdorner:
@mouse-reeve definitely!

Issue #2493, opened 2022-12-11
Let Admins manually confirm Email from frontend


jaschaurbach:
With #2492 admins can manually confirm the users email from the CLI.

Add same function into frontend

(this is a reminder for myself to do this soonish :D)

PR #2627, opened 2023-01-27
Button in admin view to manually activate pending user

alopix:
Resolves https://github.com/bookwyrm-social/bookwyrm/issues/2493

This PR adds a simple button to the admin user interface that allows quickly activating a pending user when they have not confirmed their email address yet.
No need to use the CLI anymore.

### DIFF
--- a/bookwyrm/templates/settings/users/user_moderation_actions.html
+++ b/bookwyrm/templates/settings/users/user_moderation_actions.html
@@ -15,6 +15,12 @@
             </p>
             {% endif %}
 
+            {% if not user.is_active and user.deactivation_reason == "pending" %}
+            <form name="suspend" method="post" action="{% url 'settings-activate-user' user.id  %}" class="mr-1">
+                {% csrf_token %}
+                <button type="submit" class="button is-success is-light">{% trans "Activate user" %}</button>
+            </form>
+            {% endif %}
             {% if user.is_active or user.deactivation_reason == "pending" %}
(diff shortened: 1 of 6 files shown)


bookwyrm/templates/settings/users/user_moderation_actions.html
```diff
@@ -15,6 +15,12 @@ <h3>{% trans "User Actions" %}</h3>
             </p>
             {% endif %}
 
+            {% if not user.is_active and user.deactivation_reason == "pending" %}
+            <form name="suspend" method="post" action="{% url 'settings-activate-user' user.id  %}" class="mr-1">
```
mouse-reeve, line 19: this form should probably be named "activate" rather than "suspend"

bookwyrm/views/admin/user_admin.py
```diff
@@ -95,3 +95,18 @@ def post(self, request, user):
...
+
+@method_decorator(login_required, name="dispatch")
+@method_decorator(
+    permission_required("bookwyrm.moderate_user", raise_exception=True),
+    name="dispatch",
+)
+class ActivateUserAdmin(View):
+    """activate a user manually"""
+
+    def post(self, request, user):
```
mouse-reeve, line 108: You can disable the pylint error by adding `# pylint: disable=unused-argument` above this line.

mouse-reeve:
Perfect! Thank you so much. Just for future reference, it's generally better not to do the translations in a pull request, since it can lead to complicated merge conflicts. Instead, the translations can be updated after the merge.

PR #2628, opened 2023-01-27
Improve feed header for federated generated notes

chdorner:
Partially addresses #2531.

Roughly taking #2532 but applying it to the feed templates.

This works for wants/started/finished reading generated notes, but not yet for reading goal generated notes (i.e. `<p>set a goal to read 12 books in 2023</p>`).

This does feel fairly brittle though and I wonder if we should rethink how statuses are federated (#2398), maybe making a distinction between federating to BookWyrm instances vs other ActivityPub services?


mouse-reeve:
Thank you!

PR #2629, opened 2023-01-27
Minor performance improvements

0x29a:
- Where it really saves a DB request, this PR changes `len(queryset)` to `queryset.count()`, so calculations done at a database level.
- Also, changes `model.id` to `model_id` to avoid fetching all `model`'s fields from the DB.

### DIFF
--- a/bookwyrm/models/annual_goal.py
+++ b/bookwyrm/models/annual_goal.py
@@ -52,7 +52,7 @@ class AnnualGoal(BookWyrmModel):
             user=self.user,
             book__in=book_ids,
         )
-        return {r.book.id: r.rating for r in reviews}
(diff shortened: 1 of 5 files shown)


joachimesque (approved):
Looks good to me! I coded the annual summary page without knowing too much of Django, so I guess that a lot more could be done to improve the performance 😓

0x29a:
@joachimesque, thanks for reviewing and approving. I'm not sure if these optimizations improve overall speed visibly to be honest. Initially, I was going to search for more redundant requests and include fixes there, but it's better to do this in a separate PR, I'm removing `[WIP]` from the title.

joachimesque:
My bad, I hadn't seen the `[WIP]` 🤦

Thank you for your contributions!

Issue #2625, opened 2023-01-26
Pluralization error 


jaschaurbach:
**Describe the bug**
When I set the import limit to 200 every 1 days the string for the user looks like the following:

<img width="517" alt="Bildschirm­foto 2023-01-26 um 18 28 11" src="https://user-images.githubusercontent.com/2017105/214906784-2342bec0-3b52-4b5f-aac3-973a7500d090.png">

**Expected behavior**
I expect to read

`Currently you are allowed to import 200 books every day.`

**Instance**
All since 0.5.4.

marcusyoung:
The easiest fix is just to change the text "days" to "day(s)" as the number of days is configurable.

mouse-reeve:
It wouldn't be too much harder to use a `blocktrans` element that handles both singular and plural, which would be more consistent with the rest of the app

jaschaurbach:
I agree with @mouse-reeve - it should be consistent through the whole app

chdorner:
Just a note that this applies to both books and days:
<img width="474" alt="image" src="https://user-images.githubusercontent.com/112063/215153448-ce31d078-96e9-4783-9474-81b8a6330308.png">

PR #2631, opened 2023-01-28
fixed singularisation/pluralisation

rritik772:
edited two files

### DIFF
--- a/bookwyrm/templates/import/import.html
+++ b/bookwyrm/templates/import/import.html
@@ -19,10 +19,14 @@
             <div class="notification">
                 <p>
                     {% blocktrans count books=import_size_limit%}
-                        Currently you are allowed to import {{ import_size_limit }} {{books}}
+                        Currently you are allowed to import {{ import_size_limit }} book
+                    {% plural %}
+                        Currently you are allowed to import {{ import_size_limit }} books
                     {% endblocktrans %}
                     {% blocktrans count days=import_limit_reset%}
-                         every {{ import_limit_reset }} {{days}}.
+                         every {{ import_limit_reset }} day.
+                    {% plural %}
+                         every {{ import_limit_reset }} days.
                     {% endblocktrans %}
                 </p>
                 <p>{% blocktrans %}You have {{ allowed_imports }} left.{% endblocktrans %}</p>
--- a/bookwyrm/templates/import/import.html
+++ b/bookwyrm/templates/import/import.html
@@ -17,7 +17,7 @@
     {% if site.imports_enabled %}
         {% if import_size_limit and import_limit_reset %}
             <div class="notification">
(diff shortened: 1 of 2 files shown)


rritik772:
this PR solved the #2625 error. Have a look through it.

joachimesque (commented):
Hi, thank you for your contribution!

I’m not 100% up to date with internationalization and Django templates, but I think that using this code won’t allow for the text to be translated in the right way. 

Django templates use a `count` option in its `blocktrans` tag, to specify a plural form:

```
{% blocktrans count count=customers.count %}
  One customer
{% plural %}
  {{ count }} customers
{% endblocktrans %} 
```

Because Django uses `gettext` and `ngettext` for translation, pluralization can only be applied to one variable ([source on stackoverflow](https://stackoverflow.com/a/5244999)). So basically the choice would be either:

A. split the text into multiple strings that are translated apart from each other:

> Current book import is limited to:
> - 50 books
> - every 3 days

B. keep it in one sentence but only pluralize one variable. That solution would work best if we choose the daily limit (IMO an import size limit of 1 will be less used than a reset time of one day).

```
{% blocktrans count days=import_limit_reset %}
  Currently you are allowed to import {{ import_size_limit }} books every {{ days }} day.
{% plural %}
  Currently you are allowed to import {{ import_size_limit }} books every {{ cays }} days.
{% endblocktrans %}
```

mouse-reeve:
@joachimesque is correct, the way this is written unfortunately will not work. I think what makes this complicated is that there are two sources of plurals (book or books, and day or days), but personally I think it would be a reasonable workaround to assume book will be plural because a case where you're only allowed to import 1 book a day seems like a real edge case.

rritik772:
I have edited the file as per Feasibility.

I have an another solution to this problem if you all dont like the solution. We can just replace 
`                <p>{% blocktrans %}Currently you are allowed to import {{ import_size_limit }} books every {{ import_limit_reset }} days.{% endblocktrans %}</p>`
to
`                <p>{% blocktrans %}Currently you are allowed to import {{ import_size_limit }} book/s every {{ import_limit_reset }} day/s.{% endblocktrans %}</p>
`

mouse-reeve:
I like your proposed alternative, but I think it should just assume there are multiple books, so rather than "book/s", just "books"

rritik772:
Ok, That's also reasonable.

joachimesque (commented):
Before validating this PR I'd love to have the opinion of translators for various languages for the current solution. Separating the two parts of the sentence might not work universally, though I see no problem with French:

> Actuellement vous pouvez importer un livre
> Actuellement vous pouvez importer {{ import_size_limit }} livres

> par jour.
> tous les {{ import_limit_reset }} jours.

mouse-reeve:
I think it's better to keep whole sentences together, rather than dividing them by internal clauses; it's just hard to predict problems that it could cause in different translations.

mouse-reeve:
I went ahead and did the last little bit of work on this in #2904

PR #2636, opened 2023-01-29
Remove redundant DB queries

0x29a:
This PR downs the number of DB requests from 46 to 38 for the feed page, assuming you have 3 books added to each shelf (`to-read`, `reading`, `read`). Also, it fixes one duplicated request from the book page.

Testing steps:
1. Switch to the `main` branch.
2. Cherry-pick ee3c29f9a269b8aa7dbd218be895e59b1fad3270 and fd98aa34a6ff2ae7325bea64af2463db5d5b09eb
3. Add three test books, each to it's own shelf.
4. Note the number of DB requests.
5. Rebase on `redundant-db-queries`.
6. There should be 8 less DB requests now.
7. Test that book suggestions block works well.
8. Test that book page works well too.

### DIFF
--- a/bookwyrm/templates/book/book.html
+++ b/bookwyrm/templates/book/book.html
@@ -215,10 +215,10 @@
                 {% endif %}
 
 
-                {% with work=book.parent_work %}
+                {% with work=book.parent_work editions_count=book.parent_work.editions.count %}
                 <p>
                     <a href="{{ work.local_path }}/editions" id="tour-other-editions-link">
-                        {% blocktrans trimmed count counter=work.editions.count with count=work.editions.count|intcomma %}
+                        {% blocktrans trimmed count counter=editions_count with count=editions_count|intcomma %}
                             {{ count }} edition
                             {% plural %}
                             {{ count }} editions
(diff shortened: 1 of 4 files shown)


chdorner (changes requested):
These performance improvements are great, thank you so much!

There's probably a small mistake in the logic in `get_suggested_books`, it behaves differently now when changing up the number of books on a user's shelves.

It works great with 1 book per shelf as described in the PR description, but trying to shake things up a bit I tested with 5 to-read, 2 reading, and 3 read.

On `main` I get (in order):
* 2x reading
* 2x read
* 1x to-read

on this PR I get (also in order):
* 5x to-read
(no books from reading or read)

0x29a:
Ah, right, `user_shelves` seem to be losing the original ordering. I should have tested this on a bigger set of books. :disappointed:  Thank you for the review and catching this, @chdorner, I'll fix this sometime tomorrow.

0x29a:
@chdorner, should be fixed now.

chdorner (approved):
Thank you so much for these performance improvements! 🚤 

The book suggestions are now following the same shelf ordering as on `main`, but with less queries :raised_hands:

0x29a:
Thanks for the review, @chdorner!

PR #2637, opened 2023-01-30
UI: larger dropdown size for mobile devices

alopix:
Currently the dropdown options (popup) are tiny when viewed on a phone which makes it really difficult to click.

This PR changes the size for mobile devices so the full-screen popup has larger options.

| Before | After |
| --- | --- |
| IMG_5886 | IMG_5887 |

The dropdowns I have identified are:
- front page additional/more actions on posts
- front page currently reading list shelf actions
- my books shelf actions
- user profile actions

### DIFF
--- a/bookwyrm/static/css/bookwyrm/overrides/_bulma_overrides.scss
+++ b/bookwyrm/static/css/bookwyrm/overrides/_bulma_overrides.scss
@@ -1,111 +1,117 @@
 .summary-on-open {
-    display: none;
+	display: none;
 }
 
 @media only screen and (max-width: 768px) {
-    .navbar-menu {
-        text-align: right;
-        padding-right: 1rem;
+	.navbar-menu {
+		text-align: right;
+		padding-right: 1rem;
 
-        .tags {
-            justify-content: flex-end;
-        }
+		.tags {
+			justify-content: flex-end;
+		}
 
-        #navbar-dropdown {
-            &[open] {
-                .summary-on-open {
-                    display: initial;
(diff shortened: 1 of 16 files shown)


joachimesque (approved):
Thanks for your contribution!

I'm wondering if that specific override could have been done by editing `bookwyrm/static/css/bookwyrm/overrides/_bulma_overrides.scss`. That way, you would not need to list all the affected dropdowns, but targeting `.dropdown .button` in a mobile media query would do that job. What do you think?

The code can be merged as-is if the CSS override technique doesn't work.

alopix:
> Thanks for your contribution!
> 
> I'm wondering if that specific override could have been done by editing `bookwyrm/static/css/bookwyrm/overrides/_bulma_overrides.scss`. That way, you would not need to list all the affected dropdowns, but targeting `.dropdown .button` in a mobile media query would do that job. What do you think?
> 
> The code can be merged as-is if the CSS override technique doesn't work.

@joachimesque I was actually wondering if using a global CSS override would work as well, but I was afraid it could affect places where it is not desired. But happy to try the CSS solution

alopix:
I updated the PR with CSS instead of classes. I kept the other minor fixe:
* one to `bookwyrm/templates/snippets/report_button.html` not supporting the `class` attribute

And stylelint found the comment issue which is now also fixed

genebean:
This will be nice to have… I was just lamenting the small buttons

mouse-reeve:
Thank you! This looks really nice

Issue #2609, opened 2023-01-17
A page range should be possible for a quote


iGormilhit:
**Is your feature request related to a problem? Please describe.**
When adding a quote to a book, it's possible to set at which page, or which percentage, the quote is situated. And for the page solution, only one integer is allowed. But, quite often, a quote is on several pages.

**Describe the solution you'd like**
It should be possible to set a page range for a quote.

Giebisch:
I'll work on it :)

Giebisch:
Any ideas to make it clearer to the user that you can also enter a range here ?

![page_range](https://user-images.githubusercontent.com/39874595/214946322-abf3301d-363b-4142-b97b-1f44bdd6c79f.PNG)

mouse-reeve:
You can use a span with the class `button is-static` to add an a text element between the two number fields:
<img width="551" alt="Screen Shot 2023-02-20 at 1 19 55 PM" src="https://user-images.githubusercontent.com/1807695/220200155-20e4c873-6e07-470a-9895-e3187c365a4e.png">

PR #2638, opened 2023-01-30
Page Range

Giebisch:
Implements #2609 

**Input for page range:**

page_range_1

**List quotes:**

page_range_quotes

Still missing:

- [x] Tests


hughrun:
I can take a look at this in a few days, but before I do - @mouse-reeve any thoughts about this in relation to #2640? What do we want to do?

mouse-reeve:
Ah, I hadn't connected these two in my head! Inconclusive thoughts:
 - Having a formal page range meets a common use cases and would make the display consistent
 - Free text fields could contain page ranges, which might cause display weirdness
 - Having free text and page range would also allow things like "VI - VIII", which would be good to support

I guess having written that out I'm inclined to think that we should have a start and end, and make them both be free text fields?

Giebisch:
I think the current implementation probably covers the vast majority of cases. But we could perhaps create a new issue to request a third option for free texts besides "page" and "percent".

hughrun (changes requested):
This is good, and I was scope-creeping earlier so I agree let's pull this in and deal with whether they should be `int` or a string separately.

See my comment: it would be good to ignore the second page number if it's the same as the first page. 

The other thing I noticed is that if the user does not enter an initial page, the last page is ignored and it just posts without page numbers rather than throwing an error. I think that might be a better UX to be honest, but if you'd prefer to indicate to the user that they've made a mistake, or alternatively to treat the `endposition` in that case as a single page (i.e. treat it as the `position`) that will need to be dealt with.

Giebisch:
Yep, that sounds like a good addition. From now on, the page will only be displayed once if start and end position are the same. I took the [suggestion](https://github.com/bookwyrm-social/bookwyrm/issues/2609#issuecomment-1437561065) with the design, so the input should be more understandable for the user. So I think we don't need any further distinction of the cases here.

hughrun (approved):
Looks good 😃

PR #2644, opened 2023-02-01
Add Content-Security-Policy headers and secure cookies

rrgeorge:
Add Content-Security-Policy headers using django-csp
And set set cookies to secure when using https


mouse-reeve:
Sorry about that! It looks like a new release of Black came out with changes, and it flagged some formatting in the codebase; I should really pin a version on it. Don't worry about it in the context of this PR

chdorner:
#2645 has a fix for the failing `black` checks.

chdorner (changes requested):
This looks great! We probably need to tweak the policy slightly, for example in the search results for books the covers are loaded directly from the connecting site like OpenLibrary or another BookWyrm instance. What do you think of allowing images from any source with `img-src *`?

I'm not sure if there are other examples where we legitimately load external content other than images.

rrgeorge:
> What do you think of allowing images from any source with `img-src *`?

I'm reluctant to add a wholesale exception for images, but I did put img-src * for the search view specifically. If there are any other views that need an exception, let me know.

chdorner (approved):
Only allowing all image sources on the search view sounds like a good solution 👍 

@mouse-reeve I'll leave the merge up to you, as far as I can tell the search results are the only place where we could load external content. But you might have other places in mind too!

mouse-reeve:
Search is the only place I can think of too

PR #2645, opened 2023-02-02
Pin black versions

chdorner:
This pins the black version of GitHub Actions and upgrades the development environment to the same 22.12.0.

Mainly due to `black` changing the stable style in their [newest 23.1.0 release](https://github.com/psf/black/releases/tag/23.1.0).

PR #2646, opened 2023-02-02
Improve search results

0x29a:
Search query has to be encoded. Otherwise, a query like this:
```
https://openlibrary.org/search.json?q=Rang%20&%20Dale%27s%20Pharmacology=&limit=10
```
Is equivalent to this:
```
https://openlibrary.org/search.json?q=Rang%20
```
Because the ampersand isn't encoded.

Consider search results for `Rang & Dale's Pharmacology` query without this fix:

image

And with the fix:

image

You can see that Inventaire returned more accurate results. Also, OpenLibrary returned relevant results.

### DIFF
--- a/bookwyrm/connectors/abstract_connector.py
+++ b/bookwyrm/connectors/abstract_connector.py
@@ -1,5 +1,6 @@
 """ functionality outline for a book data connector """
 from abc import ABC, abstractmethod
+from urllib.parse import quote_plus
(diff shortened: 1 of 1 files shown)


chdorner (approved):
Great find, thank you so much for this contribution!

Don't worry about the failing `black` PR check, that will be fixed soon with #2645

PR #2652, opened 2023-02-04
Always expand content status on single status view

### LINKED ISSUE #2649: Default-expand long reviews when directly linked to
When clicking on a link to a review (from RSS, a fedi client that doesn't show the full content, etc), it has the "show more" button by default. This doesn't really make sense on a page where the entire content of the page is the review — instead the content should be expanded by default.

Not sure whether it's better to still have the "Show less" button at the bottom, or to just remove it entirely and always show the entire content (probably the latter makes more since, since I can't really think of a use case for expanding / collapsing outside of a feed), but either would probably be fine.

([example review that is collapsed by default](https://bookwyrm.social/user/wesleyac/review/918780/s/pirate-enlightenment-or-the-real-libertalia))

chdorner:
Fixes #2649

On the feed view along with other statuses, the body will be trimmed, but on the single view, there's no need to trim it. This preserves the logic for spoiler alerts.

### DIFF
--- a/bookwyrm/templates/feed/status.html
+++ b/bookwyrm/templates/feed/status.html
@@ -30,7 +30,7 @@
         {% endif %}
         {% endfor %}
         <div class="is-main block">
-            {% include 'snippets/status/status.html' with status=status main=True %}
+            {% include 'snippets/status/status.html' with status=status main=True expand=True %}
         </div>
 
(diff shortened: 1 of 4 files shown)


hughrun (approved):
Looks good!

Issue #2656, opened 2023-02-09
2FA code not accepted


chris-y:
**Describe the bug**
I have 2FA set up on my account and was trying to log in but the code was not being accepted.  I've had this issue previously and tried a new code and it worked.  On this occassion I resorted to a backup code.  I disabled and re-enabled 2FA on my ccount but even the initial setup code was not being accepted.

After some head-scratching I realised that the clock on the computer I was using was out by about 1 minute.  I tried some other services which use TOTP codes and they worked OK with the incorrect clock, so I think the window of accepted codes needs to be expanded (ie. accept more codes either side of the current one).

**To Reproduce**
1. Enable 2FA
2. Set computer clock out by 1 minute
3. Try to log on.

**Expected behavior**
Bookwyrm should accept a larger range of codes.

**Instance**
wyrms.de

**Additional context**

I checked pyotp documentation:
> 
> verify(otp: [str](https://docs.python.org/3/library/stdtypes.html#str), for_time: [Optional](https://docs.python.org/3/library/typing.html#typing.Optional)[[datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime)] = None, valid_window: [int](https://docs.python.org/3/library/functions.html#int) = 0) → [bool](https://docs.python.org/3/library/functions.html#bool)[[source]](https://pyauth.github.io/pyotp/_modules/pyotp/totp.html#TOTP.verify)
> Verifies the OTP passed in against the current time OTP.
> 
> Parameters:
> otp – the OTP to check against
> 
> for_time – Time to check OTP at (defaults to now)
> 
> valid_window – extends the validity to this many counter ticks before and after the current one
> 
> Returns:
> True if verification succeeded, False otherwise

It appears that setting valid_window=2 in https://github.com/bookwyrm-social/bookwyrm/blob/e88f1f0b8c2e186c350c137a957c080e012e23a0/bookwyrm/forms/landing.py#L111 should solve this issue.

---

**Desktop (please complete the following information):**
    - OS: Windows 10
    - Browser Edge
    - Version 109.0.1518.78

marcusyoung:
Yeah, I've experienced this issue too, where a code is close to expiry and is rejected by the time I enter it.

chdorner:
Funny coincidence, I've never seen this myself until this issue was opened, now I've already experienced it twice!

hughrun:
This should be fixed in #2657 which defaults to a 60 second grace period but can be adjusted by server admins.

PR #2657, opened 2023-02-09
Expand TOTP validity window

chris-y:
This is a potential fix for #2656
I'm not sure if there are any other instances in the code where TOTP codes are checked, and I'm not in a position to check this fix works.


chris-y:
Seems sensible, I've applied another commit which should do this (again, unfortunately, I'm not in a position to be able to test).

bookwyrm/settings.py
```diff
@@ -369,6 +369,7 @@
 OTEL_SERVICE_NAME = env("OTEL_SERVICE_NAME", None)
 
 TWO_FACTOR_LOGIN_MAX_SECONDS = 60
+TWO_FACTOR_LOGIN_VALIDITY_WINDOW = 2
```
hughrun, line 372: It would be better to pull this value in from `.env` so that it doesn't get wiped out when pulling in the latest changes:
```python
TWO_FACTOR_LOGIN_VALIDITY_WINDOW = env.int("TWO_FACTOR_LOGIN_VALIDITY_WINDOW")
```

And then add `TWO_FACTOR_LOGIN_VALIDITY_WINDOW` to `env.example` with a value of `2`.
This actually also applies to `TWO_FACTOR_LOGIN_MAX_SECONDS` I guess so maybe the same for that would be good.

hughrun (changes requested):
This works well! I've left a suggestion to pull the value in `settings.py` from `.env` so that the value can be retained in the local environment when running updates. Otherwise, it looks good.

bookwyrm/settings.py
chris-y: Done 🙂

chdorner: Mostly out of curiosity, how does this behave when the env file is missing those values? 

Our upgrading instructions don't really mention anything about new env file values, and this has the potential of locking people out of their accounts, including the admins and moderators 😬

If that is the case we could just add some sensible defaults in `settings.py` when reading from the environment, just in case they're not set.

chris-y:
Yeah, I wasn't sure if I needed to add ",default" to the env() call.  Looks
like that might be a good idea?

bookwyrm/settings.py
jaschaurbach: if the .env is missing the value and no default is given it should throw an error the moment you try to create or run migrations, restart bookwyrm, etc.

jaschaurbach:
The value how long a TOTP is valid should be shown somewhere on the admin site. Maybe we need to introduce an overview site somewhere showing values from .env

hughrun:
@jaschaurbach I think that's out of scope for this PR but I agree that would be very useful. Especially for things like this where it's not obvious that each time window is 30 seconds.

hughrun (approved):

PR #2659, opened 2023-02-12
Added support for oauth 2 login and registration

rrgeorge:
I've added support for OAuth 2 login and registration to my instance, so I can login with my mastodon credentials.
Enabling this disables password login.
If an account does not exist in bookwyrm, it will automatically register a new account with the same local part, and requires the user to enter their email address. This is not provided from Mastodon, and I am unable to set it to NULL via Django, so this step is unfortunately unavoidable.
I'm not sure if this is the best way to go about things, but it does work. I am submitting this as a draft pr to see if anyone has any other input or wants to test it out for themselves.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -120,3 +120,11 @@ OTEL_SERVICE_NAME=
 # for your instance:
 # https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header
 HTTP_X_FORWARDED_PROTO=false
+
+# Enable logging in and registration with OAuth 2
+OAUTH_ACTIVE=false
+#OAUTH_NAME="OAuth Provider" # Displayed on Login Button as "Login with OAUTH_NAME"
+#OAUTH_CLIENT_ID=""
+#OAUTH_CLIENT_SECRET=""
+#OAUTH_AUTHORIZE_URL="" # For mastodon use "https://<mastodon domain>/oauth/authorize"
+#OAUTH_ACCESS_TOKEN_URL="" # For mastodon use "https://<mastodon domain>/oauth/token"
--- a/bookwyrm/views/landing/register.py
+++ b/bookwyrm/views/landing/register.py
@@ -29,7 +29,7 @@ class Register(View):
         if settings.install_mode:
             raise PermissionDenied()
 
-        if not settings.allow_registration:
+        if not settings.allow_registration and 'oauth-newuser' not in request.session:
             invite_code = request.POST.get("invite_code")
 
             if not invite_code:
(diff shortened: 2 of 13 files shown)


.env.example
```diff
@@ -120,3 +120,11 @@ OTEL_SERVICE_NAME=
 # for your instance:
 # https://docs.djangoproject.com/en/3.2/ref/settings/#secure-proxy-ssl-header
 HTTP_X_FORWARDED_PROTO=false
+
+# Enable logging in and registration with OAuth 2
+OAUTH_ACTIVE=false
+#OAUTH_NAME="OAuth Provider" # Displayed on Login Button as "Login with OAUTH_NAME"
+#OAUTH_CLIENT_ID=""
+#OAUTH_CLIENT_SECRET=""
```
alopix, line 128: This would only work with a single external source though, would it?
Since a user could use any Mastodon instance bookwyrm would need to dynamically register an OAuth client with every Mastodon server once it was used the first time + store that securely in the database.

rrgeorge: Right.
The idea is that the bookwyrm server authenticates against a *specific* mastodon (or other oauth 2.0 provider).
For mastodon specifically, you’d need to use the app registration api to generate these values for your bookwyrm server: https://docs.joinmastodon.org/methods/apps/#create

alopix: That's what I mean, exactly.
I think without the dynamic possibility of authentication users would get frustrated easily, since they wouldn't understand why only a very specific server could be used for authentication—so you'd potentially lock out people who'd like to use their Mastodon login. I'd argue that's against the spirit of the Fediverse.
(But to be clear: I think it's great you want to add external OAuth login!)

rrgeorge: It’s just like any other closed registration server. The idea is to unify login for the users of one service that wants to provide multiple fediverse platforms for their users. The way I am using it is to provide a bookwyrm server for my mastodon users.

alopix: I understand your use case but I am worried that this is such a specific use-case vs. how OAuth login (especially within the Fediverse) works, where one would generally not expect to be restricted to a single server—since at least that is how I’d interpret the idea of the Fediverse.
So maybe it’s possible to allow any Mastodon server with dynamic registration but add a server config to restrict it to a default Mastodon server as well to cover your use case?

bookwyrm/views/landing/register.py
```diff
@@ -55,7 +61,7 @@ def post(self, request):
 
         localname = form.data["localname"].strip()
         email = form.data["email"]
-        password = form.data["password"]
+        password = None if 'oauth-newuser' in request.session else form.data["password"]
```
beedell-roke: @rrgeorge, I really do not believe that password authentication should be disabled if OAuth is enabled, because that prevents users — such as myself — using my password manager to automatically authenticate. If it were the case, I would never enable OAuth, whereas I — without doubt — would, otherwise.

This would solely be (theoretically) sensible for CTAP2 key-pairs.

rokejulianlockhart (changes requested):

mouse-reeve:

PR #2681, opened 2023-02-18
Fix `SiteSettings.default_user_auth_group` FK on_delete value

chdorner:
This is a bug I introduced in #2554. The `on_delete` value in the model didn't match the one in the migration.

The migration uses `RESTRICT` instead of `PROTECT`, which is both more correct, but also those values need to be identical, otherwise Django thinks that there's a migration missing and will refuse to apply any new migrations.


joachimesque (approved):
Looks good to me 👍

Issue #2626, opened 2023-01-27
Bookwyrm Posts Not Showing on Mastodon Feed


pnutbutterprincess:
Describe the bug
My updates on bookwyrm are not showing up in my home timeline on mastodon (I am following my bookwyrm account)

To Reproduce

- Already following a bookwyrm account from mastodon, make a post on bookwyrm (in this case, a comment)
- Check home timeline on mastodon account
- Check the profile of the bookwyrm account on mastodon (confirmed I am following)
- No updates are shown

Expected behavior
I expected to see the posts starting to show up in my feed once I had followed (thanks to [this issue being resolved](https://github.com/bookwyrm-social/bookwyrm/pull/2613)) ♥

Instance
bookwyrm.social

Additional context

- my mastodon account is on kolektiva.social, meaning I was unable to follow until recently once the above issue was resolved
- I am still pretty new to federation and mastodon, as well as bookwyrm, so hoping I understand the way they should work together correctly

mouse-reeve:
Thank you for reporting! The fix you linked has been merged, but it hasn't been deployed yet, so it will probably not be until next week that it's live. There are also still some pretty big delays in federating bookwyrm content from bookwyrm.social out, which I'm working on separately.

pnutbutterprincess:
Awesome, thank you! I went through the issues, but hadn't seen it. I appreciate the quick reply!

alopix:
This is exactly what I have been experiencing. It seems like posts are delayed by 6 days since I have seen a few of my bookwyrm.social reviews and states being published this week with always a 6 day delay.

cloo:
My post is already lagging more than a week to propagate to Mastodon, I suspect it will not show up.

bluebirch:
I can search individal posts from Mastodon, but if I favourite of boost, notifications doesn't reach back to BookWyrm.

BlindMoon38:
> My post is already lagging more than a week to propagate to Mastodon, I suspect it will not show up.

I got an update propagated after 12 days to Mastodon (Hometown), and interactions from Mastodon were retrieved and relayed to me after 18 days...

I can say with certainty that the problem is getting worse.

genebean:
> There are also still some pretty big delays in federating bookwyrm content from bookwyrm.social out, which I'm working on separately.
Is there somewhere in particular this is talked about? I’d like to track progress and see if I have any ability to help.

mouse-reeve:
@genebean this is the place!

[...]

PR #2683, opened 2023-02-20
Attempt to complete inbox requests synchronously     

mouse-reeve:
When an inbox activity comes in from another fediverse instance, the behavior prior to this commit was  always to immediately give a 200 response to the external server and then create a celery activity (usually in the MEDIUM_PRIORITY queue) to complete it.
    
Instead, this would receive a request and try to complete it without making any http requests (which would make the request take too long to process). If an external request is required to complete the activity, a task is created and added to the queue.
    
Ideally, this will cause some tasks to happen very promptly, and reduce the load on celery, which would help queued tasks happen more quickly as well.
    
One downside is that this will make completing http requests from external servers slowing (since it's doing a bunch of thinking before responding).

(When I tested this, I was able to receive follow request `Accept`s and direct messages synchronously.)

Works on #2626
Works on #2614

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -100,9 +100,27 @@ class ActivityObject:
 
     # pylint: disable=too-many-locals,too-many-branches,too-many-arguments
     def to_model(
-        self, model=None, instance=None, allow_create=True, save=True, overwrite=True
+        self,
+        model=None,
+        instance=None,
+        allow_create=True,
+        save=True,
+        overwrite=True,
+        allow_external_connections=True,
     ):
-        """convert from an activity to a model instance"""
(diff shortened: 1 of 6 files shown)


hughrun:
Sorry @mouse-reeve I've been a bit busy this week. I'll take a look at this tomorrow properly, but on a first pass it looks like an elegant solution.

hughrun (approved):
Ok I have tested this against `bookwyrm.tech` (BW 0.5.4) and `ausglam.space` (Hometown Mastodon) and it looks good!

Interestingly things seem to be almost instantaneous against Mastodon, but with a delay against bookwyrm.tech - I assume this is because the recent fixes including this one for Celery tasks aren't in place on bw.tech yet.

I can't see any problems other than ones that have nothing to do with this PR. @mouse-reeve this looks good to go.

PR #2685, opened 2023-02-20
Use a separate queue for broadcasts

mouse-reeve:
I think this will go a long way to solve the federation delay problems we're seeing on b.s. I'm not sure at what point adding more queues will create more problems than it solves, but I do think in this case the queues are out of balance and moving broadcasts (which are the most common type of `medium_priority` task at the moment) to their own queue will be an improvement.

Works on #2626
Works on #2614

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -62,7 +62,7 @@ services:
     build: .
     networks:
       - main
-    command: celery -A celerywyrm worker -l info -Q high_priority,medium_priority,low_priority,imports
+    command: celery -A celerywyrm worker -l info -Q high_priority,medium_priority,low_priority,imports,broadcast
     volumes:
       - .:/app
       - static_volume:/app/static
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -21,7 +21,7 @@ from django.utils.http import http_date
 from bookwyrm import activitypub
 from bookwyrm.settings import USER_AGENT, PAGE_LENGTH
 from bookwyrm.signatures import make_signature, make_digest
-from bookwyrm.tasks import app, MEDIUM
+from bookwyrm.tasks import app, MEDIUM, BROADCAST
 from bookwyrm.models.fields import ImageField, ManyToManyField
 
 logger = logging.getLogger(__name__)
@@ -126,7 +126,7 @@ class ActivitypubMixin:
         # there OUGHT to be only one match
         return match.first()
 
-    def broadcast(self, activity, sender, software=None, queue=MEDIUM):
+    def broadcast(self, activity, sender, software=None, queue=BROADCAST):
         """send out an activity"""
         broadcast_task.apply_async(
             args=(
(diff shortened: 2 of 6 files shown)


docker-compose.yml
```diff
@@ -62,7 +62,7 @@ services:
     build: .
     networks:
       - main
-    command: celery -A celerywyrm worker -l info -Q high_priority,medium_priority,low_priority,imports
+    command: celery -A celerywyrm worker -l info -Q high_priority,medium_priority,low_priority,imports,broadcast
```
chdorner, line 65: Sorry to chime in here unannounced. I haven't worked with Celery in a long time so I was curious about the ordering of this. I thought the `-Q` argument might define the priority in which the worker should be picking stuff from the queues. But apparently [according to the docs](https://docs.celeryq.dev/en/stable/getting-started/next-steps.html#routing) the order doesn't matter:
> The order of the queues doesn’t matter as the worker will give equal weight to the queues

So depending on how you run the worker processes on b.s you might just get one massive FIFO-style queueing system, but with multiple queue names.

mouse-reeve: No need to apologize, chiming in is ideal. And yeah, the priority isn't meaningful on a technical level (I think there's a way to do this with rabbitmq but not redis? but that's a tangent). Celery seems to basically pick tasks off the top of one queue, then the top of another queue, et cetera, so the effect isn't that the high priority task queue is picked more often, but rather that it's a short queue, and when 10,000 import tasks are added, they aren't blocking a high priority task.

This approach of adding a queue did fix the problem with imports taking untenably long times to start, which is why I'm hopeful that it will improve the situation with posts not reaching remote servers in a timely manner (obviously only for posts created after this change, existing posts wouldn't be re-queued)

chdorner: So in a way, we are effectively getting the same benefit as if we were to scrap all those queues and just use one plus [Celery's message priorities](https://docs.celeryq.dev/en/stable/userguide/routing.html#redis-message-priorities)?

After reading more about scaling Celery, it seems that the way it handles queues and priorities, we would be better off running multiple workers with different `-Q` configurations, which would require each admin to tailor their setup better to the demand. i.e. in b.s's case with loads and loads of activity, having two separate worker configurations, one only on high-priority tasks, and the other on all or all other queues could help bring those high-priority queue lengths down a bit?

chdorner: Ah, there's also this config setting: https://docs.celeryq.dev/projects/kombu/en/latest/reference/kombu.transport.redis.html#kombu.transport.redis.Channel.queue_order_strategy

We would probably need to think this through properly, but using something other than the default `round_robin` might be worth trying out. Seems to me that by setting this to `priority` then the order of queues in the `-Q` would effectively be a priority list. But there are probably noticeable side-effects, lower priority queues would simply not get any tasks taken off of it if there aren't enough workers running. Instead of with round robin the whole system would just get slower under the same situation.

mouse-reeve: I think that's worth trying out (both because it sounds like a better, in-built approach to the problem I'm trying to solve, and it would be less maintenance, especially across docker and non-docker), and I think I also misunderstood those docs the first time I read them.

I'm a little unsure how much work it would be to switch systems, however, and I'm tempted to keep using this approach in this PR and re-think it as future work. Do you have thoughts on that? I might be making it out to be a bigger task in my head that it really is.

chdorner: Yeah it makes sense to give this one a go on b.s and see how it behaves.

I don't think the code changes to switch from `round_robin` to `priority` or `ordered` are going to be that big, but probably makes sense to think it through and do a bit of testing before rolling it out 👍

chdorner: I'm guessing this has already been deployed to b.s? I'm seeing updates from that instance federate within at least 17 minutes. It's likely that it's even faster, just that my timing looking at my feed was off :) It used to take a least a day or two for b.s status updates to populate in my instance's feed.

mouse-reeve: Yes, I'm using it there and federation has gotten a lot better, at least for outgoing activities.

PR #2687, opened 2023-02-21
Add config variable for additional CSP hosts

chdorner:
This came up in the developer chat yesterday. One instance had their `AWS_S3_CUSTOM_DOMAIN` redirect to another domain which then served the static files. This didn't work with our content security policy that was introduced in #2644.

This change adds a new environment variable `CSP_ADDITIONAL_HOSTS` which allows configuring additional hosts that are added to the policy after `"self"` (basically `DOMAIN`), and `AWS_S3_CUSTOM_DOMAIN`.


mouse-reeve (approved):
Looks good! And thank you for the comments in the .env file

PR #2688, opened 2023-02-22
fix: missing expand variable

0x29a:
Discover page is returning error 500 for me.

### DIFF
--- a/bookwyrm/templates/discover/large-book.html
+++ b/bookwyrm/templates/discover/large-book.html
@@ -46,7 +46,7 @@
             </div>
 
             <div class="notification has-background-body p-2 mb-2 clip-text">
-                {% include "snippets/status/content_status.html" with hide_book=True trim_length=70 hide_more=True %}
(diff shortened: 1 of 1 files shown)


0x29a:
@chdorner, as the author of https://github.com/bookwyrm-social/bookwyrm/pull/2652, would you like to take a look? I'm not sure that `expand` here should be equal to `expand` and not to either `True` or `False`.

mouse-reeve:
`expand` should be set to `False` for this view

mouse-reeve (changes requested):
This will be ready to go with expand set to false :)

0x29a:
@mouse-reeve, fixed. Thank you for the fast reply. :slightly_smiling_face:

mouse-reeve:
Perfect!

mouse-reeve (approved):
Perfect!

chdorner:
Oof! Thanks to both of you for fixing this. Great to have it covered by test cases now as well 🎉

Issue #2660, opened 2023-02-12
Importing book uses today's values for missing date parts


texttheater:
**Describe the bug**
When importing a book from OpenLibrary, month and day are set to today's month and day.

**To Reproduce**
1. Find a book on OpenLibary, e.g., one published in 2022
2. Import it into a BookWyrm instance, e.g., on Feb 12, 2023

**Expected behavior**
The book is shown on BookWyrm as published in 2022 (month and date unspecified)

**Actual behavior**
The book is shown on BookWyrm as published on Feb 12, 2022

**Additional context**
Related issue: #743

chdorner:
I just [imported a book](https://secretbearlibrary.org/book/15357/s/activities-of-daily-living) today from OpenLibrary which had its published date set correctly.

Do you have a specific book that I could check this with?

chdorner:
With the date and month yes, which is most likely related to #743. But the year is set correctly to 222, not 2023.

texttheater:
Sorry, my mistake. Fixed the description now.

chdorner:
Did a bit of digging, and this is where it originally happens: https://github.com/bookwyrm-social/bookwyrm/blob/2c2daf5fdf2d334a344e166eb3bfbcf375dbc4c5/bookwyrm/models/fields.py#L486

`value` is set to `"2022"` coming from OpenLibrary's API and `dateutil` automatically turns this into `datetime.datetime(2022, 2, 13, 0, 0)`.

In addition, and this is more related to #743, we would need to change how to store this data, it's currently using Django's `models.DateTimeField` which doesn't support missing day and month parts of a date.

PR #2691, opened 2023-02-22
Proof of concept for a loosely defined date field

chdorner:
This could be a potential solution for #743.

Essentially this adds two new classes:
* `LooseDate` value object with `year` (required), `month` (optional), and `day` (optional)
* `LooseDateField` for storing `LooseDate` value objects in the database

Things to note:
* `LooseDate` is validating that it's a valid date by replacing optionally missing values for month and day with 1, and then validating the date
* The database representation is a string in the `YYYY-MM-DD` format, missing values for month and day are represented as `00`, so the different permutations are `"2023-02-22"`, `"2023-02-00"`, `"2023-00-00"`

Things that I still need to try:
* Rendering of the value, in its current state I only changed the book editing form to work with the new field
* Queries that reference this field
  * Need to make `ORDER BY published_date` work with this new field. Could potentially work out of the box with string ordering?
  * Need to double-check that we don't do any fancy less-than, more-than, or in-between type queries. This could get tricky

Things that are outstanding for proper implementation of this:
* Replacing the existing book model fields with this, including a fairly scary database migration.
  * Unclear if we could change the type of the existing column to a string without needing to create a new column
  * This could potentially be a multi-step migration, add a new temporary column as a string, copy data over, delete the old column, rename the new column

What do you all think about this? Anything I missed that I should check?


marcusyoung:
You can change type and contents at the same time with USING:

```
ALTER TABLE table
ALTER COLUMN column TYPE data_type USING expression;

```
I'm a bit wary of ditching the date type field when in many cases it is a date being entered in the field.

I would look to allow the input of the published date without day or month. When written to the database replace anything missing with 01 and set a flag in another column. Then when displaying the date extract the required elements of the date (e.g. just year) based on this flag. That way you lose none of the benefits of storing as a date. Just my initial thoughts.

chdorner:
Yeah I need to understand more how we query those date fields, the main benefits for keeping them as date fields mostly revolve around querying, either with `WHERE` clauses or for ordering.
Date validity could be handled in the Django field implementation fairly nicely.

chdorner:
The ideal requirements on ordering could also help arguing one way or another. 

For example, should the ascending order be:
1. First: Feb 2023
2. Second: 13 Feb 2023

Or the other way around, ordering less precise dates at the end:
1. First: 13 Feb 2023
2. Second: Feb 2023

If we think the second case is more intuitive then your proposed solution @marcusyoung would make more sense, with a small twist:
For dates missing the day part we could fill in the last day of the month, or when month is missing the last month and last day of the month. Then we'd have that sort order figured out.

marcusyoung:
You'll also need to consider what happens on import too, with the date coming from OpenLibrary etc.

marcusyoung:
May also be a problem with instances being on different versions if this field is inconsistent between databases?

chdorner:
> You'll also need to consider what happens on import too, with the date coming from OpenLibrary etc.

OpenLibrary is even more flexible, either it's just a year (like mentioned [here](https://github.com/bookwyrm-social/bookwyrm/issues/2660#issuecomment-1428430993)), or if a book has a precise publishing date ([this specific audiobook edition](https://openlibrary.org/books/OL32067572M/How_to_Do_Nothing_Lib_E)) it transmits it as "Mar 01, 2021".

> May also be a problem with instances being on different versions if this field is inconsistent between databases?
Yes! That is a point that we'll have to consider. I'll add it above to not forget.

chdorner:
I added a proof of concept with @marcusyoung's proposed solution which stores an extra field for each date, i.e. `first_published_date` and `first_published_date_precision`.

Besides the extra fields in the model, we can get away with just overriding the form's `__init__` function and handling everything in there. Depending on which arguments are passed to `EditionForm.__init__` we know if it's a form render or a form submit.

On a form render we:
* check the precision field and override the corresponding date field value with a string (precision `month` will results in `f"{year}-{month}-0"`, precision `year` in `f"{year}-0-0"`).
* The date selector widget can already handle these types of strings since those are the ones that are also sent on form submit

On a form submit we:
* detect based on the individual date part fields (i.e. `published_date_year`, `published_date_month`, `published_date_day`) what the precision is, set the precision value in the data, default the missing field values to "1"
* translating those individual fields into a date is handled automatically by the form

The (ugly) implementation is in this commit: 5c9e3183317d46723c7d72cc88c2981fe87c042a

Generally, I like this approach better. We can still use the date fields for ordering, `WHERE` clauses, etc.
We could even use this same approach for federating these dates out: the date field with its month and day parts defaulting to "1", plus an extra "_precision" field. This would be a simple way for backwards compatibility, older BookWyrm versions would just import the date field and ignore the precision field, and newer BookWyrm versions would also import the precision field and would be able to support less precise dates.

Curious to know what you think of these two approaches @mouse-reeve? Feel free to compare the two commits in this PR.

dato:
We could perhaps avoid the additional AP fields because partial serialization in formats like _YYYY_ and _YYYY-MM_ are backwards-compatible with the current use of `dateutil.parser` (which parses `2022` and `2022-10` just fine).

(To be clear, I'm only talking about network-transmitted data here.)

Say such code is released in v0.7. Then, upon an incoming date like '2023':

- v0.7 instances know to parse it as "2023, precision = year"
- v0.6.x parse it as "2023-01-01", which is strictly not worse than what happens today (with _x > 6_ because of #2660; I've provided fix for that bug in #3038)

-----
P.S.: While I'm commenting, I do like the idea of a custom `PubDateField` but—given the effort and care it takes—it would be great if the existing column could be upgraded, rather than introducing an extra one (if I'm reading the code correctly). Anybody mind if I take this code and take a stab at that approach?¹

(¹) As an instance maintainer, I would _love_ if the migration either defaulted, or allowed me to pick, "set precision = year if day = month = 1".

dato:
> While I'm commenting, I do like the idea of a custom PubDateField but—given the effort and care it takes—it would be great if the existing column could be upgraded, rather than introducing an extra one (if I'm reading the code correctly). Anybody mind if I take this code and take a stab at that approach?

I submitted #3059 for review, which reuses every possible field and form, in order to keep backwards compatibility.

Turns out Django forms already support serializing partial (though possibly invalid) dates: they do it replacing `0` for the missing parts. I leveraged this in the forms, which makes for few modifications needed there.

mouse-reeve:

I think this issue has already been resolved in any case. Thank you all!

PR #2694, opened 2023-02-24
Replace Next and Previous with Older and Newer

kangangi:
Replaces "Next" and "Previous" with "Older" and "Newer"

### DIFF
--- a/bookwyrm/templates/snippets/pagination.html
+++ b/bookwyrm/templates/snippets/pagination.html
@@ -9,7 +9,7 @@
         {% endif %}>
 
         <span class="icon icon-arrow-left" aria-hidden="true"></span>
-        {% trans "Previous" %}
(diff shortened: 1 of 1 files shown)


chdorner (approved):
Thank you so much for your contribution! 🙌

mouse-reeve:
This is actually applied a little too broadly, since this snippet is used for all pagination; it's a good fix for the feed, but makes less sense for pages of a list, or books by an author, which are not sorted chronologically. The direction is also reversed – the feed is reverse chronological (newer first, older later), so the "next" page takes you to older posts and the "previous" page takes you to newer posts. Not intuitive when you're looking at the template at all!

Followup PR here, let me know if I completely brainfuzzed the logic here, and thank you both for working on this: https://github.com/bookwyrm-social/bookwyrm/pull/2696

PR #2695, opened 2023-02-25
Show notification banner on top of form when book failed to update

### LINKED ISSUE #2651: Adding new book fails 'quietly'
**Describe the bug**
I'm trying to add a new book on bookwyrm.social. After filling in the relevant parts of the form, I click 'Save'. The page reloads, but no book gets added. I'm still in the form, 'Author' is empty, everything else stays the same. No error message of any kind.

**To Reproduce**
I go to https://bookwyrm.social/create-book
I add Title, Subtitle, Language, Publisher, First published date (Year only), Published Date (Year only), URL for Cover image, Format, Pages, ISBN 13. I leave everything else blank.
I click on 'Save'
Browser loads something, page reloads, still on the same form. Author field is now empty, everything else is still there. No book gets added. No error message that I can see.

**Expected behavior**
Well..... either the book gets added, or the user somehow gets informed, what part of his input is blocking the book from getting added :)

**Instance**
bookwyrm.social

---

    - OS:: MX Linux
    - Browser: I tried both Firefox (109) and Chromium (109.0.5414.119)

chdorner:
Fixes #2651

This shows a notification banner on top of the form when there are validation errors:
image


jaschaurbach:
The moment I get an error banner I would expect the error message to be there as well and not somewhere on the page. 

I think it is bad UI to show a banner saying "Something went wrong" and without more information. Yeah thanks, something went wring. WHAT went wrong? Can I correct it by myself or do I need to contact someone? 

Y'all remember Blue Screens from Windows? This is the same: Thanks for the error message, it does not help.

mouse-reeve:
The error is also on the page, shown at the element that has the error. But since the banner is a lot more noticeable than the banner, maybe it would make sense to list the actual errors in the alert box (and use the class `is-light` if it's a larger footprint box)? I believe the django form has the errors in a list and they are already localized

chdorner:
@mouse-reeve I'm not sure I understand, you'd like to have all form errors listed in that box in addition to having them as a help paragraph next to each field?

`is-light` is a good shout!

chdorner:
To be honest I was mostly thinking of not reinventing the wheel. The main issue is that this form is fairly large, so depending on the screen size it might not be immediately obvious that the form validation failed. So I was trying to just add a banner on top so that it's obvious that form validation failed for the user to then scroll down the offending fields which have the error message next to them.

mouse-reeve:
I can see that (and thanks for doing the lighter screenshot, it looks great). Listing the errors on the top is what I was thinking, but I'm not sure if it's a good call or not -- I can see it either being large and redundant, or a helpful way to see which field needs work? I think this is good to go as is, and we can consider in the future if it's too hard to find the error fields, and if so what the fix should be, since it might be that, if that's the case, the error messages should look different rather than listing them twice.

mouse-reeve:
The tests look like a problem with main, I probably need to add a merge migration

chdorner:
One additional improvement we could do in the future is to add the `is-danger` class to each input that has a validation error so that the fields get a red border like in the [Bulma docs:](https://bulma.io/documentation/form/general/#complete-form-example)
image
This makes it a bit more obvious which field has a validation error.

I don't think we already do that, at least I couldn't find anything in the code that would hint at that. I'd probably handle this as a separate issue though, seems that we'll have to mess with Django's widgets for this.

mouse-reeve:
I agree - we should do that and it's out of scope for this PR

PR #2696, opened 2023-02-25
Only use chronological pagination sometimes

mouse-reeve:
The timeline uses chronological buttons, but other paginated pages do not (by default). This also reversed the chronology.


chdorner (approved):
Ah, you are right. Sorry for missing this when merging #2694

I added one commit to this PR for using the chronological pagination on the user profile activities lists. Feel free to merge with or without that commit :)

PR #2702, opened 2023-03-01
Add attributes to images to hint async load

Ryuno-Ki:
This was suggested on Matrix a while ago but I only found the time now to move forward with it.


mouse-reeve:
Thanks so much!

PR #2704, opened 2023-03-05
Enable Azure BlobStorage as an alternative to AWS S3

tebriel:
This adds a new storage backend for Azure BlobStorage, [I'm running](https://bookwyrm.frodux.in) this [patch](https://github.com/tebriel/bookwyrm-docker) in Azure Kubernetes
hosting all my assets in Azure BlobStorage and it's been stable enough I thought I should commit
this back upstream.

If you're willing to accept this patch I can make a PR to the bookwyrm-social/documentation repository
but I didn't want to do so before I was sure this was something you'd be willing to merge in.

Thank you!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@ __pycache__
 .git
 .github
 .pytest*
+.env
--- a/.env.example
+++ b/.env.example
@@ -82,6 +82,12 @@ AWS_SECRET_ACCESS_KEY=
 # AWS_S3_REGION_NAME=None         # "fr-par"
 # AWS_S3_ENDPOINT_URL=None        # "https://s3.fr-par.scw.cloud"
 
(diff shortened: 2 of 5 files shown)


jaschaurbach:
You can solve this issue by running ./bw-dev black before commit. 

Could you please add the Azure Settings to .env.example? Thanks!

tebriel:
@jaschaurbach anything else I should do? Happy to wait just didn't know if I was supposed to notify anyone, thanks!

jalada:
I'm not sure why; maybe it's my set up as I've installed without docker, but I had to manually `venv/bin/pip3 install django-storages[azure]` on my instance (during an upgrade) otherwise I was getting an error about a missing module `azure`.

tebriel:
@jalada I don't know how the non-docker upgrade works, but do you normally install the dependencies before the upgrade?

Issue #2157, opened 2022-06-28
CSV export results in empty file with no data


noisydeadlines:
**Describe the bug**
When using the CSV data export the resulting file is empty

**To Reproduce**
1. Go to 'Settings'
2. Click on 'CSV Export'
3. Click on 'Download the File'
4. Open the downloaded file

**Expected behavior**
CSV with data from the books

**Screenshots**
![empty CSV](https://user-images.githubusercontent.com/108202814/176283524-75f62de1-7f3d-4300-a219-72210cb1a661.JPG)

**Instance**
BookWyrm Social

---

**Desktop (please complete the following information):**
    - OS: Windows 10
    - Browser: Edge
    - Version 103.0.1264.37 (Official build) (64-bit)

noisydeadlines:
Hi! I tried exporting it again twice, I still get the same empty file every time.

Strubbl:
I can confirm this issue was happening with v0.4.3 and is still not working with v0.4.4. I was trying to export on the BookWyrm Social instance.

cubicgarden:
I can confirm I'm getting the exact same thing too.
Tired to export twice and spoke with @mouse-reeve who pointed to this issue

r3k2:
same here, is annoying only reason I wanted this is to migrate to another server, the one I am at is not been taken care off is more time down than up for many people, https://bookwyrm.social/  I keep getting issues for more than  9 months, I cant edit a group list some of us made, works when not logged in, but as soon I log in it breaks,  because a nginx bad gateway that usually happens when the service is unresponsive or the container is low on memory unresponsive etc, so got tired of waiting for the admin to reply to us, I even offer to let me ssh so I can fix it, (systems engineer for 24 years)  and to try to troubleshoot this even do I sent 2-3 times screenshots, I waited 9 months because I know this is not a job and voluntier but now and here comes to the point, I try to get the CSV to migrate to another instance and is **empty**  :(

eldang:
I'm also frustrated by this bug, and I'm wondering if the solution is to make the export asynchronous.  I have two examples in mind which might be useful as patterns:

[...]

PR #2713, opened 2023-03-07
Change CSV export to buffer instead of streaming

WesleyAC:
The idea behind a streaming CSV export was to reduce the amount of memory used, by avoiding building the entire CSV file in memory before sending it to the client. However, it didn't work out this way in practice: the query objects that were created to represent each line caused Postgres to generate a very large (~200MB on bookwyrm.social) temp file, not to mention the memory being used by the Query object likely being similar to, if not larger than that used by the finalized CSV row.

While we should in the long term run our CSV exports as a Celery task, this change should allow CSV exports to work on large servers without causing disk-space problems.

Fixes: #2157

### DIFF
--- a/bookwyrm/tests/views/preferences/test_export.py
+++ b/bookwyrm/tests/views/preferences/test_export.py
@@ -1,7 +1,7 @@
 """ test for app action functionality """
 from unittest.mock import patch
 
-from django.http import StreamingHttpResponse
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Love to do the exact opposite of what I intended to! Thank you; it will remain to be tested at scale, but in my local instance this works well

PR #2721, opened 2023-03-08
fix typo in systemd example

verymilan:
...which prevented imports from running


mouse-reeve:
Thanks so much for fixing this 🤦

PR #2724, opened 2023-03-09
Fix dbshell command

WesleyAC:
dbshell needs to be run in a already-running container, thus exec rather than run is the correct docker-compose command.


mouse-reeve:

Issue #2736, opened 2023-03-13
CSS doesn't load on server error page


WesleyAC:
When there's a 500 server error, the page doesn't load any CSS, and thus looks quite ugly. We should figure out why this is and fix it.

PR #2737, opened 2023-03-13
Use context processor for 500 page

WesleyAC:
By default, Django doesn't run any context processors for server errors, to make the error path as simple as possible. However, this has the downside that our template does not load correctly. To fix this, I added a custom 500 error handler, which will run the context processor.

Fixes: #2736


0x29a:
Yay, thanks for fixing this one!

PR #2746, opened 2023-03-18
Env var improvements

WesleyAC:
* Use typecasting env var methods
* Set default `FLOWER_PORT`
* Raise error if using default `SECRET_KEY` in production


mouse-reeve:
Looks good! Thank you :)

PR #2749, opened 2023-03-20
Add helper to refer views back to http referers safely

### LINKED ISSUE #2677: Setting A Book's List Position Should Not Re-sort List
When adjusting the list position of books in a list, every time one hits "set," the list is immediately re-sorted by list position.  This happens even if one has set the sort order as title to assign positions on the list.  One cannot set the list position of more than one book at a time.

When one presses "Set," the position of the book should be recorded but the list should not be immediately re-sorted.

Setting the position of a single book, hitting the set button, re-sorting the list manually to title, and repeating and repeating this process.

mouse-reeve:
In most cases, we want to return back to where we came from after performing an action. It's not safe to return to an arbitrary referer, so this streamlines using the util validator to verify the redirect and fall back on regular redirect params if the referer is outside our domain.

how the heck does anyone spell referer correctly??

Fixes #2677

### DIFF
--- a/bookwyrm/tests/views/test_helpers.py
+++ b/bookwyrm/tests/views/test_helpers.py
@@ -8,7 +8,7 @@ from django.test.client import RequestFactory
 import responses
 
 from bookwyrm import models, views
-from bookwyrm.settings import USER_AGENT
+from bookwyrm.settings import USER_AGENT, DOMAIN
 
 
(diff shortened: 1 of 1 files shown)


mouse-reeve:
@bookwyrm-social/code-review I made this team so I could tag that I've opened a PR that I'd like to get reviewed. Is that helpful or just creating more noise in your notifications?

Issue #2720, opened 2023-03-08
Investigate `add_status_task` performance


WesleyAC:
`add_status_task` is one of the most common tasks, so it's a good first target for improving performance.

I did some "profiling" today by adding calls to `time.time()` in the code, and discovered that almost all of the time spent on this task is in the `HomeStream.get_audience`, and specifically this call:

```python3
audience.filter(
    Q(id=status.user.id)  # if the user is the post's author
    | Q(following=status.user)  # if the user is following the author
).distinct()
```

I'm confused about why the bottleneck is there specifically — the generic `ActivityStream.get_audience`, which is called by the `HomeStream` version has very similar queries with no such problems.

One thing that I don't understand about this is that just making the `QuerySet` should have basically no impact on performance, only executing it should, and I don't think the `QuerySet` is supposed to get executed in this function. But, I change the `HomeStream.get_audience` function to:

```python3
    def get_audience(self, status):
        t1 = time.time()
        audience = super().get_audience(status)
        t2 = time.time()
        if not audience:
            return []
        x = audience.filter(
            Q(id=status.user.id)  # if the user is the post's author
            | Q(following=status.user)  # if the user is following the author
        ).distinct()
        t3 = time.time()
        print(f"HomeStream.get_audience {t2 - t1} {t3 - t2}", file=sys.stderr)
        return x
```

and get:

```
HomeStream.get_audience 0.010448694229125977 7.642062664031982
HomeStream.get_audience 0.008398294448852539 7.4949047565460205
HomeStream.get_audience 0.010053634643554688 6.8434672355651855
```

and similar results, which seem to clearly indicate that simply constructing the `QuerySet` is what's taking the time. I assume I must be missing something here — I intent to dig in more later, but if it's obvious to anyone what I'm missing, or what to do to improve this, please do chime in.

WesleyAC:
<details>
<summary>Generated SQL (for user ID 1):</summary>

```sql
[...]

PR #2750, opened 2023-03-20
Add more detailed telemetry for get_audience

WesleyAC:
This is still slow in some cases, despite #2723, so this information should give useful data about how it could be optimized more.

This also adds some abstraction around getting the tracer, just to follow the advice in the OpenTelemetry documentation not to use __name__ directly to set the tracer name. The advice is ignored in most of their examples, so it probably doesn't matter, but IDK, seems reasonable to try to follow it.

Related: #2720


mouse-reeve:
When I went to test this without having OTEL configured, I got an `Transient error StatusCode.UNAVAILABLE encountered while exporting span batch` after posting a status, and I didn't see it when I tried again on the `main` branch. I didn't get a stack trace, however

WesleyAC:
@mouse-reeve thanks for catching that! modified the telemetry code to only enable the OTLP exporter if it's actually configured, which gets rid of that warning.

Issue #2748, opened 2023-03-20
Private profiles


lostfictions:
**Is your feature request related to a problem? Please describe.**

Hi! Thanks for all your work on Bookwyrm. I was hoping to use it as a way of keeping track of what I've read and want to read for myself and a few close friends. In my profile settings, I checked "manually approve followers," and I also set "default post privacy" to "followers." Some time later, I was a little shocked to discover that all the shelves on my profile are completely public -- not just to any other Bookwyrm user, but to anyone on the internet. This is something I really don't want.

I guess I don't fully understand what effects the privacy settings are actually supposed to have, if not to make the things I'm marking as to-read or did-read private. (If I think about it as someone who has used Mastodon and knows that Bookwyrm federates and so probably conforms to some shared model I can maybe start to reason about how it's centered around some concept of "activity" and "feeds" that have their own pre-existing notions of visibility ...but to be honest I've been unchecking "post to feed" for much of my activity in the hopes that it would make it _more_ private, not that it would exist as some parallel thing that's posted for the world to see.)

**Describe the solution you'd like**

Would it make sense to allow making profiles visible only to followers? I'm loath to request an extra privacy knob when my experience of the existing privacy model has been pretty confusing so far... but I'm not sure if there's an easier solution.

**Describe alternatives you've considered**

I personally think that "manually approve followers," "hide followers and following," and a hypothetical "restrict profile visibility to followers" would, from a usability and comprehensibility standpoint, make a lot more sense to roll into a single "private profile" setting. But I know that would probably be confusing to existing users, and that some people want more fine-grained control over these things.

Alternately, maybe documentation and onboarding could just be a lot clearer about the public nature of shelves, so that users can make an informed decision about whether or not they'd be comfortable using Bookwyrm.

WesleyAC:
You can set privacy for shelves on a per-shelf basis, by going to the "Your Books" section, clicking "Edit shelf" for each shelf, and adjusting the dropdown (which is unfortunately not labelled). I agree that this is confusing, though, I ran into similar confusion at first.

I think there would be value in providing some less fine-grained privacy control here. Changing "Default post privacy" to "Default privacy level" and having it also affect shelves (maybe a popup when you change it of "would you like to change the privacy level of your existing shelves?") seems good to me.

WesleyAC:
[...]

PR #2752, opened 2023-03-21
Add message about shelf privacy in user settings

lostfictions:
Following up from #2748, here's a very small PR to add a hint about where to find shelf privacy. I think it could still be nice to further refine how privacy works in Bookwyrm, but hopefully this is a step towards alleviating some of the confusion of having privacy settings that exist in very different places.

I just did this in the Github in-browser editor, so please double-check that I'm not just proposing nonsense -- I have no idea how Django templating works!

The url comes from [`user_menu.html`](https://github.com/bookwyrm-social/bookwyrm/blob/e9d08e7424d66c9a7b45a6853ec8fae09e7f74c0/bookwyrm/templates/user_menu.html#L38-L40); the url binding and locale block format is cribbed [a little up the file in `edit_user.html`](https://github.com/bookwyrm-social/bookwyrm/blob/e9d08e7424d66c9a7b45a6853ec8fae09e7f74c0/bookwyrm/templates/preferences/edit_user.html#L83-L86) (but I don't know how scoping works and whether it's ok to redeclare `path` like this). The `notification` class is [used in `delete_user.html`](https://github.com/bookwyrm-social/bookwyrm/blob/e9d08e7424d66c9a7b45a6853ec8fae09e7f74c0/bookwyrm/templates/preferences/delete_user.html#L14-L16), though I've removed the additional colour class (`is-link`).

### DIFF
--- a/bookwyrm/templates/preferences/edit_user.html
+++ b/bookwyrm/templates/preferences/edit_user.html
@@ -131,6 +131,10 @@
                     {{ form.default_post_privacy }}
                 </div>
             </div>
+            {% url 'user-shelves' request.user.localname as path %}
+            <p class="notification is-light">
+                {% blocktrans %}Looking for shelf privacy? You can set a sepearate visibility level for each of your shelves. Go to <a href="{{ path }}">Your Books</a>, pick a shelf from the tab bar, and click "Edit shelf."{% endblocktrans %}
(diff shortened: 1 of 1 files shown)


jaschaurbach:
Thank you!

PR #2774, opened 2023-04-01
Added command 'down', to bring down containers

fwalloe:
Added the command 'down', which can be used to bring down the docker containers. Equivalent to running 'docker-compose down'

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -287,6 +287,7 @@ case "$CMD" in
         echo "Unrecognised command. Try:"
         echo "    setup"
         echo "    up [container]"
+        echo "    down [container]"        
         echo "    service_ports_web"
         echo "    initdb"
         echo "    resetdb"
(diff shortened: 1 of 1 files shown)


jaschaurbach (changes requested):
Please add it to the list of commands available (at the bottom of the file)

mouse-reeve:
I'm not sure if it's a good idea to replicate docker commands in `bw-dev` when they don't do anything different than running then from docker-compose? It's clear to me why it's useful to have commands that are longer, involve multiple steps, or have specific flags are beneficial to have in the file, but it feels brittle to me to have bw-dev also encompass regular docker commands.

nb: The one place where this is already the case is `build`, which I also think perhaps shouldn't be a command in bw-dev

fwalloe:
@jaschaurbach Added it to the list of commands

@mouse-reeve That's a fair point and I would tend to agree, but so long as there's a command to start the container in bw-dev, there should also be an equivalent command to bring it down.

jaschaurbach (approved):
Thanks.

mouse-reeve:
That seems reasonable enough to me, thank you both!

PR #2778, opened 2023-04-04
Move the search request logic into the AbstractConnector

ranok:
Per discussion in the Matrix dev chat with @mouse-reeve, here is an initial code reorganization where the code to get the search JSON is moved from the connector_manager to the AbstractConnector class. This should allow for subclasses (i.e. new connectors) to support more diverse APIs, e.g., Arxiv's which is Atom (XML) based, or Citation, which requires an API key to be set in the request headers.

### DIFF
--- a/bookwyrm/connectors/abstract_connector.py
+++ b/bookwyrm/connectors/abstract_connector.py
@@ -4,6 +4,8 @@ from urllib.parse import quote_plus
 import imghdr
 import logging
 import re
+import aiohttp
+import asyncio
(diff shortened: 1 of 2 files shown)


ranok:
Added a commit to fix the pylint issues

jaschaurbach:
@ranok Sorry that no-one started the tests. Did that now.

mouse-reeve:
Looks good! Sorry for the delay on merging this

PR #2781, opened 2023-04-05
Replace docker-compose with compose V2 call

Neriderc:
Replaced compose V1 'docker-compose' with compose V2s 'docker compose'

Docker is removing support for docker-compose, and it doesn't appear to be possible to install it anymore. Instead, it has been replaced with compose V2 which is a docker plugin called with 'docker compose' (no hyphen). [More info](https://docs.docker.com/compose/compose-v2/).

This pull request replaces all instances of 'docker-compose' with 'docker compose' exclusively within bw-dev.

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -24,20 +24,20 @@ source .env
 trap - EXIT
 
 function clean {
-    docker-compose stop
-    docker-compose rm -f
+    docker compose stop
+    docker compose rm -f
 }
 
 function runweb {
-    docker-compose run --rm web "$@"
+    docker compose run --rm web "$@"
 }
 
 function execdb {
-    docker-compose exec db $@
+    docker compose exec db $@
 }
 
 function execweb {
-    docker-compose exec web "$@"
+    docker compose exec web "$@"
 }
 
 function initdb {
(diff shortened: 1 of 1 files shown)


WesleyAC (changes requested):
Thanks for finding this!

I think it would be preferable to check if `docker compose` is installed, but fall back to `docker-compose` and give a warning if not, given that older installs are likely to not have the compose plugin.

Neriderc:
> Thanks for finding this!
> 
> I think it would be preferable to check if `docker compose` is installed, but fall back to `docker-compose` and give a warning if not, given that older installs are likely to not have the compose plugin.

I'm afraid I'm not familiar enough with how to do that. @jaschaurbach asked on the matrix chat if I could do a pull request with this change since it worked for me, but I'm no expert. Probably better for someone else to do, or if you have info or an example you could point me at I can give it a go.

WesleyAC:
No worries! I can make the change.

WesleyAC:
Sent #2814 for this, so I'm going to go ahead and close this. Thanks again for reporting it!

PR #2782, opened 2023-04-05
Small quality of life improvements to list handling

ranok:
This small PR improves the clarity of the list status, as closed lists currently say they are curated, which led me to wonder why I couldn't suggest titles to add. I also got a stack trace in debug mode/404 in production on deleting a list as it returns the user to the now-deleted list. I changed it to return back to the /list view that shows all the lists (now minus the deleted one).


mouse-reeve:
Thank you so much for fixing these!

WesleyAC (approved):
LGTM, thanks for the fixes!

PR #2785, opened 2023-04-06
Optimize get_audience

WesleyAC:
See #2720 (as well as the commit messages) for details. I haven't done detailed performance analysis on these queries (in part because doing so well would require data from #2779 in prod), but I think it's highly likely that these changes will allow Postgres to use the indices it has, and reduce these queries from ~8 seconds to hundreds of milliseconds at most, given that's what I saw for the narrower case of this that I analyzed before.

I don't have a super high level of confidence in this code, but it is to the best of my knowledge correct. This is just a fiddly part of the codebase, so I want to be careful.

### DIFF
--- a/bookwyrm/activitystreams.py
+++ b/bookwyrm/activitystreams.py
@@ -208,11 +208,11 @@ class LocalStream(ActivityStream):
 
     key = "local"
 
-    def _get_audience(self, status):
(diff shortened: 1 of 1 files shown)


WesleyAC:
@mouse-reeve conflict resolved :)

PR #2787, opened 2023-04-06
Add form to remove tasks from Celery

WesleyAC:
Screenshot 2023-04-06 at 05-54-40 BookWyrm

### DIFF
--- a/bookwyrm/templates/settings/celery.html
+++ b/bookwyrm/templates/settings/celery.html
@@ -116,6 +116,35 @@
 </div>
 {% endif %}
 
+<section class="block">
+    <div class="content"><h2>{% trans "Clear Queues" %}</h2></div>
+
+    <div class="content notification is-warning is-flex is-align-items-start">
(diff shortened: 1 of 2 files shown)


bookwyrm/templates/settings/celery.html
```diff
@@ -116,6 +116,37 @@ <h3>{{ worker_name }}</h3>
 </div>
 {% endif %}
 
+<section class="block">
+    <details>
+        <summary><div class="content"><h2>{% trans "Clear Queues" %}</h2></div></summary>
```
mouse-reeve, line 121: I believe the test is failing because you aren't supposed to nest a div inside a summary element

mouse-reeve (changes requested):
This is going to be incredibly helpful, thank you

bookwyrm/templates/settings/celery.html
WesleyAC: Yeah, decided to just give up on making it collapsible, that's too fancy I guess :sweat_smile:

mouse-reeve (approved):

PR #2789, opened 2023-04-06
Fix two places where the reading status shelf name wasn’t translated

bpeel:
- In the list of shelves on the page for a book, the name of the reading status shelf was always in English.
- The “stopped-reading” status shelf was never translated.


bpeel:
I’ve added another commit to try to fix the pylint problem. Sorry about that.

jaschaurbach:
@bpeel We all had been there forgetting pylint ;)

Issue #2794, opened 2023-04-07
Can't follow to or from GoToSocial


kvibber:
**Describe the bug**
GoToSocial can retrieve individual posts from Bookwyrm, and can retrieve a Bookwyrm user's profile, but it cannot follow that user.

Bookwyrm cannot retrieve a GoToSocial user's profile at all and displays an error message when I search for one.

**To Reproduce**
On a GoToSocial instance, search for a Bookwyrm user. Then try to follow them.

On a Bookwyrm instance, search for a GoToSocial user. You can search for my test account at @kelsonv@gts.keysmash.xyz

**Expected behavior**
The GTS user should be able to follow the Bookwyrm user and vice versa

**Instance**
bookwyrm.social

**Additional context**

Using GoToSocial 0.8.0-rc1 and bookwyrm.social.

Logs show both of the following are throwing back 401 Unauthorized HTTP responses:
GTS' POST request to https://bookwyrm.social/inbox 
Bookwyrm's GET request to https://gts.keysmash.xyz/users/kelsonv

This looks very similar to a bug I fixed in Takahe where it was rejecting signatures whose algorithm had been obfuscated by labeling it as "hs2019", which GoToSocial does according to https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12 
https://github.com/jointakahe/takahe/issues/528

In the case of Takahe it was easy to just treat hs2019 as sha-256, since that's the only signature algorithm it accepts. I took a look at bookwyrm/signatures.py and I see Bookwyrm also supports incoming sha-512 signatures, so it'll be a bit more complicated than just adding `or algorithm == 'hs2019'` in verify_digest.

I can give this a try over the weekend and see if it works and fixes either (or both) issues.

WesleyAC:
Stack trace:

```
web_1             | OK: /search
web_1             | Traceback (most recent call last):
web_1             |   File "/app/bookwyrm/views/helpers.py", line 78, in handle_remote_webfinger
web_1             |     user = models.User.objects.get(username__iexact=query)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
web_1             |     return getattr(self.get_queryset(), name)(*args, **kwargs)
web_1             |   File "/usr/local/lib/python3.9/site-packages/django/db/models/query.py", line 435, in get
web_1             |     raise self.model.DoesNotExist(
[...]

PR #2800, opened 2023-04-07
Fix Accept header for requesting ActivityPub objects

WesleyAC:
This is the header described in the ActivityPub spec, which should fix some federation problems with GoToSocial and potentially other picky services.

Related: #2794, https://github.com/superseriousbusiness/gotosocial/issues/1676

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -384,7 +384,8 @@ def get_activitypub_data(url):
         resp = requests.get(
             url,
             headers={
-                "Accept": "application/json; charset=utf-8",
+                # pylint: disable=line-too-long
+                "Accept": 'application/ld+json; profile="https://www.w3.org/ns/activitystreams"',
                 "Date": now,
                 "Signature": make_signature("get", sender, url, now),
             },


hughrun:
Hi @WesleyAC 

I've tested this and unfortunately it doesn't solve the problem - when attempting to interact with a GTS acccount I'm getting `401` errors.

I'm wondering if the issue is actually the `keyId` field in the signature. See [this note in the GTS documentation](https://docs.gotosocial.org/en/latest/federation/federating_with_gotosocial/#quirks).

I'll do some fiddling and see if I can confirm that as the issue. If it is, we then will need to decide whether to try to identify GTS servers and send amended signatures, or simply ask them to fix it at their end.

WesleyAC:
Hmm, interesting. I think this should probably still be merged, since I think we will need this to communicate with GtS in general (this does solve the issue with getting a 406 that we were seeing, there are just more errors than just that — see https://github.com/superseriousbusiness/gotosocial/issues/1676), but yeah, looks like this won't fix even our side of #2794.

I think that part of the GtS documentation is just about the signatures that they send, not the ones that they receive, so I don't see how it would cause the 401 that we're seeing, but my understanding of the whole process is shaky.

Have you set up a GtS server for testing? I think seeing the logs from their side of this interaction is probably the quickest path to fix this, but I haven't tried spinning up GtS before. From what I hear it sounds pretty simple, though.

hughrun:
Yeah I agree with all that and you're right about the docs, I realised later it was kind of irrelevant.

I did set up a GTS server to see the other side, but even with debug logging on it didn't really tell me much other than, as you say, it's returning a `401` instead of a `406`. You seem to be on the right track, it looks like an issue with the signature, but the GTS logs don't really give us much more info about why:

```
timestamp="08/04/2023 06:01:47.417" func=middleware.Logger.func1.1 level=INFO latency=2.532888ms clientIP=172.28.0.1 userAgent="python-requests/2.28.2 (BookWyrm/0.6.1; +https://74ce-14-201-0-252.ngrok-free.app/)" method=GET statusCode=200 path=/.well-known/webfinger requestID=jvf78qw7040008h0hwrg msg="OK: wrote 195B"
timestamp="08/04/2023 06:01:48.232" func=middleware.Logger.func1.1 level=INFO latency="464.607µs" clientIP=172.28.0.1 userAgent="python-requests/2.28.2 (BookWyrm/0.6.1; +https://74ce-14-201-0-252.ngrok-free.app/)" method=GET statusCode=401 path=/users/editor requestID=s3gq8qw7040013kee3g0 msg="Unauthorized: wrote 96B"
timestamp="08/04/2023 06:01:49.139" func=transport.(*transport).do level=INFO pubKeyID=https://not.exactly.news/users/editor/main-key method=GET url=https://74ce-14-201-0-252.ngrok-free.app/user/bookwyrm.instance.actor#main-key requestID=8zjq8qw704001vkny29g msg="performing request"
timestamp="08/04/2023 06:01:49.863" func=middleware.Logger.func1.1 level=INFO latency=735.905463ms clientIP=172.28.0.1 userAgent=python-requests/2.28.2 method=GET statusCode=401 path=/users/editor requestID=8zjq8qw704001vkny29g msg="Unauthorized: wrote 48B"
```

hughrun (approved):
Whilst this doesn't solve the problem with GoToSocial, we should fix the header anyway, and it doesn't appear to break anything.

PR #2805, opened 2023-04-08
Stop ignoring task results

WesleyAC:
This is essentially a revert of 9cbff312a. The commit was at the advice of the Celery docs for optimization, but I've since decided that the downsides in terms of making things harder to debug (it makes Flower nearly useless, for instance) are bigger than the upsides in performance gain (which seem extremely small in practice, given how long our tasks take, and the number of tasks we have).


mouse-reeve:
Oops I thought I already merged this :)

Issue #2797, opened 2023-04-07
The book page shows a server error if the book has a series but no author


bpeel:
**Describe the bug**

A user added (or maybe imported) a book on my instance that has a series but no author. Now when you visit the page for the book an error is displayed on the page and the server logs have this:

```
web_1             | 2023-04-07T05:13:54.906069845Z   File "/usr/local/lib/python3.9/site-packages/django/urls/resolvers.py", line 698, in _reverse_with_prefix
web_1             | 2023-04-07T05:13:54.906084666Z     raise NoReverseMatch(msg)
web_1             | 2023-04-07T05:13:54.906139381Z django.urls.exceptions.NoReverseMatch: Reverse for 'book-series-by' with arguments '('',)' not found. 1 pattern(s) tried: ['series/by/(?P<author_id>\\d+)/?$']
```

**To Reproduce**

1. Create a new book.
2. Type something in the title field, the series name and the series number.
3. Don’t add an author.
4. Submit the book
5. The server will now try to show the book page but there will be an error instead.

**Expected behavior**

The book page should be shown. It should probably be fine to have a book without an author.

**Instance**

Libroj.org (BookWyrm 0.6.1)

PR #2808, opened 2023-04-08
Don't show the series as a link if the book has no author

bpeel:
The series link needs an author so if it doesn't have one, instead of showing a server error let's just show the series details as plain text without a link.

Fixes: #2797


WesleyAC (approved):
It's a little unfortunate to have separate if statements for the opening and closing tag, but I can't figure out a way to do it that would be much better than this — the thing that comes to mind is making a variable for the title of the book with the series number, but Django's `with` tag is seriously annoying to use, and concatenation of strings with it is tricky, so this seems fine as is.

Thanks for fixing this!

PR #2811, opened 2023-04-09
accept hs2019-obfuscated http signatures

kvibber:
Partial fix for #2794 (can't follow to/from GoToSocial)

hs2019 is used by some libraries to obfuscate the real algorithm per the spec https://datatracker.ietf.org/doc/html/draft-cavage-http-signatures-12

### DIFF
--- a/bookwyrm/signatures.py
+++ b/bookwyrm/signatures.py
@@ -57,7 +57,7 @@ def make_digest(data):
 def verify_digest(request):
     """checks if a digest is syntactically valid and matches the message"""
     algorithm, digest = request.headers["digest"].split("=", 1)
-    if algorithm == "SHA-256":
+    if algorithm == "SHA-256" or algorithm == "hs2019":
         hash_function = hashlib.sha256
     elif algorithm == "SHA-512":
         hash_function = hashlib.sha512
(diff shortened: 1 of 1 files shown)


WesleyAC (commented):
I think it might be cleaner to implement this similarly to how GoToSocial does it, looping through a list of hash functions regardless of the professed algorithm, maybe prioritizing the algorithm in the header if we want to be fancy.

kvibber:
Makes sense.

I'm not sure my code is quite right here - Python's not my strongest language. Also I think I may have some issues with my dev setup, since even going back to the current main branch, it isn't picking up posts from Mastodon accounts I follow on Bookwyrm (which do show up on bookwyrm.social), so I can't test it as thoroughly as I'd like.

jaschaurbach:
@WesleyAC How do we want to proceed in this one?

mouse-reeve:

This one has zero merge conflicts, but I'm not really sure where we are with it and I don't think enough of the people who weighed in on this are still involved. But if there's still interest, a new PR would be welcome.

kvibber:
The compatibility issues I was trying to fix seem to have been fixed by something else anyway, so I was probably misdiagnosing the problem to begin with.

PR #2812, opened 2023-04-10
Fix federation with GoToSocial and inconsistent KeyId in headers

### LINKED ISSUE #2794: Can't follow to or from GoToSocial

### LINKED ISSUE #2801: Remote ActivityPub actors with fragment identifiers (IRI) are not supported
A remote fediverse actor with URL fragment identifier can be searched and even asked to follow but BookWyrm reject 'Accept' and other incoming activities with http/401, note the remote proven to communicate with Mastodon and Pixelfed based instances (for example).  

An example actor could be tried is `http://fediverse.demo.openlinksw.com/dataspace/person/demo#this`  identifier `@demo@fediverse.demo.openlinksw.com`

The problem boils down to following code in signature verification:

bookwyrm/views/inbox.py
 
```
134         key_actor = urldefrag(signature.key_id).url
135         if key_actor != activity.get("actor"):  # <<< [1]
136             raise ValueError("Wrong actor created signature.")
137     
138         remote_user = activitypub.resolve_remote_id(key_actor, model=models.User) # <<< [2]
139         if not remote_user:
140             return False

```

1. The keyId is not mandatory to be a same path as actor, it is common to be in same document, but actor is not a document URL.
2. Person/Organization etc. objects are not documents, therefore these can have fragment identifiers. 

Expected flow:
Resolve the document based on `keyId`, look at publicKey/owner relation and compare with actor, if so use publicKeyPem to construct crypto key for verification, perhaps can cache the keyId publicKey/id from user profile and look for it.

Version is main branch on Github but fails with old versions as well.

hughrun:
Fixes #2801
Fixes #2794

This PR resolves a number of problems specific to GoToSocial federation but also potentially affecting other fediverse software.

## Fix making assumptions about location of `main-key`

It is legitimate to use any url for the user's key id. We have been assuming this id is the user id plus a fragment (`{userid}#main-key`) but this is not always the case, notably in the case of GoToSocial it is at `{userid}/main-key`. This commit instead checks the remote user's information to see if the key id listed matches the key id of the message allegedly received from them.

This resolves most of the `401` errors when trying to communicate with GtS servers.

## Fix Bookwyrm using wrong keyId in signed request headers

Whilst troubleshooting this it also became apparent that there is a mismatch between Bookwyrm users' keyId and the keyId we claim to be using in signed requests:

* Actual keyId defined in `User`: `{userid}/#main-key`
* keyId stated in signed request headers: `{userid}#main-key`

This commit changes the keyId listed in request headers to align with that assigned to users on creation.

## Fix GoToSOcial DMs not appearing and single-user mentions not notifying

This error was caused by GtS sending the `tag` value for these posts as a single object (dict) rather than an array containing the object, and Bookwyrm assuming it would be an array (list).

This commit wraps the dict in a list where necessary.

## Maintain backwards compatibility with legacy Bookwyrm instances

Aligning the keyId in the signatures with the real keyId of Bookwyrm users would break compatibility with all Bookwyrm versions before this change. The PR adds some logic to re-send the request with the old (wrong) keyId in the header if the first request is rejected. As Bookwyrm instances are updated this will become less of a problem.

### DIFF
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -540,7 +540,6 @@ async def sign_and_send(
 
     digest = make_digest(data)
 
-
     headers = {
         "Date": now,
         "Digest": digest,
@@ -557,14 +556,20 @@ async def sign_and_send(
                 )
                 logger.info("Trying again with legacy keyId")
                 # try with incorrect keyId to enable communication with legacy Bookwyrm servers
-                legacy_signature = make_signature("post", sender, destination, now, digest, True)
+                legacy_signature = make_signature(
+                    "post", sender, destination, now, digest, True
+                )
                 headers["Signature"] = legacy_signature
-                async with session.post(destination, data=data, headers=headers) as response: 
+                async with session.post(
+                    destination, data=data, headers=headers
+                ) as response:
                     if not response.ok:
                         logger.exception(
-                        "Failed to send broadcast with legacy keyId to %s: %s", destination, response.reason
-                    )
-                
+                            "Failed to send broadcast with legacy keyId to %s: %s",
+                            destination,
+                            response.reason,
+                        )
+
             return response
     except asyncio.TimeoutError:
         logger.info("Connection timed out for url: %s", destination)
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -529,7 +529,7 @@ async def async_broadcast(recipients: List[str], sender, data: str):
 
 
 async def sign_and_send(
-    session: aiohttp.ClientSession, sender, data: str, destination: str
+    session: aiohttp.ClientSession, sender, data: str, destination: str, **kwargs
 ):
     """Sign the messages and send them in an asynchronous bundle"""
     now = http_date()
@@ -539,11 +539,14 @@ async def sign_and_send(
         raise ValueError("No private key found for sender")
 
     digest = make_digest(data)
+    signature = make_signature(
+        "post", sender, destination, now, digest, kwargs.get("use_legacy_key")
+    )
 
     headers = {
         "Date": now,
         "Digest": digest,
-        "Signature": make_signature("post", sender, destination, now, digest, False),
+        "Signature": signature,
         "Content-Type": "application/activity+json; charset=utf-8",
         "User-Agent": USER_AGENT,
     }
@@ -554,21 +557,15 @@ async def sign_and_send(
                 logger.exception(
                     "Failed to send broadcast to %s: %s", destination, response.reason
                 )
-                logger.info("Trying again with legacy keyId")
-                # try with incorrect keyId to enable communication with legacy Bookwyrm servers
-                legacy_signature = make_signature(
-                    "post", sender, destination, now, digest, True
-                )
-                headers["Signature"] = legacy_signature
-                async with session.post(
-                    destination, data=data, headers=headers
-                ) as response:
-                    if not response.ok:
-                        logger.exception(
-                            "Failed to send broadcast with legacy keyId to %s: %s",
-                            destination,
-                            response.reason,
+                if kwargs.get("use_legacy_key") is not True:
+                    # try with incorrect keyId to enable communication with legacy Bookwyrm servers
+                    logger.info("Trying again with legacy keyId header value")
+
+                    asyncio.ensure_future(
+                        sign_and_send(
+                            session, sender, data, destination, use_legacy_key=True
                         )
+                    )
 
             return response
     except asyncio.TimeoutError:
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -540,7 +540,7 @@ async def sign_and_send(
 
     digest = make_digest(data)
     signature = make_signature(
-        "post", sender, destination, now, digest, kwargs.get("use_legacy_key")
+        "post", sender, destination, now, digest=digest, use_legacy_key=kwargs.get("use_legacy_key")
     )
 
     headers = {
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -540,10 +540,11 @@ async def sign_and_send(
 
     digest = make_digest(data)
 
+
     headers = {
         "Date": now,
         "Digest": digest,
-        "Signature": make_signature("post", sender, destination, now, digest),
+        "Signature": make_signature("post", sender, destination, now, digest, False),
         "Content-Type": "application/activity+json; charset=utf-8",
         "User-Agent": USER_AGENT,
     }
@@ -554,6 +555,16 @@ async def sign_and_send(
                 logger.exception(
                     "Failed to send broadcast to %s: %s", destination, response.reason
                 )
+                logger.info("Trying again with legacy keyId")
+                # try with incorrect keyId to enable communication with legacy Bookwyrm servers
+                legacy_signature = make_signature("post", sender, destination, now, digest, True)
+                headers["Signature"] = legacy_signature
+                async with session.post(destination, data=data, headers=headers) as response: 
+                    if not response.ok:
+                        logger.exception(
+                        "Failed to send broadcast with legacy keyId to %s: %s", destination, response.reason
+                    )
+                
             return response
     except asyncio.TimeoutError:
         logger.info("Connection timed out for url: %s", destination)
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -563,9 +563,7 @@ async def sign_and_send(
                     "Failed to send broadcast to %s: %s", destination, response.reason
                 )
                 if kwargs.get("use_legacy_key") is not True:
-                    # try with incorrect keyId to enable communication with legacy Bookwyrm servers
                     logger.info("Trying again with legacy keyId header value")
-
                     asyncio.ensure_future(
                         sign_and_send(
                             session, sender, data, destination, use_legacy_key=True
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -540,7 +540,12 @@ async def sign_and_send(
 
     digest = make_digest(data)
     signature = make_signature(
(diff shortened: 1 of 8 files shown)


bookwyrm/signatures.py
```diff
@@ -38,8 +38,9 @@ def make_signature(method, sender, destination, date, digest=None):
     message_to_sign = "\n".join(signature_headers)
     signer = pkcs1_15.new(RSA.import_key(sender.key_pair.private_key))
     signed_message = signer.sign(SHA256.new(message_to_sign.encode("utf8")))
+    # TODO: this should be /#main-key to match our user key ids, even though that was probably a mistake in hindsight.
     signature = {
-        "keyId": f"{sender.remote_id}#main-key",
+        "keyId": f"{sender.remote_id}/#main-key",
```
tsmethurst, line 43: damn this is a sneaky bug!

bookwyrm/views/inbox.py
```diff
@@ -130,15 +130,13 @@ def has_valid_signature(request, activity):
...
-        key_actor = urldefrag(signature.key_id).url
-        if key_actor != activity.get("actor"):
-            raise ValueError("Wrong actor created signature.")
-
-        remote_user = activitypub.resolve_remote_id(key_actor, model=models.User)
+        remote_user = activitypub.resolve_remote_id(activity.get("actor"), model=models.User)
         if not remote_user:
             return False
 
+        if signature.key_id != remote_user.key_pair.remote_id:
```
hughrun, line 139: The only downside of this is that it ...completely breaks compatibility between Bookwyrm instances because of the signatures problem. 🙃 Whoops. Might need a workaround.

hughrun: Where we set the user's key does have a slash, but currently when we create the signature we state it doesn't. So the proposed fix is to add the slash when signing.

I had a similar thought but it will absolutely break backwards compatibility because user keys are created when the user is created.

On the other hand, I strongly suspect that user key ids ending in `/#key-id` wasn't originally intended because it's kind of weird. If we have to account for multiple possibilities anyway it might be ok to change new keys rather than new signatures and then make the signature a slightly different way 🤔

hughrun: Hmm ok now I have woken up properly, what I wrote makes less sense. 

If we fix the signing mechanism to match the existing key creation code it will break backwards-compatabiity of signature-creation. This eventually will resolve itself by people ugrading, and in the meantime with some additional checks.

If we fix the key-creation code so that new users _don't_ have the slash in the key id, this will break backwards-compatibility of user key IDs, which can only be resolved by running a DB query to change every existing user's keys.

bookwyrm/models/status.py
```diff
@@ -136,10 +136,18 @@ def ignore_activity(cls, activity):  # pylint: disable=too-many-return-statement
         # keep notes if they mention local users
         if activity.tag == MISSING or activity.tag is None:
             return True
-        tags = [l["href"] for l in activity.tag if l["type"] == "Mention"]
+
+        # BUG: this fixes the TypeError but if there is only one user mentioned
+        # we still don't get any notifs and DMs are dropped.
+        tags = activity.tag if type(activity.tag) == list else [ activity.tag ]
```
hughrun, line 142: This change stops Bookwyrm from throwing a TypeError but doesn't actually resolve the problem of not knowing who the post mentions if it only mentions a single person. Any tips appreciated.

bookwyrm/models/activitypub_mixin.py
```diff
@@ -554,6 +555,16 @@ async def sign_and_send(
                 logger.exception(
                     "Failed to send broadcast to %s: %s", destination, response.reason
                 )
+                logger.info("Trying again with legacy keyId")
+                # try with incorrect keyId to enable communication with legacy Bookwyrm servers
+                legacy_signature = make_signature("post", sender, destination, now, digest, True)
+                headers["Signature"] = legacy_signature
+                async with session.post(destination, data=data, headers=headers) as response:
```
WesleyAC, line 562: I think it would be cleaner to make a `sign_and_send` take `use_legacy_key` and call it recursively, to avoid duplicating some of this code.

bookwyrm/views/inbox.py
```diff
@@ -130,15 +130,16 @@ def has_valid_signature(request, activity):
...
-
-        remote_user = activitypub.resolve_remote_id(key_actor, model=models.User)
+        remote_user = activitypub.resolve_remote_id(
+            activity.get("actor"), model=models.User
+        )
         if not remote_user:
             return False
 
+        if signature.key_id != remote_user.key_pair.remote_id:
+            if signature.key_id != f"{remote_user.remote_id}#main-key": # legacy Bookwyrm
```
WesleyAC, line 140: I'm a bit confused by this, isn't the legacy one the one with the `/`? And I thought `remote_user.remote_id` didn't have a `/`?

hughrun: It is incredibly confusing.

All Bookwyrm keyIds are at `{username}/#main-key` - this is set in `models.User` so it's impractical to change that because we can't retrospectively change people's keyIds.

However, the legacy _signatures_ (falsely) assert that the keyId is at `{username}#main-key`. Both are equivalent if you want to hit a url, but are not if you want to compare strings.

So if the remote user is from another Bookwyrm instance that has not upgraded, it will send a signature asserting the remote user's keyId is at `#main-key` but when our updated instance goes and checks, it will see that the keyId is actually listed as `/#main-key`. This line is a bit of a hack where our instance says "hmm, maybe it's old Bookwyrm, what if I just pretend the last forward slash isn't there?".

chdorner:
> Fix GoToSOcial DMs not appearing and single-user mentions not notifying

Haven’t looked too much into it yet but I think the same fix should also help with any interactions like boosting or liking BookWyrm posts from Takahē and those ActivityPub objects being properly handled on BookWyrm’s side.

mouse-reeve:
Is this ready to merge?

hughrun:
I think so but I'd really like someone to review and test it because it has a lot of potential to break things.

kvibber:
I brought this up to current main and ran it on a test instance for a bit. So far:

- I'm able to look up and follow users on GoToSocial, Mastodon and other Bookwyrm instances, and they're able to look up and follow back.
- statuses are federating outward to all three types of servers. (Sorry about the review spam on bw.social - I've deleted the test posts!)
- statuses from users I follow on other Bookwyrm instances are showing up
- mentions from GoToSocial are showing up in Bookwyrm, but not regular posts from a GTS user who's being followed
- same with Mastodon, which might be a regression since I'm able to follow Mastodon users from Bookwyrm.social...but might just be something broken in my test server, because when I switch back to main, it still isn't picking up new statues from Mastodon users that it follows.

hughrun:
Thanks for testing!

> * mentions from GoToSocial are showing up in Bookwyrm, but not regular posts from a GTS user who's being followed
> * same with Mastodon, which might be a regression

This is intended behaviour - Bookwyrm has never displayed standard `Note`s from people you follow unless they mention you.

So it sounds like we're good to go @mouse-reeve !

mirabilos:
Is this released? And if so, in 0.6.3?

I have many messages like…

    current:timestamp="2023-12-20T20:15:34.663Z" func=server.glob..func1.Logger.func13.1 level=INFO latency="210.618µs" userAgent="python-requests/2.31.0 (BookWyrm/0.6.3; +https://outside.ofa.dog/)" method=POST statusCode=406 path=/users/mirabilos/inbox errors="Error #01: Content-Type application/activity+json; charset=utf-8 not acceptable, this endpoint accepts: [\"application/activity+json\" \"application/ld+json;profile=https://w3.org/ns/activitystreams\"]\n" requestID=0y4xx24c04001wfddvn0 msg="Not Acceptable: wrote 50B"

… and even one…

    current:timestamp="2023-12-20T20:11:03.869Z" func=server.glob..func1.Logger.func13.1 level=INFO latency="23.278798ms" userAgent="python-requests/2.31.0 (BookWyrm/0.6.3; +https://outside.ofa.dog/)" method=GET statusCode=401 path=/users/mirabilos errors="Error #01: AuthenticateFederatedRequest: http request wasn't signed or http signature was invalid\n" requestID=4nkxn24c04001s3y0ef0 msg="Unauthorized: wrote 104B"

… in my GtS log and can’t follow my new Bookwyrm user from my GtS user (it stays on “following requested”, doesn’t get the accept).

tsmethurst:
Former is not related to this, it's an issue with GtS perhaps too strictly enforcing POST content-type as given in the spec: https://www.w3.org/TR/activitypub/#client-to-server-interactions

We could fix this by trimming `; charset=utf-8` from incoming POST request headers on GoToSocial's side.

Latter error message looks like an unsigned GET to a user profile.

mirabilos:
Ah, okay then, thanks. I can hold off further experiments until after the GtS vacation (and until the instance I’m using runs latest BookWyrm, too). I was just having an initial look at BookWyrm.

PR #2814, opened 2023-04-12
Choose installed docker compose command in bw-dev

WesleyAC:
Docker is removing support for docker-compose, and it doesn't appear to be possible to install it anymore. Instead, it has been replaced with compose V2 which is a docker plugin called with 'docker compose' (no hyphen). See https://docs.docker.com/compose/compose-v2/

Thanks to @Neriderc for noticing this in #2781.

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -23,21 +23,27 @@ trap showerr EXIT
 source .env
 trap - EXIT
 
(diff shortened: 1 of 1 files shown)


jaschaurbach:
Thank you so much for fixing this!

Issue #2817, opened 2023-04-15
deduplicate_book_data doesn’t work if the book to be removed is in a list


bpeel:
**Describe the bug**

If you run the `deduplicate_book_data` management command when a duplicate book has been added to a list then it reports an error like the following:

```
 File "/app/bookwyrm/management/commands/deduplicate_book_data.py", line 77, in handle               
    dedupe_model(models.Edition)                                                                      
  File "/app/bookwyrm/management/commands/deduplicate_book_data.py", line 65, in dedupe_model         
    update_related(canonical, obj)                                                                    
  File "/app/bookwyrm/management/commands/deduplicate_book_data.py", line 22, in update_related       
    getattr(related_obj, related_field).add(canonical)            
…
django.db.utils.IntegrityError: null value in column "order" of relation "bookwyrm_listitem" violates not-null constraint
DETAIL:  Failing row contains (5, 2023-04-15 06:25:48.864318+00, 2023-04-15 06:25:48.864335+00, null, null, t, null, null, 139, 1).
``` 

It looks like the script tries to merge the book by adding the canonical book to the book list and then removing the old one. However, this doesn’t work with a simple call to `add` because the `ListItem` model has extra fields that can’t be null such as the order. 

**To Reproduce**

1. Create a book and add an ISBN.
2. Create a second book and add the same ISBN to it.
3. Create a book list and add the second book to it.
4. Run the `deduplicate_book_data` management command like this:

```
docker-compose run --rm web python manage.py deduplicate_book_data
```

**Expected behavior**

The second book should be removed and the first book should appear in the book list. The book list information, such as the order in the list should be preserved.

bpeel:
The same thing also happens if the book to be removed is on a shelf.

PR #2818, opened 2023-04-15
Fix deduplicating books that are on a shelf or in a list

bpeel:
Previously when the deduplicate_book_data script tried to merge an edition that was on a shelf or in a list then it would fail because when the canonical book was added to the shelf or the list then it wouldn’t set the extra fields of the linking table for the “through” model of the field. These would end up defaulting to NULL, but that is not valid for some of the fields in ShelfItem and ListItem so postgres wouldn’t accept it.

To fix that, this patch makes it skip updating fields that have a non-autogenerated linking table. The linking table would appear as a separate model anyway so the book will be moved via that instead.

Fixes: #2817


jaschaurbach:
no need to merge if #2821 is merged.

PR #2821, opened 2023-04-15
Add management commands to merge a pair of editions or authors

bpeel:
This is mostly a request for comments because I’m not sure if it’s a good idea. I made this because I wanted to remove a duplicate author and book from my instance. The existing `deduplicate_book_data` management command should be able to do the trick but it has a scary warning about proceeding with caution and I was too scared to let the whole script run on my live instance. This branch adds commands which use the same code to merge objects but the pair of objects are specified explicitly with command line arguments so that if it goes horribly wrong then at least the damage is limited to a small set of things.

If you use this branch you can for example merge two editions with a command like this:

```
docker-compose run --rm web python manage.py merge_editions --canonical=27 --other=38
```

All the information from the “other” edition that is missing in the “canoncial” edition will be copied over and all comments, lists and reviews etc will be changed to point to the canonical edition. Then the “other” edition will be deleted. You can find the numbers to set by visiting the page for a book and then looking at the number in the URL.

There is a second command to merge two authors like this:

```
docker-compose run --rm web python manage.py merge_authors --canonical=37 --other=96
```

Note that this branch includes the commit from PR #2818 because it’s not very useful without it.

### DIFF
--- a/bookwyrm/management/commands/deduplicate_book_data.py
+++ b/bookwyrm/management/commands/deduplicate_book_data.py
@@ -1,7 +1,7 @@
 """ PROCEED WITH CAUTION: uses deduplication fields to permanently
 merge book data objects """
 from django.core.management.base import BaseCommand
-from django.db.models import Count
+from django.db.models import Count, ManyToManyField
 from bookwyrm import models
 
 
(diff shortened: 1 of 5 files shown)


jaschaurbach:
@mouse-reeve @chdorner I just tried it and it works like a charm. Is there anything against merging this into main? This is a feature we really need.

Maybe someone needs to write a test for it.

mouse-reeve:
Let me take a look before merging! I should be able to this afternoon

bpeel:
I’ve added another command on a [separate branch](https://github.com/bpeel/bookwyrm/tree/merge-works-command) that can be used to merge works. This has the effect of combining editions so that they are just editions of the same work instead of being separate works. Eg, if you do this:

```
docker-compose run --rm web python manage.py merge_works --canonical=37 --other=96
```

Then it will merge the works corresponding to the editions with the numbers 37 and 96. Note that the IDs are the edition numbers, not the work numbers so you can still look up the numbers by looking at the URL for a book. I don’t think the work numbers are readily accessible via the web interface.

I’ve been recklessly using all of these commands on my instance to clean up some books that got imported badly and so far I haven’t had any problems.

mouse-reeve:
That was a longer than planned afternoon! Looks good and thank you

mouse-reeve:
You can expose the work id in the UI by going to the list of editions; the url is to the work id

Issue #2760, opened 2023-03-27
Add Book Manually Loses The Cover Image Uploaded


ghost:
**Describe the bug**
When adding a book manually through the UI as a regular user, the cover image doesn't actually get saved and associated with the book.

**To Reproduce**
1. Add book manually
2. Fill out book data, inclusive of adding a cover image
3. Save
4. If linked to existing author/book/edition, the form below clearly shows that no cover image is submitted even if there was a cover image uploaded by the user on the initial form.
5. Generic cover is created and the actual cover has to be uploaded separately once the book is saved.

**Expected behavior**
Data and files entered on the form should save to the form.

**Screenshots**
Form completed with cover image
![screenshot-bookwyrm social-2023 03 27-11_45_04](https://user-images.githubusercontent.com/103219771/228010195-b1b50abf-39d2-4b3a-b6dd-3dccd7cc568e.png)

Form retains most data if you need to select existing/new author, but does not retain the file
![screenshot-bookwyrm social-2023 03 27-11_45_42](https://user-images.githubusercontent.com/103219771/228010265-6f688fbb-4a29-47c5-b52d-81ecbab03e93.png)

Cover is not uploaded as a result.
![screenshot-bookwyrm social-2023 03 27-11_46_14](https://user-images.githubusercontent.com/103219771/228010333-25a44ef0-ba1e-4470-a5d9-d4b1fde31551.png)

**Instance**
[On which BookWyrm instance did you encounter this problem.](https://bookwyrm.social/book/935881/s/how-to-be-an-mp) this example is the one I created.

**Additional context**
I've noticed this behavior before, but just re-uploaded the cover.  This was the first time I was able to screenshot the workflow to quantify the issue.

---

**Desktop (please complete the following information):**
    - OS: Manjaro/Budge
    - Browser: FireFox
    - Version: 111.0

jaschaurbach:
I have the same issue and was too lazy to fill a bug report...

hughrun:
This is caused by browser security - see this [Stack Overflow discussion](https://stackoverflow.com/questions/5293960/resubmitting-image-in-imagefield-after-validation-error-in-django/74216827).

The problem is that the page is always reloaded either due to a validation error or because we check for matching authors.

We should look into [`django-file-resubmit`](https://github.com/un1t/django-file-resubmit) as a possible solution.

PR #2827, opened 2023-04-16
Retain subjects and authors when new book form fails validation

### LINKED ISSUE #2480: Adding new books: "Confirm Book Info" dialog resets Subjects field
When manually adding a new book with Subjects, the form will refresh with the Confirm Book Info panel open at the top. Most data is kept in the form, but the Subjects section does not retain any entered information.

**To Reproduce**
1. Manually add a new book
2. Enter minimally necessary info plus a Subject or two
3. Save
4. When the form refreshes with the Confirm Book Info panel open at the top, scroll down to see that the Subjects block has been reset and is empty.

**Expected behavior**
Subject entries should be retained.

**Instance**
This was tested on bookrastinating.com

---

**Desktop:**
    - OS: MacOS 10.15 Catalina
    - Browser Safari 15.6.1
    - BookWyrm version 0.5.2

### LINKED ISSUE #2571: When validating the form for adding book manually, some fields are emptied if a field is not valid

hughrun:
Fixes #2480
Fixes #2571

Persists `book.subjects` and `add_author` when form validation fails. The problem here was that `add_author` is not a field in the `Form`, and the template logic is a bit complicated because we use the same form for creating new books and editing existing ones.

This PR does not resolve the problem of cover image uploads being dropped because this is a broader problem and is covered separately in #2760.

### DIFF
--- a/bookwyrm/views/books/edit_book.py
+++ b/bookwyrm/views/books/edit_book.py
@@ -45,6 +45,7 @@ class EditBook(View):
         data = {"book": book, "form": form}
         ensure_transient_values_persist(request, data)
         if not form.is_valid():
+            ensure_transient_values_persist(request, data, add_author=True)
             return TemplateResponse(request, "book/edit/edit_book.html", data)
 
         data = add_authors(request, data)
@@ -147,6 +148,8 @@ def ensure_transient_values_persist(request, data, **kwargs):
     if kwargs and kwargs.get("form"):
         data["book"] = data.get("book") or {}
         data["book"]["subjects"] = kwargs["form"].cleaned_data["subjects"]
+        data["add_author"] = request.POST.getlist("add_author")
+    elif kwargs and kwargs.get("add_author") is True:
         data["add_author"] = request.POST.getlist("add_author")
 
 
--- a/bookwyrm/views/books/edit_book.py
+++ b/bookwyrm/views/books/edit_book.py
@@ -103,12 +103,12 @@ class CreateBook(View):
             }
 
         if not form.is_valid():
-            ensure_transient_values_persist(request, data, form)
+            ensure_transient_values_persist(request, data, form=form)
             return TemplateResponse(request, "book/edit/edit_book.html", data)
 
         # we have to call this twice because it requires form.cleaned_data
         # which only exists after we validate the form
-        ensure_transient_values_persist(request, data, form)
+        ensure_transient_values_persist(request, data, form=form)
         data = add_authors(request, data)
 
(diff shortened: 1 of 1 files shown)


chdorner (commented):
Sorry for the delay!

Just tried it locally and it only seems to be working for subjects, but not for authors. Not sure why exactly, could you have another look?

hughrun:
@chdorner hmm I must have missed a step testing. Is this for a brand new book or an edit to an existing one?

chdorner:
> @chdorner hmm I must have missed a step testing. Is this for a brand new book or an edit to an existing one?

Ah that might have been it. I tried editing an existing book, it already had one author stored and I tried to add a second one while making the form validation fail

hughrun:
Ok I didn't look at that workflow but it makes sense that we should fix it too in the same PR - it will be in a different class but I expect it fails for the same reason.

chdorner:
I must be doing something wrong, I now get the following behaviour:

Adding new book:
1. Filling out the form with a published date validation error, and adding 3 subjects and 2 authors
2. No subject or author values are retained, all blank fields

Editing an existing book:
1. Changing the published date to fail validation, and adding 3 subjects and 2 authors
2. All 3 subjects are retained, but the authors aren't

Is that the right way to test this, or am I completely missing something here?

hughrun:
You're testing correctly, though there are two issues to test:

1. Force an error (like a wrong date)
2. Add an author, which creates a checking stage before submitting the final form

Each of these cases is slightly different but we can lose information with both because it's not all stored as `Form` values.

I'm not sure why you are still seeing errors I thought I had fixed but presumably it means I was wrong! I'll take a more exhaustive look at this, sorry about that.

hughrun:
> I must be doing something wrong, I now get the following behaviour

Ok @chdorner I've just re-tested by making a new copy of this PR on my local machine and it works as expected. I'm a bit confused as to why you can't replicate this fix.

chdorner (approved):
Not sure what I did last time, it does seem to work now as you've described. Sorry for the back and forth!

Issue #2678, opened 2023-02-17
Sorting Should Ignore “a,” “an,” and “the” 


varanasib:
**Is your feature request related to a problem? Please describe.**
Unlike the standard manner of sorting book titles, see libraries and calibre, bookwyrm considers initial articles in titles.  So, for example, "The Yiddish Policemen's Union" appears before "Siddhartha."

**Describe the solution you'd like**
Bookwyrm should sort titles in the standard manner, ignoring “a,” “an,” and “the”

**Describe alternatives you've considered**
Struggling to find books in lists sorted by bookwyrm that cannot be placed in the same order as lists from other sources.

mouse-reeve:
I think the approach here is to create a "sort title" field for the title of the book minus any initial articles, and do a best guess at populating it in a migration, and allow users to edit it.

Txopi:
They are instances with many non English contents: Spanish, French, Italian, Portuguese, Finnish, Chinese, Basque...
The English initial articles list should not interfere with other languages or the "sort title" field's value should be set differently depending on the language of the book. For example "A" in Spanish means "To" and "The" can be "El", "La", "Lo", "Los" or "Las".

hughrun:
If the `sort_title` is set at the edition level that should make it easier to set it appropriately to the language of the edition, but the rules about what to ignore/remove will obviously have to be set for each language.

mouse-reeve:
Yes, I think it would definitely be at the edition level. And I was picturing this as a field that can be manually edited, not one that is generated automatically when a book is edited, so we don't have to algorithmically plan around every language variation or unexpected title format.

ccamara:
I believe this has been fixed after #2834. Can this issue be closed?

PR #2834, opened 2023-04-25
Add support for title sort to ignore initial article

zachflanders:
Related to #2678

This PR adds to the `save` method on the `Edition` model that populates the `sort_title` based on the `title`.  The code compiles a list of articles based on the edition's languages and then uses regex to remove those articles if they appear at the beginning of the title.  

I changed the shelf and the lists to use `sort_title` instead of `title` to sort by.

I also added a (non-reverseable) migration that filters the null and blank `Edition`s and populates the `sort_title` based on the same logic.

I thought about updating the `sort_title` field on the `Book` model to be `null=False` and `blank=False`.  This might be a good idea if `sort_title` is used for sorting throughout the app?  I also thought about adding this to the `save` method on the `Book` model, but the comment `"""can't be abstract for query reasons, but you shouldn't USE it"""` kind of scared me off from adding anything to this method.

I will need help adding articles for languages other than English.


bookwyrm/migrations/0179_populate_sort_title.py
```diff
@@ -0,0 +1,41 @@
...
+from itertools import chain
+
+from django.db import migrations, transaction
+from django.db.models import Q
+
+from bookwyrm.settings import LANGUAGE_ARTICLES
+
+
+@transaction.atomic
+def populate_sort_title(apps, schema_editor):
```
mouse-reeve, line 11: I think this will work well for smaller instances, but for bigger ones like bookwyrm.social, it's going to be too intensive to iterate over all editions. This could work as a management command which admins can run optionally, however. What do you think?

bookwyrm/models/book.py
```diff
@@ -363,6 +365,16 @@ def save(self, *args, **kwargs):
             for author_id in self.authors.values_list("id", flat=True):
                 cache.delete(f"author-books-{author_id}")
 
+        # Create sort title by removing articles from title
+        if self.sort_title is None:
+            articles = chain(
+                *(LANGUAGE_ARTICLES[language] for language in self.languages)
+            )
+            icase_articles = (
+                f"[{a[0].capitalize()}{a[0].lower()}]{a[1:]}" for a in articles
```
mouse-reeve, line 374: Would it be easier just to fully lowercase everything here, rather than trying to use title case?

bookwyrm/models/book.py
```diff
@@ -363,6 +365,16 @@ def save(self, *args, **kwargs):
             for author_id in self.authors.values_list("id", flat=True):
                 cache.delete(f"author-books-{author_id}")
 
+        # Create sort title by removing articles from title
+        if self.sort_title is None:
+            articles = chain(
+                *(LANGUAGE_ARTICLES[language] for language in self.languages)
```
mouse-reeve, line 371: You can use `LANGUAGE_ARTICLES.get(language)` here to avoid throwing an error when the language isn't known

mouse-reeve (changes requested):
Thank you so much! A few small changes here just based on looking over it, I haven't gotten a chance to run it yet.

bookwyrm/models/book.py
zachflanders: 👍 I agree, especially since the `sort_title` will not be displayed on the site.

mouse-reeve: I think a logical next step from here is to let users set custom sort titles in the edit book view, so it would show up there, but imo that's out of scope for this PR

bookwyrm/migrations/0179_populate_sort_title.py
zachflanders: Do you think something like [`bulk_update`](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#bulk-update) would work for larger instances?

mouse-reeve: That's a good question! I don't know -- on b.s, there are about 700,000 editions, if that helps contextualize it. @WesleyAC do you have a sense?

zachflanders: There are probably some edge cases were "A" is used as a noun in the title such as [A is for Activist](https://bookwyrm.social/book/257447/s/a-is-for-activist).  We could probably get fancy and exclude titles that start with "A is" or even fancier with "A [verb]", but then we open a can of worms implementing custom rules for every language.  Perhaps the planned addition of the ability to edit the sort title on the book edit page mitigates this concern?

bookwyrm/models/book.py
```diff
@@ -363,6 +365,19 @@ def save(self, *args, **kwargs):
...
+            if self.sort_title in [None, ""]:
+                articles = chain(
+                    *(
+                        LANGUAGE_ARTICLES.get(language, ())
+                        for language in tuple(self.languages)
+                    )
+                )
+                self.sort_title = re.sub(
+                    f'^{" |^".join(articles)} ', "", str(self.title).lower()
+                )
```
zachflanders, line 379: There are probably some edge cases were "A" is used as a noun in the title such as [A is for Activist](https://bookwyrm.social/book/257447/s/a-is-for-activist). We could probably get fancy and exclude titles that start with "A is" or even fancier with "A [verb]", but then we open a can of worms implementing custom rules for every language. Perhaps the planned addition of the ability to edit the sort title on the book edit page mitigates this concern?

mouse-reeve: The edge cases are endless! I think letting people manually edit is the way.

bookwyrm/templates/book/edit/edit_book_form.html
```diff
@@ -28,6 +28,15 @@ <h2 class="title is-4">
...
                 </div>
 
+                <div class="field">
+                    <label class="label" for="id_sort_title">
+                        {% trans "Sort Title:" %}
+                    </label>
+                    <input type="text" name="sort_title" value="{{ form.sort_title.value|default:'' }}" maxlength="255" class="input" required="" id="id_sort_title" aria-describedby="desc_sort_title">
+
+                    {% include 'snippets/form_errors.html' with errors_list=form.sort_title.errors id="desc_sort_title" %}
+                </div>
```
zachflanders, line 38: I added the sort title to the edit book form.  I wonder if more context is needed for users to understand what a sort title is used for?  For example I could imagine an information icon next to the label with a description that appears in a tooltip on hover?  I could also imagine this hidden in an expandable accordion type ui element under "advanced" in order to keep this form from being too cluttered?

Screen Shot 2023-04-26 at 11 02 44 PM

bookwyrm/migrations/0179_populate_sort_title.py
zachflanders: Thinking about this some more I added some logic to slice the queryset using batches.  This way we don't try to hold all the editions in memory at the same time.  The `bulk_update` should limit the # of queries.  I would love to get additional thoughts on this and whether it would work.

```python
    editions_wo_sort_title = Edition.objects.using(db_alias).filter(
        Q(sort_title__isnull=True) | Q(sort_title__exact="")
    )
    batch_size = 50000
    start = 0
    end = batch_size
    while editions_wo_sort_title[start: end]:
        Edition.objects.bulk_update((
            set_sort_title(edition)
            for edition in editions_wo_sort_title[start: end]
        ), ["sort_title"])
        start = end
        end += batch_size
```

WesleyAC: @zachflanders if it's something you're comfortable with, could you change the postgres `log_statement` setting to `all` and paste the postgres logs from running this migration here (or probably a link to a gist)? that would help in determining how impactful it's likely to be.

if not, i can take a look at it, but it might take me a little while to get to it.

zachflanders: @mouse-reeve @WesleyAC With a little help from [this SO answer](https://stackoverflow.com/a/71270291/3669465) I refactored this a bit.  Here is a link [a gist of the database logs](https://gist.github.com/zachflanders/d14a9035273626cf90d1dbbe1cb91237).  I have 522 editions in my local dev environment, and I lowered the batch size to 20 in order to mimic the number of iterations a large instance like BS would likely have. (Of course each iteration will update many more rows, but it would be the same number of queries)

bookwyrm/templates/book/edit/edit_book_form.html
mouse-reeve: I do think this (and a number of other fields) could use some clarification, but I think it's outside the scope of this PR, since I want to make sure we approach it with consistent UI and markup

mouse-reeve:
I checked this out (apologies for the delay!) and it works really well in my test instance. @WesleyAC do you think we can run this migration successfully? If you don't have capacity, another option would be to 1) fake this migration for b.s, 2) use a management command rather than a migration to set the sort titles (which would be more coding work, but is something I could do in the next couple weeks if that would be best)

WesleyAC:
I think the batch size should be smaller than 50,000 (1000 strikes me as more reasonable — 50,000 might actually be better, but I'm not sure exactly how well postgres would handle the large number of `THEN` clauses in the `UPDATE`, and I am pretty sure that it can handle 700 batches of 1000 reasonably well), but other than that this looks good in terms of DB performance, I don't see anything concerning in the `EXPLAIN` results.

mouse-reeve (approved):
I went in and changed the limit, so this is ready to go. Thank you so much for all the work on this! I will make a new release in the next week or two.

PR #2845, opened 2023-05-09
2712: Add class is-danger to fields with errors

### LINKED ISSUE #2712: Better signal which form fields have validation errors
This is coming out of a discussion in #2695.

We can add the `is-danger` class to input fields in Bulma to get these fields a red border, making it easier to see which fields need attention from the user.

Lakoja:
Fixes #2712

Draft because of the two "this is complicated" questions.

### DIFF
--- a/bookwyrm/forms/custom_form.py
+++ b/bookwyrm/forms/custom_form.py
@@ -1,5 +1,6 @@
 """ Overrides django's default form class """
 from collections import defaultdict
+from django.core.exceptions import ValidationError
 from django.forms import ModelForm
 from django.forms.widgets import Textarea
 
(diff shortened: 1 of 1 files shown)


bookwyrm/forms/custom_form.py
```diff
@@ -34,3 +35,43 @@ def save(self, request, *args, **kwargs):
         """Save and check perms"""
         self.instance.raise_not_editable(request.user)
         return super().save(*args, **kwargs)
+    
+    # def clean(self):
+    #     self.cleaned_data = super().clean()
+
+    #     print("Clean")
+    #     print(self.errors)
```
Lakoja, line 43: This is empty at this position.

Which surprises me after reading documentation...

PR #2850, opened 2023-05-12
Use the translated shelf name in the message for when another edition is shelved

bpeel:
In the page for an edition, if the user has a different edition of the same work on a shelf then a message is shown. Currently the shelf name for status shelves isn’t translated so this PR fixes that.


mouse-reeve:
Perfect, thank you for fixing this!

PR #2852, opened 2023-05-15
Use the translated shelf name in the “remove from” shelf button

bpeel:
I found another place where the shelf name wasn’t translated. Sorry to keep pestering the devs with these…


mouse-reeve:
This is the opposite of pestering, really appreciate the fix

PR #2854, opened 2023-05-16
Record report actions

### LINKED ISSUE #2505: Show history on reports
A report should show a little timeline with who reported the ticket, what actions were taken related to it, comments on it, and who closed it

mouse-reeve:
This table will now track all actions taken on a report, like resolving it, re-opening it, suspending the reported user, et cetera, in addition to comments. When there are multiple admins, this change will make it easier to understand what actions have been taken by whom on a report.

Work:
 - [x] update model
 - [x] track resolve/re-open
 - [x] track suspend/un-suspend
 - [x] track block/approve link domains
 - [ ] ~track messaging reporter and reportee~
 - [x] track deleting statuses
 - [x] display actions taken

Fixes #2505


mouse-reeve:
On second thoughts, I think recording DMs sent would be unnecessarily convoluted. It might make more sense (another time) to have a different way to updating users about actions taken on reports. @bookwyrm-social/code-review could someone take a look at this?

hughrun:
I can take a look in a couple of days unless someone else gets to it first, which they are very welcome to do! I've not used the admin interface but is there any way to store free text notes against a report? If so, that should be stored, and could possibly be included in a message/email to the reporter when the report is resolved?

mouse-reeve:
You can write plain text notes in reports (that's been true prior to this PR). Right now there's just a link to DM the reporter which autofills with "update on your report:" or something, which is kinda janky :)

Here's what a report action log (in this case, on a link domain reported as spam) might look like:
Screen Shot 2023-05-30 at 1 49 42 PM

mouse-reeve:
Now that I'm looking at it, I wonder if the chronology should be reversed 🤔

jaschaurbach:
> Now that I'm looking at it, I wonder if the chronology should be reversed 🤔

As the comment box is at the bottom it would be good to have the order reversed for reading consistency.

mouse-reeve:
Sorry for the long delay on a very small change! This is ready for review

bookwyrm/templates/settings/users/user_moderation_actions.html
```diff
@@ -49,7 +49,7 @@ <h3>{% trans "User Actions" %}</h3>
 
         {% if user.local %}
         <div>
-            <form name="permission" method="post" action="{% url 'settings-user' user.id %}">
+            <form name="permission" method="post" action="{% url 'settings-user' user.id report.id %}">
```
hughrun, line 52: The `UserAdmin` view doesn't expect a `report_id` keyword so trying to change user permissions throws a `TypeError`.

I'm guessing this is supposed to have a check for the the presence of a `report_id` kwarg and then run a `models.Report.record_action()`?

bookwyrm/templates/settings/reports/report.html
```diff
@@ -61,30 +62,59 @@ <h3 class="title is-4">{% trans "Reported links" %}</h3>
...
+                    <span class="tag">{{ comment.created_date }}</span>
+                </div>
+
+                {% if comment.note %}
+                <blockquote>{{ comment.note }}</blockquote>
+                {% endif %}
+            </li>
+            {% endfor %}
+
+            <li class="mb-2">
```
hughrun, line 96: Suggest moving this to the top, otherwise everything runs top to bottom except the initial opening of the report, which is kinda weird.

bookwyrm/views/admin/reports.py
```diff
@@ -81,41 +82,42 @@ def get(self, request, report_id):
...
-        )
+        note = request.POST.get("note")
+        report.comment(request.user, note)
         return redirect("settings-report", report.id)
 
 
 @login_required
 @permission_required("bookwyrm.moderate_user")
-def suspend_user(request, user_id):
+def suspend_user(request, user_id, report_id=None):
```
hughrun, line 92: It seems strange to me that the suspended user is not notified - they should probably get an email telling them their account is suspended due to a moderation decision?

bookwyrm/views/admin/reports.py
```diff
@@ -81,41 +82,42 @@ def get(self, request, report_id):
...
     user.save(broadcast=False)
+
+    models.Report.record_action(report_id, USER_SUSPENSION, request.user)
     return redirect_to_referer(request, "settings-user", user.id)
 
 
 @login_required
 @permission_required("bookwyrm.moderate_user")
-def unsuspend_user(request, user_id):
+def unsuspend_user(request, user_id, report_id=None):
```
hughrun, line 106: Unsuspended users should also be notified.

hughrun (changes requested):
This looks mostly good. There's a bug, and what I guess is an enhancement suggestion.

I also noticed another bug when testing this that is not caused by anything in your code (the javascript for reporting a user is a little buggy) - I'll log an Issue for that.

bookwyrm/views/admin/reports.py
mouse-reeve: I'm not sure! Almost all of the users I've suspended so far have been spam accounts, and I'm not sure if it's counterproductive to notify them? But I see how it _does_ make sense in a case of an actual person who has been behaving poorly. I think this is out of scope of this PR, however. I'll open an issue for it.

mouse-reeve:
I'm working on the prettier complaints in #2903

mouse-reeve:
@hughrun This has been sitting for a bit and I think it's ready to go so I'm going to go ahead and merge. Thank you for the feedback on it!

Issue #2860, opened 2023-05-27
Reactivation of permanently deleted user account


MQ-xz:
**Describe the bug**
Reactivation of any permanently deleted user account can be possible to restore by the user via confirmation of the email address or using the reactivation method of a deactivated user account.

Detailed report: https://huntr.dev/bounties/d294ea37-b067-49bd-a806-2a893a1c8b18/

PR #2861, opened 2023-05-27
Fix reactivation of deleted account

MQ-xz:
Fix: #2860


MQ-xz:
Hey @jaschaurbach, I can fix the lint issues but I don't what the issues with test.

MQ-xz:
> @MQ-xz Would you be able to fix both errors?

mouse-reeve:
I was still able to replicate the bug with this patch -- I've opened an alternative pull request that I think closes this issue: https://github.com/bookwyrm-social/bookwyrm/pull/2879

PR #2868, opened 2023-06-06
Export/Import user (aka "migrate")

### LINKED ISSUE #1012: Migrate Accounts To Another Instance & Import/Export
**Migrate Account To Another Instance**
Suggesting the ability to migrate a BookWyrm account, and all information associated with it (lists, reviews, etc.) to a new instance. 

Cloning/syncing instead of simply migrating. This would allow a user to have a backup account on another instance in the case one goes down. The user could input the second instance information on the other account and have it sync once a day automatically. 
The user could always delete the account on the main instance if they want to migrate, so this is migration but with optional sync functionality.

**Export/Import**
Suggesting the ability to export/import a profile (via file) to store locally, as well as shelves/lists/reading progress.
This could potentially also be offered in a .json format or similar if users would like to utilize it in other programs. 
But it can also be used to import their content into other instances, or have a local copy.

hughrun:
## Preliminary work on an Export/Import flow for users

**This is very preliminary and subject to change, potentially in major ways.**

Intended to resolve #1012 

Users can export a JSON file and import the file into another instance.
Uses `serializers.serialize` so this is just a straight Django dump, not an ActivityPub JSON format.

- [x] import user profile
- [x] import selected user settings
- [x] import goals
- [x] import books (as combined `Book` and `Edition` objects) and their authors
- [x] import shelves and ShelfBooks
- [x] import lists and ListItems
- [x] import reviews (as combined `Review` and `Status` objects)
- [ ] import readthroughs
- [ ] saved lists
- [ ] follows
- [ ] user blocks
- [ ] load book images
- [ ] flip curation on group-curating lists (see below)
- [ ] use task manager for imports

## Related future work

We will also need a way to indicate in the old instance that the user has moved. I am anticipating we'll do that in a separate PR, using a `Move` ActivityPub action similar to how Mastodon does it ("Alice has `Move`d Alice to Bob") but also possibly `Move`ing book lists.

## Exclusions

* I initially did not intend to migrate quotes and comments but we could do this fairly easily.
* as `Group`s are not _currently_ ActivityPub-aware objects, I have excluded them, though in future these need to be ActivityPub objects and therefore included. Not this also means we should flip any Group curated lists back to default. i.e. Groups can only exist between users who are all _local_ users.

### DIFF
--- a/bookwyrm/forms/forms.py
+++ b/bookwyrm/forms/forms.py
@@ -24,6 +24,7 @@ class FeedStatusTypesForm(CustomForm):
 class ImportForm(forms.Form):
     csv_file = forms.FileField()
 
+
 class ImportUserForm(forms.Form):
     json_file = forms.FileField()
 
--- a/bookwyrm/forms/forms.py
+++ b/bookwyrm/forms/forms.py
@@ -24,6 +24,9 @@ class FeedStatusTypesForm(CustomForm):
 class ImportForm(forms.Form):
     csv_file = forms.FileField()
 
+class ImportUserForm(forms.Form):
+    json_file = forms.FileField()
+
 
 class ShelfForm(CustomForm):
     class Meta:
--- a/bookwyrm/importers/__init__.py
+++ b/bookwyrm/importers/__init__.py
@@ -1,6 +1,7 @@
 """ import classes """
 
 from .importer import Importer
+from .bookwyrm_import import BookwyrmImporter
 from .calibre_import import CalibreImporter
 from .goodreads_import import GoodreadsImporter
 from .librarything_import import LibrarythingImporter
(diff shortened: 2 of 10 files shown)


hughrun:
Work for this is happening at https://codeberg.org/GuildAlpha/bookwyrm for now and will come back here eventually.

mouse-reeve:
thank you SO much for working on this and feel free to ping me here or on mastodon if I can answer any questions

skobkin:
By the way, @mouse-reeve, does bookwyrm support move activity right now? If not, it'd be good to have it with migration to be able to automatically move followers to the new instance.

hughrun:
@skobkin Not yet. My plan is to do this in 2 steps: first just enable user export/import, then add `Move` as a separate PR immediately following - for user ids and potentially lists. The second should be fairly straightforward once we sort out export and import, though I guess technically they don't rely on each other.

skobkin:
@hughrun Sounds awesome.

Sorry if it wasn't clear, but I didn't mean that it should be implemented in this PR.

hughrun:
closing this in favour of #2980

PR #2897, opened 2023-07-09
[WIP] Add typechecking using mypy

### LINKED ISSUE #2722: Consider adding types via mypy
I find that Python is significantly easier to write and maintain when it's statically typed via [mypy](https://www.mypy-lang.org/). It's fairly easy to [add types incrementally](https://mypy.readthedocs.io/en/stable/existing_code.html). [Here's a good post](https://blog.zulip.com/2016/10/13/static-types-in-python-oh-mypy/) on how Zulip started using it (also a Django project, albeit larger than Bookwyrm) talking about benefits and approaches to migrating.

The steps for this would probably look like:

* Get mypy running on the codebase without annotations, using `# type: ignore` where needed
* Add mypy to test script / CI
* Require type annotations, but exempt all existing code on a per-module basis. This will require new modules to be typed, without requiring huge changes all at once.
* Slowly, over time, add type annotations for existing modules, and remove the exemptions as we do so.

I'd be happy to drive this if it's something people are open to.

jderuiter:
- [ ] Add mypy checks to the test script
- [ ] Add mypy checks to CI
- [ ] Add type annotations

Fixes #2722

### DIFF
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -22,6 +22,8 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements.txt
     - name: Analysing the code with mypy
+      env:
+        DOMAIN: your.domain.here
       run: |
         mypy bookwyrm celerywyrm
 
--- a/.github/workflows/mypy.yml
+++ b/.github/workflows/mypy.yml
@@ -23,7 +23,29 @@ jobs:
         pip install -r requirements.txt
     - name: Analysing the code with mypy
       env:
+        SECRET_KEY: beepbeep
+        DEBUG: false
+        USE_HTTPS: true
         DOMAIN: your.domain.here
+        BOOKWYRM_DATABASE_BACKEND: postgres
+        MEDIA_ROOT: images/
+        POSTGRES_PASSWORD: hunter2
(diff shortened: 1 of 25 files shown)


mouse-reeve:
I merged in `main` to fix the failing Prettier workflow which wasn't related to your work. And thank you for opening this PR!

jderuiter:
Thank you! I've been adding type annotations to the code, but I'm wondering what the best way forward is. I'm not familiar enough with the code base and Django internals to always know the intent of the code to add the correct types. For now I just added questions I had in TODOs. What do you think would be the best way to do this?

mouse-reeve:
I think it may be a good idea to break this into a series of pull requests, doing one module at a time. That way there's less risk of merge conflicts, and we can address any fixes in more manageable chunks. I'm on my phone right now but when I'm back at my computer I can take a closer look at the custom types. But I think starting with one or two relatively simple files might be a nice way to get going

jderuiter:
Sounds good. I created a smaller pull request with just the mypy tooling and some first annotations ( #2916). Once that is merged we can start looking at the other annotations I already added.

mouse-reeve:
I'm closing this since you've been adding incremental typing fixes

Issue #2907, opened 2023-07-20
Split Celery queues into more functional categories


WesleyAC:
Currently, we have five celery queues: Low priority, Medium priority, High priority, Imports, and Broadcasts. I propose (as described in https://github.com/bookwyrm-social/bookwyrm/issues/2874#issuecomment-1643027240), splitting these up into more functionality-related categories (seeing as the different "priority" queues are not actually prioritized differently):

> In general, switching from the current queue "priority" system to a more functionality-based system would I think probably be preferable. The current status quo is that everything in the "high-priority" queue will be behind by the same amount of wall-clock time, which makes everything in it equally slow, when it might be better to have some particular tasks fall behind, while other (less common or compute intensive) tasks are operating in real time.

In particular, I propose starting with the following queues:

* Suggested users
  * `bookwyrm.suggested_users.bulk_add_instance_task`
  * `bookwyrm.suggested_users.bulk_remove_instance_task`
  * `bookwyrm.suggested_users.remove_suggestion_task`
  * `bookwyrm.suggested_users.remove_user_task`
  * `bookwyrm.suggested_users.rerank_suggestions_task`
  * `bookwyrm.suggested_users.rerank_user_task`
* Images
  * `bookwyrm.preview_images.generate_edition_preview_image_task`
  * `bookwyrm.preview_images.generate_site_preview_image_task`
  * `bookwyrm.preview_images.generate_user_preview_image_task`
  * `bookwyrm.preview_images.remove_user_preview_image_task`
  * `imagekit.cachefiles.backends._generate_file`
* Email
  * `bookwyrm.emailing.send_email`
* Broadcast
  * `bookwyrm.models.activitypub_mixin.broadcast_task`
* ActivityStreams
  * `bookwyrm.activitystreams.add_status_task`
  * `bookwyrm.activitystreams.add_book_statuses_task`
  * `bookwyrm.activitystreams.add_user_statuses_task`
  * `bookwyrm.activitystreams.handle_boost_task`
  * `bookwyrm.activitystreams.populate_stream_task`
  * `bookwyrm.activitystreams.remove_book_statuses_task`
  * `bookwyrm.activitystreams.remove_status_task`
  * `bookwyrm.activitystreams.remove_user_statuses_task`
* Imports
  * `bookwyrm.models.import_job.import_item_task`
  * `bookwyrm.models.import_job.start_import_task`
* Connection Manager
[...]

PR #2908, opened 2023-07-20
Switch from priority queues to function-based queues

WesleyAC:
Fixes: #2907


WesleyAC:
@mouse-reeve it would be nice to get this merged and deployed while we still have a bunch of imports in the queue (in order to see what kind of effect it has), if that's possible.

PR #2912, opened 2023-07-21
fix pointed ids of labels in invite request form in admin

categulario:
I noticed while using the admin interface that clicking on the "Set a question for invite requests" label didn't toggle the checkbox even though the input is inside the label. Turns out the reason is the `for=""` attribute that is not properly set.


joachimesque (approved):
Looks good, but it could also work without the `for` attribute: the `<input>` element is wrapped inside the `<label>`.

categulario:
most certainly.

joachimesque:
The simplest code is the better in my opinion. If we don’t need the id, there’s less risk of error if the Form has to change :)

categulario:
I updated the code based on this conversation

categulario:
Let me know if mor changes are necessary, or if a rebase is desirable

mouse-reeve:
Perfect, thank you so much!

PR #2916, opened 2023-07-22
mypy integration

jderuiter:
- Added mypy to bw-dev and workflow
- Some first annotations for celerywyrm

When this is merged we can start gradually adding types for the other modules


mouse-reeve:
Looks good! Thank you so much, I'm really excited to have this

PR #2918, opened 2023-07-23
Include Dutch locale

jderuiter:
The translation seems to be complete so this includes the Dutch locale. The translation is probably not perfect (at least not the part I did), but hopefully this will also help to get some feedback on that.


axiomizer (approved):
Looks good to me, I changed my preferred language to Dutch and the translated strings seemed to be displayed correctly.

Issue #2892, opened 2023-07-05
UX improvement: ISBN "Copy" button and dashed formatting 


soshial:
**Describe the proposition**
1. I propose a small COPY icon button close to ISBN number to copy it (see a red area on the screenshot):
2. Also, I would love to see formatted ISBNs for easier reading. This can be [done with JavaScript](http://inventaire.github.io/isbn3/?isbn=9781781682135).

**Expected behavior**

**Screenshots**
<img width="293" alt="Screenshot 2023-07-05 at 12 48 22" src="https://github.com/bookwyrm-social/bookwyrm/assets/100644/a7278583-6e50-48d4-a4b3-afe4872da529">

**Instance**
Main instance.

**Additional context**

mouse-reeve:
This should be in the next release! Thank you both so much :)

PR #2921, opened 2023-07-25
Isbn hyphenation

axiomizer:
For #2892 

Here is information about the range message: https://www.isbn-international.org/range_file_generation. That page also has a link to download the range message, and provides the URL where you can send a request for the range message. I happened to notice that the range message had 1 very minor update over a period of about 3 days. This could help inform a decision about how often, or if, we need to update the range message in production. In this PR I included a function that can be used to update the range message, but it isn't called anywhere.


axiomizer:
As detailed [here](https://en.wikipedia.org/wiki/ISBN), the ISBN-13 has 5 parts: GS1 prefix, registration group, registrant, publication, and checksum. The GS1 prefix is always 3 digits, and the checksum is always 1 digit. The other parts are variable length, and the range message helps determine their length for a given ISBN-13.

axiomizer:
Tests:

`>>> from bookwyrm.isbn.isbn import hyphenator_singleton as hyphenator
>>> hyphenator.hyphenate('9780439554930')  # 978-0 (English language) 3700000-6389999
'978-0-439-55493-0'
>>> hyphenator.hyphenate('9782070100927')  # 978-2 (French language) 0000000-1999999
'978-2-07-010092-7'
>>> hyphenator.hyphenate('9783518188125')  # 978-3 (German language) 2000000-6999999
'978-3-518-18812-5'
>>> hyphenator.hyphenate('9784101050454')  # 978-4 (Japan) 0000000-1999999
'978-4-10-105045-4'
>>> hyphenator.hyphenate('9786269533251')  # 978-626 (Taiwan) 9500000-9999999
'978-626-95332-5-1'
>>> hyphenator.hyphenate('9798627974040')  # 979-8 (United States) 4000000-8499999
'979-8-6279-7404-0'
>>> hyphenator.hyphenate('9786268533251')  # 978-626 (Taiwan) 8000000-9499999 (unassigned)
'9786268533251'
>>> hyphenator.hyphenate('9786769533251')  # 978 range 6600000-6999999 (unassigned)
'9786769533251'
>>> hyphenator.hyphenate('9798311111111')  # 979-8 (United States) 2300000-3499999 (unassigned)
'9798311111111'
>>>`

bookwyrm/views/books/books.py
```diff
@@ -90,6 +91,7 @@ def get(self, request, book_id, **kwargs):
             "rating": reviews.aggregate(Avg("rating"))["rating__avg"],
             "lists": lists,
             "update_error": kwargs.get("update_error", False),
+            "hyphenated_isbn13": hyphenator.hyphenate(book.isbn_13),
```
mouse-reeve, line 94: I wonder if this would make more sense as a property on the book model? There are some generated-on-the-fly fields there already: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/models/book.py#L75

axiomizer: Sure, I don't know where it makes the most sense but I don't mind changing that

axiomizer: Yeah I think that makes a lot of sense

mouse-reeve:
In my naivety I would have thought you could solve this with a simple regex! Thank you for such a thorough PR. I think it may make the most sense to update the ranegfile from a management command? But that's a question for another day.

Issue #2671, opened 2023-02-16
Edit Page Should Say What Is Being Edited


varanasib:
**Is your feature request related to a problem? Please describe.**
All edits occur on a page entitled "Edit status."  This is confusing since one might be editing a review, quote, or comment.

**Describe the solution you'd like**
The title of the edit page should reflect the type of thing one is editing.

Issue #2668, opened 2023-02-16
Indicate That Edit Is Recorded/Saved


varanasib:
When one edits a post and hits the Post button, there is no indication that the edit has been recorded.  Nothing on the edit page changes.

**Describe the solution you'd like**
At a minimum, the Post button (which should probably be labeled Save, see #2642) should disappear or become greyed out.  Ideally, a message saying something like, "Thanks for the edit.."

chdorner:
This is partially a regression of #2544. The behaviour before that PR wasn't ideal either, it redirected to the home page, but at least it didn't stay on the form.
I'd say ideally after successfully editing a status it should redirect to the single status page, i.e. `/user/:username/:status_type/:status_id`

PR #2922, opened 2023-07-27
fix button text when editing posts

### LINKED ISSUE #2642: Change "Post" to "Save" on edit page
On the page to edit a quote (and possibly a review/status/etc as well, I haven't checked), the button to save the edit says "Post" — this is confusing, since it implies (to me) that a new post is being made.

Instead, the button should say "Save" or something similar.

tararoshan:
Indicate edited post is saved

Should close #2642 and #2671. Gives a minimal solution to #2668.


hughrun (approved):
Thanks for this contribution @tararoshan! 

This does fix #2642

For #2671 we need to change the heading so that is says "Edit Comment", "Edit Review" etc so that will be a separate PR.

#2668 really needs to move the user out of the edit screen, though I agree this is a partial fix/improvement.

Issue #2901, opened 2023-07-14
Hashtag behaviour not compatible with Mastodon


arjanboltjes:
I noticed two things:
1. Hashtags in Bookwyrm are not formatted the same as in Mastodon.
2. When clicking the hashtag in Mastodon, my Bookwyrm comment does not show up.

Ad 1.
I added `#cycling-related` to my comment. In Mastodon, the hashtag would then be `#cycling`, the hashtag being terminated at the dash. Bookwyrm does not do this. In Mastodon, you now see the complete hashtag, but it doesn't work: it yields a "The page you are looking for isn't there". On Bookwyrm, the `#cycling-related` hashtag is created, and actually works. I had expected it to be `#cycling`.

Ad 2.
I added other hashtags that were the same between Mastodon and Bookwyrm, e.g. `#roadbike`, but when clicking this hashtag on Mastodon does not find my Bookwyrm comment, despite the comment finding its way to Mastodon - I follow my Bookrastinating account. A Mastodon comment to the Bookwyrm comment carrying the same hashtag is found.
This suggests to me incomplete federation of Bookwyrm content into Mastodon. Is this expected behaviour? 

The other way around I do see behaviour that I had expected: when clicking the hashtag in Bookwyrm, I do not expect Mastodon posts to pop up since Mastodon content is not generally visible in Bookwyrm, and this is indeed the case. I only found my own comment and the Mastodon comment to this comment. Apparently, no one else on Bookrastinating has used this hashtag. 

---

**Desktop:**
    - OS: Linux - Kubuntu
    - Browser: Firefox
    - Version: 114.0.1 (64-bit)

jderuiter:
For point 2: My guess is that something is not going quite right with the serialisation for the hashtags (or at least not the same as with Mastodon). In the ActivityPub message I see the following for tags:
```
'tag': [{'type': 'Hashtag', 'href': 'https://your.domain.here/hashtag/1', 'name': '@#content'}]
```
It looks like there is a superfluous `@` in the name, which I expect would explain why Mastodon does not pick up the tag.

arjanboltjes:
Recently I posted on Mastodon concerning hashtag behaviour that very much seems to resemble the point 2 of my original post above.

Since I did not receive any reaction from bookwyrm or bookrastinating Mastodon accounts, I'll post it here, too:

[...]

PR #2923, opened 2023-07-28
Remove '@' from the beginning of hashtag names in activity

jderuiter:
All tag names got a '@' as prefix when converting to activity, which you would not expect for hashtags as this results in the name '@#hashtag'. I added an exception for hashtags. This addresses the second point in #2901

Should the default behaviour for tags actually be that the name is preceded by '@' or should this maybe just be the exception for Users?


mouse-reeve (approved):
Thank you so much! Looks good

PR #2925, opened 2023-07-28
Type annotations and related changes for connectors

jderuiter:
Mostly type annotations, but a couple of changes to the code to handle potential undefined or None values. Probably wouldn't hurt if someone a bit more familiar with the code would go over it to see whether it makes sense like this.

The `create_edition_from_data` method seems a bit funny as it expects different arguments in the inventaire connector compared to its parent `AbstractConnector`, which results in a Union type.


mouse-reeve (approved):
Looks good to me, thank you so much for digging into what is one of the gnarlier parts of the codebase

PR #2927, opened 2023-07-30
Reduce activity in streams queue from boost tasks

mouse-reeve:


WesleyAC:
I don't have a good enough understanding of this code to validate this very well (I don't understand the reason that boosts trigger saving twice, for instance), but from your description of the problem this looks correct. If you want I can look into this more / try to test it, but I don't feel that I need to unless you're wary of this change.

WesleyAC (approved):

mouse-reeve:
It's a pretty safe change imo and I tested it locall

PR #2929, opened 2023-07-30
Disable Spellcheck on header search inputs

joachimesque:
The header search input shouldn't accept automatic spellcheck. Searching for a book or an author means entering words that are not in the dictionary (like author names, invented places, etc.), and I shouldn't have my computer or mobile phone automagically change these words just before I submit the search

On the technical side, here's the doc for the `spellcheck` attribute on `input` elements: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#spellcheck

(This PR hasn't been tested yet)


hughrun (approved):
Tested in Safari, looks good.

joachimesque:
Thank you!

PR #2934, opened 2023-08-01
Only trigger add_status_task when status is first created

mouse-reeve:
I think the reason I didn't do this initially was so that related users and books, which are added necessarily after the model instance is crated, will be part of the object when the task runs, but I have investigated this and because of the transaction.atomic statement in the to_model method in bookwyrm/activitypub/base_activity.py and in the status view (added in this commit), this is not an issue.


WesleyAC (approved):

WesleyAC:
Looks good, thank you for finding these! Let me know if you'd like me to cherrypick this onto b.s tonight, or you can do it yourself if you'd like.

PR #2935, opened 2023-08-01
Convert description from Markdown when importing from Open Library

### LINKED ISSUE #2930: Importing from OpenLibrary breaks markdown quotes
**Describe the bug**
When importing a book from OpenLibrary, if the description includes a quote as indicated in markdown (with ">"), the quote is not formatted/Bookwyrm doesn't recognize the code.

**To Reproduce**
1. Find a book with a description on OpenLibrary that uses a quote (example to try: OL27172905W)
2. Import the book to a Bookwyrm instance
3. View the description

**Expected behavior**
The quote should appear formatted correctly like markdown links, italics, etc. do.

**Screenshots**

**Instance**
bookwyrm.social

**Additional context**

---

    - OS: Windows

    - OS: Android

jderuiter:
The description from Open Library is in Markdown format. When importing from Open Library the description gets sanitised as if it is HTML. This breaks Markdown blockquote as the `>` gets converted to `&gt;`. Other Markdown syntax seems to work as the description field seems to be able to handle both Markdown and HTML. Is it the intention to support a mix of Markdown and HTML in the database?

Fixes #2930


bookwyrm/connectors/openlibrary.py
```diff
@@ -235,8 +237,19 @@ def ignore_edition(edition_data):
 def get_description(description_blob):
     """descriptions can be a string or a dict"""
     if isinstance(description_blob, dict):
-        return description_blob.get("value")
-    return description_blob
+        description = markdown(description_blob.get("value"))
+    else:
+        description = markdown(description_blob)
```
mouse-reeve, line 242: You'll want to pass the markdown parsed results into the `sanitizer.clean` function -- something like:
```
from bookwyrm.utils import sanitizer
...
sanitizer.clean(description)
```

This will remove any HTML characters that aren't on our allow-list and removes properties on elements that could cause problems (think onclick javascript events or marquee tags), and remove any malformed HTML (like unterminated tags).

mouse-reeve (changes requested):
One small change to sanitize the input, but otherwise looks good!

jderuiter:
clean is already called when the data is converted to HtmlField (which actually caused the original issue), but I agree it would be good to call it here as well to sanitize the data closer to input.

bookwyrm/connectors/openlibrary.py
jderuiter: added `clean` now at the end of the function

mouse-reeve (approved):
I see! Thank you for clarifying and for the changes!

PR #2937, opened 2023-08-01
Fixes for create-book

jderuiter:
This PR addresses a couple of smaller issues when using `create-book`:
- When manually adding a new edition and you do not add a new author, you're asked whether you want to add the author `['']`
- When creating a new book manually and leaving the author blank you're asked whether you want to create the author `[]`
- When adding a new book and possible matches are found these are not displayed as `data["book"]` is always set for the subjects (even when none are provided) and when a possible match is selected it is not actually set as parent_work as `parent_work` was sent twice in the POST request (once the selected value and once an empty hidden field)

### DIFF
--- a/bookwyrm/forms/books.py
+++ b/bookwyrm/forms/books.py
@@ -111,6 +111,7 @@ class EditionFromWorkForm(CustomForm):
         model = models.Work
         fields = [
             "title",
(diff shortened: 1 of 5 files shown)


mouse-reeve:
Thank you so much! Looks good

PR #2938, opened 2023-08-01
On readthrough, progressupdate or status delete return to previous page

jderuiter:
When deleting, for example, a readthrough on a bookpage you are redirected to the main page. I expected to stay on the bookpage in that case. This PR changes the redirect to `/` to `redirect_to_referer`.

### DIFF
--- a/bookwyrm/views/status.py
+++ b/bookwyrm/views/status.py
@@ -8,7 +8,7 @@ from django.core.validators import URLValidator
 from django.core.exceptions import ValidationError
 from django.db.models import Q
 from django.http import HttpResponse, HttpResponseBadRequest, Http404
-from django.shortcuts import get_object_or_404, redirect
+from django.shortcuts import get_object_or_404
 from django.template.response import TemplateResponse
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


bookwyrm/views/status.py
```diff
@@ -176,7 +176,7 @@ def post(self, request, status_id):
 
         # perform deletion
         status.delete()
-        return redirect("/")
```
mouse-reeve, line 179: If you delete a status from the status' own page, I think this is going to redirect you to a 404?

jderuiter:
You're right, I missed that case. Sorry about that. That will make things a bit more complicated. Probably need to include the desired redirect in a hidden field in the form (and carefully validate it). At least it might be a construct that could be useful for reuse.

bookwyrm/views/status.py
jderuiter: I reverted this change. I'd still like to look at this, but I'll do that in a different PR as I don't have time for this at moment.

mouse-reeve:
Thank you!!

PR #2946, opened 2023-08-06
Adds breadcrumbs and better titles to followers/following pages

mouse-reeve:
This makes it a bit easier to get back to the profile page and clarifies the hierarchy a bit. It also gives these pages distinct titles which is nice. This design pattern mirrors what you see on the user's books pages:
Screen Shot 2023-08-06 at 3 10 45 PM
Screen Shot 2023-08-06 at 3 11 04 PM

### DIFF
--- a/bookwyrm/templates/user/layout.html
+++ b/bookwyrm/templates/user/layout.html
@@ -21,6 +21,8 @@
     {% endblock %}
 </header>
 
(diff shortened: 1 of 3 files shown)


mouse-reeve:
@bookwyrm-social/code-review should be a simple one

hughrun (approved):
Looks good, nice simple improvement.

PR #2947, opened 2023-08-06
Small UI fixes

mouse-reeve:
A couple tiny things I noticed that were bugging me:
 - Follower/following counts don't use the intcomma filter -- now you will have "1,234" instead of "1234" followers, for example
 - The default filter values weren't being shown in the directory page, now they will be
 - The page title for the sub-pages of user profiles were all just the username, which was confusing. Now they say what page you are on
 - Added null state text for when you have no lists/groups/et ceters
 - Updates the formatting of null state text to all match (italics)

### DIFF
--- a/bookwyrm/templates/lists/list_items.html
+++ b/bookwyrm/templates/lists/list_items.html
@@ -46,8 +46,7 @@
             </div>
         </div>
     </div>
-    {% endfor %}
-    {% if not lists or not lists.exists %}
+    {% empty %}
     <p class="column"><em>{% trans "No lists found." %}</em></p>
-    {% endif %}
+    {% endfor %}
 </div>
--- a/bookwyrm/templates/lists/list_items.html
+++ b/bookwyrm/templates/lists/list_items.html
@@ -47,4 +47,7 @@
         </div>
     </div>
     {% endfor %}
+    {% if not lists or not lists.exists %}
+    <p class="column"><em>{% trans "No lists found." %}</em></p>
+    {% endif %}
 </div>
--- a/bookwyrm/templates/user/groups.html
+++ b/bookwyrm/templates/user/groups.html
@@ -35,10 +35,6 @@
     </div>
 
     {% include 'groups/user_groups.html' with memberships=memberships %}
-
-    {% if not memberships or not memberships.exists %}
-    <p><em>{% trans "No groups found." %}</em></p>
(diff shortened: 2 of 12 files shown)


mouse-reeve:
@bookwyrm-social/code-review should be a simple one

bookwyrm/templates/lists/list_items.html
```diff
@@ -47,4 +47,7 @@ <h4 class="card-header-title is-clipped">
         </div>
     </div>
     {% endfor %}
+    {% if not lists or not lists.exists %}
```
hughrun, line 50: `lists.exists` is always false so even when a list is found it still says "No lists found".

Consider using `{% empty %}` instead, above `{% endfor %}`

bookwyrm/templates/user/groups.html
```diff
@@ -30,6 +35,10 @@ <h1 class="title">
     </div>
 
     {% include 'groups/user_groups.html' with memberships=memberships %}
+
+    {% if not memberships or not memberships.exists %}
```
hughrun, line 39: Same as for lists, `membership.exists` is always false. This also should be in `groups/user_groups` rather than here. From line 34 you can do:

```
{% empty %}

<p><em>{% trans "No groups found." %}</em></p>

{% endfor %}
```

hughrun (changes requested):
Looks mostly good!

Couple of things to fix for when there are no lists or groups, as noted.

This may be scope-creep so feel free to ignore, but I also noticed that when looking at a brand-new user profile as that user, the default (Activity) tab looks kinda weird:

Screenshot 2023-08-07 at 12 21 19 pm

Ideally in the case of no activities, we'd also hide the "User Activity" header and in the absence of any books hide the "View all books" link - but these are minor problems that could go into a different issue/PR.

bookwyrm/templates/lists/list_items.html
mouse-reeve: I was genuinely thinking that I wished something like `{% empty %}` existed when I wrote this 🤦. Amazing, thank you

mouse-reeve:
I fixed the `empty` situations and removed the "all books" link and added some null state text there. I think removing the activity header is a bigger problem than this PR, since it contextualized what behavior the page will have (like an RSS feed) in a way that might be useful to a new user?

PR #2948, opened 2023-08-07
Pre-populate sort title in edit book form if not provided

### LINKED ISSUE #2944: Save-button seems unresponsive when adding a book in Firefox for Android when no sort title is given
**Describe the bug**
When I try to add a new book or a new edition of a book in Firefox for Android and I add no sort title, the save-button seems unresponsive. Nothing happens when I press the button.

**To Reproduce**
1. Open Firefox on an Android device.
1. Go to "Add Book" (https://bookwyrm.social/create-book on the instance I use.)
1. Add a title, don't add a sort title
1. Scroll down to the save-button
1. Press save-button
1. Nothing happens

**Expected behavior**
When I go through the steps above on my laptop, the browser scrolls to the missing field and shows a hint which data is missing. I would expect the same behavior in Firefox for Android.

**Instance**
https://bookwyrm.social/

---
**Smartphone:**
    - Device: Fairphone 4
    - OS: Android 12
    - Browser: Firefox
    - Version 116.0

mouse-reeve:
It's confusing to edit a book when this isn't set, so this provides the best-guess version of the sort title if there isn't one provided, and allows the user to change it as needed.

Fixes #2944

### DIFF
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -217,6 +217,13 @@ class Book(BookDataModel):
         """editions and works both use "book" instead of model_name"""
         return f"https://{DOMAIN}/book/{self.id}"
 
+    def guess_sort_title(self):
+        """Get a best-guess sort title for the current book"""
+        articles = chain(
+            *(LANGUAGE_ARTICLES.get(language, ()) for language in tuple(self.languages))
+        )
+        return re.sub(f'^{" |^".join(articles)} ', "", str(self.title).lower())
+
     def __repr__(self):
         # pylint: disable=consider-using-f-string
         return "<{} key={!r} title={!r}>".format(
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Fixes #2944

mouse-reeve:
@bookwyrm-social/code-review I think this one is also pretty straightforward, if a little weird to test. On bookwyrm.social, I ended up skipping the migration that populated default sort titles, so this issue is specific to that instance. It's a little hard to replicate it because if you save a book with a blank sort title, one is [populated by the save function](https://github.com/bookwyrm-social/bookwyrm/pull/2948/files#diff-878c30e588114931c96c12df3c7a8682857e59c0bdb5e9d3005591f2f0db5c7fR37). You can comment that out, or use `models.Edition.objects.filter(id=<book_id>).update(sort_title=None)` to update the field without triggering the save function.

bookwyrm/models/book.py
```diff
@@ -375,15 +382,7 @@ def save(self, *args: Any, **kwargs: Any) -> None:
         # Create sort title by removing articles from title
         if self.sort_title in [None, ""]:
```
hughrun, line 383: We probably don't need to check this twice?

hughrun (changes requested):
Seems to work as expected, I can't see any obvious problems other than we are checking for null values inside a check for null values.

bookwyrm/models/book.py
mouse-reeve: haha oops -- the famous "wait, did I check that already" programming philosophy

PR #2953, opened 2023-08-16
Add spanish articles

MaggieFero:
Hi! Our fork added Spanish-language articles to get us one step closer to the dream of Issue #2678 . It's just a little change, but probably other people would like it, too!


mouse-reeve:
This is a great change and the only thing holding me back from merging it is that I'm not sure if the articles should be separated out by language. I need to take a look at how those languages are actually being used, and haven't done that yet.

MaggieFero:
Thanks! Keeping them separated out by language is really useful for our multi-lingual server because some languages' articles

MaggieFero:
(Apologies for both the early-enter on the comment and the closed-then-reopened pull request, both were a single misclick)

mouse-reeve:
That makes sense! I'm going to merge and we can always revisit as time goes on. Thanks so much for adding to this.

Issue #2844, opened 2023-05-09
Gunicorn causes signed GET requests (e.g. AUTHORIZED_FETCH) to fail in Production


futzle:
**Describe the bug**
On production branch, dockerless, still getting Unauthorized: /user/bookwyrmuser/inbox on follow from a Mastodon instance.

**To Reproduce**
1. Install production branch on Debian 11, Dockerless instructions.
2. Create a regular user on bookwyrm. Allow automatic follows without approval.
3. Try to follow this user from a Mastodon 4.0.2 (Hometown 1.1.1) instance with AUTHORIZED_FETCH turned on.
4. Error appears in bookwyrm log: Unauthorized: /user/bookwyrmuser/inbox
5. On Mastodon instance, follow degrades to a request and becomes "Withdraw follow request".

**Expected behavior**
The follow should complete, and it should be visible at both ends.

**Instance**
https://outside.ofa.dog/
https://old.mermaid.town/

**Additional context**
This issue looks just like closed issues #2734, #1964, #2626, which are all supposed to be fixed by #2613. Yet I'm pretty sure that my installation, new on 8 May 2023, includes #2613 and I'm still getting the exact same behaviour.

Users on other instances can follow me on Bookwyrm, even users on other Mastodon instances.

I can't seem to follow any Bookwyrm instances from old.mermaid.town.

I tried two users on Mastodon instance old.mermaid.town and had the same result.

Mastodon server has errors in sidekiq logs:
```
"{\"@context\":\"https://www.w3.org/ns/activitystreams\",\"id\":\"https://old.mermaid.town/4fb96c8a-614f-478c-b4c4-24ad957527b3\",\"type\":\"Follow\",\"actor\":\"https://old.mermaid.town/users/Kymberly\",\"object\":\"https://outside.ofa.dog/user/futzle\"}",  109376236586682428, "https://outside.ofa.dog/user/futzle/inbox"

Mastodon::UnexpectedResponseError: https://outside.ofa.dog/user/futzle/inbox returned code 401
```

```
[...]

Issue #2717, opened 2023-03-08
Stop making HTTP requests in views


WesleyAC:
It's bad practice to make blocking HTTP requests in views, since that means that the response can be delayed indefinitely, which is bad both for the user and for the worker thread that is hung. I think this is part of why requests often take so long to load on bookwyrm.social — most of the workers are blocking on various things that should be done async. I've worked around this somewhat by significantly increasing the number of workers, but it's better to not do this kind of work in views at all.

We currently do this in a handful of places in Bookwyrm — I've made a (possibly incomplete) list below. HTTP requests in these views should be moved into Celery tasks or otherwise re-architected where at all feasible. In cases where it's particularly difficult to avoid, we should make sure both that there is a relatively short timeout, and that errors are handled correctly.

- [ ] `bookwyrm/views/admin/dashboard.py:Dashboard.get`
  * via `get_data`
- [ ] `bookwyrm/views/follow.py:remote_follow`
  * via `get_data` via `subscribe_remote_webfinger`
- [ ] `bookwyrm/views/follow.py:ostatus_follow_request`
  * via `get_data` via `handle_remote_webfinger`
- [ ] `bookwyrm/views/status.py:CreateStatus.post`
  * via `get_data` via `handle_remote_webfinger` via `find_mentions`
- [ ] `bookwyrm/views/search.py:Search.get`
  * via `get_data` via `handle_remote_webfinger` via `user_search`
- [ ] `bookwyrm/views/admin/federation.py:refresh_server`
  * via `get_data` via `get_or_create_remote_server`
- [ ] `bookwyrm/views/books/books.py:upload_cover`
  * via `get_image` via `set_cover_from_url`
- [ ] `bookwyrm/views/books/books.py:resolve_book`
  * via `get_data` via `resolve_remote_id` via `get_or_create_book`
- [ ] `bookwyrm/views/books/books.py:update_book_from_remote`
  * via `get_data` via `resolve_remote_id` via `Status.ignore_activity` via `ActivityObject.to_model` via `create_edition_from_data`
- [ ] `bookwyrm/views/books/edit_book.py:EditBook.post`
  * via `get_image` via `set_cover_from_url`
- [ ] `bookwyrm/views/books/edit_book.py:ConfirmEditBook.post`
  * via `get_data` via `resolve_remote_id` via `Status.ignore_activity` via `ActivityObject.to_model`
  * via `request_isni_data` via `get_author_from_isni` via `build_author_from_isni`
[...]

PR #2958, opened 2023-08-19
temporarily fix signed get requests failing

hughrun:
Various issues have been raised regarding HTTP requests that seem to be dropped. The real fix for this in #2717 - "Stop making HTTP requests in views"

This commit is a quick fix in some circumstances - particularly #2844 - signed GET requests failing in production. It adds an extra thread to Gunicorn to make the default configuration multi-threaded. This will now match the multi-threaded nature of the development environment, which is why we have not been able to replicate these errors in dev.

### DIFF
--- a/contrib/systemd/bookwyrm.service
+++ b/contrib/systemd/bookwyrm.service
@@ -6,7 +6,7 @@ After=network.target postgresql.service redis.service
 User=bookwyrm
 Group=bookwyrm
 WorkingDirectory=/opt/bookwyrm/
-ExecStart=/opt/bookwyrm/venv/bin/gunicorn bookwyrm.wsgi:application --threads=2 --bind 0.0.0.0:8000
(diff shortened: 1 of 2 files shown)


WesleyAC:
@hughrun I think more like 8 or 16 threads would be a better default. In experimenting on b.s, more threads didn't seem to cause problems, and avoid falling over in times of load. On b.s we have ~30x more threads than CPU cores and it's totally fine.

hughrun:
> @hughrun I think more like 8 or 16 threads would be a better default. In experimenting on b.s, more threads didn't seem to cause problems, and avoid falling over in times of load. On b.s we have ~30x more threads than CPU cores and it's totally fine.

Sounds good, I'll update. This is definitely on the very edge of my expertise and I'm aware some people just want to run with the bare minimum overhead so I was being conservative.

PR #2970, opened 2023-08-29
Add `Move` activity for user migration

hughrun:
This is both a stand-alone PR and works in partnership with #2868 - migrating user data from one server to another is Step 1, but we then need a way to tell Fediverse servers that the user has moved: that's what this PR does.

Additionally, this will cover a gap where we're currently ignoring incoming `Move` notifications, including when e.g. a Mastodon user migrates from one server to another. This means those users effectively become orphaned from BookWyrm because we don't follow them to the new account.

This largely follows the precedent set by Mastodon, since there is no ActivityPub standard for account migration, however there are a few differences.

## Features:

* Sends notification to local followers when a user `Move` action is received from any ActivityPub server;
* Sends `Move` activity to followers when a user moves from one AP account to another;
* Requires the target account to have the source account listed in `alsoKnownAs`/`also_known_as` aka an alias;
* adds `movedTo`/`moved_to` value to the source account
* redirects source account webfinger to the target account
* Adds prominent notice to source user profile page
* stops source user acting other than within preferences section
* move can be undone at any time

## Following behaviour:

Unlike Mastodon I've set this up so that followers are _notified_ of the move but there is no action taken regarding following unless initiated by the follower (Mastodon auto-unfollows the old user and auto-follows the new user when receiving a `Move` activity). Notifications have a button to follow the new account, but the user remains following the old account. 

This makes moves less high-stakes and provides more autonomy to the follower. It also spreads out load on servers when a popular user moves their account as follows/unfollows will happen at the speed other users see and act on notifications.

We could change this so that the button in the notification unfollows the old user as well - happy to take advice on that.

Context:

* https://www.w3.org/TR/activitystreams-vocabulary/#dfn-move
* https://docs.joinmastodon.org/spec/activitypub/#Move


bookwyrm/activitypub/verbs.py
```diff
@@ -231,3 +231,28 @@ class Announce(Verb):
...
         """boost"""
         self.to_model(allow_external_connections=allow_external_connections)
+
+@dataclass(init=False)
+class Move(Verb):
+    """a user moving an object"""
+
+    # note the spec example for target and origin is an object but
+    # Mastodon uses a URI string and TBH this makes more sense
+    # Is there a way we can account for either format?
```
mouse-reeve, line 241: `set_related_field` can take either an object or a string `remote_id`, which it then looks up by querying the string as a url. I'm not sure that that's directly helpful here, but it is the case that the value of a field will sometimes be a string and sometimes an actual object in a number of places.

bookwyrm/models/move.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class Move(ActivityMixin, BookWyrmModel):
+    """migrating an activitypub user account"""
+
+    user = fields.ForeignKey(
+        "User", on_delete=models.PROTECT, activitypub_field="actor"
+    )
+
+    # TODO: can we just use the abstract class here?
```
mouse-reeve, line 18: I don't think you can set a field to reference an abstract class, unfortunately. Since I didn't design a good Actor base model and haven't managed that large a refactor, I'm not sure how to handle arbitrary types besides subclassing the Move object for different object types?

bookwyrm/models/move.py
```diff
@@ -0,0 +1,50 @@
...
+        # TODO how do we check this for any conceivable object?
+        pass
+
+    def save(self, *args, **kwargs):
+        """update user active time"""
+        self.user.update_active_date()
+        super().save(*args, **kwargs)
+
+    # Ok what else? We can trigger a notification for followers of a user who sends a `Move` for themselves
+    # What about when a book is merged (i.e. moved from one id into another)? We could use that to send out a message
```
mouse-reeve, line 49: We might be able to get away with just creating a redirect, so the old url is still resolvable?

bookwyrm/models/move.py
```diff
@@ -0,0 +1,50 @@
...
+    # to other Bookwyrm instances to update their remote_id for the book, but ...how do we trigger any action?
```
mouse-reeve, line 50: The two approaches currently in use are to either trigger a task in the `save` function of the model, or to add a `receiver` that listens on `post_save`, like in `bookwyrm/models/notifications.py`

bookwyrm/templates/notifications/items/move.html
```diff
@@ -0,0 +1,28 @@
...
+{% endspaceless %}{% endblock %}
+
+{% block icon %}
+    <span class="icon icon-local"></span>
+{% endblock %}
+
+{% block description %}
+    <!--
+        TODO: a user has a 'name' but not everything does, notably a book.
+        On the other hand, maybe we don't need to notify anyone if a book
```
mouse-reeve, line 17: I don't think we would want to notify about books being merged, in part because it *ought* to be a routine administrative task and also it would potentially be a huge number of notifications

mouse-reeve (commented):
This is just a quick pass because I haven't completely forgotten about you despite having Quite A Week Or Two right now! Let me know if anything is helpful? I'm not sure I totally answered what you were asking in some places

hughrun:
Ok this more or less works outbound (Mastodon will accept the `Move`, unfollow the old user and follow the new user), and within a single instance (updates user details and sends notification to followers).

However...

When a Bookwyrm user is following a Mastodon user who moves to another account, Mastodon sends not just a `Move` but also an `Update` for that user, and then Bookwyrm panics about trying to save a new user with a duplicate username.

```python
# bookwyrm.activitypub.verbs.py
class Update(Verb):
    """Update activity"""

    to: List[str]
    type: str = "Update"

    def action(self, allow_external_connections=True):
        """update a model instance from the dataclass"""
        if not self.object:
            return
        # the line below now seems to trigger an attempt to save a duplicate user
        self.object.to_model(
            allow_create=False, allow_external_connections=allow_external_connections
        )
```

I'm struggling to understand what's happening here, any tips would be appreciated. I haven't included migrations in this PR yet because I suspect they'll change, but I did make and run them.

jaschaurbach:
This one is great. @mouse-reeve I reviewed it and I would merge this. As this is a big one please do a review as well. Thanks!

bookwyrm/models/move.py
```diff
@@ -0,0 +1,87 @@
...
+
+            for follower in self.user.followers.all():
+                if follower.local:
+                    MoveUserNotification.objects.create(user=follower, target=self.user)
+
+        else:
+            raise PermissionDenied()
+
+
+class MoveUserNotification(models.Model):
```
mouse-reeve, line 72: I'm not understanding what this model does that can't be handled in the notification model? It looks like saving here triggers creating a Notification, but does it serve an additional function?

bookwyrm/templates/feed/layout.html
```diff
@@ -1,9 +1,26 @@
 {% extends 'layout.html' %}
 {% load i18n %}
+{% load utilities %}
 
 {% block title %}{% trans "Updates" %}{% endblock %}
 
 {% block content %}
+{% if user.moved_to %}
```
mouse-reeve, line 8: Is it desirable, in the bigger picture, to have this panel only obscure the feed page? I would think it makes more sense for the whole interface to be blocked, but I realize that's a more complicated endeavor, and I also might be missing some reason to keep other views active.

If it would be better to block the entire UI, I think that could also fall outside the purview of this PR

bookwyrm/templates/feed/layout.html
```diff
@@ -1,9 +1,26 @@
...
+<div class="container my-6">
+    <div class="notification is-info has-text-centered">
+        <p>
+            {% trans "You have have moved to" %}
```
mouse-reeve, line 12: It would be better to include the link in a blocktrans, so that languages with different word order can place the link in different places

bookwyrm/templates/notifications/items/move_user.html
```diff
@@ -0,0 +1,24 @@
...
+    {{ notification.related_object.local_path }}
+{% endspaceless %}{% endblock %}
+
+{% block icon %}
+    <span class="icon icon-local"></span>
+{% endblock %}
+
+{% block description %}
+    {% if related_user_moved_to %}
+        {{ related_user }} {% trans "has moved to" %} <a href="{{ related_user_moved_to }}">{% id_to_username related_user_moved_to %}</a>
```
mouse-reeve, line 17: This trans tag should also include the username and link in a `blocktrans`

bookwyrm/templates/notifications/items/move_user.html
```diff
@@ -0,0 +1,24 @@
...
+        <div class="row shrink my-2">
+            {% include 'snippets/move_user_buttons.html' with group=notification.related_group %}
+        </div>
+    {% else %}
+        {{ related_user }} {% trans "has undone their move" %}
```
mouse-reeve, line 22: and here

bookwyrm/templates/preferences/move_user.html
```diff
@@ -0,0 +1,41 @@
...
+{% block panel %}
+<div class="block">
+    <h2 class="title is-4">{% trans "Migrate account to another server" %}</h2>
+    <div class="box">
+        <div class="notification is-danger is-light">
+            <p>
+                {% trans "Moving your account will notify all your followers and direct them to follow the new account." %}
+            </p>
+            <p>
+                <strong>{{ user.username }}</strong> {% trans "will be marked as moved and will not be discoverable or usable unless you undo the move." %}
```
mouse-reeve, line 19: Another place to use `blocktrans`

mouse-reeve (commented):
This works really smoothly! My main question is about the additional notification model. I know that I've been the blocker on this PR, so I'm happy to consider my translation items non-blockers on merging this, and I can change them this weekend. 

The bigger question (which again, I don't want to make a blocker), is how functional a moved account should be. I mentioned showing other parts of the UI in the comments, but I believe a moved account will also still receive updates to its activitystreams, and otherwise behave like a fully operational account. I suspect that's the correct behavior, because I think it is supposed to be possible to, for example, follow a moved account, and I think it would be challenging to re-build the activitystream if the user decided to un-move. However, it seems odd that a moved account can post statuses.

bookwyrm/models/move.py
hughrun: This is a good question. I think I didn't understand how the notifications work and copied this pattern from somewhere else...

So this should just be an ordinary notification with the `target` as `related_user` and type of `MOVE`?

bookwyrm/templates/feed/layout.html
hughrun: Happy to take advice on this. I think it's important that people have the ability to change their mind and not lose too much. Mastodon, for example, makes moves fairly catastrophic if you change your mind because it automatically triggers an unfollow for all your followers, and has fairly draconian cool-off periods.

On the other hand it needs to be clear that you shouldn't be doing anything with a moved account other than perhaps "unmoving" it.

To be honest I wasn't quite sure what to do for this aspect, and was feeling a bit overwhelmed by the various options. I tend to agree that the better option is to exclude the user from doing anything much other than changing their mind about the move. It's probably not much work, I guess I wasn't confident about making the call one way or the other on my own.

hughrun:
> However, it seems odd that a moved account can post statuses.

I definitely didn't mean to leave that ability in, oops.

Basically I agree. See comment on your review - I really just wasn't sure what to leave available and what to block, and if I'm honest it seemed a bit complicated and I took an easy route.

I agree: restoring accounts should be reasonably painless but other than that moved accounts should be essentially non-functional. I'll take a look at this and adjust the UI changes - it may make sense to just create a new template and redirect to that.

bookwyrm/models/move.py
mouse-reeve: The way it's created [in the listener](https://github.com/bookwyrm-social/bookwyrm/pull/2970/files#diff-ac98bf3171c4c7025f6ea238b24ed7e1056f52f0f0f77709e21462ba6df215d4R345) is correct -- you can just call `Notification.notify(...)` instead of [creating the MoveUserNotification](https://github.com/bookwyrm-social/bookwyrm/pull/2970/files#diff-0b203a68eea4a30f195d7ccd947b1557775264146348ee1f00a13c9ac5eed255R66) object

hughrun: yeah honestly I have no idea why I did it this way, it's kind of obvious now that I just created a totally unnecessary step in the middle.

hughrun:
With the changes I've made post-review, it should be that the only thing a `move`d user can now do when they are logged in is either log out, or cancel the move. I also took the opportunity to make the UI a little nicer for such users.

mouse-reeve (approved):
I can't for the LIFE of me figure out how to create a PR that merges into this PR and I don't want to keep making you make stupidly tiny changes to the blocktrans tag, so I'm going to merge this into a new branch and then commit my changes and then merge THAT branch in, which seems like a ridiculous way to do it but at least it will be merged

PR #2972, opened 2023-08-30
fix opensearch template

### LINKED ISSUE #2971: Unable to add bookwyrm opensearch on firefox
When I tried to [add bookwyrm to my list of search engines](https://support.mozilla.org/en-US/kb/add-or-remove-search-engine-firefox) in firefox, I hit an error. 

The console had the following error message:
`Content-Security-Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).`

hughrun:
* "method" is not a valid attribute of the `Url` element
* "ShortName" cannot be empty - fixes `site_name` being used before it is assigned

fixes #2971


mouse-reeve:
Thank you!

Issue #2928, opened 2023-07-30
Pin version of docker containers


WesleyAC:
We use the container `redis` directly, for instance, which means that different installs run different versions. This can cause unexpected problems. Instead we should pin versions of our dependencies.

ilkka-ollakka:
Currently `docker-compose` and `Dockerfile` contain version pinning on images, I think it should resolve this issue?

mouse-reeve:
I'm closing this issue because it's very old, and the codebase has changes substantially in the subsequent years.

PR #2987, opened 2023-09-08
Pinned versions of docker containers

JJimenez71:
This is a quick push to fix issue #2928. I pinned these containers with the latest version. I left the postgres image in version 13.


mouse-reeve:
Thank you for doing this!

PR #2992, opened 2023-09-13
parent_work was not always included when needed

### LINKED ISSUE #2973: Authors removed when creating a new edition
When adding a new edition of a book, existing authors are removed even though the checkboxes for that are not marked.

**To Reproduce**
1. Go to a book, click on editions (e.g. https://buecher.pnpde.social/book/46494/editions)
2. Click to add a new edition.
3. Enter data for new edition, leave the existing authors as is, safe the entry.
4. New edition has no authors at all.

**Expected behavior**
I expect authors to be carried over, unless expressly specified to remove them.

**Screenshot**
This is what it looks like in German: Checkboxes unmarked, existing authors linked. This is what I save.
Create-Edition-Authors

**Instance**
https://buecher.pnpde.social/about
version 0.6.4 (German localization)

**Additional context**
I tried removing them (checking the checkboxes), which works as expected: Authors are removed. (Adding a new author works, too.)
My current workaround is to create the new edition, then edit it to add the authors which had already been listed for the existing edition.

---

    - OS: Windows 11
    - Browser: Firefox v. 116.0.3 (64bit)

### LINKED ISSUE #2978: Creating new Edition fails
When creatin new edition it asks me what book is this edition for. But when there is a bunch of similar booktitles you cant link both edition. It also happened when you create a new edition from the link in the previous edition. It should automaticaly link both edition without asking because its created from the link in the edition view.

**To Reproduce**
https://bookwyrm.social/book/63057/editions
klick on create new edition
type anything in and save
there comes the question about the previous edition but only a few titles but not the real one so you cant properly link the new edition

**Expected behavior**
It should automaticaly link both edition without asking because its created from the link in the edition view.

**Instance**
bookwyrm.social

jderuiter:
parent_work was only included in the edit book form when ` form.parent_work` was set and not `book.parent_work.id`.

Fix #2973
Fix  #2978

### DIFF
--- a/bookwyrm/templates/book/edit/edit_book_form.html
+++ b/bookwyrm/templates/book/edit/edit_book_form.html
@@ -10,7 +10,7 @@
 {% csrf_token %}
 
 <input type="hidden" name="last_edited_by" value="{{ request.user.id }}">
-{% if form.parent_work %}
+{% if book.parent_work.id or form.parent_work %}
 <input type="hidden" name="parent_work" value="{% firstof book.parent_work.id form.parent_work %}">
 {% endif %}
 


mouse-reeve:
Thank you!!

PR #2998, opened 2023-09-17
add reverse proxy for flower

rrgeorge:
This adds a reverse proxy to flower from the admin dashboard so that you don't need to have a separate login for flower. Unfortunately, it relies on `django-revproxy`, and the current version (`0.10.0` on pypi is outdated and does not work with the current version of Django. A fork has been updated to work [django-revproxy 0.11.0](https://github.com/jazzband/django-revproxy) and can manually be installed via `venv/bin/pip install git+https://github.com/jazzband/django-revproxy.git`
Because of this complication, I'm only submitting this as draft pr, maybe someone has an idea on how to make this work without manual intervention?


Thovi98:
It seems to have been updated since then?
https://pypi.org/project/django-revproxy/

mouse-reeve:

PR #2999, opened 2023-09-17
Use ornament quotations

dato:
The result is visually similar to what the web UI uses, and
avoids any ambiguity with common quote characters used by the
quotation itself.

-----

I would love to have this for visuals aspects on the Fediverse, but
I'm marking it as draft because I'm unsure whether this is 100%
accessible.


dato:
Closing this because it probably makes things worse for screen readers

PR #3000, opened 2023-09-17
Minor improvements to Quotation pure content

dato:
- 25fd7276ea23d69a707414eb874363601cdcc433 `pure_content()` refactor: shorter conditionals
- 1322a0c6939f364a023aa0df1c01e9685d4c31f3 Substitute “p.” for “page” in page progress serialization
- ce3885d4f6bed6761cac6761682a8982a4170188 Use `endposition` when serializing Quotation
- cc05cabcb57d340f87a3ae44856fe6574bec7eb7 Note content: use italics for book titles + em-dash for Quotation
- fadf30b94216a407aceb3d089595fc40e8bc446e Also use italics for book title in editions.html template

-----

### DIFF
--- a/bookwyrm/models/status.py
+++ b/bookwyrm/models/status.py
@@ -270,7 +270,7 @@ class GeneratedNote(Status):
         """indicate the book in question for mastodon (or w/e) users"""
         message = self.content
         books = ", ".join(
-            f'<a href="{book.remote_id}">"{book.title}"</a>'
+            f'<a href="{book.remote_id}"><i>{book.title}</i></a>'
             for book in self.mention_books.all()
         )
         return f"{self.user.display_name} {message} {books}"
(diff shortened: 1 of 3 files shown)


dato:
Also, while were on this code, any reason why the title is wrapped in quotes, instead of `<i>`?

The quotes seem fine for a plaintext serialization, but since I see `<p>` and `<a>`, I wondered if there was a particular reason, or if I could just change it.

mouse-reeve:
No reason I can think of, as long as it's in the content of the serialization rather than an Article title, which it shuldn't be since this is a Note not an Article. You're welcome to change it. Chances are, at the time I simply didn't realize that Mastodon would render italics (but it does so that's a non-issue)

dato:
Thanks, will do and ask for re-review.

dato:
Done. I placed each item in a separate commit.

mouse-reeve:
This looks good, can you run the Python formatter? With that, I will be ready to merge

dato:
Oops, my bad. Fixed.

PR #3002, opened 2023-09-18
Serve static files in debug mode

dato:
I develop without Docker (plain gunicorn), and this would be very useful.

(For this reason I would also add gunicorn to requirements.txt, if you're okay with that.)

P.S.: Here's the [documentation](https://docs.djangoproject.com/en/4.2/ref/contrib/staticfiles/#django.contrib.staticfiles.urls.staticfiles_urlpatterns) for `staticfiles_urlpatthers`, where it's clearly documented that it's DEBUG-only.

Thanks!

### DIFF
--- a/bookwyrm/urls.py
+++ b/bookwyrm/urls.py
@@ -1,6 +1,7 @@
 """ url routing for the app and api """
 from django.conf.urls.static import static
 from django.contrib import admin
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you for this (and the documentation link!)

PR #3003, opened 2023-09-19
Fallback to showing first published date

dato:
First publication date is only shown when the editor is not known
(rationale: if the editor for the first edition is known, the same
date should be set in both fields?).

### DIFF
--- a/bookwyrm/templates/book/publisher_info.html
+++ b/bookwyrm/templates/book/publisher_info.html
@@ -40,16 +40,13 @@
     </p>
 {% endif %}
 
-{% with date=book.published_date|naturalday publisher=book.publishers|join:', ' %}
-{% if date or book.first_published_date or book.publishers %}
-{% if date or book.first_published_date %}
+{% if book.published_date or book.first_published_date %}
     <meta
         itemprop="datePublished"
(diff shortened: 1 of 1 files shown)


mouse-reeve:
This is good, but I think there's a better way to handle this that's a slightly more complicated change. We can use `{% with %}` to set a variable to `firstof` the two available date options, and then proceed with all the boolean logic using one variable, which will be a heck of a lot simpler. Does that make sense?

dato:
Sure, done that in 2c968e94cbf7452437303789746b3fa8f40a87d0.

In my initial version, I didn't want to allow combining publisher with first_published_version. Let me know if we should rescue that, which is [easy enough to do](https://github.com/dato/bookwyrm/commit/ec2c5cb54600ae24105f434e3097f3780180e671).

mouse-reeve:
Ah I didn't even think of that but it would look extremely weird to say a book was released by a publisher before the publisher existed. I think your additional change is a good solution

dato:
Done.

As a plus, that part of the template didn't grow, but shrank by (net) three lines:

```
❯ g diff --stat @^^
 bookwyrm/templates/book/publisher_info.html | 15 ++++++---------
 1 file changed, 6 insertions(+), 9 deletions(-)
```

Thanks!

PR #3036, opened 2023-10-15
Minor vocabulary fixes and structured data improvements

dato:
- 4a96a9333f7b5e49b53ac0c58f5d781c6edb53ba Remove duplicate Review object under `rating` property
- b1887081058a3ca1579c811d21008c6b53cc5b0e Drop `bestRating` property in ratings, since it defaults to 5
- 80879411a0a0ecfe604687726b0c3813b2819d24 Create Rating object (and its enclosing Review) in book/rating.html
- 02409a4700ae6f4a2a7c3eb993e56b9ed5d7c602 Use `position` property for Book objects in a series
- 1be6f7957439ec45cb835852a0935cd5a5c364cd Give URL of book series when setting `isPartOf`
- d368ea51b76e824617c2398144516376fba7a16a series.html: Create empty BookSeries object

-----

Most of these are fixes (as in, validator.schema.org warned about them); the last two are minor improvements.

The validator is actually run by Google, and they have additional checks of their own (for example, they don't allow _reviewCount_ to be zero in an AggregateRating). I could include a fix for those as well.

### DIFF
--- a/bookwyrm/templates/book/book.html
+++ b/bookwyrm/templates/book/book.html
@@ -45,7 +45,7 @@
 
                     {% if book.series %}
                         <meta itemprop="isPartOf" content="{{ book.series | escape }}">
-                        <meta itemprop="volumeNumber" content="{{ book.series_number }}">
+                        <meta itemprop="position" content="{{ book.series_number }}">
 
                         {% if book.authors.exists %}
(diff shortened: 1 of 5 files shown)


jaschaurbach:
> The validator is actually run by Google, and they have additional checks of their own (for example, they don't allow _reviewCount_ to be zero in an AggregateRating). I could include a fix for those as well.

It would be great if you could ad a fix @dato

PR #3039, opened 2023-10-16
Fix creation of covers for ActivityPub imports

dato:
-----
Imports for other BookWyrm instances were showing without covers for me, and I think this is the cause?

Code seemed correct, as in, I _thought_ `value` here would be a document. But then the comment refers to receiving JSON, in which case the code is not correct. For this reason I've left both.

### DIFF
--- a/bookwyrm/models/fields.py
+++ b/bookwyrm/models/fields.py
@@ -482,10 +482,12 @@ class ImageField(ActivitypubFieldMixin, models.ImageField):
         image_slug = value
         # when it's an inline image (User avatar/icon, Book cover), it's a json
         # blob, but when it's an attached image, it's just a url
-        if hasattr(image_slug, "url"):
-            url = image_slug.url
-        elif isinstance(image_slug, str):
+        if isinstance(image_slug, str):
             url = image_slug
+        elif isinstance(image_slug, dict):
+            url = image_slug.get("url")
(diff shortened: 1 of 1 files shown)


prolibre:
Oh great! Thank you so much !

hughrun:
I can check this tomorrow.

The comment is confusing, because it refers to a "JSON blob" which initially made me think it meant a [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob).

`value` here is only the image field listed within the ActivityPub JSON for a given activity. It's called from [`set_field_from_activity`](https://github.com/bookwyrm-social/bookwyrm/blob/3251ef0bf5f88630bb0efcc172687ae0bc6069aa/bookwyrm/models/fields.py#L71).

I think what the comment is meant to convey is that sometimes `value` is just the url as a string (if it's pulled from an "attached" image), and sometimes it's originally a JavaScript _object_ (if the image is in intrinsic part of the ActivityPub object, as in the case of a user or Edition).

mouse-reeve:
This seems like it works to me -- is there any reason it's still a draft?

I believe I meant that the value can either be a dict or a string, but I wrote it to support objects, possibly because the activitypub could be passed in still as an ActivityPub object, even though evidently it's coming in as a dict

PR #3034, opened 2023-10-14
introduce .prettierrc

jaschaurbach:
This fixes #3032 

The produced is in fact correct (see [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)).
I do not know why prettier suddenly enforces this rule, I could not find it out.

To tell prettier to NOT enforce this rule I introduced `.prettierrc` wit one rule:
`'trailingComma': 'none'`

By enforcing this rule, prettier changes some long used .js which used trailing commas bringing inconsistency to the code. 

Now it is all consistent.

PR #3043, opened 2023-10-17
Final prettier re-run

dato:
After #3042, the prettier config is the one we want ('es5', compatible with ESLint), but an update to the tree is necessary (since trailing commas were removed as part of #3034).

From what I've checked now: 3034 + 3042 + this PR = no diff in JS files, just prettier config.

So I think we're all set now!

/cc @jaschaurbach


mouse-reeve:
This looks good to me! Thank you both for all the work to get this functional

Issue #3028, opened 2023-10-13
Dates wrong in ActivityPub Stream


cloo:
Viewing a quote on Mastodon dispays a shifted publication date. In the screenshot - 2022 turned into 2021.

![image](https://github.com/bookwyrm-social/bookwyrm/assets/651760/5e7cb29b-fe76-4db8-91dd-2e61e7828fa3)
![image](https://github.com/bookwyrm-social/bookwyrm/assets/651760/7b1023d9-06cb-4956-b36e-68ede752d7fb)

**To Reproduce**
1. Create a quote for a book
2. Open the feed in Mastodon
3. Observe the publication date

**Instance**
Bookwyrm.social, v. 0.6.6

**Additional context**
I suspect it could also be related to timezones and January 1st problem. 
There is a harmful requirement to put in precise day for publication, which forces users to type in January 1st when the day is not provided (which is the usual case). See bug https://github.com/bookwyrm-social/bookwyrm/issues/743
Additionally, in the screenshot you can observe shifted diacritics in the title - bug https://github.com/bookwyrm-social/bookwyrm/issues/2661

PR #3048, opened 2023-10-20
Shelter stable dates from USE_TZ

dato:
- 28d1d466b08c81d33369ae1dee0d22bd251c7cb4 Add failing test case for "January 1st" offset bug
- 85f04f19ba8f5a71e15d51ff55bbd3b452832000 Shelter stable dates from USE_TZ

-----

Some dates (publication dates, author dates) are meant as _literals_. What
the user inputs through a `SelectDateWidget` should be preserved as-is.
Django's otherwise-excelent support for timezones interferes with it (see
#3028).

Until a better fate of these columns is determined (do we migrate them to a
DateField?), and as a stop-gap measure, we can start being faithful to the 
data by storing them in the Western-most timezone.

This is particularly important because `1/1/YYYY` is a common pattern in
publication dates, given #743.

Fixes: #3028.

-----

#3028 is somewhat intertwined with #743, because both work with the same columns. In general, I try to structure work in chunks that do something small and can be meaningfully applied.

With no need for a migration, I think this is a net win wrt the "shifting year" issue that was recently reported.

Furthermore, I believe in the future we might be able to deduce the correct tz conversions to apply. With this PR, I just offer a small tuning to affect/fix _observable_ behavior.

Many thanks in advance for any reviews!

### DIFF
--- a/bookwyrm/models/author.py
+++ b/bookwyrm/models/author.py
@@ -32,8 +32,8 @@ class Author(BookDataModel):
         max_length=255, blank=True, null=True, deduplication_field=True
     )
     # idk probably other keys would be useful here?
-    born = fields.DateTimeField(blank=True, null=True)
-    died = fields.DateTimeField(blank=True, null=True)
+    born = fields.StableDateField(blank=True, null=True)
+    died = fields.StableDateField(blank=True, null=True)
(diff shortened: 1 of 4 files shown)


dato:
Dropping in favor of #3059

Issue #2996, opened 2023-09-15
Book covers not deleted (with accumulation of old images)


prolibre:
**Describe the bug**
I've realised that book images are not physically deleted after a book has been deleted, or even after the cover has been changed.
I'm afraid that as time goes by the disk will become full of useless images.

**To Reproduce**
- Add an image to a book. The image is found in the "covers" folder.
- Modify the image in the same book: the new image is in the "covers" folder, but the old image is also there.
- Delete an image from the same book: both images are still in the "covers" folder.

**Expected behavior**
Unnecessary images should be deleted to avoid wasting disk space.

Perhaps the best thing would be to give each book cover a unique file name to simplify image management. In this way, a new image would overwrite the previous one and deleting images would be simple to implement.
It's a solution that also prevents images from being lost in old mastodon publications (for example).

hughrun:
This is not a bug in BookWyrm, but rather [standard Django behaviour](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.fields.files.FieldFile.delete):

> Note that when a model is deleted, related files are not deleted. If you need to cleanup orphaned files, you’ll need to handle it yourself (for instance, with a custom management command that can be run manually or scheduled to run periodically via e.g. cron).

That doesn't mean it isn't a problem, however!

We could potentially manage this going forward by [overriding](https://docs.djangoproject.com/en/5.2/topics/db/models/#overriding-model-methods) the `delete()` method on `Book` and `User` classes, and also the `save()` method on covers and avatars (so that images that have been replaced are cleaned up at the time they are overwritten on the object).

That would leave us with existing files that are not associated with any model object. I have a pull request waiting (#3669) that looks for objects where the filepath doesn't resolve. I guess there could be a management command that reverses the logic and deletes files that aren't associated with a model object.

PR #2980, opened 2023-09-05
Complete Migrations of Bookwyrm Accounts across instances

CSDUMMI:
This PR aims to implement [#1012](https://github.com/bookwyrm-social/bookwyrm/issues/1012). It introduces new export and import functionality, that allows users to export their entire account and all associated media in a self-contained tar file.

## Archive structure
The tar.gz archive contains three parts:
1. `archive.json`: Containing the user settings, names and books used by the user.
2. `avatar.png`/`avatar.jpg`: A user's avatar image if present
3. `covers/`: the book covers used by the user

## Implementation
This project was implemented in collaboration during the first Guild Alpha Sprint.

We decided to use a custom JSON archive as the basis of this proposal, as @hughrun had already done work on that, which we used as a base. 

On this we added: 
- [X] Improved serialization of the `archive.json`
- [X] Move exporting and importing to celery tasks
- [X] Wrap the `archive.json` into a tar archive
- [X] Add media files to the archive
- [X] Unit Tests of the import and export process

Due to time constraints we couldn't implement everything needed. Here are the tasks, that we know of, that still need work:
- [ ] Cool-down period between creating archives to prevent overloading the server
- [ ] Regularly scheduled task to remove old archives according to an environment variable
- [ ] Notification on import & export completion
- [ ] Error handling & retry mechanism. Currently when an import or export fails the Task status remains active. The status doesn't show "failed" and no error is given.
- [ ] Admin controls

Hopefully with our work as a base, the remainder can be undertaken by BookWyrm itself to bring this PR to completion.

## Screenshots
import-example

export-example

#### `test_export_user.py` Errors
- `ExportUserViews.test_download_export_user_file` is throwing: `bookwyrm.models.bookwyrm_export_job.BookwyrmExportJob.DoesNotExist: BookwyrmExportJob matching query does not exist.`
- `ExportUserViews.test_export_user_get` is throwing: `Exception: line 367 column 5 - Warning: inserting implicit <p>`

### Contributors:
- @CSDUMMI
- @dannymate
- @hughrun 
- @tomat0

Special thanks to @Ryuno-Ki and [@circlebuilder](https://codeberg.org/circlebuilder/)

PR #3054, opened 2023-10-22
User migration via export file

### LINKED ISSUE #1012: Migrate Accounts To Another Instance & Import/Export

### LINKED ISSUE #2666: Create an option to import BookWyrm CSVs into a new BookWyrm account
I have an old BookWyrm ID that I want to move away from for painful, personal reasons. I created a new BookWrym account (same instance, bookwyrm.social), and I'd like to import my booklist from the old account to the new one. Exporting the CSV from the old account works just fine, but when I try importing the CSV to the new account, I get a "this is not a valid CSV" error no matter which option I use (e.g., Goodreads, OpenLibrary) I've attached my CSV export for reference:
[bookwyrm-export.csv](https://github.com/bookwyrm-social/bookwyrm/files/10745520/bookwyrm-export.csv)

An option to import BookWyrm CSVs into a BookWyrm account.

hughrun:
This PR aims to implement https://github.com/bookwyrm-social/bookwyrm/issues/1012. It introduces new export and import functionality, that allows users to export their entire account and all associated media in a self-contained tar file.

Resolves #1012 
Resolves #2666 

## Archive structure

The tar.gz archive contains three parts:

- `archive.json`: Containing the user settings, names and books used by the user.
- `avatar.png/avatar.jpg`: A user's avatar image if present
- `covers/`: the book covers used by the user

### Included

- user profile info
- basic user settings
- reading goal
- books
- readthroughs
- shelves
- book lists
- reviews
- comments
- quotes
- saved book lists
- follows (who you follow)
- blocks (who you block)

### Not included

- user setting relying on changeable server settings (e.g. themes)
- replies to statuses
- followers (these are managed in #2970)
- Groups (these are currently not federated so can't be migrated)

Also note that for reasons of expediency the JSON export is a custom structure and not in the form of ActivityPub JSON-LD.

## Implementation

This project was mostly implemented in collaboration during the first Guild Alpha Sprint.

We decided to use a custom JSON archive as the basis of this proposal, as @hughrun had already done work on that, which we used as a base.

Contributors:

- @CSDUMMI
- @dannymate
- @hughrun
- @Tomat0

Special thanks to @Ryuno-Ki and [@circlebuilder](https://codeberg.org/circlebuilder/)

Guild Alpha added:

- [x] Improved serialization of the archive.json
- [x] Move exporting and importing to celery tasks
- [x] Wrap the archive.json into a tar archive
- [x] Add media files to the archive
- [x] Unit Tests of the import and export process

Due to time constraints we couldn't implement everything needed. Subsequently @hughrun has added:

- [x] Cool-down period between creating archives to prevent overloading the server
- [x] Notification on import & export completion
- [x] Error handling & retry mechanism. Currently when an import or export fails the Task status remains active. The status doesn't show "failed" and no error is given.
- [x] Admin controls to monitor and stop user imports

## Related work

This was originally offered as #2980 - it was pulled into a new branch so the work could be completed
This is a sibling PR with #2970

## Remaining work - help wanted

- `mypy` doesn't like `bookwyrm/utils/tar.py` and I'm not sure what to do or whether I can just tell mypy to ignore types for now. There's also a related test that is failing
- there are some pylint errors in `test_bookwyrm_import_job` which indicate I probably need to refactor a test there but I can't figure out what I should be doing
- this is a big PR and needs a couple of people to thoroughly review it

## Other work - possibly in scope

- It may make sense to review how this works with #2970 as we can probably consolidate some menus and definitely will need some clear instructions. Possibly we could pull them both into `main` but check for potential streamlining before releasing as production-ready.

## Future work - not in scope

- We should have a scheduled task to remove old archives older than a certain amount of time. This is also needed to clean up failed book exports and orphaned image files from books and users, so it makes sense to deal with that separately. See #2996


bookwyrm/utils/tar.py
```diff
@@ -0,0 +1,46 @@
...
+    """Create tar files for user exports"""
+
+    def write_bytes(self, data: bytes) -> None:
+        """Add a file containing bytes to the archive"""
+        buffer = io.BytesIO(data)
+        info = tarfile.TarInfo("archive.json")
+        info.size = len(data)
+        self.addfile(info, fileobj=buffer)
+
+    def add_image(self, image: Any, filename: str = None, directory: Any = "") -> None:
```
dato, line 19:
```suggestion
    def add_image(self, image: Any, filename: Optional[str] = None, directory: Any = "") -> None:
```

bookwyrm/utils/tar.py
```diff
@@ -0,0 +1,46 @@
...
+            filename = f"{directory}{image.name}"
+
+        info = tarfile.TarInfo(name=filename)
+        info.size = image.size
+
+        self.addfile(info, fileobj=image)
+
+    def read(self, filename: str) -> Any:
+        """read data from the tar"""
+        with self.extractfile(filename) as reader:
```
dato, line 38: > `error: Item "None" of "IO[bytes] | None" has no attribute "__enter__"`

```suggestion
        if reader := self.extractfile(filename):
```

(`extractfile` can return None, plus if it doesn't, it's just a regular object that doesn't need enter/exit with `with`)

bookwyrm/utils/tar.py
```diff
@@ -0,0 +1,46 @@
...
+            return reader.read()
+
+    def write_image_to_file(self, filename: str, file_field: Any) -> None:
+        """add an image to the tar"""
+        extension = filename.rsplit(".")[-1]
+        with self.extractfile(filename) as reader:
+            filename = f"{str(uuid4())}.{extension}"
+            file_field.save(filename, File(reader))
```
dato, line 46: Same as above. 

```suggestion
        if buf := self.extractfile(filename):
            filename = f"{str(uuid4())}.{extension}"
            file_field.save(filename, File(buf))
```

bookwyrm/tests/models/test_bookwyrm_import_job.py
```diff
@@ -0,0 +1,567 @@
...
+                openlibrary_key="OL28216445M",
+                parent_work=self.work,
+            )
+
+        archive_file = pathlib.Path(__file__).parent.joinpath(
+            "../data/bookwyrm_account_export.tar.gz"
+        )
+        self.tarfile = BookwyrmTarFile.open(
+            mode="r:gz", fileobj=open(archive_file, "rb")
+        )
```
dato, line 70: > there are some pylint errors in test_bookwyrm_import_job which indicate I probably need to refactor a test there but I can't figure out what I should be doing

Not sure if this was about this warning, but here you could do?:

```suggestion
        with open(archive_file, "rb") as fileobj:
            self.tarfile = BookwyrmTarFile.open(mode="r:gz", fileobj=fileobj)
```

dato (commented):
Hi!

I took a look at the mypy/pylint errors you mentioned, please see below.

hughrun:
Thanks @dato those fixes do indeed seem to resolve most of the problems. And thanks @mouse-reeve for picking up my over-eager linting "fix" with the tuple.

I've sorted these out on my local, I'll probably push them up shortly.

Now that's resolved, I've done another manual test and realised there is something here creating duplicate IDs in the database. I think it's probably not the `BookwyrmImport` functions, because it's happening every time I do a normal add of a book from another bookwyrm site: just a chaos of duplicate user IDs all over the place. Having said that, we also seem to be getting duplicate statuses on import which is obviously not great. We have reports of duplicates on production sites now, so it's possible this is something that wasn't introduced in this PR but is manifesting more obviously for some reason.

I'll have to dig in and see if I can work out what's going on. Any extra help would be appreciated, for now I'm going to change this PR back to draft.

bookwyrm/models/bookwyrm_export_job.py
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    return DjangoJSONEncoder().encode(data)
+
+
+def get_books_for_user(user):
+    """Get all the books and editions related to a user
+    :returns: tuple of editions, books
+    """
+    all_books = models.Edition.viewer_aware_objects(user)
```
mouse-reeve, line 207: Does this need to be viewer aware? I don't see the prefetched data being used

bookwyrm/models/bookwyrm_export_job.py
```diff
@@ -0,0 +1,217 @@
...
+    except Exception:  # pylint: disable=broad-except
+        readthroughs = []
+
+    # books
+    editions, books = get_books_for_user(user)
+    final_books = []
+
+    for book in books.values():
+        edition = editions.filter(id=book["id"])
+        book["edition"] = edition.values()[0]
```
mouse-reeve, line 118: It took me a second to understand this, but it looks like you're loading the object as a queryset, and then using `queryset.values()` to give you an object representation of the edition, is that right?

I've been wondering if it's possible to use the activitypub formulation of books and authors, because then you could use the pre-existing code that converts activitypub editions/authors into model objects (that code also does deduplication against existing model entries). I think the only meaningful field related to the book itself that you'd miss was the cover? But I'm not sure if it would be prohibitively hard to relate the user's data (like readthrough) to the activitypub serialization.

bookwyrm/models/bookwyrm_export_job.py
```diff
@@ -0,0 +1,217 @@
...
+            books__in=[book["id"]], user=user
+        ).distinct()
+        book["lists"] = list(book_lists.values())
+        book["list_items"] = {}
+        for blist in book_lists:
+            list_items = models.ListItem.objects.filter(book_list=blist).distinct()
+            book["list_items"][blist.name] = list(list_items.values())
+
+        # reviews
+        reviews = models.Review.objects.filter(user=user, book=book["id"]).distinct()
```
mouse-reeve, line 157: Would it work to use `models.Status.objects.select_subclasses.filter(...)` to collect all the subclasses, rather than doing them separately?

bookwyrm/templates/import/import_user.html
```diff
@@ -0,0 +1,156 @@
...
+        {% csrf_token %}
+
+        <div class="columns">
+        <div class="column is-half">
+            <div class="field">
+                <label class="label" for="id_archive_file">{% trans "Data file:" %}</label>
+                {{ import_form.archive_file }}
+            </div>
+            <div>
+                <p class="block"> {% trans "Importing this file will overwrite any data you currently have saved." %}</p>
```
mouse-reeve, line 42: I think this needs some clarification -- it looks to me like this adds additional data, and _can_ change or override existing data in surprising ways, but doesn't, for example, delete a readthrough that already exists if there's a new one. The wording as is makes it sounds like your current data would be completely cleared (which some people really really want and would be sad when it didn't happen)

mouse-reeve (commented):
This seems like it works super well from a basic test. I will try to test edge cases, but for now some questions :)

bookwyrm/models/bookwyrm_import_job.py
```diff
@@ -0,0 +1,515 @@
...
+
+    common = data.keys() & values
+    new_data = data
+    for val in common:
+        del new_data[val]
+    return new_data
+
+
+def find_existing(cls, data):
+    """Given a book or author, find any existing model instances"""
```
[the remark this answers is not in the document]
hughrun: Happy to take advice on this and make things cleaner, but the primary reason we didn't use `model.find_existing` is that we absolutely do NOT want to match on ids. The incoming ids are likely from a different database, so they will match on a different model instance.

bookwyrm/models/__init__.py
```diff
@@ -26,6 +26,7 @@
 from .group import Group, GroupMember, GroupMemberInvitation
 
 from .import_job import ImportJob, ImportItem
+from .bookwyrm_import_job import BookwyrmImportJob
```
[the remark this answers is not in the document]
hughrun: From memory this was because we have `from bookwyrm import models` in `BookwyrmExportJob` and we have up trying to resolve a horrendous loop of infinitely-self-referencing references. If you can see a way to get around this I agree it would be better.

bookwyrm/models/bookwyrm_export_job.py
hughrun: Honestly I think I just copied this from somewhere else when I was initially seeing whether it might work. So this should be something like this instead:

`all_books = models.Edition.objects.filter(user=user)`?

bookwyrm/models/bookwyrm_export_job.py
hughrun: > It took me a second to understand this, but it looks like you're loading the object as a queryset, and then using queryset.values() to give you an object representation of the edition, is that right?

Correct.

I'd be very happy to adjust this with advice. Initially I thought we probably should do exports and imports as ActivityPub objects for the reason you've noted - most of the code is already there - but it was proving to be quite challenging. Then I was trying to do something kinda silly to get the Django model instances out as JSON, and then we ended up doing it this way. Images end up kind of having to be pulled out separately anyway because all we can get at this point is a filepath, so if you can walk me through what you're thinking a bit more, definitely happy to keep it simpler.

bookwyrm/models/bookwyrm_export_job.py
hughrun: Ooh. I forgot that existed. Uh, probably? We'd need to make sure we still store them separately or at least retain knowledge of the underlying model though.

bookwyrm/templates/import/import_user.html
hughrun: Yes. Actually I'd really like to go over that with you to make sure we're clear on what we're overwriting and what we're adding (and that what's happening is intentional). It would be good to make it less surprising.

What it _should_ do, I believe, is:

* always overwrite relevant user profile and settings values
* **not** create new editions or authors if there is a matching one
* **not** overwrite existing identical readthroughs, goals, statuses, or lists
* only follow or block users that aren't currently followed or blocked

So the intention is that most things are "no-op" on matching things that already exist, except user settings and profile. My thinking on this is that people might want to migrate in to an account that they've been running as an alt, and don't want to wipe out everything they did previously. Bulk-deleting is really a separate question.

bookwyrm/models/bookwyrm_import_job.py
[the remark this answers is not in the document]
mouse-reeve: It shouldn't ever match on id, since that's not a deduplication field. I wrote a couple test just to make sure I was right and I seem to be: https://github.com/bookwyrm-social/bookwyrm/pull/3084

mouse-reeve:
Oops, I did the merge conflicts so I could do some stuff on the branch forgot that it's going to need a merge migration too

bookwyrm/models/__init__.py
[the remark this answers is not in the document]
mouse-reeve: I think this will do it: https://github.com/bookwyrm-social/bookwyrm/pull/3085

bookwyrm/models/bookwyrm_export_job.py
mouse-reeve: Yes, that looks right. In this case, `viewer_aware` preloads some user data about the books that makes the templates render more efficiently.

bookwyrm/models/bookwyrm_export_job.py
mouse-reeve: I'm not surprised that using activitypub ended up having a lot of challenges, and I'm satisfied knowing that you tried it and it turned out to be a major headache.

bookwyrm/models/bookwyrm_import_job.py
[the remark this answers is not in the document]
hughrun: Ah, I see where I confused myself. I was thinking [this referred to the database id](https://github.com/bookwyrm-social/bookwyrm/blob/a93519ec3e45bf91dba96e48288e81fbc9bb3e18/bookwyrm/models/activitypub_mixin.py#L112C1-L114C54), but it's actually the ActivityPub id

bookwyrm/models/bookwyrm_export_job.py
hughrun: I took a look at this but it seemed to make things more complicated? I can see that it would reduce the database queries, but since we still need to know what class everything is and get it into JSON, I'm struggling to see how to use it in a way that makes things simpler rather than more complex. Advice welcome.

bookwyrm/templates/import/import_user.html
hughrun: @mouse-reeve I see what you mean now and updated this, but it may be too verbose 😬

bookwyrm/models/bookwyrm_import_job.py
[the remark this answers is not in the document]
hughrun: Ok @mouse-reeve I've tried this out. It's not going to work because `find_existing` from `activitypub_mixin` is looking for activitypub fieldnames, and we just have Bookwyrm class fieldnames.

This is only used for authors and editions, and those are relatively unproblematic to work with as ActivityPub JSON. So notwithstanding my comment that dealing with AP JSON was challenging, maybe we should take another look at how that might be done at least for book data?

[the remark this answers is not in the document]
mouse-reeve: I think if it's possible to do with book data, that would have the most benefit, since there's pretty well tested code in place for converting an activitypub version of a book to a model entry. At minimum, this deduplicator function should use deduplication fields rather than enumerating them here

bookwyrm/models/bookwyrm_export_job.py
mouse-reeve: I think it was just not a helpful suggestion on my part and you can skip it

bookwyrm/templates/import/import_user.html
mouse-reeve: Do you think it would work to combine the category-by-category description in with the checkboxes? That might make it easier to parse

bookwyrm/models/bookwyrm_import_job.py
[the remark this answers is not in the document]
hughrun: ok ok ok @mouse-reeve this can work! _almost_.

It's totally broken right now but [see this PR in my repo](https://github.com/hughrun/bookwyrm/pull/1) so you can possibly find what's happening. Basically I now realise how we can pretty easily grab almost everything other than Readthroughs as AP JSON. I haven't looked at the import side yet but I'm reasonably confident about how it will be done.

One small problem: I'm getting an Exception from `DjangoJSONEncoder`:

```
bookwyrm-celery_worker-1   | User Export Job 37 Failed with error: Object of type _MISSING_TYPE is not JSON serializable
bookwyrm-celery_worker-1   | Traceback (most recent call last):
bookwyrm-celery_worker-1   |   File "/app/bookwyrm/models/bookwyrm_export_job.py", line 46, in start_export_task
bookwyrm-celery_worker-1   |     json_data = json_export(job.user)
bookwyrm-celery_worker-1   |   File "/app/bookwyrm/models/bookwyrm_export_job.py", line 214, in json_export
bookwyrm-celery_worker-1   |     return DjangoJSONEncoder().encode(data)
bookwyrm-celery_worker-1   |   File "/usr/local/lib/python3.9/json/encoder.py", line 199, in encode
bookwyrm-celery_worker-1   |     chunks = self.iterencode(o, _one_shot=True)
bookwyrm-celery_worker-1   |   File "/usr/local/lib/python3.9/json/encoder.py", line 257, in iterencode
bookwyrm-celery_worker-1   |     return _iterencode(o, 0)
bookwyrm-celery_worker-1   |   File "/usr/local/lib/python3.9/site-packages/django/core/serializers/json.py", line 105, in default
bookwyrm-celery_worker-1   |     return super().default(o)
bookwyrm-celery_worker-1   |   File "/usr/local/lib/python3.9/json/encoder.py", line 179, in default
bookwyrm-celery_worker-1   |     raise TypeError(f'Object of type {o.__class__.__name__} '
bookwyrm-celery_worker-1   | TypeError: Object of type _MISSING_TYPE is not JSON serializable
```

Looks like this is coming from the edition after it's exported via `to_activity()`:

```json
{
    "id": "https: //example.com/book/1",
    "type": "Edition",
    "links": <dataclasses._MISSING_TYPE object at 0xffff80d0faf0>,
    "fileLinks": [],
    "title": "Test book 1",
    "sortTitle": "Test book 1"
    ...
}
```

Not sure what's going on here, it feels like there's something wrong in `to_activity()` for `Link` fields.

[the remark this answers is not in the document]
mouse-reeve: It's very weird that that's causing an error, but `links` _are_ in the activitypub spec for editions, even though they're not used, and I don't think are even in the model. Would removing it from `bookwyrm/activitypub/book.py` fix it?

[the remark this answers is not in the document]
hughrun: Aha! So they are. That explains it. The model has `file_links` but not `links`, so the `BookData` model is wrong (it has both). When I get home I'll try taking that out and see if it works.

hughrun:
@mouse-reeve I'm having some issues with `to_model` throwing errors when optional values are missing from the AP object - can you please look at this and let me know if you have any bright ideas? https://github.com/hughrun/bookwyrm/pull/1

hughrun:
@mouse-reeve I think this is ready.

All the tests are failing because of a giant migration leaf node mess. I tried deleting the migrations that aren't in `main` and running them again but Django's not happy and I'm too scared to try anything else! Can you please take a look and advice/push a fix?

The pylint is just a an extra blank line from when I fixed a merge conflict, I think.

hughrun:
hmm Ok I'm not sure what's happening with that test, I didn't touch those migration files.

hughrun:
idk what is going on here, everything passes locally.

mouse-reeve:
I can't get ANY of the tests to run locally. It looks like migration 0187 is referencing `0186_alter_notification_notification_type`, which should actually be `0185_alter_notification_notification_type`? I tried deleting the migrations and making new ones with no merge migrations at all.

That doesn't fix the failing migration test, but I think that's because that test is really brittle -- it looks like it's failing because it's using the database schema pegged to a specific migration but the current code, which is in a different state.  I propose deleting it.

mouse-reeve:
On reflection, I wonder if you have a local copy of migration 0186 that was allowing your tests to pass? At any rate, if the commit I just made passes, I think that's good? If not (or if you don't think we should delete that test) I'm happy to nix it.

hughrun:
Looks like that worked!

I think I messed up a merge somewhere, but also this PR has taken so long with so many different merges it's a bit trickier to keep all the migrations aligned.

I would recommend testing a `migrate` on a pre-existing DB once this is merged into `main`, before releasing it to `production`.

mouse-reeve:
I made a small change to fix a bug in how notifications were being created, and changed the wording to (hopefully) clarify the difference between CSV import/export and user import/export, since user import/export does include both user and book data:
Screenshot 2023-12-09 at 8 19 26 AM

Otherwise this looks good to me. I'd like to get it merged and do a release early next week 🥳 -- it would also be great to get someone else to test this (@bookwyrm-social/code-review) both for bugs and for clarity.

dato:
Hello! (I apologize that I didn't run these tests much earlier.)

I tested this on my development instance, so I exported data for my main user A (id=2), then created an empty user B (id=90) and imported there.

While books in shelves where imported correctly:

```
bookwyrm_devel=> SELECT user_id, book_id, shelf_id
  FROM bookwyrm_shelfbook
  WHERE user_id in (2, 90)
  ORDER BY by 1, 2, 3;

 user_id | book_id | shelf_id 
---------+---------+----------
       2 |       8 |        6
       2 |       8 |        9
       2 |       9 |        6
      90 |       8 |       19
      90 |       8 |       22
      90 |       9 |       19
```

something weird (bad?) happened with status: it would seem they simply got reassigned to user id 90 (in the web UI, indeed user A has no statuses in their Activity tab after the import):

```

bookwyrm_devel=> SELECT id as status_id, user_id, created_date, updated_date
  FROM bookwyrm_status
  WHERE user_id in (2, 90)
  ORDER BY 3;

 status_id | user_id |         created_date          |         updated_date          
-----------+---------+-------------------------------+-------------------------------
         1 |      90 | 2023-09-24 19:17:48.920904-03 | 2023-12-11 05:04:34.542547-03
         3 |      90 | 2023-10-14 22:13:13.460129-03 | 2023-12-11 05:04:34.660894-03
         2 |      90 | 2023-09-24 19:51:24.496355-03 | 2023-12-11 05:04:34.949896-03
         4 |      90 | 2023-10-14 22:13:35.197837-03 | 2023-12-11 05:04:35.06195-03
         5 |      90 | 2023-11-15 18:52:25.785505-03 | 2023-12-11 05:04:35.16605-03
(5 rows)
```

Importing into the export instance is arguably a corner case, albeit one I think people might need to use for username changes?

I haven't had a chance to look into the code yet, though I suspect the fix might be easier than this write-up!, heh.

hughrun:
@dato thanks for being so thorough.

> it would seem they simply got reassigned to user id 90 

I hadn't expected this, but it makes sense. In the original way I wrote this it wouldn't have happened, but making export files as close to ActivityPub objects has a lot of advantages so we use `to_model()` on import.

The issue here is in `bookwyrm_import_job`:

```python
instance = parsed.to_model(model=cls, save=True, overwrite=True)
````
The only way I could get this to work is to use `overwrite=True`, which ...does what you would expect. This is what I was struggling with in https://github.com/hughrun/bookwyrm/pull/1 So we swap out the `User` to avoid assigning the status to, to take your example, the user with an pk of `2` on a new server. But if it's the same server then the effect is to just keep the statuses that already existed but change the user id. This is actually fine if everyone is doing the right thing, but as you point out the import file can edited or created from scratch.

> I was wondering if this was minor enough that the PR could be merged anyway, but I just realized this means a malicious user could (perhaps?) craft an archive.json that deletes another user’s data.

I admit this is something I hadn't thought of, and you're right, it would be possible for a malicious user - though you would also need to know the user's primary key, which isn't necessarily obvious, and craft an export file targeted at their post IDs. This PR was originally started before #3076, but now that's merged in we could potentially add a check to insist that the source account has to have `movedTo` set to the target account. That would however make it impossible to migrate an account after the source server is shut down or if the source account is deleted rather than 'moved', for example. Perhaps that's a price we have to pay to remove a potential vector for malicious activity?

**Comment by dato** [comment]
> so we use `to_model()` on import

I took a closer look because of this comment, and there's an issue with `remote_id`, because the new statuses preserve the `remote_id` field, pointing to the old URLs. Clicking on their permalink (_“15 hours ago...”_) fails:¹

```
bookwyrm_devel=> SELECT s.id, username, s.remote_id
  FROM bookwyrm_status s
  JOIN bookwyrm_user u
  ON u.id = s.user_id
  WHERE user_id = 90
  ORDER BY 1;

 id |           username           |                        remote_id                        
----+------------------------------+---------------------------------------------------------
  1 | import@libros.dev.rizoma1.ar | https://libros.dev.rizoma1.ar/user/admin/comment/1
  2 | import@libros.dev.rizoma1.ar | https://libros.dev.rizoma1.ar/user/admin/comment/2
  3 | import@libros.dev.rizoma1.ar | https://libros.dev.rizoma1.ar/user/admin/reviewrating/3
  4 | import@libros.dev.rizoma1.ar | https://libros.dev.rizoma1.ar/user/admin/review/4
  5 | import@libros.dev.rizoma1.ar | https://libros.dev.rizoma1.ar/user/admin/review/5
(5 rows)
```

(¹) It fails because those URLs don't exist any more. But if this happens in cross-instance imports (I'm not sure that it does), it could be considered worse since those permalinks should stay in-instance.

**Comment by hughrun** [comment]
dang. I missed that, I think I assumed it would create a new remote_id but logically, why would it? I think I see what needs to happen here, will take a look tonight my time (and also remove a few rogue `print` statements).

**Comment by mouse-reeve** [comment]
I’m so glad you caught that, it does make sense that the old remote is remains. Every model should have a function to generate the remote id which we can re-run and save

**Comment by hughrun** [comment]
I think I've got a working solution for this. This solution will mean the import still works for everything else, but statuses won't be created if the source account hasn't been set as `movedTo` the account that we're importing into: i.e. you have to export, then "move" the account, _then_ you import the user data to the new account.

I'm not sure of the best way to present any errors to the user though. If it fails to import statuses that will probably be because they're a legit user who forgot to do things in the correct order, or maybe doesn't have access to the old account any more, so we need to provide clear info about why statuses haven't imported rather than just silently  not importing them.

Any suggestions @mouse-reeve ?

**Comment by hughrun** [comment]
@dato 

1. Yes. I momentarily thought maybe I'd done something wrong because I was getting duplicated statuses, but overnight I realised that the only thing that BookWyrm matches on for statuses is the `remote_id`. When that is changed a new status will _always_ be created. This is a very sensible default usually.

However, that creates a new problem in that we don't always want to do that when importing (e.g. if for some reason a user imports the same file twice). So I'll have to add some custom logic there.

2. This is a good suggestion and would be easy to do. If the target user is set as `alsoKnownAs` OR `movedTo` by the source, it's ok.

**Comment by hughrun** [comment]
@dato @bookwyrm-social/code-review please take a look at this now. 

Generally, does anything look problematic? 

Specifically:

- does the latest commit resolve the bug @dato identified?
- does it introduce any new problems?
- are the messages about setting an alias sufficient?

File: bookwyrm/models/bookwyrm_import_job.py [hunk 11]
```diff
@@ -0,0 +1,459 @@
...
+        remote_id=remote_id, model=models.User, save=False
+    )
+
+    if remote_user:
+
+        if remote_user.moved_to:
+            return user.remote_id == remote_user.moved_to
+
+        if remote_user.also_known_as:
+            return user in remote_user.also_known_as.all()
```
dato, line 446: The documentation talks about "either moved or alias" being enough, so perhaps do not do a hard return? (Potentially one could set an alias to X, then move to Y, then want to import into X.)

dato (commented):
> - does the latest commit resolve the bug @dato identified?
> - does it introduce any new problems?
> - are the messages about setting an alias sufficient?

I believe the commit is good, yes!

One tiny comment below, but absolutely not a blocker, is a corner case (again).

Thanks!

bookwyrm/models/bookwyrm_import_job.py
hughrun: `moved_to` [will only ever have a value if there is an active `move`](https://github.com/bookwyrm-social/bookwyrm/blob/000e5e614532a87b79d21a3b266713dff8c186db/bookwyrm/views/preferences/move_user.py#L103C1-L108C1). So if `remote_user.moved_to` is truthy then we know the remote user is currently moved to somewhere. In this scenario, if our local user is an alias but the user they have moved to is _not_ our local user, we definitely don't want to import statuses. Hence the hard return.

dato:
👏🏼👏🏼👏🏼

mapperfr:
Hi, 

I've been trying to use the "Move account" and/or "Import/Export account" functions between two different up to date instances, with no success so far. 

I could describe the issues on this thread on open a new issue. Any preferences?

PR #3055, opened 2023-10-22
formatting and linting fixes

hughrun:


bookwyrm/models/bookwyrm_export_job.py
```diff
@@ -49,21 +51,22 @@ def start_export_task(**kwargs):
...
+    file.open("wb")
+    with BookwyrmTarFile.open(mode="w:gz", fileobj=file) as tar:
         tar.write_bytes(json_data.encode("utf-8"))
 
         # Add avatar image if present
         if getattr(user, "avatar", False):
             tar.add_image(user.avatar, filename="avatar")
 
-        editions, books = get_books_for_user(user)
+        editions = get_books_for_user(user)
```
mouse-reeve, line 64: Isn't this going to return a tuple stored in `editions`, rather than the original behavior that stores just the first value of the tuple?

hughrun: hmm, yes. Looks like I was a little hasty "fixing" pylint complaints.

PR #3068, opened 2023-10-29
make options consistent in celery

### LINKED ISSUE #3066: Celery queue names inconsistent
In the admin interface, the Celery Status tab shows all the current queues.
They seemed to be only ever-increasing recently, so I had to look into why that happened.
After a lot of wasting of time, I figured out how to properly get into Flower, (and how to start it...)
Here, I discovered that the Celery worker was ignoring all but four of the queues, so I had to manually add them.
This did not really seem to be a problem. However, when I entered `broadcasts`, as that is what the queue is named in the admin interface, nothing happened. An investigation later I found out that the queue was actually called `broadcast`, singular.

So, please fix that. Either change the name of the queue, or change it in the admin interface, so that someone can look at the admin interface to know what the queue names are...

N/A

**Additional context**
N/A

hughrun:
- changes 'broadcasts' to 'broadcast' in Celery page in admin section
- re-orders celery queues on admin page to be in English alphabetical order (other than priority levels) - this makes them consistent with the Flower interface
- fixes #3066


mouse-reeve:
Thank you!

PR #3070, opened 2023-10-30
File resubmit Option 1

### LINKED ISSUE #2760: Add Book Manually Loses The Cover Image Uploaded

hughrun:
This fixes #2760 by using [django-file-resubmit](https://github.com/un1t/django-file-resubmit).

Pro: this is an off-the-shelf import

Con: `django-file-resubmit` hasn't been updated for 4 years, is written for Python 2, and leaves a confusing file input indicating that no file has been selected, when it has actually cached the file.

I will send a second PR shortly with an alternative option which pulls the code directly into Bookwyrm. It's MIT licensed so we can do this.

I believe ~~Option 1~~ Option 2 is the better option, but am providing both to compare.

### DIFF
--- a/bookwyrm/forms/books.py
+++ b/bookwyrm/forms/books.py
@@ -3,6 +3,7 @@ from django import forms
 
 from bookwyrm import models
 from bookwyrm.models.fields import ClearableFileInputWithWarning
+from file_resubmit.admin import AdminResubmitImageWidget
 from .custom_form import CustomForm
 from .widgets import ArrayWidget, SelectDateWidget, Select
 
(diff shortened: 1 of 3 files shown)


mouse-reeve:
option 2 sounded better to me, since the library appears to be abandoned, so I'm glad we're now in agreement

PR #3071, opened 2023-10-30
File resubmit Option 2 - Retain uploaded images when creating new books

### LINKED ISSUE #2760: Add Book Manually Loses The Cover Image Uploaded

hughrun:
django-file-resubmit works ok, but it hasn't been updated for 4 years and seems to be abandoned. It also:

- leaves a file selector visible indicating no file is selected, which is confusing
- is written for Python 2

This PR brings the relevant parts of django-file-resubmit into Bookwyrm directly, under utils, and fixes the issue with the file selector.

Fixes #2760

### DIFF
--- a/bookwyrm/forms/books.py
+++ b/bookwyrm/forms/books.py
@@ -3,6 +3,7 @@ from django import forms
 
 from bookwyrm import models
 from bookwyrm.models.fields import ClearableFileInputWithWarning
+from file_resubmit.admin import AdminResubmitImageWidget
 from .custom_form import CustomForm
 from .widgets import ArrayWidget, SelectDateWidget, Select
 
@@ -70,9 +71,7 @@ class EditionForm(CustomForm):
             "published_date": SelectDateWidget(
                 attrs={"aria-describedby": "desc_published_date"}
             ),
-            "cover": ClearableFileInputWithWarning(
-                attrs={"aria-describedby": "desc_cover"}
-            ),
+            "cover": AdminResubmitImageWidget(attrs={"aria-describedby": "desc_cover"}),
             "physical_format": Select(
                 attrs={"aria-describedby": "desc_physical_format"}
             ),
--- a/bookwyrm/forms/books.py
+++ b/bookwyrm/forms/books.py
@@ -2,8 +2,7 @@
 from django import forms
 
 from bookwyrm import models
-from bookwyrm.models.fields import ClearableFileInputWithWarning
-from file_resubmit.admin import AdminResubmitImageWidget
+from bookwyrm.utils.file_resubmit import AdminResubmitImageWidget
 from .custom_form import CustomForm
 from .widgets import ArrayWidget, SelectDateWidget, Select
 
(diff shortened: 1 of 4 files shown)


mouse-reeve:
Would it potentially make sense to maintain a fork of the library (or a copy of it) in a separate repo under the bookwyrm project rather than including it as a file?

hughrun:
Yes, it would be more sensible! I hadn't thought of doing it as a BookWyrm thing and was reluctant to make a fork I was individually responsible for, since I'm not confident I completely understand all the code.

I can work on that, it would be more like Option one from the perspective of the main project.

mouse-reeve:
Sweet, that all sounds good to me

mouse-reeve:
Oh, let me know if you need me to create anything in the org if there are permissions issues

hughrun:
@mouse-reeve I don't have permission to create a new repo. I actually ran into this problem last week which is why I had to create so many PRs to update the branch for user migration.

A separate issue/question: would this be something we'd want to add to pypi? If so we'll need to work out logistics and workflow for that.

mouse-reeve:
Okay, should I fork it or create a new repo?

hughrun:
> Okay, should I fork it or create a new repo?

I can't see any advantage to a "fork", since the reason we're doing this is that the current repo seems to be abandoned. If I'm understanding correctly, all prior contributors will appear in our new repository. I intend to pull as many of the outstanding PRs as we can do without too much fuss. Probably I'll drop an "Issue" in to the existing repository pointing to ours, which is a bit cheeky but I can't think of any other way to indicate that we've created an active fork.

hughrun:
Sorry didn't actually answer your question. We should do a new repo as this allows us to change the name if we want, and also to change the main branch to `main` instead of `master`.

mouse-reeve:
Perfect -- you should have full control over https://github.com/bookwyrm-social/django-file-resubmit (and you're welcome to change the name)

PR #3086, opened 2023-11-05
Erase user data and statuses on account deletion

### LINKED ISSUE #2420: GDPR requires data to be deleted on request
GDPR requires that user data is deleted on request or on withdrawal of consent (inferred by a user deleting their account). However, the delete option only marks an account as deleted and as far as I can tell retains all the data for that user in the database. This needs to be addressed to be compliant with GDPR in EU/UK.

### LINKED ISSUE #2911: Unable to make new account using same email after deleting old account
**Describe the bug**
I deleted my Bookwyrm account some time ago, and I now want to make a new account. However, I am unable to register for a new account. Bookwyrm complains: "User with this Email already exists." Could it be that I am attempting to register using the same email that was associated with the deleted account? That is weird behavior. Why can't I register using the same email address that was used in a deleted account?

I see that there is #1445, but the "solution" there is atypical and completely surprising for web applications.

>re-registration can still be done by using the + syntax supported by most email providers

**Instance**
https://bookwyrm.social

### LINKED ISSUE #2951: Question: can deleted accounts be deactivated? And if so should they?
I get emails quite often from people who have permanently deleted their account and would now like to un-delete it so they can use their username. I tried to make it very clear from the UI and provide deactivation as an alternative, but it still comes up quite a bit. 

My impression is that it would be effectively impossible to properly reinstate a user action a Delete activity has been broadcast, because the deletion has propagated to various instances, and any subsequent activities from that user would understandably be rejected.

But I wanted to make sure that I'm actually correct about this!

mouse-reeve:
Fixes #2420 
Fixes #2911 
Fixes #2951 

This updates how we store deactivated users and deleted users, giving deleted users their own `is_deleted` field rather than relying on the deactivation reason to determine if a user is deleted or just suspended/deactivated.

It also purges user profiles, emails, and statuses when a user is deleted (and retroactively for existing deleted users). 

As a little bonus, I also refactored the box in the admin panel that shows an account's status.


mouse-reeve:
@bookwyrm-social/code-review It would be great to get a thorough set of eyes on this

joachimesque (approved):
It looks good to me :)

PR #3089, opened 2023-11-06
UI changes for notification and download link in import/export flow

mouse-reeve:
The first change adds a link to the import notification, and changes the export notification to link to the page, rather than the file itself (I found that unexpected behavior). I also changed the table to only show one date (I'm not sure if this is a good change, but I thought both seemed like more info than needed?) and to use an explicit download link rather than having people click the date, which I think is clearer. I was looking at the mastodon export UI for ideas, and I also like that they should file size, but I don't know if that's easy to get.

Screenshot 2023-11-06 at 8 41 26 AM

### DIFF
--- a/bookwyrm/templates/notifications/items/user_export.html
+++ b/bookwyrm/templates/notifications/items/user_export.html
@@ -1,11 +1,15 @@
 {% extends 'notifications/items/layout.html' %}
 {% load i18n %}
 
+{% block primary_link %}{% spaceless %}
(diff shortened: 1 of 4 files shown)


hughrun (commented):
This looks nice! I added file size and a small-but-important fix in #3091 so if you're happy with that, you can pull that in to this branch and then pull this branch into `user-migration`. Looks like this:

user-downloads

PR #3090, opened 2023-11-06
User migration instructions

mouse-reeve:
What do you think about this as an alternative way of presenting the user import form, @hughrun ? It's very long on the page, but more explicit

One thing I specifically wanted to add is that you can only import from a bookwyrm export, because I can imagine people trying to use mastodon exports.

Screenshot 2023-11-06 at 9 34 22 AM

### DIFF
--- a/bookwyrm/templates/import/import_user.html
+++ b/bookwyrm/templates/import/import_user.html
@@ -29,85 +29,124 @@
             <p>{% blocktrans %}You will next be able to import a user file at {{ next_available }}{% endblocktrans %}</p>
         </div>
     {% else %}
-    <form class="box" name="import-user" action="/user-import" method="post" enctype="multipart/form-data">
+    <form class="box content" name="import-user" action="/user-import" method="post" enctype="multipart/form-data">
         {% csrf_token %}
 
-        <div class="columns">
-        <div class="column is-half">
-            <div class="field">
+        <div class="block">
+            <div class="notification">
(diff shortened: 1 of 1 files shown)


hughrun:
Oh wow this looks awesome! So much better 🙂

The other thing I meant to check with you is whether the boxes should be checked by default or unchecked. Checked by default is an _easier_ user experience, and assumes the user wants to import everything. But unchecked by default is a _safer_ option.

And as a slight tangent: We probably need a little more info on the export side. For example we don't export DMs - this was a deliberate (though admittedly somewhat arbitrary) decision, primarily because you'd moving third-party messages expected to be _somewhat_ private onto another server, and partially because it would add extra complexity I wasn't ready to handle. Conveniently Mastodon won't do this either. And also - importantly for some users probably - we don't export any other third-party statuses (i.e. we don't export replies to your statuses). We _could_ do that, and the main reason we didn't is because it was already a really big bunch of code and I didn't want to make it even more complicated - and _also_ because I have some misgivings about publishing replies to posts that were originally published under a different account on a different server. But anyway since we're not, we probably need to explain that. Then the question is whether that all needs to be explained in the UI, or it would be better to write some clear documentation at `docs.joinbookwyrm.com`

mouse-reeve:
Thanks! Those are all good thoughts on the export. I think it would be good to have it on the page to the degree it's possible, since it's a lot more likely to actually get read there. I think the limitations you've placed seem pretty reasonable to me (I have similar misgivings), and at worst can be reconsidered in the future

Issue #2066, opened 2022-03-27
Notification for admin/moderator when someone makes an invite request


Stoori:
**Is your feature request related to a problem? Please describe.**
At the moment the only way to know if there are invite requests is to go to the admin page and look. If invite requests are happening with long intervals, it's sometimes hard to remember to do that, and as a result the prospective users may have to wait for a long time before the admin remembers to check them.

**Describe the solution you'd like**
A notification in the UI for admins and moderators (whoever have the right to accept invite requests) whenever a new invite request is made. Opt-in/opt-out setting for those admins and moderators who don't want to see these notifications.

**Describe alternatives you've considered**
Email notification is another option, but I'd prefer UI notifications. Perhaps admins and moderators could have the choice of how they want to receive these?

PR #3099, opened 2023-11-10
Create NotificationType as class, not through API

dato:
This way, we need not list every value again to create the enum.

N.B.: Enum values are now accessed as `models.NotificationType.FOO`,
instead of `models.Notification.FOO`.


dato:
I was taking a look at the code for #2066, and saw this improvement could be easily done.

mouse-reeve:
omg THANK YOU this is so much better, I know the way I was doing this couldn't be right

PR #3110, opened 2023-11-13
Do not create AggregateRating if review_count is zero

dato:
This is a follow-up to #3036 ("Minor vocabulary fixes and structured data improvements"), where @jaschaurbach suggested including this fix.


dato:
templates check fails because of [curlylint#14](https://github.com/thibaudcolas/curlylint/issues/14) (open since Aug. 2020). It doesn't seem it's going to be fixed any time soon, and I'm not sure this micro-fix is worth making the template code even less readable for it. Feel free to close!

PR #3112, opened 2023-11-13
Check no missing migrations in django-tests workflow

dato:
-----

I missed a migration from #3099 (I missed it because, AFAICS, it's a no-op migration, yet Django wants it).

In this PR, I add the missing migration, plus a test so that this doesn't happen in the future. (Adding the test alone first so as to see it fail.)


mouse-reeve:
Good call, thank you!

Issue #2926, opened 2023-07-29
Ratings aren't switched when switching editions


rmcminds:
I've imported many books from Goodreads, but the editions are usually wrong. When I use the 'switch to this edition' button, it correctly switches the version that's on my shelf and that has 'read dates' attached to it - but then it looks like I haven't rated it, because the rating stays with the old edition.

joric3:
This has bothered me, too. An option to move user data (shelves, read dates, comments and reviews...) from one edition to another would be great.

PR #3117, opened 2023-11-16
Move ratings and reviews when switching editions

mattlehrer:
Fixes [#2926](https://github.com/bookwyrm-social/bookwyrm/issues/2926)

This is my first contribution to Bookwyrm and, more importantly, my first time looking at Django and pytest.   I am not (yet) a python person. The code and tests are pretty simple but something is not making sense. 

Most importantly, the ratings and reviews do move to the new edition in manual testing. If I comment out the for loop or just the line `readthrough.book = new_edition` then the associated test fails. Same for `review.book = new_edition` and my new test for reviews. All good so far. However, if I remove the new code for ratings or just the `rating.book = new_edition` line, the new test for ratings still passes. I feel crazy. 

I hope this is at least a step in the right direction. Thanks!


bookwyrm/views/books/editions.py
```diff
@@ -103,4 +103,18 @@ def switch_edition(request):
         readthrough.book = new_edition
         readthrough.save()
 
+    ratings = models.ReviewRating.objects.filter(
+        book__parent_work=new_edition.parent_work, user=request.user
+    )
+    for rating in ratings.all():
+        rating.book = new_edition
+        rating.save()
```
dato, line 111: If the above above is correct, perhaps it means this block could be completely dropped. I'm sure one of the maintainers will be able to confirm.

Thanks again!

dato (commented):
> This is my first contribution to Bookwyrm and, more importantly, my first time looking at Django and pytest. I am not (yet) a python person.

Welcome! And thanks for contributing. I'm a fairly new contributor myself, and interacting with the community has been a pleasure so far. <3

> The code and tests are pretty simple but something is not making sense.

Ah, I cloned your PR and then followed your steps... (behaved as you described), then started reading about the models a bit and noticed...

> However, if I remove the new code for ratings or just the `rating.book = new_edition` line, the new test for ratings still passes. I feel crazy.

... I believe this weird behavior could be accounted for by the fact that ReviewRating inherits from Review, and hence the loop over `Review.objects` would cover ReviewRating as well?

bookwyrm/views/books/editions.py
mattlehrer: Yes, this is it! I confirmed this by creating a ReviewRating and see that the Review loop picks it up. It makes sense. I will update the PR now. Thank you!

dato (approved):
Hi again!

Apologies for the incredible delay. I sat down to re-review and test this, and everything looks good. I'm going to trigger a re-run of the tests against the latest _main_, and merge once they pass.

Thanks again for your contribution!

Issue #2079, opened 2022-04-08
Let a user search within their books


mouse-reeve:
It would be cool if you had an option to search within just the books that you have shelved, to make it easier to find something on your shelves. This would be similar to the [existing book search](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/book_search.py) in that it would need to search on identifiers and use the title/author fuzzy matching with `SearchRank`, but would only need to look at editions in the user's shelf set, and would not need to worry about deduplication or producing the "default edition", which is most of what makes that code so complicated.

JoeyBryson:
I might have a go at this one.

mouse-reeve:
Please do!

todrobbins:
@JoeyBryson let me know if I can pair with you! I'm not very proficient with Python, but would love to learn more/help.

david-code:
Like I said in the Matrix chat, I'd like to take a stab at this one.

@todrobbins  if you still want to try pairing, let me know, it might be helpful.

david-code:
> @todrobbins , @JoeyBryson , @david-code I wish I knew enough coding to help on this. It seems like a really useful feature. Were any of you able to make any progress?

Still working on it! I've been a bit busy lately but I've definitely made some progress.

placoderm:
> I've definitely made some progress.

That's awesome! I look forward to seeing this in the future.

david-code:
@mouse-reeve I have a question about this when you have a minute.

I've made a function that searches the users shelves. Now I'm not sure exactly the best way to integrate it.

I assume I can edit the file [search.py](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/search.py) to give the frontend the option to pass a parameter to limit the search to user's shelves. However I assume we would also like to give the option the ability to check some sort of box to only search their shelves. Or maybe just adding the backend functionality would be enough for this PR?

Thanks :)

todrobbins:
@david-code I've been inactive on GitHub, but I'm still interested in working with you. You can DM me on [Twitter](https://twitter.com/todrobbins) or [Mastodon](https://social.coop/@todrobbins)

david-code:
> Somehow I totally missed that you tagged me in this, sorry! I'm not sure off the top of my head if editing search.py is a better approach than writing a separate view. But for the UI, it wouldn't need to be a checkbox, but rather a different search box entirely that is placed on the shelf view.

[...]

PR #3118, opened 2023-11-17
Let a user search books within their shelves

rosschapman:
## Description
Search within your own books! 🔍 👤 📚  
For issue background and details: https://github.com/bookwyrm-social/bookwyrm/issues/2079

## Goals
- Users should be able to search their own shelves, including _All Books_
- Users should expect that this search control behaves similar to the global book search
	- They can search by author or title -- which will fuzzy match with a `min_confidence` of 0-- or by exact match (against identifiers, see https://github.com/bookwyrm-social/bookwyrm/pull/3118#issuecomment-1815881202)
	- They will be presented with a list of books that match the query or the existing _empty message_

## What's changed? 
1. **Adds shelf search query handling to Shelf view**
In preparation for this work, the assumption was that we might reuse the /search route and update the  Search view with the new code. But it quickly became obvious that the Shelf view was a better place to add search behavior for the Your Books pages. 
1. **Extends `search()`**
To leverage existing system search behavior to search _within_ a user's shelves, the [`search()`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/book_search.py) function parameters have been extended with a kwarg for `books` of type `QuerySet`. Now book search can be invoked on a preloaded query or not.
1. **Uses *filtering* components for shelf book search***
To differentiate from the global, federated book search and be consistent with the filtering controls that exist for other list interfaces, I've reused the `filter_panels` components

## Tests

View cases are being worked out over at: https://github.com/rosschapman/bookwyrm/pull/1

## Screenshots

image

image

image

---
###  PRD and implementation notes (for posterity)
```
- 🙋 Questions: 
	- Do we want to allow users to search across
		1. All their books
		4. All their books *within* a shelf identifier (eg: just *Read* books)
		5. Both
		- Implementing # 1 is probably a good MVP
- Presumably we don't have to touch the core book search `search` method ([`bookwyrm/book_search.py`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/book_search.py#L40))
	- 🙋 Questions:
		- What is a `search identifier` and why do we search on that first?
		```python
		# first, try searching unique identifiers
	    # unique identifiers never have spaces, title/author usually do
    	if not " " in query:
        	results = search_identifiers(query, *filters, return_first=return_first)
		```

### Speculative task list (large, can be broken up)
- [ ] [BE] Create receiver (view class method) for new user books search request
	- **Option 1:** Modify the existing Search View [`views/search#book_search`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/search.py) with additional logic to restrict the search query to the current user's books
	- **Option 2:** Create a separate view method - named something like `def book_search_user` or `def book_search_own` etc... -- which might reuse some of the existing view logic but will give 
	- My preference is Option # 2 to start. We can always consolidate after if collaborators determine the branching logic doesn't create more headaches than it's worth. 
		- For example, we don't need the `remote_search` capability. Perhaps we forego ISBN check?
- [ ] [FE] Add search control to User's Shelf page ([`bookwyrm/templates/shelf/shelf.html`](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/shelf/shelf.html))
	- **Option 1:** Duplicate the search control code from the chrome header ([`bookwyrm/templates/layout.html`]([bookwyrm/templates/layout.html](https://github.com/bookwyrm-social/bookwyrm/blob/main/)))
	- **Option 2:** Create a reusable Search control component
		- My preference is option # 1. We can always consolidate after if collaborators determine the branching logic doesn't create more headaches than it's worth>)
```


hughrun:
Great stuff @rosschapman awesome to have someone working on this.

> What is a search identifier and why do we search on that first?

The [`search_identifiers` function](https://github.com/bookwyrm-social/bookwyrm/blob/06568aab88b3b099c42c8b7c9d426f523db4ffdb/bookwyrm/book_search.py#L100) searches for anything that might be an ISBN or a `deduplication_field` on an `Edition`. 

So what is a `deduplication_field`? If you check `models/book.py` you'll see that pretty much everything in `BookDataModel`, plus a few other fields in models that inherit it, are `deduplication_field`s. These are standard identifiers like ISBNs, `librarything_key`, `wikidata` identifiers, and so on. What they have in common is that they generally don't duplicate - i.e. a `librarything_key` will only ever refer to one particular edition of a particular book. So if the term someone searches by is a `librarything_key` for example, if you get a match on that you pretty much know that's the book they meant. This is not the case for an author or a title, where many books can have the same title, and multiple authors can have the same name. So we look for a match on a `deduplication_field` first and if we find one we just return that because it's very likely to be the match, but if we don't find one, we keep searching knowing we'll return stuff that may or may not be what they were looking for.

Hopefully that makes sense.

mouse-reeve:
(I just went in and added spaces in places where you mentioned an option number and GitHub thought it was a reference to an issue, it had me baffled for a second 😂)

rosschapman:
(@mouse-reeve 🙈 😅 githuh markdown fail)

rosschapman:
@hughrun Thank you, that's the context I needed to understand that part. 💪

bookwyrm/templates/shelf/shelf.html
```diff
@@ -124,6 +124,10 @@ <h2 class="title is-3">
                 {% endif %}
                 {% endwith %}
             </h2>
+            {% if books|length > 0 %}
```
hughrun, line 127: If the user filters a shelf with keywords that find nothing, this will remove the filter box, forcing them to refresh the page to see the unfiltered results.

You could fix this by doing something like this:

1. change the line here to `{% if books|length > 0 or is_filtered %}`
2. add a hidden input to the `search_filter_field.html` file: `<input hidden name="is_filtered" value="true">`
3. Add a value to the shelf View data: `"is_filtered": is_filtered = request.GET.get("is_filtered")`

bookwyrm/templates/shelf/search_filter_field.html
```diff
@@ -0,0 +1,9 @@
+{% extends 'snippets/filters_panel/filter_field.html' %}
+{% load i18n %}
+
+{% block filter %}
+  <div class="control">
+    <label class="label" for="search_query">{% trans 'Search by keyword' %}</label>
```
hughrun, line 6: "search by keyword" is a bit confusing when this is a "filter" box. Technically we're doing both, but I wonder if it would be clearer to make this "filter by keyword" or "enter keywords to filter your books" or something similar.

hughrun (changes requested):
This is really nice!

Works well, just a small change needed, and a question about terminology, as noted.

bookwyrm/templates/shelf/search_filter_field.html
```diff
@@ -0,0 +1,9 @@
...
+    <input aria-label="Search by keyword" id="my-books-search" class="input" type="text" name="shelves_q" placeholder="Enter text here" value="{{ shelves_search_query|default:'' }}" spellcheck="false" />
```
mouse-reeve, line 7: You'll want to wrap that placeholder text in `{% trans 'Enter text here' %}` here as well

rosschapman:
@hughrun I tried a different approach of showing a hint when the results are empty. See: https://github.com/bookwyrm-social/bookwyrm/pull/3118/commits/4a4046a704d9d2abfa72b51f1da33e6f6d0c908a. (Also renamed things from "search" to "filter" so it's more obvious to developers looking at this code that we are _filtering_ -- as in, search _within_ my books -- rather than _searching_ -- search _for_ a book.

Here's a screenshot of new hint: 
image

hughrun:
> @hughrun I tried a different approach of showing a hint when the results are empty. 

This is a much better idea than my suggestion 😄 Looks great! 

My only suggestion at this point is that in some ways you now have the opposite problem. Previously there was no way to clear the filter if the results were empty. Now it tells the user two opposing things: that we couldn't find any books using their filter, _and_ that the filter has been applied - but then we're showing them an unfiltered list (i.e. the filter has NOT been applied, which is deliberate).

I can see that this may confuse some users. Is there a way to do this where the filters/search term isn't passed back through to the view if the search set is empty, so we can say "we couldn't find anything" and automatically clear the filter?

rosschapman:
@hughrun for some reason it made sense in my head to still show _all books_ without a match. But yeah, with only the placement of a small hint above it seems confusing. What about my latest commit which updates the template to show the not found message in the same place as the books list, while also keeps the Filters panel and _Clear Filters option in view_: 

image

rosschapman:
Btw, it seems the template logic for setting the panel to "open" is a little wonky since it depends a `filters_applied` context variable being set. But if I set this variable it actually doubles up the `Filters Applied` token on the page. I was thinking about keeping the panel open in the case of no matches found.

hughrun:
> I was thinking about keeping the panel open in the case of no matches found.

That would be nice, but if it's too hard I don't think it matters too much. Your latest screenshot looks good, let me know when you're happy to call it "ready" and I'll take a proper final look later.

rosschapman:
> That would be nice, but if it's too hard I don't think it matters too much. 

`filters_applied` is only used for the `Feed` view but I wonder if it was a bit of a hack to hide the "Clear Filters" button. Anywho, I don't think it would be that hard to suss out and fix the logic to work more generically for different kinds of filter fields. But out of scope of this PR 😅

bookwyrm/views/shelf/shelf.py
```diff
@@ -103,6 +112,7 @@ def get(self, request, username, shelf_identifier=None):
             "page_range": paginated.get_elided_page_range(
                 page.number, on_each_side=2, on_ends=1
             ),
+            "shelves_filter_query": shelves_filter_query,
```
hughrun, line 115: Sorry this is really pedantic but the "Filters" label looks kind of goofy to me because it's a level 5 `title` in the filters_panel template. Can we add `"size": "small",` here to make it a bit smaller?

hughrun (changes requested):
Looks great other than a teeny change as per my comment.

bookwyrm/views/shelf/shelf.py
rosschapman: Cool with me. Although that is a very ambiguously named value to pass in from a view model 😅 . Took me a minute to figure out how "small" was used.

hughrun: yeah i think it's generally passed within the templates but I couldn't work out how you would do it when `include`ing - you effectively want to pass it up to the parent.

bookwyrm/templates/shelf/shelf.html
```diff
@@ -209,7 +210,12 @@ <h2 class="title is-3">
             </tbody>
         </table>
         {% else %}
-        <p><em>{% trans "This shelf is empty." %}</em></p>
+        <p><em>{% if shelves_filter_query %}
+                    {% trans "We couldn't find any books that matched" %} "{{ shelves_filter_query }}"
```
mouse-reeve, line 214: This will need to be a blocktrans element with the query included, so that languages with different word order can translate it correctly. It should look something like:
```
{% blocktrans trimmed %}
We couldn't find any books that matched {{ shelves_filter_query }}
{% endblocktrans %}
```

rosschapman: Done!

mouse-reeve (approved):
Amazing! Thank you, looks good to me

todrobbins:
Thank you for implementing this @rosschapman!

PR #3124, opened 2023-11-19
Allow removing followers and fix follow rejections

### LINKED ISSUE #2635: Allow for soft-blocking followers
(unless I've missed it) there is currently no ability to "unapprove" (remove from followers list) followers once they are following you. I'd like the option to silently remove other users from my followers, rather than having to block them. This will help with cleaning up defunct accounts, but also creates a non-confrontational and less strict way of removing followers as compared to blocking.

At `domain.tld/user/{username}/followers` instead of the button being a follow/unfollow button, make it "remove follow" (with a confirmation screen before confirming).

Alternatively, the dropdown when viewing a user could have an option to "remove from followers" or similar wording.

**Additional context**
Whilst Mastodon should not be the measure of all things, this is an option available on Mastodon.

hughrun:
* adds the ability to remove a user from your followers list
* fixes `verbs.Reject` to process "reject" activities for previously accepted follows in both directions

To explain the second point - previously when e.g. a Mastodon user deleted a follow by a BookWyrm user, BookWyrm errored out when received the `reject` activity so the Mastodon user was still listed in "following". This PR fixes this so that users are removed from the "followers" list of the user deleting the follow, and also deleted from the "following" list of the target user.

fixes #2635


hughrun:
@bookwyrm-social/code-review this can be checked with a Mastodon account, but I probably should add one or two tests?

bookwyrm/templates/snippets/follow_button.html
```diff
@@ -12,6 +12,7 @@
 
 <div class="field{% if not minimal %} has-addons{% else %} mb-0{% endif %}">
     <div class="control">
+        {% if not followers_page %}
         <form action="{% url 'follow' %}" method="POST" class="interaction follow_{{ user.id }} {% if relationship.is_following or relationship.is_follow_pending %}is-hidden{%endif %}" data-id="follow_{{ user.id }}">
```
dato, line 16: Implemented like this, it is no longer possible to follow anybody from the followers page.

Since I think following/unfollowing are going to be more likely than removing follows... about keeping the current Follow/Unfollow button here, and have "Remove follow" move to the menu? (With an `is-danger` class, probably below _Block_.)

P.S.: I tried to create a commit that moves it but, unlike the `/block` endpoint, `/remove-follow` does not take any parameters, and I wasn't sure how to proceed.

dato (commented):
Hi! I tested this on my instance. Found one tiny error, below, and then a comment about UX.

bookwyrm/activitypub/verbs.py
```diff
@@ -171,8 +171,36 @@ class Reject(Verb):
...
+                obj = self.object.to_model(
+                    model=model,
+                    save=False,
+                    allow_create=False,
+                    allow_external_connections=allow_external_connections,
+                )
+        else:
+            # it's something else
+            obj = self.object.to_model(
+                model=model,
```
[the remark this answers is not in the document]
hughrun: This is neat, I like it! Though I assume we need to wrap it in a list here:
```python
for model_name in ["UserFollowRequest", "UserFollows", None]:
```

Regarding the order, it depends what we think is more likely to happen most: rejecting a follow request, or deleting an already-approved follow. I would guess there will be a larger number of rejected requests, so I put that first.

bookwyrm/templates/snippets/follow_button.html
hughrun: Oh yeah, somehow that hadn't occurred to me, oops!

> Since (I assume) following/unfollowing are going to be more common than removing follows... about keeping the current Follow/Unfollow button here, and have "Remove follow" move to the menu? (With an is-danger class, probably below Block.)

This is a good suggestion, I'll adjust.

bookwyrm/templates/snippets/user_options.html
```diff
@@ -20,4 +20,9 @@
 <li role="menuitem">
     {% include 'snippets/block_button.html' with user=user class="is-fullwidth" blocks=False %}
 </li>
+{% if followers_page %}
+<li role="menuitem">
+    {% include 'snippets/remove_follower_button.html' with user=user class="is-fullwidth" blocks=False %}
```
dato, line 25:
```suggestion
    {% include 'snippets/remove_follower_button.html' with user=user class="is-fullwidth" %}
```

dato (approved):
> @dato better?

Works beautifully! Tested against Mastodon, and BookWyrm as well.

PR #3128, opened 2023-11-20
Give admins option to test if a theme loads correctly

mouse-reeve:
If a theme is uploaded incorrectly or has errors in it, users can still select the theme but it will cause a 500 error on every page, making the app unusable and also making it impossible for them to switch to a functional theme.

A better fix would be to fail gracefully, but in lieu of that, this will at least let admins confirm if a theme is broken safely.

---

Screenshot 2023-11-20 at 9 55 50 AM

Screenshot 2023-11-20 at 9 55 58 AM


bookwyrm/tests/views/admin/test_themes.py
```diff
@@ -86,3 +86,25 @@ def test_themes_post_forbidden(self):
...
+    def test_test_theme_broken(self):
+        """Testing test for testing when it's a bad theme"""
+        theme = models.Theme.objects.create(name="bad theme", path="dsf/sdf/sdf.sdf")
+        self.assertIsNone(theme.loads)
+        request = self.factory.post("")
+        request.user = self.local_user
+
+        views.test_theme(request, theme.id)
+        theme.refresh_from_db()
+        self.assertFalse(theme.loads)
```
dato, line 110:
```suggestion
        self.assertIs(False, theme.loads)
```

(otherwise _loads_ could still be None, and pass)

PR #3134, opened 2023-11-25
Support trigger migrations

dato:
- 0299f2e2352773c53155ae815cd0a525d6408fe3 Add functional tests for search_vector triggers
- e4d688665c7d54912b73752a270c4d4ead2a63b0 Remove index for `author.search_vector`, which is never used
- 44ef928c3ceb8ce9b4426113551baa53563630f7 Alter object row IDs to force test failure in original code
- 416a6caf2d4124b884fd1a04841908e572d92c61 Define `author_search_vector_trigger` via Author.Meta.triggers
- bcb3a343d4514fd9bc2674a918755aaa832c8886 Fix JOIN in `author_search_vector_trigger`, add missing WHERE clause
- 8df408e07eccefb934373f5c69b5068c1bee5046 Define `search_vector_trigger` via Book.Meta.triggers
- 9bcb5b80ea9dd4f11cae26ad1fc752a7cb71dcde Further simplify bookwyrm_author trigger
- bbfbd1e97ab201cd87c6414127c603f615923ce5 Add tests for trigger code (i.e. how search_vector is computed)
- b5805accacb497be62cb4e7bb54a52258a1d12e3 Minor improvements to bookwyrm_book trigger code
- d6eb390ceed8e6f63926aa8b54a4ef9314b93c75 Add test that forces `book_authors_search_vector_trigger` to execute

-----

This PR uses [django-pgtrigger](https://django-pgtrigger.readthedocs.io/en/stable/) to define triggers as part of model metadata. With this, Django's migration mechanism can be used to keep track of changes.

Done:

- [x] tests for existing triggers (functional and structural)
- [x] migrate trigger on bookwyrm_book table
- [x] migrate trigger on bookwyrm_author table

In a later PR:

- migrate trigger on bookwyrm_book_authors table

     (pgtriggers docs say the latter probably requires giving `Book.authors` its own "through" table, which we need anyway for things like #1120. But that's a more pervasive change, and I don't believe it belongs in this PR.)

    Also, about this trigger, see last commit.)


bookwyrm/tests/test_book_search.py
```diff
@@ -134,3 +135,188 @@ def parse_isbn_search_data(self, data):
         # there's really not much to test here, it's just a dataclass
         self.assertEqual(result.confidence, 1)
         self.assertEqual(result.title, "Title")
+
+
+class SearchVectorTest(TestCase):
+    """check search_vector is computed correctly"""
```
dato, line 141: Uhm I'm not sure how, but I managed to completely miss tests/test_progress.py.

I think everything tested there is tested here, though triggers are tested functionally, rather than structurally (i.e. for trigger behavior we look at search results, not at search_vector). I can bring those missing tests from there to here, and have it all in this file.

In any case I apologize for the confusion; it wasn't intentional. :)

dato:
fwiw, I think it might be a good idea to delay merging this until after having tagged the next release. I’m completely fine with that if you agree it makes sense.

mouse-reeve:
I agree! I think I'm going to make a release tomorrow morning with just what's already merged unless you think there's anything that should get in under the wire

dato:
Only #3096 comes to mind (very optional, just to get it out?; but I didn’t quite understand what it fixes, so I have no technical opinion).

mouse-reeve:
I considered that one and decided to wait because I have zero recollection of why I had it set to the current way, but I know I did it on purpose, and I don't want to have to re-debug it tomorrow. But another day, I will have the willpower to do so 😆

dato:
💯

Thank you for preparing this release.

bookwyrm/migrations/0190_book_search_updates.py
```diff
@@ -0,0 +1,16 @@
...
+class Migration(migrations.Migration):
+    dependencies = [
+        ("bookwyrm", "0188_theme_loads"),
+    ]
+
+    operations = [
+        migrations.RemoveIndex(
+            model_name="author",
+            name="bookwyrm_au_search__b050a8_gin",
+        ),
```
Minnozz, line 15: I added searching for authors in https://github.com/bookwyrm-social/bookwyrm/pull/3274, but it doesn't use `search_vector`. Should it?

bookwyrm/migrations/0190_book_search_updates.py
```diff
@@ -0,0 +1,16 @@
+# Generated by Django 3.2.20 on 2023-11-24 17:11
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("bookwyrm", "0188_theme_loads"),
```
Minnozz, line 8: Could you rebase this PR on the latest `main`?

bookwyrm/tests/test_book_search.py
Minnozz: I think consolidating them here is best

Minnozz (approved):

bookwyrm/models/author.py
```diff
@@ -67,9 +68,28 @@ def get_remote_id(self):
...
+        triggers = [
+            pgtrigger.Trigger(
+                name="reset_search_vector_on_author_edit",
+                when=pgtrigger.After,
+                operation=pgtrigger.UpdateOf("name"),
+                func=format_trigger(
+                    """WITH updated_books AS (
+                         SELECT book_id
+                         FROM bookwyrm_book_authors
+                         WHERE author_id = new.id
```
Minnozz, line 83: Now that this bug is fixed, I think the migration should re-calculate the search vector for all books because it may have missed some author name changes.

bookwyrm/migrations/0190_book_search_updates.py
Minnozz: (I can also do this if you'd like)

dato: Thanks, I performed a merge (just a preference), and produced 0196_merge_pr3134_into_main.py to take care of leaf migrations.

bookwyrm/tests/test_book_search.py
dato: Will do.

bookwyrm/models/author.py
dato: First of all, thanks for the review! I addressed most outstanding items (there's still the question about #3274 pending).

Secondly: if I understand correctly, your suggestion here is to include in the migration something like: `UPDATE bookwyrm_book SET search_vector = ''`?

Minnozz: Yes, exactly

bookwyrm/migrations/0190_book_search_updates.py
Minnozz: I will change author search to use `search_vector` instead of `TrigramSimilarity`; would you remove the `RemoveIndex` here?

Minnozz: Interestingly, the `Author` model inherits the `search_vector` field from `BookDataModel`, but there is no trigger that actually fills the `bookwyrm_author.search_vector` column with (for instance) the author name, so this column is empty for every record.

Feel free to leave in this `RemoveIndex` and I'll re-add it in a separate PR that also adds the actual trigger that fills the column and starts using it for author search.

Or if you feel like it, you could add that trigger in this PR.

dato: > Feel free to leave in this RemoveIndex and I'll re-add it in a separate PR that also adds the actual trigger that fills the column and starts using it for author search.

Let's do this; thank you!

bookwyrm/models/author.py
dato: I'll admit I've failed to add it so far because it's a destructive operation that I wanted to double-check. I'll take a stab after reviewing Hugh's pending PR, but help welcome in the meantime, if you feel so inclined! (no obligation either way)

Minnozz: Added to the migration and and tested

PR #3135, opened 2023-11-25
csv import and export fixes

### LINKED ISSUE #2411: Generic CSV import
In addition to the import from other platforms, I would like a generic CSV import option. Yes, you can use a none exported CSV file of your own, but this would make it clearer to users that they can import their own CSV file (e.g. from a database they may be keeping at home). In addition, documentation would advise on the required structure of this CSV file.

### LINKED ISSUE #2666: Create an option to import BookWyrm CSVs into a new BookWyrm account

### LINKED ISSUE #2846: Exported CSV file does not seem to contain information on what shelf my books are on
**Describe the bug**
I exported the CSV-file from my account on bookwyrm.social, and it does not seem to contain information on what books I've read and what I want to read and what I am currently reading (Shelves)

**To Reproduce**
1. Go to settings
2. click on  CSV export
3. Click on "Download file"
4. No apparent status-column in the exported csv-file

**Expected behavior**
A column indicating what shelf a book is on

**Instance**
Exported from bookwyrm.social

    - OS: MacOS
    - Browser Firefox
    - Version 113b

### LINKED ISSUE #3004: Importing books from another Bookwyrm instance fails
**Describe the bug**
I'm running into a pair of issues importing a CSV exported from Bookwyrm: rows with a review or rating fail, and the books that get successfully imported don't seem to get added to my books.

**To Reproduce**
1. Export from bookwyrm.social
2. Try importing to outside.ofa.dog
3. Any entry with a review or star rating will fail, and the import will hang at a percentage that corresponds to the share of rows in the export file that had neither a review nor a rating
4. Remove all of those entries from the bookwyrm export CSV and retry
5. The import succeeds, but none of the books actually get added to my shelves or profile.

Other things I've tried:
* Breaking the export down into small chunks: even down to 25 rows in a file, I get the same behaviour pattern.
* Unchecking the "import reviews" box in the import: it gets through the file, but books still don't get added to my shelves or profile.
* Each of the different CSV format options in the dropdown: seems to get the same behaviour, though I haven't tested all permutations thoroughly.
* Retrying failed books: none of them work on a retry.

**Expected behavior**
The books import with reviews, and get shelved the same as in the source instance.

**Instance**
outside.ofa.dog

**Additional context**
I am aware that https://github.com/bookwyrm-social/bookwyrm/pull/2980 may make this issue redundant by providing an alternative way to migrate books between instances.  I can't tell how close that work is to completion, which should probably determine whether this bug is worth working on in parallel or not.

---

    - OS: Mac OS 13.5.2
    - Browser Firefox
    - Version 117.0.1

hughrun:
Adds shelved and published dates for books and their imported reviews. Provides option to create new (custom) shelves when importing books.

fixes #3004
fixes #2411

@bookwyrm-social/code-review I'm looking for some clues as to why my tests fail when run sequentially but pass when run individually.


mouse-reeve:
I'm not sure how to handle the merge conflicts on this one

mouse-reeve:
I wasn't sure how to resolve the merge conflicts on this

hughrun:
yeah I got distracted with other things, I need to come back to this.

hughrun:
There's something wrong with my tests here (`test_create_new_shelf` and `test_handle_imported_book_review`) - they pass when run in isolation but not when running the whole `TestCase`. Any ideas @bookwyrm-social/code-review ?

dato:
Hello @hughrun!

I took a look at the failures, here's the root cause (fwiw took a while to figure out!):

The `__init__` method of BookwyrmBooksImporter is used to extend the `row_mappings_guesses` attribute:

https://github.com/bookwyrm-social/bookwyrm/blob/06d6360082b6164526dee75d8fb03bc5674c85ab/bookwyrm/importers/bookwyrm_import.py#L38-L40

But `row_mappings_guesses` is a class level attribute, not instance level; so what happens is that the pair `("shelf_name", ["shelf_name"])` is appended _once per created object_.

The tests fail when such `__init__` runs more than once (more than one importer created), because the duplication results in `shelf_name=None` in the import item.

To fix the issue, you can remove the `__init__` method in favor of, for example:

```python
class BookwyrmBooksImporter(Importer):
    # ...
    row_mappings_guesses = Importer.row_mappings_guesses + [
        ("shelf_name", ["shelf_name"]),
    ]
```

to avoid the bogus duplication.

hughrun:
Ok a couple of things here:

1. I amended the test for retry jobs in `bookwyrm/tests/importers/test_importer.py` to order the jobs, because they were running in the reverse order when running multiple tests (but not a single test). My assumption is that this happened because pylint runs tests in parallel, and that ultimately it doesn't matter which order they run in, but it may have been indicating an actual problem - so some eyes on that would be helpful.
2. All tests pass locally, but there seems to be some kind of issue with the GitHub runners. Not sure what I can do about that.

@dato thanks for your advice.

mouse-reeve:
I think this just needs a merge migration in order to be ready to go

hughrun:
Ok there seems to be a conflict in the tests (causing some imports to finish out of order) because I'm using the same books as in `generic.csv`. Presumably this is because tests run in parallel? I'll adjust the test data to fix this.

hughrun:
@bookwyrm-social/code-review I think this is ready.

mouse-reeve:
Worked neatly for me! Thank you so much

eldang:
Thank you!  I'm excited for when this makes out to instances in the wild.

mouse-reeve:
It seems like since merging this tests have been failing on new pull requests:
```
FAILED bookwyrm/tests/importers/test_importer.py::GenericImporter::test_create_job - AssertionError: 1 != 0
```

hughrun:
ugggghhhh that's the error I was getting originally, I thought I'd fixed it by not using the same books to test importing.

@mouse-reeve from my own troubleshooting on this it seems that the tests sometimes return books in a slightly different order, which is what leads to this assertion error.

All tests pass individually but not together, so there's something I'm not understanding that leads to this difference: perhaps tests running concurrently leading to celery tasks returning in different orders?

mouse-reeve:
Ah how frustrating! I tried re-writing that test to ignore the order of the items: https://github.com/bookwyrm-social/bookwyrm/pull/3419. It's hacky, but if it works...?

PR #3144, opened 2023-12-04
Update page formatter on ordered collection

villasv:
Hi 👋 

CleanShot 2023-12-03 at 20 04 05@2x

I noticed this prev URL is a bit suspicious. Is the %d in there intentional? If not, here it goes a CR removing it!


mouse-reeve:
Suspicious indeed! This was breaking pagination; thank you for fixing :)

PR #3146, opened 2023-12-05
Faster tests with setUpTestData

dato:
Django's [TestCase.setUpTestData()](https://django.readthedocs.io/en/3.2.x/topics/testing/tools.html#django.test.TestCase.setUpTestData) allows us to create database fixtures once per test class, instead of once per test method. This speeds up tests considerably.¹

As per the documentation, database fixtures are created once, and roll'd-back to after each test. (Additionally, it _is_ safe to store the resulting models as class variables, since the Python objects are deepcopy()'d from the original before each test.)

I think this could be a nice addition to the BookWyrm codebase.

(¹) On my old laptop, running the whole test suite goes down from ~20min to ~9min (on my newer desktop, it's down to 2min, which is a joy). It seems to go from ~8min to ~4min in the Github runners.

### DIFF
--- a/bookwyrm/tests/activitypub/test_author.py
+++ b/bookwyrm/tests/activitypub/test_author.py
@@ -6,7 +6,8 @@ from bookwyrm import models
 class Author(TestCase):
     """serialize author tests"""
 
-    def setUp(self):
+    @classmethod
+    def setUpTestData(self):  # pylint: disable=bad-classmethod-argument
         """initial data"""
         self.book = models.Edition.objects.create(
             title="Example Edition",
--- a/bookwyrm/tests/activitypub/test_base_activity.py
+++ b/bookwyrm/tests/activitypub/test_base_activity.py
@@ -28,8 +28,8 @@ from bookwyrm import models
 class BaseActivity(TestCase):
     """the super class for model-linked activitypub dataclasses"""
 
-    # pylint: disable=invalid-name
-    def setUp(self):
+    @classmethod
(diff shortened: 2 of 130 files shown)


.pylintrc
```diff
@@ -3,7 +3,8 @@ ignore=migrations
 load-plugins=pylint.extensions.no_self_use
 
 [MESSAGES CONTROL]
-disable=E1101,E1135,E1136,R0903,R0901,R0902,W0707,W0511,W0406,R0401,R0801,C3001,import-error
+# XXX remove `bad-classmethod-argument`
+disable=E1101,E1135,E1136,R0903,R0901,R0902,W0707,W0511,W0406,R0401,R0801,C3001,import-error,bad-classmethod-argument
```
dato, line 7: This here is the main reason the PR is marked as draft.

As `setUpTestData()` is a class method, pylint (rightfully) complains that the argument to the method should be `cls`, and not `self`. This involves substantially many more changes (for little benefit?).

I tried to find a way to disable this warning only in the bookwyrm/tests subdirectory, but it doesn't seem to be possible. I offer, for the time being, adding override directives to each test file, rather than leaving the global disable here.

mouse-reeve: I think overrides in the file is fine, that's what I had been doing with invalid name. The CI I believe is not checking the test files, so it won't block PRs. This is a great change and having quicker test will be a big win :)

dato: Thank you for taking a look. I applied your suggestion!

Issue #3173, opened 2023-12-31
`bw-dev build` fails 


ccamara:
**Describe the bug**
Running `bw-dev build` fails, throwing the error below

**To Reproduce**

I followed the steps on https://docs.joinbookwyrm.com/install-dev.html

1. Installed docker and docker compose
2. Clone my Forked repo
3. Edit `.env` as described
4. Run `cp nginx/development nginx/default.conf`
5. Run `./bw-dev build` and I get this error:

```
[...]

=> CANCELED [flower 6/6] RUN pip install -r requirements.txt --no-cache-dir                            45.8s

=> ERROR [dev-tools 5/6] RUN apt-get update            &&     apt-get install -y nodejs &&     pip in  44.2s

[...]

failed to solve: process "/bin/sh -c apt-get update            &&     apt-get install -y nodejs &&     pip install -r requirements.txt &&     npm install ." did not complete successfully: exit code: 127

```

It can reproduce the same in two different machines, one running KDE Neon, and another one running Mac OS

**Expected behavior**
It should correctly install all dependencies.

---

**Desktop (please complete the following information):**
    - OS: KDE Neon / Mac OS

bernaborrero:
It looks like it can't find the `npm` command specified on dev-tools/Dockerfile.
I tried to install explicitly `npm` on the Dockerfile, just to try, and it solved the issue, but I don't know what other implications this could have.

Apparently, installing `nodejs` should include `npm`, except in old distros, but that shouldn't happen when installing using the setup script. The setup script was deprecated, and thus replaced by the apt source in `sources.list.d/` in [this commit](https://github.com/bookwyrm-social/bookwyrm/commit/1937177e1a5df853d2cea1ba1436b4baa2290e14) by @dato  as a `.sources` file. I've never used the Deb822 format, but it should do the same as before, I understand.

Lastly, when executing the command, we can see that `apt` is suggesting the installation of `npm`, but it's not installing it, so perhaps https://deb.nodesource.com/node_18.x no longer includes it?

```
> [dev-tools 5/6] RUN apt-get update            &&     apt-get install -y nodejs &&     pip install -r requirements.txt &&     npm install .:
0.725 Get:1 http://deb.debian.org/debian bookworm InRelease [151 kB]
0.785 Get:2 http://deb.debian.org/debian bookworm-updates InRelease [52.1 kB]
0.787 Get:3 http://deb.debian.org/debian-security bookworm-security InRelease [48.0 kB]
[...]

PR #3175, opened 2024-01-01
dev-tools: ensure we install Node from upstream

dato:
Fixes: #3173 ("`bw-dev build` fails")


dato:
The test failures happen because today's date (1/1/24) triggers a "yesterday" condition for some tests. I think we can safely ignore them (or re-run after tomorrow).

However the failures exposed a bug in that code, that I address in #3177.

PR #3177, opened 2024-01-01
Adjustments to naturalday_partial

dato:
In particular:

- naturalize date and datetime objects
- never  naturalize partial dates with no day


mouse-reeve:
Ah this fixes #3179 doesn't it?

dato:
To all practical effects, yes. (The tests fail only on 2023, wouldn’t fail in 2024 even without these changes.)

I’ll update the description.

PR #3183, opened 2024-01-02
Adds management command to clear all deleted user data

mouse-reeve:
This replaces the migration step removed in https://github.com/bookwyrm-social/bookwyrm/pull/3182

### DIFF
--- a/bookwyrm/management/commands/erase_deleted_user_data.py
+++ b/bookwyrm/management/commands/erase_deleted_user_data.py
@@ -38,3 +38,6 @@ class Command(BaseCommand):
         for user_id in deleted_users.values_list("id", flat=True):
             erase_user_data.delay(user_id)
             self.stdout.write(".", ending="")
+
+        self.stdout.write("")
+        self.stdout.write("Tasks created successfully")
--- a/bookwyrm/management/commands/erase_deleted_user_data.py
+++ b/bookwyrm/management/commands/erase_deleted_user_data.py
@@ -1,0 +1,40 @@
+""" Erase any data stored about deleted users """
+import sys
+from django.core.management.base import BaseCommand, CommandError
+from bookwyrm import models
+from bookwyrm.models.user import erase_user_data
+
+# pylint: disable=missing-function-docstring
+class Command(BaseCommand):
+    """command-line options"""
+
+    help = "Remove Two Factor Authorisation from user"
+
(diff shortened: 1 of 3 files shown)


mouse-reeve:
@bookwyrm-social/code-review

hughrun:
checking this now.

bw-dev
```diff
@@ -246,6 +246,9 @@ case "$CMD" in
     remove_remote_user_preview_images)
         runweb python manage.py remove_remote_user_preview_images
         ;;
+    erase_deleted_user_data)
```
hughrun, line 249: This has a function but no command: `./bw-dev erase_deleted_data` shows an error.

bookwyrm/management/commands/erase_deleted_user_data.py
```diff
@@ -0,0 +1,40 @@
...
+        if options["dryrun"]:
+            self.stdout.write("\n".join(u.username for u in deleted_users[:5]))
+            if deleted_users.count() > 5:
+                self.stdout.write("... and more")
+            sys.exit()
+
+        self.stdout.write("Erasing user data:")
+        for user_id in deleted_users.values_list("id", flat=True):
+            erase_user_data.delay(user_id)
+            self.stdout.write(".", ending="")
```
hughrun, line 40: This produces a kind of weird stdout message:
```
Found 1 deleted users
Erasing user data:
.%
```
Not sure where the '%' comes from, and it might be good to output a message when the loop is finished indicating that it's done successfully.

hughrun (changes requested):
Seems to work 😄 

I created a user with a fresh DB on the latest `main`, then pulled in this PR - so I can't guarantee no problems with a DB from previous versions.

Couple of comments as shown, nothing too complicated to resolve.

bw-dev
mouse-reeve: Huh it seems to work for me, what error are you getting?

bookwyrm/management/commands/erase_deleted_user_data.py
mouse-reeve: Ah the % is your terminal I think, because the command ends with no line break. Fixed!

bw-dev
hughrun: scrap that, I was a little hasty with my testing and turns out I entered the wrong command 🙃

hughrun (approved):

PR #3185, opened 2024-01-02
Check version number asynchronously

mouse-reeve:
Admins will now need to schedule daily checks for updated releases. The UI for that looks like:
Screenshot 2024-01-02 at 1 23 13 PM

This does a few other cleanup tasks to remove unused version code and create a view for seeing a table of scheduled tasks:

Screenshot 2024-01-02 at 1 19 34 PM

Works on #2717


mouse-reeve:
@bookwyrm-social/code-review

bookwyrm/templates/settings/schedules.html
```diff
@@ -0,0 +1,116 @@
...
+                    {{ task.last_run_at }}
+                </td>
+                <td>
+                    {% firstof task.interval task.crontab "None" %}
+                </td>
+                <td>
+                    {{ task.interval.id }}
+                </td>
+                <td>
+                    {{ task.enabled|yesno }}
```
hughrun, line 64: Currently there does not seem to be a way to _disable_ `celery.backend_cleanup` or `check-for-updates`.

Now, one might argue that they shouldn't be disabled, but it seems a bit odd that you have to manually turn on `check-for-updates` but then you can't turn it off.

bookwyrm/templates/settings/schedules.html
```diff
@@ -0,0 +1,116 @@
...
+
+{% block panel %}
+
+<div class="block content">
+    <h3>{% trans "Tasks" %}</h3>
+    <div class="table-container">
+        <table class="table is-striped is-fullwidth">
+            <tr>
+                <th>
+                    {% trans "Name" %}
```
hughrun, line 22: A nice-to-have (possibly in a subsequent PR) would be to link the task name to some sort of explanation of the purpose of the task and/or what it does.

hughrun (commented):
This looks really nice and works well (I tested by adjusting the frequency to 1 minute, and got the alert box in the dashboard). I think it's a real improvement to show all the scheduled tasks on a dashboard.

A couple of comments, but they're not show-stoppers, just suggestions really.

bookwyrm/templates/settings/schedules.html
mouse-reeve: yeah, I agree, I think it should be a subsequent PR because celery.backend_cleanup might be the kind of the thing that it's Not Good to turn off

bookwyrm/templates/settings/schedules.html
mouse-reeve: Periodic tasks have a `description` field which we're currently not using, but presumably would be perfect for this

bookwyrm/templates/settings/schedules.html
mouse-reeve: actually but now that I saw that, if someone has it running every minute, they're probably going to be displeased not to be able to stop it.

hughrun: I guess we could trigger a big scary banner if it's ever turned off. I don't think this needs to be a blocker for this PR, just wanted to note it for future improvement.

VERSION
```diff
@@ -1 +1 @@
-0.7.1
```
Minnozz, line 1: @mouse-reeve I think the version was accidentally changed in this PR

mouse-reeve: Oh my bad, I was changing it for testing and forgot to revert

mouse-reeve: https://github.com/bookwyrm-social/bookwyrm/pull/3276

PR #3188, opened 2024-01-03
Federate reports

### LINKED ISSUE #962: Reports should create Flag activities
Right now, the reporting and moderation system is internal to the instance. It should also send Flag activities to notify other services of bad behavior and spam.

mouse-reeve:
Reports should be sent (if the user wants) to the home instance for remote users. I'm using mastodon's Flag activity as a template: https://docs.joinmastodon.org/spec/activitypub/#Flag

Fixes #962 

Components:
 - [x] Fix the report model so that `user` means the user doing the reporting, not the reported user
 - [x] Make the report model activitypub-y
   - [x] Produce the easy parts of the activitypub object
   - [x] Get the "objects" part of the Flag activity in place
 - [x] Allow users to opt out of federating reports
 - [x] Broadcast reports
 - [x] Receive broadcasted reports 
 - [ ] Test between bookwyrm instances
 - [ ] Test between bookwyrm and mastodon instances


bookwyrm/models/report.py
```diff
@@ -22,28 +24,61 @@
...
+    def to_activity_dataclass(self, **kwargs):
+        """generate the objects field"""
+        activity = super().to_activity_dataclass(**kwargs)
+        activity.object = self.object()
+        return activity
+
+    def object(self):
+        """Generate a list of reported objects in a format Mastodon will like"""
+        items = [self.reported_user.remote_id]
+        if self.statuses:
```
hughrun, line 64: When I test this `self.statuses` is always `bookwyrm.Status.None` here. So the status is never sent to Mastodon. _However_ `statuses` obviously _isn't_ empty because in the BookWyrm admin interface I can see it. So I have no idea what is going on but it's weird and means statuses aren't passed through to federated servers with the report.

bookwyrm/models/report.py
```diff
@@ -22,28 +24,61 @@
...
+        "User",
+        related_name="reported_user",
+        on_delete=models.PROTECT,
+        null=True,
+        blank=True,
+        activitypub_field="to",
+    )
+    statuses = models.ManyToManyField(
         "Status",
         null=True,
```
hughrun, line 48: No need to set `null` on a `ManyToManyField`.

mouse-reeve:
@hughrun Would you be able to take another look at this? I haven't meaningfully changed anything besides resolving merge conflicts, but I'm puzzled as to why Mastodon wouldn't be able to parse it. This is an Activity from my local:
```
{
  "id": "https://--.ngrok-free.app/settings/reports/4",
  "type": "Flag",
  "actor": "https://--.ngrok-free.app/user/rat2",
  "object": [
    "https://bookwyrm.social/user/mouse",
    "https://bookwyrm.social/user/mouse/review/8092282"
  ],
  "to": "https://bookwyrm.social/user/mouse",
  "links": [],
  "content": "",
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "Hashtag": "as:Hashtag"
    }
  ]
}
```

And it seems to match the expected format according to the [docs](https://docs.joinmastodon.org/spec/activitypub/#Flag):
```
{
  "@context": "https://www.w3.org/ns/activitystreams",
  "id": "https://mastodon.example/ccb4f39a-506a-490e-9a8c-71831c7713a4",
  "type": "Flag",
  "actor": "https://mastodon.example/actor",
  "content": "Please take a look at this user and their posts",
  "object": [
    "https://example.com/users/1",
    "https://example.com/posts/380590",
    "https://example.com/posts/380591"
  ],
  "to": "https://example.com/users/1"
}
```

hughrun:
Hi @mouse-reeve I took another look at this.

There is a problem, but I think I misunderstood what it was earlier.

First of all, reports are definitely sent **TO** Mastodon if you report a user from Mastodon.

However, Mastodon is not sending `Flag` activities according to your example above. Notably and most importantly, _despite what their docs say, Mastodon doesn't set a `to` field in Flag activities_:

```py
{
    "@context": "https://www.w3.org/ns/activitystreams",
    "id": "https://ausglam.space/bd5b93f7-177c-4418-abc1-7b703fe508fc",
    "type": "Flag",
    "actor": "https://ausglam.space/actor",
    "content": "testing 123",
    "object": [
        "https://example.com/user/h",
        "https://example.com/user/h/status/2"
    ]
}
```

In this regard they are kind of compliant with [the actual spec](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-flag), but also noncompliant because they're sending `object` as an array which frankly makes no sense.

The fact that there's a compulsory field missing basically breaks everything: reports are received from Mastodon, but I can't even load the Reports page because none of the external Report objects have user ids.

I do wonder a bit about the sense of sending a report about a user "to" that user's inbox address, but I guess if we have the correct logic, it won't actually hit their feed for them to see it.

As you can see above, Mastodon sends the user as the first item in `object`. But it also _sends the object to the user inbox_. So we have two options I think: 

1. Give `to` a default value of `None` and do something funky I can't work out how to do in terms of mapping `reported_user` to the first value in "object" if "to" is `None`.
2. Grab `request.path` right up top in `activitypub.parse` and run some kind of check to see whether there is a missing "to" field, and swap the request path in if it's missing.

There's probably a third less hacky option but those are two off the top of my head.

As a much less problematic issue, I noticed that reports _from_ external servers have a message that they were not reported _to_ an external server which, uh obviously not so maybe that needs to be adjusted.

mouse-reeve:
I tried reporting from Mastodon and got this error:
```
web-1                          | Internal Server Error: /user/mouse/inbox              
web-1                          | Traceback (most recent call last):                              
web-1                          |   File "/venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner
web-1                          |     response = get_response(request)                 
web-1                          |                ^^^^^^^^^^^^^^^^^^^^^                  
web-1                          |   File "/venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response
web-1                          |     response = wrapped_callback(request, *callback_args, **callback_kwargs)
web-1                          |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
web-1                          |   File "/venv/lib/python3.11/site-packages/django/views/generic/base.py", line 105, in view
web-1                          |     return self.dispatch(request, *args, **kwargs)              
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  
web-1                          |   File "/venv/lib/python3.11/site-packages/django/utils/decorators.py", line 48, in _wrapper
web-1                          |     return bound_method(*args, **kwargs)              
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                        
web-1                          |   File "/venv/lib/python3.11/site-packages/django/views/decorators/csrf.py", line 65, in _view_wrapper
web-1                          |     return view_func(request, *args, **kwargs)       
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^        
web-1                          |   File "/venv/lib/python3.11/site-packages/django/utils/decorators.py", line 48, in _wrapper
web-1                          |     return bound_method(*args, **kwargs)            
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^              
web-1                          |   File "/app/bookwyrm/decorators.py", line 13, in wrap          
web-1                          |     return function(request, *args, **kwargs)   
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^       
web-1                          |   File "/venv/lib/python3.11/site-packages/django/views/generic/base.py", line 144, in dispatch
web-1                          |     return handler(request, *args, **kwargs)           
web-1                          |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^       
web-1                          |   File "/app/bookwyrm/views/inbox.py", line 72, in post         
web-1                          |     sometimes_async_activity_task(activity_json)    
web-1                          |   File "/app/bookwyrm/views/inbox.py", line 118, in sometimes_async_activity_task
web-1                          |     activity.action(allow_external_connections=False)           
web-1                          | TypeError: Flag.action() got an unexpected keyword argument 'allow_external_connections'
```

mouse-reeve:
Confirmed that incoming reports are working from Mastodon! Woo!

mouse-reeve:
This works in both directions! But I see that I [added back a line you removed](https://github.com/bookwyrm-social/bookwyrm/pull/3188/changes/5563400fc38506207aa2aebf37439e6724059696#diff-9aae236670b715ef0dd44eeaf20fa1f6205bd0f132898721edb10aac80b52739R285) @hughrun -- is there a good reason you took that out that I'm missing?

hughrun:
I think i just got confused when I was tidying up and then maybe didn't double check. It's crucial that external connections are allowed.

mouse-reeve:
Gotcha! It's not used in the function so it may have tripped a linter up.

hughrun:
@mouse-reeve I remember now. I was thinking it's already `True` in `Verb.action` so there's no need to set it. But I had forgotten that when overriding the method the arguments still need to be specified because they'll be used by external calls. So without it, Inbox calls `Flag.action` with an arg that isn't defined. Since you added it back, that should now work - the failing test seems to be a missing merge migration.

mouse-reeve:
Unreal level of satisfying to finally merge a two year old PR. Thank you so much for getting it over the line

PR #3207, opened 2024-01-09
Hide "year in the books" for newly registered users

### LINKED ISSUE #3187: "year in the books" should not appear for newly registered users
**Describe the bug**

**To Reproduce**
1. Register a new account during the time "{year} in the books" appears
2. Get a messsage encouraging you to see all the books you read in the year, which will be none.

**Expected behavior**
This should not be displayed for users who have no readthroughs.

**Screenshots**

Screenshot 2024-01-03 at 11 08 20 am

**Instance**
dev site

rsk2:
Fixes #3187


bookwyrm/views/feed.py
```diff
@@ -66,6 +68,7 @@ def get(self, request, tab, filters_applied=False):
                 "path": f"/{tab['key']}",
                 "annual_summary_year": get_annual_summary_year(),
                 "has_tour": True,
+                "has_read_throughs": True if len(readthroughs) else False,
```
hughrun, line 71: You could write this a little more simply as:

```py
"has_read_throughs": len(readthroughs) > 0,
``` 
which will do the same thing as it's implicitly returned. This should fix the pylint error.

bookwyrm/views/feed.py
```diff
@@ -52,6 +52,8 @@ def get(self, request, tab, filters_applied=False):
 
         suggestions = suggested_users.get_suggestions(request.user)
 
+        readthroughs = models.ReadThrough.objects.filter(user=request.user)
```
hughrun, line 55: This works, but as written it means if a new user "starts reading" a book they will suddenly have a `ReadThrough` and "year in the books" will appear. You could do something like below to ensure we are only checking for readthroughs that were "last year" or earlier:

```py
from datetime import date # line 2

cutoff = date(get_annual_summary_year(), 12, 31) # line 55
readthroughs = models.ReadThrough.objects.filter(
    user=request.user, 
    finish_date__lte=cutoff
    )
```

hughrun (changes requested):
This is great! Just a couple of little things to fix and a suggestion to make it a little more specific.

bSolt:
I'd like to help you out and add a test for this. Would that be alright with you? I'll probably try to create a PR that merges into your branch. :)

bookwyrm/views/feed.py
```diff
@@ -66,6 +72,7 @@ def get(self, request, tab, filters_applied=False):
                 "path": f"/{tab['key']}",
                 "annual_summary_year": get_annual_summary_year(),
                 "has_tour": True,
+                "has_read_throughs": len(readthroughs) > 0,
```
bSolt, line 75: Since we are filtering on the date as well, can we name this property `"has_summary_read_throughs"` instead?

hughrun: That does actually make sense and will avoid potential confusion in future.

I've also just realised the value could simply be `len(readthroughs)` because `0` is falsy.

@rsk2 do you mind making those two small changes? Then we'll stop bike-shedding this and I'll pull it in.

rsk2: OK, done!

rsk2:
> I'd like to help you out and add a test for this. Would that be alright with you? I'll probably try to create a PR that merges into your branch. :)

Thanks :)

bSolt:
Need to get better familiarized with the tests and patterns used here myself, actually. I am also new to the project and I had some trouble with how to test this specific feature. Getting this merged to fix the bug sooner is probably a better priority IMO.

bSolt (approved):
Let me know if you would like a primer on the tests anyway, but this is looking fine to me for now. Thanks!

hughrun:
> Need to get better familiarized with the tests and patterns used here myself, actually. I am also new to the project and I had some trouble with how to test this specific feature. Getting this merged to fix the bug sooner is probably a better priority IMO.

I'm planning to add some info to the docs about writing and updating tests, as it can be very confusing for people new to the project and/or Django (as I know from my own experience!) There have been some fairly urgent other things to resolve recently but I'll try to get back to that soon.

hughrun:
Hmm, I think I have led you astray here with `get_annual_summary_year`. I'm looking into it.

hughrun:
Ok @rsk2 sorry about this, I gave you some bad advice.

`get_annual_summary_year` returns _this_ year if the current date is after 15 Dec and before 1 January, or _last year_ if the current date is after 31 Dec but before 16 January. But if it's any _other_ date, it returns `None`. I momentarily forgot this is the point of [this utility function](https://github.com/bookwyrm-social/bookwyrm/blob/fd0b1d90b070ee1653140b9c5f3b79d2d21104b3/bookwyrm/views/annual_summary.py#L160). It was all working because we were looking at this within that period of time. But now time has passed!

So that's why the test is failing: because you can't make a date with a year value of `None`.

However, you can fix this reasonably easily because we only need to know whether the user `has_summary_readthroughs` when `get_annual_summary_year` is `True` (otherwise the annual summary isn't displayed anyway):

```py
cutoff = (
    date(get_annual_summary_year(), 12, 31)
    if get_annual_summary_year()
    else None
)
readthroughs = (
    models.ReadThrough.objects.filter(
        user=request.user, finish_date__lte=cutoff
    )
    if get_annual_summary_year()
    else []
)
```

rsk2:
@hughrun no worries. Thanks for the fix.

hughrun (approved):

Issue #2706, opened 2023-03-07
Use a github action to update locales


mouse-reeve:
Right now, locales have to be updated manually using `./bw-dev update_locales`. It would be better to have this get updated using a scheduled github task (similar to [how the join bookwyrm site is deployed](https://github.com/bookwyrm-social/join-bookwyrm/blob/main/.github/workflows/update-and-deploy.yml)) which runs that script and creates a PR.

jaschaurbach:
I would suggest to use something like this to merge the locales direct into the main branch as described here: https://stackoverflow.com/questions/35445186/can-github-automatically-merge-branches

tararoshan:
I'll try working on this today!

tararoshan:
I'm trying `./bw-dev update_locales` to see what the expected behavior is, but I'm getting this error:

```{shell}
Execution of msgfmt failed: /app/locale/pl_PL/LC_MESSAGES/django.po:6341: 'msgstr' is not a valid Python format string, unlike 'msgid'. Reason: The character that terminates the directive number 2 is not a valid conversion specifier.
/app/locale/pl_PL/LC_MESSAGES/django.po:6346: 'msgstr' is not a valid Python format string, unlike 'msgid'. Reason: The character that terminates the directive number 2 is not a valid conversion specifier.
msgfmt: found 2 fatal errors
CommandError: compilemessages generated one or more errors.
```
The lines in locale/pl_PL/LC_MESSAGES/django.po:
<img width="523" alt="image" src="https://github.com/bookwyrm-social/bookwyrm/assets/44852921/cf7ad203-6741-43cf-bcb6-9c898ff0730b">

And it's the same in the l10n_main branch [here](https://github.com/bookwyrm-social/bookwyrm/blob/l10n_main/locale/pl_PL/LC_MESSAGES/django.po#L6346). But I couldn't find the line which is giving me trouble, `msgstr " - %(endpercent)s% %"`, in the other locales files, like [Galician](https://github.com/bookwyrm-social/bookwyrm/blob/l10n_main/locale/gl_ES/LC_MESSAGES/django.po#L6288). Instead, the space is removed, making it `msgstr " - %(endpercent)s%%"`.

Is this normal? For now, I'll delete the space and keep chugging along! I'm hoping to submit a PR by the end of this week. 🙂

I won't implement automerging since I'm trying to keep my changes as small as possible, but we can open a new issue with the "good first issue" label for it! 😀

[...]

PR #3211, opened 2024-01-10
Add update locales workflow to l10n_main branch

tararoshan:
Relates to issue #2706. The workflow requires GitHub Actions to have access
to create and approve PRs in order for the workflow to successfully create a PR.
</br>
I did this in my fork in the project settings like so:

image
</br></br>
I noticed, however, that changes to files outside of the `locale` folder are added
to the PR as well. Let me know if that's an issue and I'll try to find a workaround!

### DIFF
--- a/.github/workflows/update_locales.yml
+++ b/.github/workflows/update_locales.yml
@@ -1,0 +1,32 @@
+name: Update locales
+
+on:
+  push:
+    # run this workflow when someone pushes to the l10n_main branch
+    branches: [ l10n_main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    # Create the mock file for the update_locales command to work, otherwise
+    # we'll get an 'no such file' error
+    - name: Create mock .env file
+      run: echo "DEBUG=true" > .env
+
+    - name: Run update_locales command
+      run: ./bw-dev update_locales
+
+    - name: Create pull request
+      run: |
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+        git config user.name "github-actions[bot]"
+        git fetch origin main:main
+        git switch main
+        git fetch origin l10n_main:l10n_main
+        git checkout l10n_main locale/*
+        git commit -m "[GitHub Action] Update locales"
+        gh pr create -B main -H l10n_main --title 'Merge l10n_main into main' --body 'Created by Github action using the update-locales.yml workflow'
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/update_locales.yml
+++ b/.github/workflows/update_locales.yml
@@ -18,15 +18,15 @@ jobs:
     - name: Run update_locales command
       run: ./bw-dev update_locales
 
-    - name: Create pull request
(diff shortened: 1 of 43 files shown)


mouse-reeve:
I set up the update locales command to only pull in locales that are manually added to a list; this is more work but it means that only locales that are in use are shown, and it means less files. I'm no longer sure that that's worth the downsides of making things more complicated, but that might not be a question for today :)

I think if you remove the git commands that merge in `l10n_main`, and just do a git add and commit, the right files will be in the commit and the additional ones will remain excluded. The `./bw-dev update_locales` command includes a step that pulls files over from that branch, so it shouldn't be necessary to do it again.

tararoshan:
Got it! I removed the merge command and set the workflow up as you recommended. I think it works, based on the new commit to the main branch from GitHub Actions on my fork! Let me know if anything seems fishy. :)

dato (commented):
Hello! This is an excellent improvement! I tested it and it works for me.

In addition to the simplifications @mouse-reeve offered, I think we could drop the `git fetch` commands too by making sure the checkout@v3 action always uses the main branch. See [this commit](https://github.com/tararoshan/bookwyrm/compare/l10n_main...dato:update_locales_workflow?expand=1) to see what I mean. (That commit also includes two other minor improvements which you may like to include.)

I wish there was a way for the workflow file itself to live in main and not and l10n_main, but it doesn't seem possible (not easily, anyway). @mouse-reeve Does Crowdin fetch l10n_main and then commits on top of it, or does it just force-push, do you know? (Asking because the later would definitely be a problem.)

tararoshan:
Thank you! I didn't know the `with: refs:` option was available for actions instead of manually setting the upstream, that's neat!

Also, I was looking into your second question and found [this](https://developer.crowdin.com/configuration-file/#configuration-file-for-vcs-integrations) article about Crowdin's version control integrations. Turns out there's already a (GitHub Crowdin Action)[https://github.com/marketplace/actions/crowdin-action] which we could use, if it's more convenient. I'm still not sure if Crowdin fetches updates before making a PR, though.

mouse-reeve:
I'm confused about why this has to live on the l10n branch -- what is the block you're hitting when running this from main? When I use the update locales shortcut, I run it from the main branch, so I would have assumed that the script would operate successfully from main as well.

I suspect the reason not to use the github action is the same as the reason I haven't been merging the PR that crowdin creates automatically -- it brings in _all_ the translation files, rather than the just the ones for locales that are actually in use.

tararoshan:
I think the issue is that to have a GitHub action be triggered by a change to a branch, it needs to live on that branch. When I tried testing with the workflow in the main branch (of my local fork), the workflow was never triggered. As I understand from [this SO issue](https://stackoverflow.com/questions/76344170/trigger-workflow-from-different-branch) and the [GitHub documentation about triggering workflows](https://stackoverflow.com/questions/76344170/trigger-workflow-from-different-branch), Actions need to live in the branch that they're triggered by. In this case, that would be `l10n_main`.

I agree though that this isn't the best for future maintenance. We could also turn this into a scheduled workflow, but it may run unnecessarily (eg. no locales were updated) or run late (eg. hours after a locale was updated).

Thank you for your patience with this, I know it took me a while to reply.

Issue #3190, opened 2024-01-03
bw-dev update fails with "Invalid value set for option Signed-By regarding" nodesource


nesges:
**Describe the bug**
When updating from v0.6.6 to v0.7.1 `docker-compose build` fails with the following error:
```
 => ERROR [6/7] RUN apt-get update            &&     apt-get install -y nodejs &&     pip install -r requirements.txt &&     npm install .                                                                                              0.5s
------
 > [6/7] RUN apt-get update            &&     apt-get install -y nodejs &&     pip install -r requirements.txt &&     npm install .:
0.458 E: Invalid value set for option Signed-By regarding source https://deb.nodesource.com/node_18.x/ nodistro (not a fingerprint)
0.458 E: The list of sources could not be read.
------
Dockerfile:13
--------------------
  12 |
  13 | >>> RUN apt-get update            && \
  14 | >>>     apt-get install -y nodejs && \
  15 | >>>     pip install -r requirements.txt && \
  16 | >>>     npm install .
  17 |
--------------------
ERROR: failed to solve: process "/bin/sh -c apt-get update            &&     apt-get install -y nodejs &&     pip install -r requirements.txt &&     npm install ." did not complete successfully: exit code: 100
ERROR: Service 'dev-tools' failed to build : Build failed
```
please see https://pad.pnpde.social/p/_srhy8QBL0hP5Dsy8IDG for a full log of all output

**To Reproduce**
1. Go to https://docs.joinbookwyrm.com/updating.html
2. Run the single steps of `bw-dev update`  as described

**Expected behavior**
no error

**Instance**
https://buecher.pnpde.social/

**Additional context**
Here's a full log of all output of the above command: https://pad.pnpde.social/p/_srhy8QBL0hP5Dsy8IDG

---

**Desktop (please complete the following information):**
    - OS: Ubuntu 22

mouse-reeve:
As a workaround while this is being fixed, you can skip building the dev tools image, which isn’t needed in production. I believe commenting out the apt-get node steps and then rerunning the update should work: https://github.com/bookwyrm-social/bookwyrm/blob/main/dev-tools/Dockerfile#L13

nesges:
Commenting out lines 13 to 16 helped with running `docker-compose build`, but the next step (`docker-compose run --rm web python manage.py migrate` failed with a new one:

```
root@oghma:/opt/bookwyrm# docker-compose run --rm web python manage.py migrate
[...]

PR #3216, opened 2024-01-13
Ensure dev-tools uses bookworm

dato:
Fixes: #3190.

### DIFF
--- a/dev-tools/Dockerfile
+++ b/dev-tools/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM python:3.9-bookworm
 WORKDIR /app/dev-tools
 
 ENV PATH="/app/dev-tools/node_modules/.bin:$PATH"


dato:
This is just a precaution, we could as well close the bug with the issue itself acting as documentation. Should be very rare. But, if you prefer, we can apply this instead.

dev-tools/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9
+FROM python:3.9-bookworm
```
mouse-reeve, line 1: is there a python version coincidentally called "bookwyrm" or am I misunderstanding this??

mouse-reeve (commented):
I think it's a good idea to fix it here, I'd rather things Just Work rather than needing to be documented wherever possible

dev-tools/Dockerfile
dato: Debian 12 is code-named "bookworm". :)

So these are the Python images based on Debian 12.

mouse-reeve: wild!

Issue #3167, opened 2023-12-25
Switching edition doesn't work


pivic:
**Describe the bug**
I try and switch edition of a book and get an error message.

**To Reproduce**
1. I go to https://bookwyrm.social/book/1215675/s/albert-camus-and-the-human-crisis
2. I see the UI state I have another edition available and I select 'switch to this edition' (the URL to which is https://bookwyrm.social/book/1215676/s/albert-camus-and-the-human-crisis).
3. The error message 'Server Error: Something went wrong! Sorry about that.' appears on the page https://bookwyrm.social/switch-edition

**Expected behavior**
To be able to switch editions.

**Screenshots**
Screencast of the issue: https://drive.proton.me/urls/VH0EG2VWAM#lxVtv1kf0N8-

**Instance**
BookWyrm.social

---

This occurs both via desktop and mobile.

oculos:
Same here. The error I get on the web container is this one: 

```
OK: /switch-edition
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "bookwyrm_shelfbook_book_id_shelf_id_7e945535_uniq"
DETAIL:  Key (book_id, shelf_id)=(1009, 3) already exists.


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/app/bookwyrm/views/books/editions.py", line 91, in switch_edition
    models.ShelfBook.objects.create(
  File "/usr/local/lib/python3.9/site-packages/django/db/models/manager.py", line 85, in manager_method
[...]

PR #3217, opened 2024-01-13
Invalidate `active_shelf` when switching editions

dato:
Fixes: #3167.


mouse-reeve:
Thank you so much for looking into this!

PR #3224, opened 2024-01-17
Pass correct user id in Move notification

### LINKED ISSUE #3196: Notitications was not showing followers, now it's not showing anything and the notification page shows "System error"
**Describe the bug**
The notification page is not working. When I see that there's a new notification (with a number close to the bell icon), I click on it. Then I get "System error" 
Screenshot 2024-01-05 at 10 30 20
Screenshot 2024-01-05 at 10 30 08

**Instance**

https://books.babb.no

**Additional context**
Before this happened, the notifications for new following requests were not showing on the list of notifications. A follow request would trigger the update on the number of new notifications, but the notification itself wasn't showing on the list of notifications

hughrun:
We were passing the *requesting* user's `moved_to` value to the Move notification template, instead of the id of the user that they are being notified about. Additionally, the `id_to_username` template tag had no fallback for if the `user_id is` None.

This resolves both problems and removes an unnecessary space in a template for when the logged in user made the move.

Fixes #3196

### DIFF
--- a/bookwyrm/templatetags/utilities.py
+++ b/bookwyrm/templatetags/utilities.py
@@ -125,7 +125,8 @@ def id_to_username(user_id):
         name = parts[-1]
         value = f"{name}@{domain}"
 
-    return value
+        return value
+    return "a new user account"
 
 
 @register.filter(name="get_file_size")
--- a/bookwyrm/templates/moved.html
+++ b/bookwyrm/templates/moved.html
@@ -23,7 +23,7 @@
 
           <div class="notification is-warning">
             <p>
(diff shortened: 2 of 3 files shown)


bookwyrm/templatetags/utilities.py
```diff
@@ -125,7 +125,8 @@ def id_to_username(user_id):
         name = parts[-1]
         value = f"{name}@{domain}"
 
-    return value
+        return value
+    return "a new user account"
```
mouse-reeve, line 129: I'm going to merge this since it's an active bug, but this should be localized as `_(a new user account)`

PR #3226, opened 2024-01-17
Disable user exports

### LINKED ISSUE #3210: default `DATA_UPLOAD_MAX_MEMORY_SIZE` is too small for user imports
**Describe the bug**
The `DATA_UPLOAD_MAX_MEMORY_SIZE` and `FILE_UPLOAD_MAX_MEMORY_SIZE` default to 2.5MB which is big enough for reasonable image uploads but not for user imports.

**To Reproduce**
1. Export user
2. Import user file larger than 2.5M
3. Get error

**Expected behavior**
We should change the defaults to something bigger (20MB?) and enable admins to adjust it in the admin interface.

See https://docs.djangoproject.com/en/5.0/ref/settings/#data-upload-max-memory-size 

**Instance**

testing

hughrun:
- add setting to enable and disable user exports (defaults to `False`)
- do not allow user exports when using s3 storage
- do not serve non-image files from `/images/` (requires admins to update nginx settings)
- increase default file upload limit to 100MB to enable user exports to be imported (can be changed in .env)

Fixes #3210


prolibre:
Hello @hughrun It's me again
sorry, I don't mean to pester you :-)

I think there's a new variable missing in .env.example to allow (or not) export.
The variable that would correspond to user_exports_enabled

hughrun:
@prolibre it's a new value in the user exports model. You need to run `./bw-dev migrate` or for dockerless I think it's `python manage.py migrate`

prolibre:
@hughrun 
I misspoke (because of the translation).
What variable should I put in my .env file (I don't use docker) to activate the export ? I use the .env file as a configuration file.

hughrun:
My apologies @prolibre this has not been explained properly. You do not use the `.env` - there is a setting in the Admin panel. Go to `/settings/dashboard` and then click on "Imports" in the menu. That shows various settings for book and user imports, and also now has a setting for enabling or disabling user _exports_. I will add an Issue to make this a bit clearer.

PR #3228, opened 2024-01-18
Fix user exports to deal with s3 storage

hughrun:
(updated 29 Jan (AEDT) )

User exports are failing on instances using s3 storage (i.e. most) because we're treating all image files as local files.

This PR:

- separates export database queries into multiple tasks
- uses `S3Tar` library to combine s3 stored images and JSON data into a `tar.gz` file when using s3 storage
- uses custom `Storage` to secure export files from public access
- uses a signed s3 url with 5 minute timeout when using s3 storage

The main guts of the changes here are in a new class called `AddFileToTar` in the export job.

This has been manually tested with local and s3 storage and appears to work with both.

**OUTSTANDING ISSUE**

- Azure blob storage is not considered here, I'm not really sure how it works but I assume it won't work for the same reason s3 wasn't working.


hughrun:
Ok I think I've worked this out. Hopefully will have a fix and a cleaner PR tomorrow.

.env.example
```diff
@@ -81,6 +81,7 @@ AWS_SECRET_ACCESS_KEY=
 # AWS_S3_CUSTOM_DOMAIN=None       # "example-bucket-name.s3.fr-par.scw.cloud"
 # AWS_S3_REGION_NAME=None         # "fr-par"
 # AWS_S3_ENDPOINT_URL=None        # "https://s3.fr-par.scw.cloud"
+# S3_ENDPOINT_URL=None            # same as AWS_S3_ENDPOINT_URL - needed for non-AWS for user exports
```
skobkin, line 84: Couldn't it be duplicated in the code itself and not by a user if it has the same value? Leaving this in the configuration could potentially lead to more misconfiguration problems.

hughrun: I certainly agree it would be cleaner but I'm not sure how to do it. [s3-tar](https://pypi.org/project/s3-tar/) doesn't provide a way to configure credentials, it just picks them up from environment variables, and it doesn't seem to be possible to make one env value equal another within a `.env` file.

skobkin: > needed for non-AWS for user exports

Do you mean [this](https://github.com/xtream1101/s3-tar/blob/master/s3_tar/utils.py#L25-L29)?

I'm not a Python developer at all and I do not know best practices here, but I have an idea.

What if Bookwyrm will set this in `os.environ` when reading it's settings? Of course it would probably be a good idea to also set it with [`os.putenv()`](https://docs.python.org/3/library/os.html#os.putenv) just to be consistent.

`s3-tar` reads `os.environ` only when `_create_s3_client()` is called and when it's called, Bookwyrm would've already set that.

I understand that I'm suggesting a dirty hack, but considering that `s3-tar` is written in a such way that it doesn't allow to override this value, for me personally using a dirty hack in the environment I as a developer of a product control is better than leaving the hack to the end-user especially if I'm completely sure that both variables should have the SAME value and user shouldn't use different values. This way I'm not giving my user an opportunity to mess things up by changing one and forgetting to change another or not setting one of them at all for example.

I don't like this and if I were `s3-tar` developer, I'd solve the problem in the library itself. But since we can't do that...

What do you think?

dato:
> I certainly agree it would be cleaner but I'm not sure how to do it.

`os.putenv()`/`os.environ`, as @skobkin suggests, could be made to work, yes; though lines like [this one](https://github.com/xtream1101/s3-tar/blob/fffacac31fbddb6c326ad42de527655c08dd65f3/s3_tar/s3_tar.py#L25) in the s3-tar source makes me think it's a fragile approach (that style of coding could easily lead to the variable be set at import time).

Another approach would be to subclass `boto3.Session`, as [done here](https://github.com/hughrun/bookwyrm/compare/user-export...dato:bookwyrm:export_job_inject_aws_endpoint_setting?expand=1) (i.e. hughrun/bookwyrm#3; untested, sorry).

Thoughts?

.env.example
hughrun: I don't love changing ENV values on the fly (it's definitely frowned upon) but this works, and I agree it's better than adding another user-defined value that is just identical in value to one we already have.

hughrun:
> Another approach would be to subclass `boto3.Session`, as [done here](https://github.com/hughrun/bookwyrm/compare/user-export...dato:bookwyrm:export_job_inject_aws_endpoint_setting?expand=1) (i.e. [hughrun#3](https://github.com/hughrun/bookwyrm/pull/3); untested, sorry).
> 
> Thoughts?

I can get overriding `os.environ` to work, but it feels very hacky. I like this better. I'll test it out and see what happens.

hughrun:
Ok almost done, I just need to disable azure storage, which I forgot about.

hughrun:
@bookwyrm-social/code-review this is ready to check.

bookwyrm/views/preferences/export.py
```diff
@@ -152,6 +154,36 @@ def get(self, request):
...
+
+            if settings.USE_S3:
+                # make custom_domain None so we can sign the url
+                # see https://github.com/jschneier/django-storages/issues/944
+                storage = S3Boto3Storage(querystring_auth=True, custom_domain=None)
+
+                # for s3 we download directly from s3, so we need a signed url
+                export["url"] = S3Boto3Storage.url(
+                    storage, f"/exports/{job.task_id}.tar.gz", expire=900
+                )  # temporarily downloadable file, expires after 5 minutes
```
dato, line 170: comment says 5 minutes, but 900 is 15 minutes in seconds. Does either figure need updating?

hughrun: hmm, honestly I'm not sure which I intended. 5 minutes seems reasonable?

mouse-reeve: 5 minutes seems tight to me, I can imagine getting up for a glass of water and missing it.

skobkin: @mouse-reeve @hughrun Especially if the process is asynchronous and we don't exactly know when it ends. For users with large archive it could be easy to miss even probably even in 15 minutes.

mouse-reeve: Would it make sense to have a settings.py/env variable for this with a reasonable default?

skobkin: Also I've checked Mastodon's source. Didn't find the lifetime of the link there, but I've found out that they're also adding some randomness to the archive filename. Which seems reasonable in case of local storage for example.

[app/services/backup_service.rb:64](https://github.com/mastodon/mastodon/blob/81a04ac25cb5a25d8501e59a16b272a6b4e177f0/app/services/backup_service.rb#L64)

hughrun: Ok yes 5 mins is too short but I think 7 days is excessive. The url will use a random uuid but we have to remember this is still a file potentially containing private statuses.

I'll fix this up probably tomorrow and make it configurable.

skobkin: I'm just pointing at a fact that as this is asynchronous process, user may not even be able to download their archive the same day 🤷

I didn't see an entire code (sorry), but if users just gets new URL when trying to download their archive it's fine. But if they need to wait for new archive generation due to expired link, it may become an endless loop for them.

- request archive
- it takes too long
- I'll check out later
- link expired
- requesting new archive
- ...

hughrun: Ok folks I think there is some confusion here.

This is a completely separate function from the creation of a user export file. That process is triggered when the user requests the export. It's a series of background jobs: making it a synchronous process would be bonkers.

What we are looking at here is the function that runs _every time the user loads the user exports page_. That's why I originally was thinking 5 minutes would be reasonable: if it has expired you can reload the page and get a fresh link for every available download file you've generated.

Minnozz: If you want to avoid ever letting the user face an expiring link, another option is to make the link in the HTML point to an internal (authenticated) endpoint that redirects to the signed S3 link. That way the link has to be valid only for a very short time, since it is created immediately before redirecting the user to it.

(But with @hughrun's explanation, generating a download link that is valid for a few minutes when the user opens the exports page is definitely fine too)

skobkin: @hughrun 
> What we are looking at here is the function that runs every time the user loads the user exports page

Ah. Thanks for clearing this out! Then my comments are mostly non-relevant 😞

> make the link in the HTML point to an internal (authenticated) endpoint that redirects to the signed S3 link

I double that. Requesting new links form S3 every time user loads the page seems to have at least two downsides:

- You can't send such link via email for example like Mastodon does
- It'll cause useless S3 requests which could be non-free in most of object storage providers

hughrun:
Uh, not sure what's going on with this error for tests - might be [the read-only problem](https://github.com/bookwyrm-social/bookwyrm/commit/864304f128fc897348342ba83506441469e8fe53)?

Minnozz:
> Uh, not sure what's going on with this error for tests - might be [the read-only problem](https://github.com/bookwyrm-social/bookwyrm/commit/864304f128fc897348342ba83506441469e8fe53)?

The migrations are not up to date, because they seem to depend on the configuration in `.env`:

```diff
❯ diff -u bookwyrm/migrations/019{8,9}*
--- bookwyrm/migrations/0198_alter_bookwyrmexportjob_export_data.py	2024-03-24 11:31:31
+++ bookwyrm/migrations/0199_alter_bookwyrmexportjob_export_data.py	2024-03-24 11:35:39
@@ -1,4 +1,4 @@
-# Generated by Django 3.2.25 on 2024-03-24 08:53
+# Generated by Django 3.2.25 on 2024-03-24 10:34

 import bookwyrm.storage_backends
 from django.db import migrations, models
@@ -7,7 +7,7 @@
 class Migration(migrations.Migration):

     dependencies = [
-        ("bookwyrm", "0197_merge_20240324_0235"),
+        ("bookwyrm", "0198_alter_bookwyrmexportjob_export_data"),
     ]

     operations = [
@@ -16,7 +16,7 @@
             name="export_data",
             field=models.FileField(
                 null=True,
-                storage=bookwyrm.storage_backends.ExportsS3Storage,
+                storage=bookwyrm.storage_backends.ExportsFileStorage,
                 upload_to="",
             ),
         ),
```

I don't think the database structure should depend on the runtime configuration. Maybe we should have separate fields for the S3 backend and File backend, and only fill the field that matches the runtime configuration?

bookwyrm/views/preferences/export.py
Minnozz: I implemented the redirect for S3.

Minnozz:
Before this PR, archives not using S3 storage were stored inside the `/images/` folder (Docker `media_volume`).

After this PR, they are stored in the `/exports/` folder. I added a Docker volume called `exports_volume` for that directory.

Because the location has changed (unrelated to Docker, just the filesystem path), this means that exports made before this PR are no longer available for download unless they are moved to the `/exports/` directory when upgrading:
```sh
❯ ./bw-dev bash
root@336b620fa3e4:/app# mv images/*-*-*-*-* exports/
```
(Do this after the new `exports_volume` is created, when using Docker).

hughrun:
Thanks for your assistance with this @Minnozz! I still have a lot to learn :)

I've tested your changes with s3, looks good.

@mouse-reeve I think this is ready now, and @Minnozz's comment above needs to go in release notes, as well as a note to say we have not yet managed to deal with Azure storage for exports.

Since this was originally my PR I'm not going to merge it but @dato, @Minnozz and myself have all had a look and made adjustments as needed so I'm reasonably confident it's good to go.

bookwyrm/views/preferences/export.py
```diff
@@ -162,7 +188,7 @@ def get(self, request):
         next_available = (
             jobs.first().created_date + timedelta(hours=hours) if not allowed else False
         )
```
Minnozz, line 190: If this check exists for security/DoS reasons, it should also happen in the `post` method, otherwise a malicious actor could simply craft the request themselves.

bookwyrm/models/bookwyrm_export_job.py
```diff
@@ -1,213 +1,445 @@
...
+
+
+def url2relativepath(url: str) -> str:
+    """turn an absolute URL into a relative filesystem path"""
+    parsed = urlparse(url)
+    return unquote(parsed.path[1:])
+
+
+class AddBookToUserExportJob(ChildJob):
+    """append book metadata for each book in an export"""
```
Minnozz, line 107: An export now consists of many separate child jobs. Is there a mechanism to prevent data races between different child jobs? Otherwise, the following can happen:

1. Start task for book B, read current JSON containing book A
2. Start task for book C, read current JSON containing book A
3. Task for B adds itself and saves; archive contains A + B
4. Task for C adds itself and saves; archive contains A + C

This means the data about book B gets lost

Minnozz (changes requested):

Minnozz:
I've refactored the creation of the tar file a bit, by not re-using the same `File` (the one behind the file field) as a temporary file every time one is needed, but just creating it directly.

bookwyrm/models/bookwyrm_export_job.py
hughrun: hmm, I see what you mean. I guess we need something in `TriggerBookJobs` to wait for each `AddBookToUserExportJob` to finish before triggering the next one?

bookwyrm/views/preferences/export.py
hughrun: Ah, you're right. I wasn't really thinking about DDoS to be honest, just preventing impatient users from making requests in the UI - we might have the same issue in book exports. It needs to be in both GET and POST because we need to tell users when they are next allowed to do an export.

Minnozz:
I noticed that we are re-using the `IMPORTS` queue for exports too (even before this PR). Should those get a separate queue?

Minnozz (approved):

Minnozz:
I just realised we may need to fix the acl of the temporary export json that is uploaded with the "manual" connection

Minnozz:
@hughrun Thanks for testing and fixing my additions! Do you consider this PR ready to merge?

hughrun:
I do. I've tested the latest changes on s3 and local.

PR #3229, opened 2024-01-18
nginx: fix missing ttf static files

verymilan:
in order to fix broken icons after the .2 release :)

### DIFF
--- a/nginx/development
+++ b/nginx/development
@@ -64,7 +64,7 @@ server {
     # directly serve images and static files from the
     # bookwyrm filesystem using sendfile.
     # make the logs quieter by not reporting these requests
-    location ~ \.(bmp|ico|jpg|jpeg|png|tif|tiff|webp|css|js)$ {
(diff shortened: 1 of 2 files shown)


prolibre:
Hello @verymilan !

So I'm working on it because I'm using Apache and I need to adapt the code.

I've put this :
`"(?i)\.(gif|bmp|ico|jpg|jpeg|png|tif|tiff|webp|css|js|ttf|svg|woff|scss)$"`

based on what I found in images and static files. The (?i) is because I also found capital letters.

hughrun:
@verymilan sorry this was my fault. Do you mind adding `svg` as well? See #3230

marcusyoung:
I also had to add woff|scss|gif|webp to prevent 404 errors. These don't seem to have made it into the nginx config file in the production branch, which still only has: `location ~ \.(bmp|ico|jpg|jpeg|png|svg|tif|tiff|webp)$`

PR #3238, opened 2024-01-20
fix multiple issues from user exports config changes

### LINKED ISSUE #3227: fix localization in #3224
See #3224

Also check whether this is part of the translations documentation.

### LINKED ISSUE #3231: Add link for admins to turn on user exports from user exports page
Now that user exports are turned off by default it's not obvious for new admins how to turn them on, especially given they have to go to a menu labelled "imports". It's a bit confusing because some config settings are in `.env` and others are controlled from the admin panel.

Either a link to the relevent admin config page or simply a button on the user exports page that will turn exports on - only visible to users with admin privileges.

Alternatively we could just document this somewhere - which we should do anyway.

**Additional context**

### LINKED ISSUE #3232: NGINX : Allowed (static) files
I'm opening a ticket because it's possible that some of my previous comments have gone unnoticed (my fault, I've posted messages here and there).

I've had a look at the various files in static and I've come up with this configuration for my apache2 server (and therefore the equivalent under NGINX) :

"(?i)\.(**gif**|bmp|ico|jpg|jpeg|png|tif|tiff|webp|css|js|ttf|svg|**woff**)$"

On my server, some cover images are in **gif** format. Sometimes there are also capital letters in the extension (JPG) => (?i)
I add **woof** for the font used in administration.

`
SELECT DISTINCT SPLIT_PART(SPLIT_PART(cover, '.', ARRAY_LENGTH(STRING_TO_ARRAY(cover, '.'), 1)), '?', 1) AS extension FROM "public"."bookwyrm_book" WHERE cover!='';`

gif
jpg
png
webp
JPG
jpeg

Well... I'll leave you in peace for the weekend :-)

EDIT: I'm not sure about the woff fonts... I seem to have seen a 404 somewhere.

### LINKED ISSUE #3236: Flower crash : ValueError: Couldn't import 'celerywyrm': Environment variable "DATA_UPLOAD_MAX_MEMORY_SIZE" invalid: Not a valid integer.`
Since updating my .env (following the .env.example) the flower application no longer worked.
Here's a bit of the error :

`master celery[2501464]: ValueError: Couldn't import 'celerywyrm': Environment variable "DATA_UPLOAD_MAX_MEMORY_SIZE" invalid: Not a valid integer.`

I changed :
DATA_UPLOAD_MAX_MEMORY_SIZE = (1024**2 * 100)
to
DATA_UPLOAD_MAX_MEMORY_SIZE = 204800

and it seems to work again

hughrun:
- improve nginx config
- fix DATA_UPLOAD_MAX_MEMORY_SIZE default not being an int
- translate fallback value in id_to_username template tag
- make location of setting to turn on user exports easier to locate for admins

fixes #3227
fixes #3231
fixes #3232
fixes #3236


hughrun:
I'm not at all confident about the nginx changes (at the very least there's surely a more succinct way to do this) so advice on that would be welcome.

@mouse-reeve I'm also not entirely confident I'm using `blocktrans` properly here. I couldn't work out how to use the `{% url 'blah' %}` tag inside a translated string.

mouse-reeve:
I found some other translation issues, so I went through and fixed the lot. To your specific question, you can set the url variable using `as`: `{% url 'settings-imports' as url %}`. Then the `blocktrans` tag should contain as little HTML as possible, so the `<br>` tag should live outside the translation block, and there shouldn't be line breaks or trailing whitespace, so I added `trimmed` to remove it:

```django
        <br/>
        {% url 'settings-imports' as url %}
        {% blocktrans trimmed %}
            User exports settings can be changed from <a href="{{ url }}">the Imports page</a> in the Admin dashboard.
        {% endblocktrans %}
```

hughrun:
@bookwyrm-social/code-review this is ready for final review.

There is still an issue with the reverse-proxy nginx config however I'm going to log that as a separate issue rather than bog down this PR, as I'm probably not the best person to fix it.

Thanks @dato and @mouse-reeve for your help!

mouse-reeve (approved):

PR #3245, opened 2024-01-23
Add redis automatic rewrite configuration.

WesleyAC:
This should hopefully prevent the AOF file from growing too large.

Should be tried out on b.s prod before we merge this.

### DIFF
--- a/redis.conf
+++ b/redis.conf
@@ -2,6 +2,9 @@ bind 127.0.0.1 ::1
 protected-mode yes
 port 6379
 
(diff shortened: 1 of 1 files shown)


WesleyAC:
`BGREWRITEAOF` on b.s reduced the redis_activity AOF size from ~29GB to ~800MB. Haven't tested this config change, but I suspect it should do the trick for eliminating this problem in the future.

mouse-reeve:
Thank you for figuring this one out!

Issue #3082, opened 2023-11-03
Make sessions last longer


WesleyAC:
I feel like I get logged out of bookwyrm.social all the time. I think we should default to much longer/more persistent sessions than we do. In the modern era short sessions are generally not considered a security advantage and are quite a nuisance.

Looks like this is as simple as setting [`SESSION_COOKIE_AGE`](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-SESSION_COOKIE_AGE), which defaults to two weeks. I propose a default value of 4 years (`60*60*24*365*4` == `126144000`). Thoughts?

mouse-reeve:
I think this would be very nice. My only question is if having longer-lasting cookies would cause GDPR issues?

dato:
I think a year should be fine in any case.

Rationale (but I am not a lawyer):

- as I understand it, log-in cookies/authentication cookies fall in the “strictly necessary cookies” category (plus are first-party cookies), and hence do not require informing consent.
- on the other hand, confirmed consent for cookies that require is only valid for 12 months, so using the same time span for log-in seems a nice form of renewing such “implicit consent.”

Still, I think that 4 years is compliant as well, though I could see how it could raise an eyebrow. (Was there any particular reason for that number, and not 2 or 6? 1 or 2 seem best to me.)

WesleyAC:
https://law.stackexchange.com/questions/44173/cookie-expiration-time-per-gdpr-guidelines has a good summary of GDPR rules, I don't think there's anything really to worry about — we need to keep cookies around for the minimum amount of time that is necessary to achieve their purpose, but there aren't any restrictions on what the purposes may be, so keeping the user logged in for a given amount of time is the purpose. I think for login cookies this essentially means that the expiry date of the cookie needs to be the same as (or less than) the expiry date of the session on the backend, otherwise the cookie would be around for longer than is necessary, but there does not appear to be a restriction on what that amount of time is. This is rather silly, but many parts of GDPR are silly in this kind of way.

I suggested four years since that's a normal period of time for someone to own a device for, the idea being that people should be able to log on once and never have to again until they get a new device. However, I realize I was misinterpreting how session expiry works when I wrote that — the session expiry time is reset every time the session is used. Given that, a year seems fine — that would mean people would get logged out after a year of not accessing bookwyrm, which feels like a reasonable tradeoff between not being annoying to infrequent users and not having sessions kicking around for lokger than they need to be.

[...]

Issue #2278, opened 2022-08-14
Add an option to view and terminate activity sessions


yuvalne:
**Current situation**
There is no way to view the active sessions for one's account.

**Desired situation**
A list of all active sessions with a descriptive title for each, their IP addresses/location, time of last connection (possibly time of first connection) and a button to revoke them, not dissimilar to the way Mastodon does it:
<img width="952" alt="Sessions section from Mastodon's settings" src="https://user-images.githubusercontent.com/97384040/184544810-6f03d642-c73d-4dd4-8e7e-79b5a79c7157.png">

**Importance of this change**
In case of logging in from a browser no longer in your control, from a no longer used client etc there is no way to revoke access other than changing the password. This is a big security risk.

Thank you 😊

mouse-reeve:
This is a great suggestion -- I've been investigating [how Django sessions work](https://docs.djangoproject.com/en/3.2/topics/http/sessions/) and unfortunately it looks like this will be a pretty involved change (I wasn't able to find any way in the current set up, for example, to find other sessions for the current request user in a view).

PR #3249, opened 2024-01-24
Set SESSION_COOKIE_AGE from environment

dato:
While we do wish for a longer maximum age (up to one year, see #3082), we only want to do that after termination of active sessions is implemented (see #2278).

In the meantime, by reading and setting the variable from settings, we allow site admins to alter the default already.


skobkin:
Maybe also update default `.env` example so it'll be there for admins to find? It could be commented by default though.

PR #3251, opened 2024-01-24
Support DATA_UPLOAD_MAX_SIZE_MiB in .env

dato:
Since arithmetic is not allowed in .env files, a change in unit for the variable seems most usable.

-----

I missed #3226 during review, but while I was editing .env.example for another PR, I noticed the `(1024**2 * 100)` syntax, which I don't think would be allowed in the actual `.env`file and hence could lead to confusion?

In cases like this, I think it's best to overshadow Django's own setting with a Bookwyrm-specific variable name, in another unit (MB in this case). Thoughts?

NOTE: I mark this PR as draft because, as implemented, it's not strictly backwards compatible (if, following .env.example, an admin set `DATA_UPLOAD_MAX_MEMORY_SIZE` in the environment for 0.7.2, this would no longer be honored). I do believe this will merit a mention in the release notes, but I don't think it would make sense to allow two variables in different units, from .env.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,6 +137,7 @@ TWO_FACTOR_LOGIN_MAX_SECONDS=60
 # and AWS_S3_CUSTOM_DOMAIN (if used) are added by default.
 # Value should be a comma-separated list of host names.
 CSP_ADDITIONAL_HOSTS=
-# The last number here means "megabytes"
+
 # Increase if users are having trouble uploading BookWyrm export files.
-DATA_UPLOAD_MAX_MEMORY_SIZE = (1024**2 * 100)
+# Default value is 2.5MB, this value is 100MB
+DATA_UPLOAD_MAX_MEMORY_SIZE=104857600
--- a/.env.example
+++ b/.env.example
@@ -138,9 +138,9 @@ TWO_FACTOR_LOGIN_MAX_SECONDS=60
 # Value should be a comma-separated list of host names.
 CSP_ADDITIONAL_HOSTS=
 
-# Increase if users are having trouble uploading BookWyrm export files.
-# Default value is 100MB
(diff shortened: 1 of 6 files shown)


.env.example
```diff
@@ -137,6 +137,7 @@ TWO_FACTOR_LOGIN_MAX_SECONDS=60
...
 # Value should be a comma-separated list of host names.
 CSP_ADDITIONAL_HOSTS=
-# The last number here means "megabytes"
-# Increase if users are having trouble uploading BookWyrm export files.
-DATA_UPLOAD_MAX_MEMORY_SIZE = (1024**2 * 100)
\ No newline at end of file
+
+# Maximum allowed memory for file uploads (increase if users are having trouble
+# uploading BookWyrm export files).
+# DATA_UPLOAD_MAX_MEMORY_MiB=100
```
dato, line 143: (Preferring to leave this commented out so that settings.py can update the default and have it affect to default installs.)

mouse-reeve:
I think this is going to be in conflict with #3238?

dato:
Good catch, thank you! I've rebased it on top of #3238 now.

I've also opened hughrun#4 in case @hughrun wants to merge it there directly. (Too bad Github doesn't allow chained PRs when forks are involved.)

hughrun:
Thanks I'll take a look soon. We're likely to have a few PR conflicts because I left a bit of a mess earlier and lots of people noticed 🙃. I'll see if there's a way to pull it all in to one PR.

hughrun:
I've merged this into #3238.

That will be ready for review again once I sort out some issues with the Reverse-Proxy nginx settings.

dato:
Great! Closing this one then.

PR #3253, opened 2024-01-24
Adding FictionBook format ("FB2", "FB3") to autocomplete options in "get a copy" block.

skobkin:
> [FictionBook](https://en.wikipedia.org/wiki/FictionBook) is an open XML-based e-book format which originated and gained popularity in Russia. FictionBook files have the `.fb2` or `.fb3` filename extension, regarding to their version. All FB2/FB3 capable readers also support ZIP-compressed FictionBook files (`.fb2.zip` or `.fbz`).

It's also supported by a lot of book-reading software, especially by mobile apps.

I propose to add it to the autocomplete list.


mouse-reeve:
Thanks for adding this!

PR #3257, opened 2024-01-26
Use shared inboxes for mentions too

dato:
This avoids duplicate submissions to remote instances when mentioning followers (i.e., `POST /user/foo/inbox` followed by `POST /inbox`, which results in two separate `add_status` tasks, and might generate duplicates in the target instance).

-----

See, for example, https://lectura.social/user/ArianeDeTroil/comment/58073, where a single comment from my instance generated three copies of the note (because there were two mentions to the same instance). I believe the first commit fixes the issue.

### DIFF
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -153,7 +153,7 @@ class ActivitypubMixin:
         mentions = self.recipients if hasattr(self, "recipients") else []
 
         # we always send activities to explicitly mentioned users' inboxes
-        recipients = [u.inbox for u in mentions or [] if not u.local]
+        recipients = [u.shared_inbox or u.inbox for u in mentions if not u.local]
 
         # unless it's a dm, all the followers should receive the activity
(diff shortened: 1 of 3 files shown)


bookwyrm/models/activitypub_mixin.py
```diff
@@ -152,8 +152,9 @@ def get_recipients(self, software=None) -> list[str]:
         # find anyone who's tagged in a status, for example
         mentions = self.recipients if hasattr(self, "recipients") else []
 
-        # we always send activities to explicitly mentioned users' inboxes
-        recipients = [u.inbox for u in mentions or [] if not u.local]
+        # we always send activities to explicitly mentioned users (using shared inboxes
+        # where available to avoid duplicate submissions to a given instance)
+        recipients = {u.shared_inbox or u.inbox for u in mentions if not u.local}
```
mouse-reeve, line 157: I didn't realize you could create sets this way, this is great! Thank you

mouse-reeve (approved):

PR #3258, opened 2024-01-26
Fix detection of `unlisted` posts

dato:
If `followers_url` is found in `to`, the post may still be _unlisted_ if `"https://www.w3.org/ns/activitystreams#Public"` appears in `cc`. Hence this should be checked earlier.


dato:
I will add a test and undraft if there's agreement the fix is good.

mouse-reeve:
This seems like a good fix to me!

Issue #2023, opened 2022-03-16
Confirmation code page is not tablet friendly


MaybeThisIsRu:
![image](https://user-images.githubusercontent.com/3974019/158655808-3fa795e4-7d87-44af-8733-ee8a4c9ae10a.png)

I will PR this sometime in the next few days.

bSolt:
I'll take a look at this later. More reactive html is right up my alley

bSolt:
Hm would just making the image smaller at this width be best?

bSolt:
Or it could be hidden like at narrower widths

![image](https://github.com/bookwyrm-social/bookwyrm/assets/36494925/0d5e8db3-b0da-4d9b-a974-bfe6bf22a372)

bSolt:
I'm not so confident in these solutions because this element is also used in other places I'll have to check

bSolt:
I probably like this best, just make the first part full width then stack the second with the image

![image](https://github.com/bookwyrm-social/bookwyrm/assets/36494925/a2c1654c-2a36-411a-9b4f-2f2746c40a4e)

This should probably apply to some other pages with similar layouts like /invite/... and possible /login. I'll work on a PR

bSolt:
@mouse-reeve @hughrun any preference on the layout solution? I am thinking either make the form column full width except on desktop, or use a smaller logo image in all cases (maybe "md"). Also I have determined that this same issue applies to the invite page, login/reactivate (without registrations enabled), and password reset pages.

mouse-reeve:
I think the full-width solution seems the nicest, since it prioritizes and gives more space to the thing that people are on the page to actually use

mouse-reeve:
Fixed by #3260

PR #3260, opened 2024-01-26
Fix awkward layout for tablets on /confirm-email, /login, /invite, and /preferences/reactivate

bSolt:
Fixes issue #2023 

I'm not actually sure what's going on with the password-reset page as I don't actually see the about snippet show up even though it is in the template.  On all of the other pages, at tablet widths (anything between the mobile and desktop cutoffs) the main element will grow to the full width and the about snippet will move below it.

e.g.

image

### DIFF
--- a/bookwyrm/templates/landing/login.html
+++ b/bookwyrm/templates/landing/login.html
@@ -6,7 +6,7 @@
 {% block content %}
 <h1 class="title">{% trans "Log in" %}</h1>
 <div class="columns is-multiline">
-    <div class="column is-half">
+    <div class="column {% if site.allow_registration %}is-half{% else %}is-full is-half-desktop{% endif %}">
         {% if login_form.non_field_errors %}
         <p class="notification is-danger">{{ login_form.non_field_errors }}</p>
         {% endif %}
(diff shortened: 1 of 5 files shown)


bSolt:
https://github.com/bookwyrm-social/bookwyrm/pull/3260/files#diff-3c51df4554b0aa768eef4aa9e834557c349d6f5db09d48e242437f5365f460a7R59-R64

This is the element that doesn't show up for me locally, but it can't really have awkward spacing if it is hidden so I guess its fine? I just don't understand why I don't see it. Maybe I'm missing an if condition

bookwyrm/templates/landing/login.html
```diff
@@ -6,7 +6,7 @@
 {% block content %}
 <h1 class="title">{% trans "Log in" %}</h1>
 <div class="columns is-multiline">
-    <div class="column is-half">
+    <div class="column {% if site.allow_registration %}is-half{% else %}is-full is-half-desktop{% endif %}">
```
bSolt, line 9: I decided that this works fine on tablets when registration is enabled to remain as half width, because then the other half is the registration form vs the about snippet (when it is disabled)

e.g.

image

mouse-reeve:
> https://github.com/bookwyrm-social/bookwyrm/pull/3260/files#diff-3c51df4554b0aa768eef4aa9e834557c349d6f5db09d48e242437f5365f460a7R59-R64
> 
> This is the element that doesn't show up for me locally, but it can't really have awkward spacing if it is hidden so I guess its fine? I just don't understand why I don't see it. Maybe I'm missing an if condition

wow that is a mystery to me too! weird

mouse-reeve:
Thanks so much for fixing these views, this is a big improvement.

PR #3261, opened 2024-01-27
Add book series by title in feed posts

### LINKED ISSUE #3256: Show series name and number after title
I read a lot of book series, and often think of books as "# 4 in that series".

Goodreads shows the series name and number after a book title:
image

BookWyrm also shows this information, but only on the [book detail page](https://bookwyrm.social/book/177870/s/field-of-dishonor):
image

and not, for instance, in the feed:
image

I would like to see the book series and number in more places where currently the book title is shown.

- A user setting could be added to configure this behavior. The default could go either way.

**Additional context**
n/a

bSolt:
This adds the book series if it exists and the series number if it exists to the header of feed posts which are: started reading, reading, done reading, review, stopped reading, quotes, and when by the title where shows up in many places around the site.

Fixes #3256 

We may or may not want to hide this behind a preference. LMK if I should add any additional css class or anything. Maybe the translation block?

image


Minnozz:
Thanks for implementing this!

bookwyrm/templates/snippets/book_titleby.html
```diff
@@ -9,12 +9,19 @@
...
+by
 {% endblocktrans %}&nbsp;{% include 'snippets/authors.html' with book=book limit=3 %}
 
 {% else %}
 <a href="{{ book.local_path }}">{{ book|book_title }}</a>
 {% endif %}
 
+{% if book.series %}
+({{book.series}}{%if book.series_number %}, #{{book.series_number}}{% endif %})
+{% endif %}
```
Minnozz, line 23: Maybe we want to put this piece in a separate snippet that we include here and in all the headers? That way there is a single point of definition for how the book series is rendered.

Minnozz:
I've factored out the snippet and made one additional change: to render the series number the same way (separated by a comma) on the book detail page:
image
I've also places the series number outside of the link to make it clear which is which (many books currently incorrectly include the series number in the series name).

PR #3282, opened 2024-02-09
Manually delete unused keys from redis activity cache

mouse-reeve:
This adds an admin UI to delete keys from the redis activity cache that are no longer in use.

Remaining work:
 - [ ] Separate out the type of keys to delete (cache vs activitystreams)
 - [ ] Allow deletion of entire django cache, separately from cache items with an outdated prefix
 - [ ] Ensure that the activitystreams code isn't re-creating invalid keys (ie, for deleted users)
 - [ ] Maybe add a schedule-able task to remove unused keys?

### DIFF
--- a/bookwyrm/templates/settings/redis.html
+++ b/bookwyrm/templates/settings/redis.html
@@ -1,0 +1,68 @@
+{% extends 'settings/layout.html' %}
+{% load humanize %}
+{% load i18n %}
+
+{% block title %}{% trans "Redis Status" %}{% endblock %}
+
+{% block header %}{% trans "Redis Status" %}{% endblock %}
+
+{% block panel %}
+
+{% if info %}
+<section class="block content">
+    <h2>{% trans "Info" %}</h2>
+    <div class="columns has-text-centered is-multiline">
+        <div class="column is-4">
+            <div class="notification">
+                <p class="header">{% trans "Used memory" %}</p>
+                <p class="title is-5">{{ info.used_memory_human }}</p>
+            </div>
+        </div>
+        <div class="column is-4">
+            <div class="notification">
+                <p class="header">{% trans "Total system memory" %}</p>
+                <p class="title is-5">{{ info.total_system_memory_human }}</p>
+            </div>
+        </div>
+        <div class="column is-4">
+            <div class="notification">
+                <p class="header">{% trans "Keys" %}</p>
+                <p class="title is-5">{{ info.db0.keys | intcomma }}</p>
+            </div>
+        </div>
+    </div>
+</section>
+
+<section>
+    {{ dead_key_count }}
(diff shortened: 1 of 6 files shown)


mouse-reeve:
Despite being two years old, it seems like this actually worked fine and just needed a few test mocks. I would love to get some review on this (@bookwyrm-social/code-review)

This should be fundamentally non-destructive -- completely clearing redis_activity would log everyone out and empty their streams, which isn't great but is repairable. If this were redis_broker, it would be a different story, as that would drop celery tasks.

bookwyrm/templates/settings/redis.html
```diff
@@ -0,0 +1,125 @@
...
+        </form>
+        {% endif %}
+    </div>
+</section>
+
+<section class="block content">
+    <h2>{% trans "Advanced" %}</h2>
+    <details class="details-panel box">
+        <summary>
+            <span class="title is-5" role="headind" aria-level="3">
```
dato:
```suggestion
            <span class="title is-5" role="heading" aria-level="3">
```

bookwyrm/views/admin/redis.py
```diff
@@ -0,0 +1,78 @@
...
+from django.utils.decorators import method_decorator
+from django.views import View
+import redis
+
+from bookwyrm import models, settings
+
+r = redis.from_url(settings.REDIS_ACTIVITY_URL)
+
+
+# pylint: disable= no-self-use
```
dato:
```suggestion
```

bookwyrm/views/admin/redis.py
```diff
@@ -0,0 +1,78 @@
...
+            data[f"{data_key}_deleted"] = deleted_count
+        return TemplateResponse(request, "settings/redis.html", data)
+
+
+def view_data():
+    """Helper function to load basic info for the view"""
+    data = {"errors": [], "prefix": settings.CACHE_KEY_PREFIX}
+    try:
+        data["info"] = r.info()
+    # pylint: disable=broad-except
```
dato:
```suggestion
```

bookwyrm/tests/views/admin/test_redis.py
```diff
@@ -0,0 +1,100 @@
...
+            patch("bookwyrm.activitystreams.populate_stream_task.delay"),
+            patch("bookwyrm.lists_stream.populate_lists_task.delay"),
+        ):
+            cls.local_user = models.User.objects.create_user(
+                "mouse@local.com",
+                "mouse@mouse.mouse",
+                "password",
+                local=True,
+                localname="mouse",
+            )
```
dato:
```suggestion
        cls.local_user = models.User.objects.create_user(
            "mouse@local.com",
            "mouse@mouse.mouse",
            "password",
            local=True,
            localname="mouse",
        )
```

Issue #2898, opened 2023-07-13
Add UI for merging different works into editions of the same work


WesleyAC:
As described in #562, there should be a UI to merge multiple different works into the same work as different editions:

> I think the highest value is letting users manually merge works -- the script does more more quickly, but it's more important that people have editorial agency and the ability to fix the problems that bug them, and there's a lot you can't automate.

Pulling this out into a separate issue to track it more easily.

Probably on the "Editions" page ([example](https://bookwyrm.social/book/231995/editions)) we should have a second button next to "Add another edition" to "Merge existing book as edition", which would let you search for a existing book (or maybe paste a URL would be a nicer UI?) to merge.

I'm not sure how the ID of the previous work should be handled, but presumably the existing merge script handles that somehow?

mouse-reeve:
Right now, the "non-canonical" entry is simply deleted (this does not federate) and all foreign keys associated with it are relocated to the canonical entry

hughrun:
The "proper" way to do this probably would be via an ActivityPub `Move` activity. This would potentially allow for a setting as per #2841 at each instance to control behaviour when a `Move` activity is received for a book - i.e. does it auto-merge, or does it go into a queue to be approved, is it dropped etc.

WesleyAC:
@mouse-reeve Are existing URLs maintained? For instance, commenting on a book has a link to a book page on non-bookwyrm-compatible fedi services, does that go to the correct place after a merge? I guess maybe part of what I'm asking is, are the "book IDs" that I see in URLs IDs of editions, or IDs of works? Or a mix of both?

mouse-reeve:
No, the urls aren't maintained, it just becomes a 404 -- not ideal! The IDs in urls are usually the edition ID, but if you enter the work ID, it will display the page for the default edition

Minnozz:
I would love to contribute to adding this feature. Do you have any suggestions for where to start?

PR #3293, opened 2024-02-22
Track which Author/Work/Edition a duplicate has been merged into (mark)

Minnozz:
Instead of deleting the duplicate, store a reference to the canonical entity.

- [x] Add `merged_into` field to `BookDataModel`
- [x] Move merging code from management command to `BookDataModel`
- [x] Change the delete into a "soft delete" that sets `merged_into`
- [ ] Exclude merged entities from all queries, except when loading by ID
- [ ] Redirect web requests to the canonical entity when a request for the duplicate entity is received
- [ ] Return the correct ActivityPub response for a duplicate entity
- [ ] Verify whether `edition_rank` still picks a good edition after merging Works
- [ ] Find out what happens when a duplicate Edition contains spelling errors

First step towards solving #2898


Minnozz:
Went for a different approach in #3293

PR #3299, opened 2024-03-01
Track which Author/Work/Edition a duplicate has been merged into

Minnozz:
Delete the duplicate, but also store its ID and the one it has been merged into for future reference.

- [x] Add `MergedBookDataModel`
- [x] Move merging code from management command to `BookDataModel`
- [x] Create an instance of `MergedBookDataModel` before deleting a duplicate
- [x] Redirect web requests to the canonical entity when a request for the duplicate entity is received
- [x] Return the correct ActivityPub response for a duplicate entity
  - The data of the canonical object is returned, without any redirecting or `Move` responses
- [x] Verify whether `edition_rank` still picks a good edition after merging Works
  - A suboptimal Edition may be chosen after merging Works, but that is unrelated to this PR
- [x] Add tests

First step towards solving #2898


Minnozz:
Note: this PR currently includes #3313 because that bug surfaced in the tests here, but we may merge #3134 instead of #3313, which also fixes that bug.

PR #3300, opened 2024-03-02
Add timeouts to requests.get

MaggieFero:
Two instances of requests.get are currently missing timeouts. Without an explicit timeout, requests.get will never time out, which can cause code to hang. 

Some instances of requests.get in other parts of the code did have timeouts, in one instance configurable and in all others 15. The missing instances felt more analogous to the place where it was already 15 than the configurable one, so this PR adds timeouts of 15 in the two places that currently lack a timeout.

The phildini/bookwyrm fork is already successfully running with these changes in production, but we've done a bunch of infrastructure stuff recently as well, so I made a new fork to submit a cleaner PR to yall.

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -423,6 +423,7 @@ def get_activitypub_data(url):
                 "Date": now,
                 "Signature": make_signature("get", sender, url, now),
             },
+            timeout=15
         )
     except requests.RequestException:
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Thank you, these should definitely have timeouts! I think we should move _all_ instances of timeouts to the one set by the environment variable, but since that would be a bigger change than this, it seems like a good call to make that a separate task.

PR #3303, opened 2024-03-03
Upgrade Python Version and Several Other Packages for Security

MaggieFero:
Hello! Our fork is now successfully running in production with significant security-relevant package upgrades, and is otherwise mostly equivalent to bookwyrm-social/main (with some small changes that are based on preference for our instance, and that I wouldn't expect to affect upgrade feasibility). 

This PR proposes the same set of changes for the upstream bookwyrm-social repo, now that we're confident they're working in prod. 

The upgrades include:

    Python 3.9->3.11 (and equivalent -bookworm versions for devtools)
    Celery 5.2.7->5.3.1
    django-celery-beat 2.4.0->2.5.0
    django-compressor 4.3.1->4.4
    flower 1.2->2.0
    psycopg2 2.9.5->2.9.7
    pytest 6.1.2->6.2.5
    pylint 2.14.0->2.15.0

We also added the following new pins:

    grpcio version at 1.57.0
    tornado at 6.3.3
    setuptools at 65.5.1

Finally, we alphabetized requirements.txt within each section and removed a duplicate types.requests dependency for developer convenience.

Note that one linter exclusion was added as a result of this Python upgrade; imghdr is deprecated in the 3.11 version of Python. It won't be removed until 3.13 and I didn't see any obvious replacement for now, so I'll file an issue for removing it (and the linter exception) but for now I've set that one instance of the rule to ignore.


MaggieFero:
I had forgotten we also set a second linter exclusion, for 
`bookwyrm/activitypub/base_activity.py:23:0: C0103: Type variable name "TBookWyrmModel" doesn't conform to predefined naming style (invalid-name)`

We didn't name that variable on our fork, and we didn't want to change variable names from upstream, so we added an exclusion. For this repo, I'm adding the same exclusion for now, but feel free to adjust as desired (though that's probably a separate PR).

MaggieFero:
@mouse-reeve This is the PR I had emailed you before submitting. Is there anything else I can do to help make this easier to review and merge? I'm a little concerned that it becomes more likely to cause issues the more other things people write against the Python version this changes.

Minnozz:
I'll fix the conflicts with https://github.com/bookwyrm-social/bookwyrm/pull/3318 once this one is merged.

mouse-reeve:
Ah I apologize @MaggieFero that's all on me; I'm taking a look at merging now

Issue #3295, opened 2024-02-26
Show that link to the book is being moderated


skobkin:
**Is your feature request related to a problem? Please describe.**
When adding a link to the book it's very unclear why the link doesn't shows in the list after being added. I often see federated books from other instances with two or more identical links. Most likely users try to add the link twice or more until they're getting frustrated and stop trying.

I did that myself when I added link first time on bookwyrm.social. Only when I set up my own instance I finally understood how it works.

**Describe the solution you'd like**
We need to show a clear warning that a link is being moderated if it's not going to be shown right away.

prolibre:
I think these duplicates are linked to [a bug](https://github.com/bookwyrm-social/bookwyrm/issues/2988) (when books circulate between instances). But that doesn't mean that the user shouldn't be warned (I didn't know that he didn't have the moderation information).

skobkin:
@prolibre Here's how it works for links to new domains:

1. User adds the link.
2. User do not see anything.
3. Administrator gets a notification about pending link domain review.

![Peek 2024-02-26 19-03](https://github.com/bookwyrm-social/bookwyrm/assets/967576/e08c8865-efcd-4b70-ba82-32e2269559fe)

skobkin:
And guess what happens if users tries to add link again?

![image](https://github.com/bookwyrm-social/bookwyrm/assets/967576/99424854-33d7-473e-9ff0-3e5b6c70a4fd)

MaggieFero:
I know on my instance I've seen a bunch of what looks to be a user being frustrated and not just the bug causing duplicates, this warning sounds very nice!

MaggieFero:
I went to see if I could handle this quickly, and there is a line in the FileLinkForm that I would expect to do this (added by Mouse two years ago, per git blame). 

I confirmed on my own instance by deliberately adding two copies of a link to the last book I read. I never got this warning, and when I approved the domain both appeared. So it seems like this may just be a small bug? 

Line 49 of https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/links.py is the exact warning I'd expect to see here, in case somebody else has a chance to get to it before I do!

skobkin:
#3311 probably fixes accidental duplication. Not sure if it helps to know WHY the link is not being shown yet.

PR #3311, opened 2024-03-09
FileLinkForm: fix duplicate check

Minnozz:
Reported @ https://github.com/bookwyrm-social/bookwyrm/issues/3295#issuecomment-1975020477

Tested locally; adding links and duplicate checking works.

### DIFF
--- a/bookwyrm/forms/links.py
+++ b/bookwyrm/forms/links.py
@@ -1,4 +1,5 @@
 """ using django model forms """
+
 from urllib.parse import urlparse
 
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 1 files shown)


dato (approved):
Thank you for this, the fix looks good to me (tested on my instance as well).

bookwyrm/forms/links.py
```diff
@@ -37,10 +38,9 @@ def clean(self):
                     ),
                 )
         if (
-            not self.instance
-            and models.FileLink.objects.filter(
-                url=url, book=book, filetype=filetype
-            ).exists()
+            models.FileLink.objects.filter(url=url, book=book, filetype=filetype)
```
dato, line 41: Not for this PR, but I would propose dropping _filetype_ from this check. Thoughts?

(If file type is wrong and needs updating, the proposed change would ensure the pre-existing link is deleted first.)

Minnozz: Agreed; I'll change that in a separate PR after this one.

PR #3313, opened 2024-03-11
Fix author_trigger() recalculating search_vector of the wrong books

Minnozz:
The join that was used to match books to a changed author used the id column instead of the author_id column to join on authors.

This migration re-calculates all book search vectors to fix any missed author name changes.

### DIFF
--- a/bookwyrm/migrations/0196_fix_author_trigger.py
+++ b/bookwyrm/migrations/0196_fix_author_trigger.py
@@ -1,0 +1,52 @@
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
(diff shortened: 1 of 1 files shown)


dato:
Great catch!

There's a pending PR by me that introduces migrations for triggers for easier  maintenance (#3134), and fixes this bug too, I believe (if it does, it would be in [`bcb3a34`](https://github.com/bookwyrm-social/bookwyrm/commit/bcb3a343d4514fd9bc2674a918755aaa832c8886): _Fix JOIN in author_search_vector_trigger, add missing WHERE clause_).

Since you already looked at trigger code… I'd be most grateful if at some point you can read, or review, that PR. The more eyes on such a tricky change, the better!

Minnozz:
Closed in favor of #3134

PR #3323, opened 2024-03-18
Docker: make bind mounts of source code read only

Minnozz:
The services don't need write access to the application source, except for the `dev-tools` container.

Make sure the `static` and `images` mount points exist as "empty" directories, since Docker can't create them any more in the read only volume. This should not cause problems for non-Docker setups, since only a `.gitkeep` file is added.

When running `./bw-dev up`, no longer start the `dev-tools` container. It is only used with `docker compose run`, which spins up a new container anyway.


mouse-reeve:
I learned something new about docker! Thank you

PR #3326, opened 2024-03-20
black: specify major version 22 only

Minnozz:
This changes some places where black (the Python formatter) is referenced:
- The GitHub Action version itself is set to `stable` (as recommended)
- The GitHub Action version constraint for black is relaxed to `22.*` (formatting is guaranteed to be stable within a major version)
- The required black major version is set in `pyproject.toml` to avoid formatting with a different major version. This improves consistency between different contributors (eg, me) having different black versions set up in their IDE
- Add black as a dependency so IDEs (like VSCode) can be configured to use black from the local environment


Minnozz:
I'm not well-versed enough with Python development setups to know if the last point (adding black as a dev dependency) is acceptable. Input is welcome!

requirements.txt
```diff
@@ -41,6 +41,7 @@ setuptools>=65.5.1 # Not a direct dependency, pinned to get a security fix
 tornado==6.3.3 # Not a direct dependency, pinned to get a security fix
 
 # Dev
+black==22.*
```
dato, line 44: Would it make sense to update dev-tools/requirements.txt with the same wildcard?

dato (approved):
> I'm not well-versed enough with Python development setups to know if the last point (adding black as a dependency) is acceptable. Input is welcome!

Sounds acceptable to me, we just need to drop it in the merge to the _production_ branch (i.e. when resolving conflicts there).

requirements.txt
Minnozz: Definitely, I forgot that place

Minnozz:
> Sounds acceptable to me, we just need to drop it in the merge to the _production_ branch (i.e. when resolving conflicts there).

Cool! (I wasn't aware a production branch existed until now)

PR #3328, opened 2024-03-21
Escape search query in generated URLs

Minnozz:
Otherwise, a query containing '&' or other special characters results in a broken URL.

As reported on https://matrix.to/#/#bookwyrm:matrix.org


mouse-reeve:
Thank you!!

PR #3332, opened 2024-03-24
Add exports_volume to docker-compose.yml

Minnozz:
Exports should be written to a Docker volume instead of to the bind mount (= source directory). This way they are shared between different containers even when they run on different machines.

After updating to the setup using the exports_volume, the following command can be used to copy existing exports that have ended up in the source directory to the new Docker volume:

```sh
docker-compose cp exports web:/app
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 .env
 /images/
 /static/
+/exports/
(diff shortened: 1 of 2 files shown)


Minnozz:
Added to #3228

PR #3336, opened 2024-03-27
Support AWS_S3_URL_PROTOCOL

Minnozz:
- Allow setting in `.env`
- Default to `PROTOCOL` (same as before)
- Propagate to django-storages so it generates the correct URLs in `sass_src`

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -386,18 +386,30 @@ if USE_S3:
     AWS_S3_ENDPOINT_URL = env("AWS_S3_ENDPOINT_URL", None)
     AWS_DEFAULT_ACL = "public-read"
     AWS_S3_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}
+    AWS_S3_URL_PROTOCOL = env("AWS_S3_URL_PROTOCOL", f"{PROTOCOL}:")
     # S3 Static settings
     STATIC_LOCATION = "static"
-    STATIC_URL = f"{PROTOCOL}://{AWS_S3_CUSTOM_DOMAIN}/{STATIC_LOCATION}/"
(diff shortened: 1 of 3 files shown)


Minnozz:
@bookwyrm-social/code-review this PR is ready for review

bookwyrm/settings.py
```diff
@@ -386,21 +386,32 @@
     AWS_S3_ENDPOINT_URL = env("AWS_S3_ENDPOINT_URL", None)
     AWS_DEFAULT_ACL = "public-read"
     AWS_S3_OBJECT_PARAMETERS = {"CacheControl": "max-age=86400"}
+    AWS_S3_URL_PROTOCOL = env("AWS_S3_URL_PROTOCOL", f"{PROTOCOL}:")
     # S3 Static settings
     STATIC_LOCATION = "static"
-    STATIC_URL = f"{PROTOCOL}://{AWS_S3_CUSTOM_DOMAIN}/{STATIC_LOCATION}/"
+    STATIC_URL = f"{AWS_S3_URL_PROTOCOL}//{AWS_S3_CUSTOM_DOMAIN}/{STATIC_LOCATION}/"
```
dato, line 392: Is it necessary that AWS_S3_URL_PROTOCOL include the colon, in contrast with PROTOCOL?

dato (approved):
I've never used S3, but tried to review this to get things moving.

It isn't clear now, from .env.example, when AWS_S3_URL_PROTOCOL should be set. I tried to find some appropriate wording ("should be set if there's a mismatch with USE_HTTPS"?), but wasn't satisfied.

I only have one real question, below.

bookwyrm/settings.py
dato: Ah! I now see django-storages requires it that way. :+1:

Minnozz:
@dato Thanks for the review! I added a comment in `.env.example` that hopefully clarifies when you should set it.

PR #3342, opened 2024-03-30
[FIX] make sure to get Pillow>=10 compatible pilkit

hbrunn:
On my system, pilkit 2.0 was pulled, which results in

```
     img = img.resize((self.width, self.height), Image.ANTIALIAS)
                                                 ^^^^^^^^^^^^^^^
 AttributeError: module 'PIL.Image' has no attribute 'ANTIALIAS'
```

with the version of Pillow the requirements file pulls.


Minnozz (approved):
Thanks!

This version incompatibility caused issues for more instances

PR #3348, opened 2024-04-03
Define more indexes for slow queries

Minnozz:


bookwyrm/activitystreams.py
```diff
@@ -156,7 +156,7 @@ def get_audience(self, status):
         status_author = models.User.objects.filter(
             is_active=True, local=True, id=status.user.id
         ).values_list("id", flat=True)
-        return list(set(list(audience) + list(status_author)))
+        return list(set(audience) | set(status_author))
```
dato, line 159: Apologies for commenting in an already-merged PR, but if you'd like, ttbomk this construct can also be written as:

```python
return list(set(audience).union(status_author))
```

PR #3350, opened 2024-04-07
Correctly handle serving BookWyrm on custom port

Minnozz:
- [x] Include correct protocol and port in remote IDs
- [x] Store correct protocol and port in BookwyrmConnector
- [x] Construct absolute URLs with correct protocol and port
- [x] Test custom port by running pytest with it


Minnozz:
Hmm, CI itself is broken, unrelated to my change (probably because CI always uses the latest version of eslint instead of the one in `dev-tools`)

mouse-reeve:
Thank you for the housekeeping work along with this change!

PR #3356, opened 2024-04-10
Install same version of eslint in CI as in dev-tools

Minnozz:
I've also commented out the stylelint dependencies since the actual check is commented out too.


dato (approved):
👍🏼

PR #3378, opened 2024-06-09
possible fix for #3372 - user export timeouts

### LINKED ISSUE #3372: Account export times out when trying to generate export json
I've tried creating account exports on b.s, even with accounts that are basically empty, and the `create_export_json_task` is taking so long that it times out and fails. It doesn't get to the point of generating a `create_archive_task`. I suspect that there's an inefficiency with the queries in the various data collection functions, but I'm surprised that it's hitting up against this even with essentially empty accounts.

Tagging you @hughrun if you still have any energy left on this ❤️

hughrun:
### Linters

- [ ] I have checked my code with `black`, `pylint`, and `mypy`, or `./bw-dev formatters`

### Tests

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Description

This definitely needs to be tested on a large DB but I believe it may fix the timeouts b.s. gets when running user exports.

Instead of a gigantic single DB query with heaps of joins, we instead just do a series of simple queries and then use `union()` to pull them into a de-duped queryset.

If I understand the results from `explain()` correctly, this is a massive reduction in DB work:

`Unique  (cost=195899.15..198201.71 rows=11808 width=19220)`

vs

`Unique  (cost=150.28..153.44 rows=16 width=19220)`

- Related Issue #
- Closes #3372

## Documentation

### DIFF
--- a/bookwyrm/models/bookwyrm_export_job.py
+++ b/bookwyrm/models/bookwyrm_export_job.py
@@ -7,7 +7,6 @@ from boto3.session import Session as BotoSession
 from s3_tar import S3Tar
 
 from django.db.models import BooleanField, FileField, JSONField
-from django.db.models import Q
 from django.core.serializers.json import DjangoJSONEncoder
 from django.core.files.base import ContentFile
 from django.core.files.storage import storages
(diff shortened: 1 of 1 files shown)


mouse-reeve:
I tried this with the query time test system in #3380 and it when from 0.06 to 0.003 which is a huge improvement!

Issue #3379, opened 2024-06-09
Merged book code is creating long-running queries


mouse-reeve:
**Describe the bug**
I'm looking at queries that are taking too long on b.s, and I hit https://github.com/bookwyrm-social/bookwyrm/pull/3299/files#diff-d0c61e914db77e19e340b5cf217a89c2032d6288d40deb444d0923a17f52f8d0R51

Just to keep us out of OOM errors, I removed the `Q(absorbed__deleted_id=book_id)` filter and have had much better results. I wouldn't have expected that to much such a big difference! But here we are.

mouse-reeve:
(cc @Minnozz, but no pressure to take this on)

mouse-reeve:
This seems to be resolved

PR #3380, opened 2024-06-09
Adds test class for checking query durations

### LINKED ISSUE #3376: Test for inefficient queries
I’m running into long running database queries that cause worker timeouts when deploying code on b.s. The queries that end up being problematic aren’t obvious to me, and don’t cause problems at a smaller scale so they don’t show up in basic tests.

If possible, I’d like to check the query time in the Django tests. This would probably require creating a pre populated test dataset, which would cause issues for other tests that assume the state of the database. I also don’t know if this would work very well? But right now I keep breaking production whenever I do a release and that isn’t a great situation

mouse-reeve:
### Tests

- [x] Secret other option: I wrote the tests specifically so they would fail

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Description
This is a first pass and I wrote it specifically so that it would fail on the query in test_book.py

- Related Issue #3379 
- Closes #3376

## Documentation


mouse-reeve:
I tried to get fixtures working to include test data, but got stuck trying to generate them on models that inherit with `select_subclasses`

mouse-reeve:
@bookwyrm-social/code-review I think this is a helpful change, but I'd love to get some outside input both on the concept and the execution. Since the query runtime isn't always exactly the same on the same machine, it could cause unexpected test failures, and since the runtime can vary wildly on different machines, tests will likely fail or pass locally differently than they do in CI. 

The duration value that I picked is one that failed prior to the fix in https://github.com/bookwyrm-social/bookwyrm/pull/3378 and passed after it was merged in. 

I made a real effort to get fixtures with serialized test data and gave up trying to wrangle the multi-table inheritance into serializing in a usable way, which is why I just created 10k edition entries in the one test I really wanted to get to fail.

dato:
This is a great initiative, thanks! 💯 

I've been thinking about the challenges that you describe here. I have no code to offer (yet), but for now, regarding this problem:

> the runtime can vary wildly on different machines

For me, this hints at the following: instead of having the tests _raise_ on "perceived slowness" a-la-unit testing, have them just _emit_ timing information. Then, never deal in absolutes, and instead compare the emitted information before and after the change, on the same machine.

I'm not sure what the developer UX would be for this, but a Github Action seems a good start: for every PR, run the test suite as normal. Then, also run, _on the base branch,_ the tests that emit timing information. And compare the two.

That said: this is just an idea. Although I'd like to make time to prepare a mockup, I'm okay if you want to move forward with `raise_long_query_runtime` as well. But if you think this idea makes sense, I can move forward with it as well.

mouse-reeve:
Ah yeah I think that's a great idea! I'd be super happy for you to make a mock-up of that

Issue #3164, opened 2023-12-22
Reading progress bug


prolibre:
A small bug (minor but a little annoying). When you add a comment with a progression (for example 25% or 100 pages) it deletes the date on which you started reading. You therefore have to re-enter it later.

Bookwyrm : 0.6.6 (test with firefox)
Instance : https://bw.heraut.eu

hughrun:
Just came to report this for "finish reading". I definitely have a readthrough with a start date but when the modal pops up only a finish date is displayed.

dato:
@prolibre Does this bug still happen, and can you reliably reproduce it (how?)? I tried, but I failed.

To those reading along, posting progress invalidates the last_readthrough cache entry:

https://github.com/bookwyrm-social/bookwyrm/blob/7f773b3dbddd4d1dc262991bdb97f2f12d39a1f1/bookwyrm/models/readthrough.py#L46

This cache item is recomputed in templates via:

https://github.com/bookwyrm-social/bookwyrm/blob/7f773b3dbddd4d1dc262991bdb97f2f12d39a1f1/bookwyrm/templatetags/shelf_tags.py#L80-L94

So there are two possibilities to begin with (non-exhaustive): in some scenarios the cache entry is not being recreated, thus `start_date` is not available; or, perhaps, some step along the way is setting `is_active=False` in the readthrough, hence it's not retrieved.

prolibre:
@dato Sorry

So I've just noticed this bug again on version 0.7.4 of my instance.

- I started my book on 21 July (I have the progress date indicated)
- I make a comment (just a comment, no quotes) on the book (with a percentage).
- The start date disappears. All that remains is my current progress: today: 39%.

So I have to enter my start date again. Once this has been done, both dates are shown in my progress.

Have a good day.

dato:
@prolibre Thank you!! With your updated description I can now reproduce it.

(For those reading along: I was trying to reproduce the issue using, in the homepage carousel, the shelf selector's "update progress" item but, for this bug at least, the behaviour occurs with a Comment + Progress field. Doing some guesswork, I assume this form doesn't have a "start_date" hidden field, but it makes its way to `edit_readthrough()`, which unconditionally sets start_date from request.POST.)

https://github.com/bookwyrm-social/bookwyrm/blob/f7958bc0f309cd80927b1c5b8c2d429f44261743/bookwyrm/views/status.py#L204-L206

prolibre:
merci @dato !

Issue #3397, opened 2024-07-26
Start date blank when finishing a book


graue:
Recently, when I tag a book I'm reading as finished, the "Started reading" date shows as blank and has to be re-entered. This didn't use to happen.

**To Reproduce**
1. Start reading a book, and set a date you started reading
2. Select the book from the "Currently reading" carousel on the homepage
3. Click "Finish reading"
4. See a blank "Started reading" input box

**Expected behavior**
The date I previously told Bookwyrm that I started the book should be pre-filled.

**Screenshots**
![image](https://github.com/user-attachments/assets/4cf05285-4c4c-40ea-9d6e-271dd5be7c82)

**Instance**
bookwyrm.social

**Desktop (please complete the following information):**
    - OS: Linux
    - Browser: Firefox
    - Version: 128

prolibre:
I remember a bug (https://github.com/bookwyrm-social/bookwyrm/issues/3164), but it concerned books whose progession had changed during reading. For example, did you comment on or quote the book by specifying a page?

graue:
That might be it. This might be a dupe of that bug. I'm not 100% sure that these are all books I quoted/commented on, but it's possible.

hughrun:
Marking this as fixed in #3401 - it's been a year and as far as I can tell it resolved the problem.

PR #3401, opened 2024-07-28
Fix reading progress `start_date` bug

dato:
- 812221456b7ce37f2344f8c2c3da613b1a6378f5 Add failing test case for comment progress clobbering start_date
- 041f2fc41304ccbd8199f57c297a9858cfe9e218 edit_readthrough: set start_date/finish_date iff present in request

-----

- Closes: #3164 
- Related issue (?): #3397
- [x] I have checked my code with `black`, `pylint`, and `mypy`

### DIFF
--- a/bookwyrm/views/status.py
+++ b/bookwyrm/views/status.py
@@ -201,12 +201,11 @@ def edit_readthrough(request):
     # TODO: remove this, it duplicates the code in the ReadThrough view
     readthrough = get_object_or_404(models.ReadThrough, id=request.POST.get("id"))
 
-    readthrough.start_date = load_date_in_user_tz_as_utc(
-        request.POST.get("start_date"), request.user
-    )
-    readthrough.finish_date = load_date_in_user_tz_as_utc(
-        request.POST.get("finish_date"), request.user
-    )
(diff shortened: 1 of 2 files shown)


bookwyrm/views/status.py
```diff
@@ -201,12 +201,11 @@ def edit_readthrough(request):
...
     readthrough = get_object_or_404(models.ReadThrough, id=request.POST.get("id"))
 
-    readthrough.start_date = load_date_in_user_tz_as_utc(
-        request.POST.get("start_date"), request.user
-    )
-    readthrough.finish_date = load_date_in_user_tz_as_utc(
-        request.POST.get("finish_date"), request.user
-    )
+    if start_date := request.POST.get("start_date"):
+        readthrough.start_date = load_date_in_user_tz_as_utc(start_date, request.user)
```
dato, line 205: Another option would be to ensure the Comment form has the ReadThrough's start_date (as with the "Update progress" modal). I don't have a clear idea if it'd be a better approach here.

mouse-reeve: I don't know either, which to me makes your solution seem equally good :)

mouse-reeve (approved):

PR #3419, opened 2024-08-26
Uses workaround for test that fails ~sometimes~

mouse-reeve:
## Description

This test sometimes return with items in the wrong order, so hopefully this will make the order irrelevant and allow the test to pass reliably.

- Related Issue #3135

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/tests/importers/test_importer.py
+++ b/bookwyrm/tests/importers/test_importer.py
@@ -63,30 +63,29 @@ class GenericImporter(TestCase):
         self.assertEqual(import_job.include_reviews, False)
         self.assertEqual(import_job.privacy, "public")
 
-        import_items = models.ImportItem.objects.filter(job=import_job).all()
+        import_items = (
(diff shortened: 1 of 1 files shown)


bookwyrm/tests/importers/test_importer.py
```diff
@@ -65,25 +65,28 @@ def test_create_job(self, *_):
 
         import_items = models.ImportItem.objects.filter(job=import_job).all()
```
hughrun, line 66: This seems to work but it is a bit whacky. Would it work better/as well to just order the queryset and keep the existing asserts?

```py
import_items = models.ImportItem.objects.filter(job=import_job).all().order_by("id")
``

mouse-reeve: oh yeah that is way simpler!

hughrun (approved):

PR #3420, opened 2024-08-26
Suggestion list

mouse-reeve:
## Description

This PR is meant to complete @joachimesque's work in #2560 -- it lets users suggest "if you like that, try this" suggestions for books using the existing lists functionality.

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

I'm hoping this won't be breaking per se, but the federation of these lists is going to be a change and that will impact how instances that have the feature interact with those that don't. But I haven't written that part yet so tbd.

## Documentation

### Tests

### Screenshots
This is what the suggestions panel looks like:

Screenshot 2026-05-23 at 9 29 00 AM


joachimesque:
🎉  I’ve just seen the PR notification so I haven’t read much, I’ll have to reacclimate myself to the codebase, but I’m very happy that my contribution is useful! I’ll check it out in details when I have more time :)

joachimesque:
From my first read-through your changes look great, but I haven’t had the time/infra to test the branch :)

mouse-reeve:
It turned out to be so simple! The one thing I'm dithering on that you might have thoughts about is whether recommendations should be associated with an `Edition` or with a `Work`. 

The big downside to using `Edition` is that books with multiple, functionally interchangeable editions (like books with a paperback, hardcover, alternative cover, et cetera) will have separate recommendations. I think that's a pretty big negative, and would make the experience for users who happen to look at the paperback edition one day and the hardcover edition the next pretty weird.

But if it's associated with `Work`, all the recommendation lists for books that have editions in various languages will get inundated with English-language recommendations, and it will make the experience worse for non-English-language readers. 

A third option would be to keep the list with `Edition` and display recommendations for all `Edition`s under the parent `Work` -- this is how reviews are displayed. That means it would be technically possible to extend or modify the UI to do something like prioritize suggestions with the same language, or give users the option to filter recommendations to that edition. At this point the experience would be the same as if it was associated with a `Work`, so it doesn't directly solve any of those problems, but it does give a little more flexibility in the future.

@joachimesque Do you have any opinions or alternative ideas?

mouse-reeve:
The more I think about my third suggestion, the harder it sounds to implement, and I'm not sure it's actually feasible

joachimesque:
(reading that after a night out, so I might not be as sprightly as I could)

I don't know if I understood the third option—it sounded complicated even before I read that it could be _very_ complicated.

Using `Work` could be the best solution, I hadn’t thought about the restrictions around `Edition` when I wrote the basis of this PR. My point of view is: even if the suggestions are in another language, I would welcome it. BUT, I’m saying that as a French guy who reads a LOT of English books, so I understand my experience is not most people’s experience.

If `Work` is used, would it be possible to select the `Edition` that’s in the language of the person/interface (if it exists)? (Or is this proposal it the third option?)

mouse-reeve:
I had a similar thought, and yes it would be possible to filter or prioritize editions based on language. Outside the scope of this PR, but it gives me some confidence that the potential downsides to using Work could be mitigated in the future.

mouse-reeve:
@bookwyrm-social/code-review This is ready for testing

hughrun (commented):
I've had a quick look at this and have some initial thoughts:

1. This is cool! I love the concept, especially that this is real human recommendations rather than an algorithm.
2. I'm a bit confused by the test in `test_inbox_add.py`. Are followers supposed to get some kind of notification when someone adds an item to a suggestion list? If so, this doesn't seem to work.
3. Is there a way to see all of my own suggestion lists? This doesn't seem obvious and it took me a while to realise how to get to my own suggestions in order to remove one
4. Could we consider adding a third tab to `/list` so that it's "All lists | Saved lists | Suggestion lists"?
5. We should definitely update the guided tour once this is implemented – I figured out how to "endorse" a recommendation but it's not necessarily obvious.
6. Consider adding a button to add each suggested book to your `To Read` shelf immediately
7. If I'm understanding correctly, we're suggesting `Editions` for a given `Work`. This seems surprising - this basically means if I'm looking at an edition I'll get a bunch of editions recommended based on the parent work? I think `Work` is the right level, so I'd expect to see recommendations of `Work` for another `Work` – am I missing something?

mouse-reeve:
>     2. I'm a bit confused by the test in `test_inbox_add.py`. Are followers supposed to get some kind of notification when someone adds an item to a suggestion list? If so, this doesn't seem to work.

That is testing recommendations federating between instances -- so if a recommendation is added to a book on one instance, it would also appear as a recommendation on the linked instances. Right now there are no notification related to any suggestions. It might make sense to notify people when their suggestions are endorsed? And being able to follow a suggestion list would be cool but out of scope for this.

- [x] It also occurs to me that "Create suggestion list" button implies that the user who clicks it has ownership over the suggestion list, which isn't the case. I think I'll re-work that UI so that creating the list happens in the background.

>     3. Is there a way to see all of my own suggestion lists? This doesn't seem obvious and it took me a while to realise how to get to my own suggestions in order to remove one

- [x] There isn't, and this is a good idea. Right now (as I'm sure you figured out) you have to click through to "View all" to see the list page version of the suggestion list, where the delete button is shown, and you would have to do that for every book you've made a recommendation on. A page that aggregates all the suggestions you made on any book would be very sensible.

>     5. We should definitely update the guided tour once this is implemented – I figured out how to "endorse" a recommendation but it's not necessarily obvious.

- [ ] 👍 

>     6. Consider adding a button to add each suggested book to your `To Read` shelf immediately

- [x] Love this

>     7. If I'm understanding correctly, we're suggesting `Editions` for a given `Work`. This seems surprising - this basically means if I'm looking at an edition I'll get a bunch of editions recommended based on the parent work? I think `Work` is the right level, so I'd expect to see recommendations of `Work` for another `Work` – am I missing something?
This is a weird one and I'm not 100% sure it's the right design choice, but I'll walk you through how I got there.

I used `work` so that, for example, a paperback edition vs a hardcover one don't have separate suggestion lists. The main downside I foresee to that is that different language editions may have suggestions for other language editions.

I went with `edition` for the suggestion because it gives the user who makes the suggestion more control of what they are recommending (you can choose a translation, for example, or the cover art you prefer), and it works with the existing List code, which allows it to be a subclass of the List model, and also use the same templates. 

I feel pretty confident that it doesn't make sense to use Edition for both, but there would be advantages to using Work for both (the main one I can see is that you could show a user an edition in a particular language (in theory, in practice we're not there and there're barriers to knowing what language a book is in))

hughrun:
Makes sense.

My 2c is that it should match `Work` to `Work`. My thinking:

1. It would make possible future discovery options where we can link together chains of recommendations (A is recommended for people who like B; C is recommended for people who like A; therefore C might also be a good match for people who like B) - we _could_ do this at Edition level but it would be cleaner and stronger to do it at Work level
2. It would prevent us having recommendation lists with e.g. 24 different editions of LotR
3. Whilst it's complex, we need to do the work of making languages more tightly structured/controlled anyway. In future we would ideally be able to default to the user's chosen language if an Edition exists in that language in all situations, including recommendation lists
4. If I'm recommending a book to people who like some other book, I'm not convinced that the format (paperback, audiobook) is all that important - it's the content I'm recommending

joachimesque:
(I haven’t followed this PR much because of life stuff, but I love you all for keeping working on it, it means a lot and I’m happy to see the little spark I sent to the world being made into real fire by you all, thanks)

bookwyrm/models/list.py
```diff
@@ -24,15 +26,107 @@
...
+    )
+    activity_serializer = activitypub.SuggestionList
+
+    @property
+    def collection_queryset(self):
+        """list of books for this shelf, overrides OrderedCollectionMixin"""
+        return self.works.order_by("suggestionlistitem")
+
+    @property
+    def editions(self):
```
mouse-reeve, line 83: I need a sanity check on this property and https://github.com/bookwyrm-social/bookwyrm/pull/3420/changes#diff-c12ba291e28f888cebd8acb81af9281bdbd19e15f39c445e3b087445f7a4d1b0R141 -- are these queries going to be problems, performance-wise?

mouse-reeve:
@hughrun The only item I haven't addressed yet from above is the guided tour, but it also occurred to me while I was working on the tests for this that it will also need to be added to the user export/import flow. I'm going to take a stab at that, but in the mean time, can you take a look at the PR as it stands? 

I changed how both the `List` and `SuggestionList` models work, so that rather than items having a generic `book` field, they have either `edition` (for List) or `work` (for SuggestionList) and then each supply the one that isn't stored as a `@property`. That way even though they have different schema, they can be called in similar ways. For example, you can display a list item using the same block of markup, regardless of whether the model uses a work (in which case it displays the `default_edition` for that work), or the edition itself. 

This change has wide-ranging side effects because it effects how all lists are used, but in most cases it should be a one to one swap of "book" for "edition."

bookwyrm/models/list.py
```diff
@@ -43,30 +137,25 @@ class List(OrderedCollectionMixin, BookWyrmModel):
...
-        return f"{BASE_URL}/list/{self.id}"
+    class Meta:
+        """default sorting"""
+
+        indexes = [Index(fields=["privacy", "-updated_date"])]
+
+    @property
+    def works(self):
+        work_model = apps.get_model("bookwyrm", "Work")
+        return work_model.objects.filter(editions__in=self.editions.all()).distinct()
```
hughrun, line 151: Would it be possible to dedupe the work IDs before calling the DB query, and would that maybe make this slightly more efficient? That is, we already have the Editions, and they all have a Work, so we have the Work IDs already. i.e. something like this:

```py
works = self.editions.all().values_list("work", flat=True).order_by("work").distinct()
work_model = apps.get_model("bookwyrm", "Work")
return work_model.objects.filter(id__in=works)
```

mouse-reeve: It looks like `self.editions.all().values_list("parent_work", flat=True)` produces a queryset so it's even simpler than that

mouse-reeve:
This is how a suggestion looks if it was made by a deleted or suspended user:
Screenshot 2026-05-22 at 10 08 05 AM

bookwyrm/forms/lists.py
```diff
@@ -17,7 +17,19 @@ class Meta:
 class ListItemForm(CustomForm):
     class Meta:
         model = models.ListItem
-        fields = ["user", "book", "book_list", "notes"]
+        fields = ["user", "edition", "book_list", "notes"]
```
hughrun: Need to update `edit_item_form.html` to match this. It still has:

```html
    <input type="hidden" name="book" value="{{ book.id }}">
```
...which throws an error when you try to add a note to a book list item.

bookwyrm/views/list/list.py
```diff
@@ -100,24 +104,32 @@ def post(self, request, list_id):
         return redirect_to_referer(request, book_list.local_path)
 
 
-def get_list_suggestions(book_list, user, query=None, num_suggestions=5):
+def get_list_suggestions(
+    book_list, user, query=None, num_suggestions=5, ignore_book=None
+):
     """What books might a user want to add to a list"""
```
hughrun, line 114: This is merely a suggestion you can ignore (especially if you think the overhead isn't worth it), but I'm wondering about a few things:

1. Should we exclude things on the `stopped-reading` and `to-read` shelves? By definition these are likely to be things the user isn't really into, or has not yet read and so isn't in a position to recommend
2. Conversely, it seems to me if a user has already added something to another list, it might be something they're likely to add to a Suggestion List and, perhaps, another book list. Should that be filtered in as well, maybe as a first fallback?
3. Can we follow suggestions backwards to get likely candidates? i.e. if I am looking at Book A, and I already suggested Book A for Book B, then should Book B come up in my suggested suggestions?
4. Maybe also push books in the same series to the front of the list too?

bookwyrm/views/books/books.py
```diff
@@ -140,6 +143,23 @@ def get(self, request, book_id, **kwargs):
...
+                    .annotate(endorsement_count=Count("endorsement"))
+                    .order_by("-endorsement_count")[:3]
+                )
+
+                data["suggested_books"] = get_list_suggestions(
+                    data["suggestion_list"],
+                    request.user,
+                    query=request.GET.get("suggestion_query", ""),
+                    ignore_book=book.parent_work,
+                )
```
hughrun, line 168: The default of 5 suggestions looks a bit weird because they line up in pairs on a standard sized screen, so there's only one book on the last line. I suggest for Suggestion lists we set `num_suggestions` to 6.

bookwyrm/templates/user/layout.html
```diff
@@ -78,29 +78,41 @@ <h2>{% trans "Follow Requests" %}</h2>
...
             {% endif %}
+
             {% if is_self or user.list_set.exists %}
             {% url 'user-lists' user|username as url %}
             <li{% if url in request.path %} class="is-active"{% endif %} id="tour-lists-tab">
                 <a href="{{ url }}">{% trans "Lists" %}</a>
             </li>
             {% endif %}
+
+            {% if is_self %}
```
hughrun: Do we really want to restrict this to only seeing your own suggestions? It would be really useful to be able to see someone's suggestions from their profile. Given you can see suggestions on the relevant book publicly, is there a privacy concern here? (and if there is, should suggestions have privacy levels?)

bookwyrm/models/user.py
hughrun, line 129: Saving a Suggestion List doesn't work because `save_list()` expects a `List`, and `saved_lists` here also expect a `List`. Since `AbstractList` is indeed abstract, either we need to add an equivalent to all these for `SuggestionList`, or maybe use multi-table inheritance instead - though that might cause its own problems.

bookwyrm/templates/book/suggestion_list/list.html
```diff
@@ -0,0 +1,54 @@
...
+    </a>
+    {% endif %}
+</h2>
+
+<p class="subtitle">
+    {% blocktrans trimmed with title=book.title %}
+    Readers who liked <em>{{ title }}</em> recommend giving these books a try:
+    {% endblocktrans %}
+</p>
+
```
hughrun: I suggest putting the subtitle inside `{% if suggestion_list %}` - if there are no suggestions, it will then just have the heading "Suggestions" and invite the user to add one.

hughrun (changes requested):
This is awesome!

Only a few little issues I can see so far, plus some suggestions and questions.

bookwyrm/views/list/list.py
mouse-reeve: I like these ideas but I think they're out of scope

bookwyrm/models/user.py
mouse-reeve: I'm just going to remove that button. We already aggregate suggestions on the user page, and have ways to save books, so I don't think its worth the effort to add a save option analogous to what List has

bookwyrm/templates/user/layout.html
mouse-reeve: The technical problem of privacy levels on list items seems... bad to me. I'm down to make this view public, and I think I tend to err on the side of being too private about user book interactions, but I suspect that granular privacy controls would be prohibitive. What do you think?

hughrun: I think we need a whole separate piece of work around "private books" something like the inverse of blocked books.

I also think that's not in scope here.

As a default I would expect my brilliant recommendations for related reading to be public 😀

mouse-reeve: Updated.

pilotMike:
From what I see, it looks like adding a suggested book to another book creates or adds to a list for that book. Are the same operations done in reverse? Meaning, if I suggest checking out "Star Trek Wars: The Fandom Menace" to "Star Trek", I will then see "Fandom Menace" under Star Trek, but will going to "Fandom Menace" show "Star Trek",

ie: Adding B to A shows B under A, but does it also show A under B?

I might be misreading, it looks like the db structure is One to Many of Book -> Suggestion List -> ...Item? In this case it should be fine, just that the reverse direction would be a new list?

I think that it makes sense to have a suggested book apply both ways, if it isn't already.

Edit: I think that other places where I've seen something similar, such as MyAnimeList likely use Many to Many relationships and like check both directions on the FROM and TO for seeing the suggestions.

mouse-reeve:
The suggestions are indeed one way. I did this intentionally because it's my preference -- for example, I can think of books that I'd recommend because I think they're better than the original book, or because they are based on or heavily reference the original book. But I can see that this would be a hassle for recommendations in the spirit of "both these books will hit the spot if you're looking for <x>", and that might be a more common type of recommendation...

pilotMike:
Oh, I see, that makes sense. Would it also be a good idea to include an option or a check box to also add the reverse, then? That could keep the spirit of what you were going for while also making it convenient by not having to add twice if that's what the user recommends?

mouse-reeve:
I think that's a good idea @pilotMike, but out of scope for this PR. It would be a good addition and I don't think it requires a fundamental restructure of what's already here, so I'd like to merge this as is and add the option to make recommendations bi-directional in the (ideally near-)future. Thank you for thinking it through with me, and for pointing this out!

mouse-reeve:
@hughrun I think this is ready for review -- #3889 and #3868 were pretty substantial merges in terms of code overlap so another set of eyes on that aspect in particular would be much appreciated

hughrun:
Unsurprisingly there are a few small problems with blocked books. I'll send through a PR on the `suggestion-list` branch soon, I'm just checking a few more things.

mouse-reeve:
is this..... ready???

joachimesque:
> is this..... ready???

!!!!!!!!!!!!!!!!!

hughrun:
@mouse-reeve this will now need `PrivateProfileMixin` for book suggestions on user profiles.

mouse-reeve:
hoisted by my own petard.

joachimesque:
> hoisted by my own petard.

!!!!!!!!!!!!!!!!! (with sadness)

mouse-reeve:
OKAY how about now 😅

hughrun (approved):
LGTM!

mouse-reeve:
unreal. thank you so much @joachimesque and @hughrun for making this actually happen

codeinabox:
Should book suggestions be bidirectional? I added one for [Not Working](https://ramblingreaders.org/book/356362/s/not-working) to [How to Do Nothing](https://ramblingreaders.org/book/397796/s/how-to-do-nothing), however it doesn't show the suggestion in reverse.

mouse-reeve:
I think it should at least be an option! This came up earlier and we decided it was out of scope for this PR. But I'd like to have it as a feature -- can you open an issue requesting it?

Issue #2841, opened 2023-05-07
Add options for federating book data


Flameborn:
I often find that users of other instances edit existing book data to match what they've read, e.g. change a paperback book to an audio book, rather than creating a new edition. It can also happen that Bookwyrm does not detect that a book is an existing edition, and an author's page has multiple versions of the same book, rather than one version with multiple editions.

As a result, my yearly book reads lists books with missing page numbers, unfilled series name or ISBN, etc, even though I have added this data when creating the book.

To avoid situations like these, it would be great if instance admins had control over whether they accept book data federation from other instances. I see having a *enable/disable* and perhaps *approve only* options to do this.

PR #3421, opened 2024-08-27
Allow an instance to disable federation

mouse-reeve:
## Description

This allows an instance to turn off all contact with other ActivityPub services. The instance would still be able to connect to OpenLibrary and Inventaire. If there's existing federated data, there can be some odd behavior (like blocking an existing remote user will "work" but not propagate the block to the other server, and following that user would fail silently), but since I expect this to either be used from the outset or as a stopgap for some unforeseen issue, I didn't bother to clarify those situations in the UI (besides putting up a banner on the user page). 

- Related Issue #
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


Flameborn:
Thank you for implementing this. If federation also includes book edits from remote instances, this might be a partial solution for https://github.com/bookwyrm-social/bookwyrm/issues/2841

mouse-reeve:
It does prevent book data federation, but it’s not granular at all

mouse-reeve:
@bookwyrm-social/code-review It would be great to get a check-over on this one

hughrun (commented):
This seems to work as advertised.

I take your point about following of external users probably being an unusual case, but it feels like something that might be reasonably easily fixable. e.g. you could just add this to `snippets/follow_button.html`:

```django
{% if request.user == user or not request.user.is_authenticated %}
{# nothing to see here -- either it's yourself or your logged out #}
{% elif site.disable_federation and not user.local %}
{# can't follow remote users without federation #}
{% else %}
```

In line with your comment about following external users silently failing, I also noticed that the `get-started` flow still includes external users when federation is turned off. This doesn't provide any indication that following will fail, so it's not ideal.

We could fix this fairly easily by adjusting the signature for `get_suggestions`, something like this but ideally a little less verbose:

```py
def get_suggestions(self, user, local=models.SiteSettings.get().disable_federation):
```

We only have the `local` arg because I needed it to be `False` for groups (which still don't federate 🙁) - so aligning this with the value of `disable_federation` makes sense to me.

PR #3425, opened 2024-08-29
Make average rating property of Edition model

### LINKED ISSUE #3424: Make average rating a property of the Edition model
### Description

Currently when displaying the ratings of a given book, we calculate the average rating of the book from within the book view as shown [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/books/books.py#L101)

It would make more sense for us to have an `average_rating` function which is a `@property` function on the `Edition` model. 

Then `average_rating` would return the query (currently `....aggregate(Avg("rating"))["rating__avg"]`) which produces the average rating of the book.

lenikadali:
## Description
Makes the query that returns the average rating of a book a `@property` function of the `Edition` model and changes the view code to use this instead.

- Closes #3424

- [ ] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -494,8 +494,9 @@ class Edition(Book):
 
     @property
     def average_rating(self, user):
+        """generate average rating of a book"""
         reviews = Review.privacy_filter(user).filter(book__parent_work__editions=self)
-        reviews.aggregate(Avg("rating"))["rating__avg"]
+        return reviews.aggregate(Avg("rating"))["rating__avg"]
 
     def get_rank(self):
         """calculate how complete the data is on this edition"""
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -9,7 +9,7 @@ from django.contrib.postgres.search import SearchVectorField
 from django.contrib.postgres.indexes import GinIndex
 from django.core.cache import cache
 from django.db import models, transaction
-from django.db.models import Avg, Prefetch, ManyToManyField
+from django.db.models import Avg, Prefetch, ManyToManyField, Q
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
 from model_utils import FieldTracker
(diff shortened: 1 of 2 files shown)


bookwyrm/models/book.py
```diff
@@ -491,6 +492,12 @@ def hyphenated_isbn13(self):
         """generate the hyphenated version of the ISBN-13"""
         return hyphenator.hyphenate(self.isbn_13)
 
+    @property
+    def average_rating(self, user):
+        """generate average rating of a book"""
+        reviews = Review.privacy_filter(user).filter(book__parent_work__editions=self)
+        return reviews.aggregate(Avg("rating"))["rating__avg"]
```
lenikadali, line 499: `@property` functions cannot take arguments as mentioned [here](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/property-with-parameters.html)

However the [original implementation](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/books/books.py#L101) is on run a `reviews` value that has the class method `privacy_filter` run on it _**before**_ the rating calculation is done as shown [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/books/books.py#L68-L70)

Should we remove the `privacy_filter` call or switch to making this a function on the `Edition` model or a `classmethod`?

lenikadali: Also, other calculations of the `rating` use the `user` argument, strengthening the idea that this should be either a method on the model or a class method. Examples are [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/annual_summary.py#L74-L79), and [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templatetags/rating_tags.py#L19-L23). The latter also looks for ratings greater than 0.

The calculation [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/preview_images.py#L455-L459) doesn't use the `user` argument.

Another idea could be to have different methods that keep the behaviour of the methods mentioned but have them be on the `Edition` model rather than in the different places they are in right now.

mouse-reeve: Using a `classmethod` seems like a solid choice to me

bookwyrm/models/book.py
```diff
@@ -492,10 +492,34 @@ def hyphenated_isbn13(self):
...
+        else:
+            base_reviews = Review.objects.filter(Q(deleted=False) & Q(rating_gt=0))
+        if user:
+            reviews = base_reviews.privacy_filter(user).exclude(rating=None)
+        elif privacy:
+            reviews = base_reviews.filter(Q(privacy=privacy)).exclude(rating=None)
+        else:
+            reviews = base_reviews.exclude(rating=None)
+
         return reviews.aggregate(Avg("rating"))["rating__avg"]
```
lenikadali, line 524: > I think using a method that takes a `Review` queryset (which can already be filtered with things like `privacy_filter` or `privacy="public"`) and returns an average rating would be a good solution.

The filter calls here have overlap with the initial filter in the `Book` view [here](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/views/books/books.py#L68-L70)

At this stage of the refactor, I am okay with leaving the `Book` view filter as is and then allowing a more elegant solution to come out of our iterations on this.

@mouse-reeve what do you think?

PR #3431, opened 2024-09-08
User import and export improvements

### LINKED ISSUE #3381: Account import fails with error: 'attributedTo'
I'm trying to import an account, but the import fails. Only 60 out of ~500 books are imported and see this in the logs:
```
celery_worker-1   | [2024-06-09 11:57:38,222: ERROR/ForkPoolWorker-4] User Import Job 4 Failed with error: 'attributedTo'
celery_worker-1   | Traceback (most recent call last):
celery_worker-1   |   File "/app/bookwyrm/models/bookwyrm_import_job.py", line 63, in start_import_task
celery_worker-1   |     process_books(job, tar)
celery_worker-1   |   File "/app/bookwyrm/models/bookwyrm_import_job.py", line 92, in process_books
celery_worker-1   |     upsert_statuses(
celery_worker-1   |   File "/app/bookwyrm/models/bookwyrm_import_job.py", line 185, in upsert_statuses
celery_worker-1   |     user, status["attributedTo"]
celery_worker-1   |           ~~~~~~^^^^^^^^^^^^^^^^
celery_worker-1   | KeyError: 'attributedTo'
``` 

I'm attaching archive.json from the export tarball
[archive.json](https://github.com/user-attachments/files/15753908/archive.json)

Any idea what might be the reason?

hughrun:
## Description

This fixes a number of bugs in User imports and exports, and improves the UX.

* User import makes each relationship, book or status a subtask of the import job
* Failures of any of the above now do not make the whole job fail
* improved troubleshooting since import items are separated
* Tombstones no longer cause an import to fail (#3381)
* Users can now see average recent import and export times
* Users can now see status of imported books, relationships (follows/blocks) and statuses as they are imported
* Fixed a bug where authors were not added to works, likely leading to import failures
* Expanded test suite

- Related Issue #
- Closes #3381

- [x] Bug Fix
- [x] Enhancement
- [x] Refactor

## Documentation

### Tests


hughrun:
@bookwyrm-social/code-review Some eyes on this would be good, there's a lot of new code.
Also any tips on the template error would be good. I see I've missed a closing div tag, is that likely to be it, or have I missed something else?

mouse-reeve:
In the course of testing I found an unrelated error in how the status was being composed, so I went in and changed that

bookwyrm/models/bookwyrm_import_job.py
```diff
@@ -22,23 +33,133 @@ class BookwyrmImportJob(ParentJob):
...
+        return math.floor((item_count - self.pending_item_count) / item_count * 100)
+
+
+class UserImportBook(ChildJob):
+    """ChildJob to import each book.
+    Equivalent to ImportItem when importing a csv file of books"""
+
+    book = ForeignKey(models.Book, on_delete=SET_NULL, null=True, blank=True)
+    book_data = JSONField(null=False)
+    fail_reason = TextField(null=True)
```
mouse-reeve, line 98: Right now, the fail reason is never showing up in the UI, even though it is being set correctly. That's because these classes inherit from a non-abstract class (ChildJob), so when they're called as that super-class, as they are in the `user_troubleshoot` view, they don't have the attributes that only exist in the child classes accessible. 

There are two ways to fix this; one is to make `fail_reason` a property of the parent class, `ChildJob` (I suspect this is the right solution for fail reason). If there's a reason not to, or for a field that shouldn't be present on all the child classes (like `status_type`), you have to make the parent model use the inheritance manager and then call `select_subclasses()` when loading all the children in the view. This is a workaround for what you're doing when you write `item.userimportpost`, for example. 

I know that's confusing, but all you have to do is add to `bookwyrm/models/job.py`:
```python3
from model_utils.managers import InheritanceManager
...

 class ChildJob(Job):
    ...
    objects = InheritanceManager() 
```

And then call `items = job.child_jobs.all().select_subclasses().order_by("task_id").filter(status="failed")` on line 27 of bookwyrm/views/imports/user_troubleshoot.py

Once you do that, you don't have to use `item.userimportpost.book`, you can just do `item.book` in the template, and fail reason will start working correctly. 

(Sorry does that make any sense?)

bookwyrm/templates/import/user_troubleshoot.html
```diff
@@ -0,0 +1,91 @@
...
+                {% trans "Relationship" %}
+            </th>
+            <th>
+                {% trans "Reason" %}
+            </th>
+        </tr>
+        {% for item in items %}
+        <tr>
+            <td {% if item.userimportpost.json.name %}class="is-italic"{% endif %}>{{ item.userimportpost.book.title }}</td>
+            <td>{{ item.userimportpost.json.name }}</td>
```
mouse-reeve, line 74: Since `name` is only present for reviews, maybe `type` here?

bookwyrm/templates/import/user_troubleshoot.html
```diff
@@ -0,0 +1,91 @@
...
+            <th>
+                {% trans "Relationship" %}
+            </th>
+            <th>
+                {% trans "Reason" %}
+            </th>
+        </tr>
+        {% for item in items %}
+        <tr>
+            <td {% if item.userimportpost.json.name %}class="is-italic"{% endif %}>{{ item.userimportpost.book.title }}</td>
```
mouse-reeve, line 73: I think this can be italic in all situations, since it either displays the title or nothing

mouse-reeve (changes requested):
These are all comparatively minor UI changes, but as far as I tested this seemed to work very well!

bookwyrm/templates/import/user_troubleshoot.html
hughrun: Oops, I think this was some copypasta

mouse-reeve:
Thank you! What a relief to have this merged.

PR #3432, opened 2024-09-08
add test for resolving users with aliases

### LINKED ISSUE #3426: `activitypub.resolve_remote_id` throws `IntegrityError` if remote user has unknown alias
**Describe the bug**
Resolving remote users who have unknown users set as an alias may fail.

**To Reproduce**
1. call `activitypub.resolve_remote_id` to retrieve details of remote user A and add them to the local DB
2. remote user A has set remote user B as an alias in their `also_known_as` list
3. remote user B is not known to the local server, and therefore their ID cannot be resolved
4. error: `null value in column "to_user_id" of relation "bookwyrm_user_also_known_as" violates not-null constraint`

**Expected behavior**
`activitypub.resolve_remote_id` should deal with unknown alias users, probably by recursively calling `activitypub.resolve_remote_id` on every alias

**Screenshots**

**Instance**
Development instance

**Additional context**
I stumbled across this when I was doing some work on improving user imports. So caveat on this: I _think_ there is a broader bug, but it's possible this is only relevant for user imports.

Thinking about what is actually happening here, it's possible this is actually even worse because if the user ID does exist in the local DB, would it just set a totally different user as the alias? Need to do some more testing.

---

    - OS: [e.g. iOS]: Docker on Mac
    - Browser [e.g. chrome, safari]: Firefox
    - Version [e.g. 22]: `main` branch

hughrun:
## Description

Adds a test to ensure resolving remote IDs when a user has an `alsoKnownAs` entry is resolved properly.

In #3426 I thought there was a problem, but this test shows that there is not. So I think what I was seeing is when the remote id can't be reached.

- Related Issue #
Closes #3426 

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/tests/activitypub/test_base_activity.py
+++ b/bookwyrm/tests/activitypub/test_base_activity.py
@@ -46,6 +46,13 @@ class BaseActivity(TestCase):
         # don't try to load the user icon
         del self.userdata["icon"]
 
+        alias_datafile = pathlib.Path(__file__).parent.joinpath(
+            "../data/ap_user_aliased.json"
+        )
+        self.alias_userdata = json.loads(alias_datafile.read_bytes())
+        # don't try to load the user icon
+        del self.alias_userdata["icon"]
+
         image_path = pathlib.Path(__file__).parent.joinpath(
             "../../static/images/default_avi.jpg"
         )
@@ -110,6 +117,13 @@ class BaseActivity(TestCase):
             status=200,
         )
 
+        responses.add(
+            responses.GET,
+            "https://example.com/user/moose",
+            json=self.alias_userdata,
+            status=200,
+        )
+
         with patch("bookwyrm.models.user.set_remote_server.delay"):
             result = resolve_remote_id(
                 "https://example.com/user/mouse", model=models.User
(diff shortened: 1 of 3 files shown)


hughrun:
LOL well I guess all the failing tests kind of prove the point.

Will fix this in a bit.

mouse-reeve:
This is failing in main (it looks like the test is correctly finding that the alias is created incorrectly). I guess what I'm wondering is why it passes here?

hughrun:
Hmm, I suspect this might be because we're running tests concurrently (`pytest -n 3`), and I guess which ones happen to still be running together is slightly different between the two branches?

Within `test_base_activity` we do indeed create 4 users, because we test to see if the default server user is created in one test, but with sequential tests that wouldn't be a problem because each test is supposed to be isolated.

@mouse-reeve using the `--dist=loadfile` flag and value in our pytest runner might help: it makes sure tests within each test file are all assigned to the same worker.

hughrun:
Actually thinking about this more, it's probably because `main` includes the change I made so that all requests are signed, which will force the creation of the server representative user?

I'll see if I can work this out locally and push a fix if possible.

mouse-reeve:
Yes I think that’s it - I was able to recreate the test failure locally running single threaded and I introspected on the user list and iirc the instance user was there so it all checks out

PR #3437, opened 2024-09-14
Fix post dates being inconsistent

### LINKED ISSUE #3365: Use consistent timestamp style
Bookwyrm inconsistently uses relative dates ("10 months ago") and exact dates ("Jul 4") in the same interface. At a glance, I couldn't tell what order these comments were in. I had to do mental math to know which comment was posted first. Also, hovering the cursor didn't show the full date, which would have helped.

(Now I see that one of them says "started reading" and the other says "finished reading," but that escaped my notice before.)

image

**The solution I'd like**
Consistently use relative or absolute dates within an area of the UI, and if a date shown isn't the full date, allow hovering (or tapping on mobile) to see the full date.

**Alternatives I've considered**
While not my preference, the display could be kept as is but allow hovering/tapping to see the full date.

hughrun:
Makes all dates fixed except if in the last day, in which case they are relative times.

Prior to this if a post was published on the same day of the month as today, the date was shown as relative (e.g. "10 months ago").

## Description

- Related Issue #
- Closes #3365

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/templatetags/status_display.py
+++ b/bookwyrm/templatetags/status_display.py
@@ -1,6 +1,7 @@
 """ template filters """
 from dateutil.relativedelta import relativedelta
 from django import template
+from django.conf import settings
 from django.contrib.humanize.templatetags.humanize import naturaltime, naturalday
 from django.template.loader import select_template
 from django.utils import timezone
@@ -61,7 +62,7 @@ def get_published_date(date):
     if delta.years:
         return naturalday(date)
     if delta.days or delta.months:
-        return naturalday(date, "M j")
+        return naturalday(date, settings.MONTH_DAY_FORMAT)
     return naturaltime(date)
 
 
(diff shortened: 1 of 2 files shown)


dato (approved):
Thanks for investigating! The fix looks good to me.

We should perhaps consider changing `"M j"` to `"MONTH_DAY_FORMAT"`?, since the former is not localized. (Not necessarily in this PR!)

(Though at the same time, the change would worsen the English localisation, since abbreviated months sound appropriate for status timestamps. `MONTH_DAY_FORMAT` uses full month names in English.)

hughrun:
Ah, I did think about that but wasn't sure how to do locale-specific dates. This is a good solution and admins can override the default. I'll update the PR shortly.

dato (approved):
Great, thank you!

If you’d like, you can squash this test into your PR: https://github.com/hughrun/bookwyrm/pull/5

(I have a branch with extended tests for get_published_date, but I’m having errors mocking datetime.now(), so I’ll send them in a separate PR later.)

PR #3451, opened 2024-10-11
Fix IntegrityError caused by duplicate periodic task creation

Guanchishan:
## Description

**Summary**
This PR addresses an issue with the creation of periodic tasks in the `settings-dashboard` view, where using `PeriodicTask.objects.get_or_create()` could lead to an IntegrityError due to duplicate primary keys when the task already exists.

**Technical Details**
The change involves replacing `PeriodicTask.objects.get_or_create()` with `PeriodicTask.objects.update_or_create()`. This modification ensures that if a periodic task with the specified name already exists, it will be updated instead of creating a new entry, thus preventing the IntegrityError that occurs due to a unique constraint violation on the primary key.

**Changes Made**
- Updated `PeriodicTask.objects.get_or_create()` to `PeriodicTask.objects.update_or_create()` to prevent duplicate key errors when the task already exists.

**Testing and Impact**
This change has been tested to verify:
- If the task with the given name already exists, the task is updated without creating duplicate entries.
- If the task does not exist, a new entry is created correctly.

This update helps to prevent issues when multiple attempts are made to create a periodic task, especially in production environments, where such collisions can cause instability.

**Additional Context**
No other changes were made beyond this single line, to minimize potential side effects.

- Related Issue #
- Closes #

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/views/admin/dashboard.py
+++ b/bookwyrm/views/admin/dashboard.py
@@ -83,7 +83,7 @@ class Dashboard(View):
             schedule, _ = IntervalSchedule.objects.get_or_create(
                 **schedule_form.cleaned_data
             )
-            PeriodicTask.objects.get_or_create(
+            PeriodicTask.objects.update_or_create(
                 interval=schedule,
                 name="check-for-updates",
                 task="bookwyrm.models.site.check_for_updates_task",


mouse-reeve:
Thank you so much for all these bug fixes! The nuance here totally escaped me :)

PR #3459, opened 2024-10-18
Adds hCaptcha support

mouse-reeve:
## Description

We're having a spam-wave! This will allow admins to require a captcha during registration. I used [hCaptcha](https://www.hcaptcha.com/), which requires the admin to register an account, and does cost money at a certain usage point (one million requests per month, which is anyone is anywhere near that it's news to me). HCaptcha seems both easy to use, well supported, and [accessible](https://www.hcaptcha.com/accessibility). 

- Related Issue #
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

hCaptcha is installed, and .env variables are added for it

## Documentation

### Tests

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,10 @@ DEFAULT_LANGUAGE="English"
 
 MEDIA_ROOT=images/
 
+# hCaptcha configuration
+HCAPTCHA_SITEKEY=
+HCAPTCHA_SECRET=
+
 # Database configuration
 PGPORT=5432
 POSTGRES_PASSWORD=securedbypassword123
--- a/bookwyrm/forms/landing.py
+++ b/bookwyrm/forms/landing.py
@@ -4,6 +4,7 @@ from django.contrib.auth.password_validation import validate_password
 from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
+from hcaptcha_field import hCaptchaField
 import pyotp
 
 from bookwyrm import models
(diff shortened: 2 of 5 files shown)


Flameborn:
While HCaptcha is easy to use and it works reasonably well, I have to say their stance on accessibility is quite appalling.

* If you are unable to see their captcha, you need to create a specific account via an email address. They monitor this account for abusive behavior.
* Then, you need to go to a specific page they send out once to the registered email. Here, you need to store a cookie on your device. This only works if you deliberately weaien your browser's security, for example by disabling cross origin protection, IP tracking protection in Safari, allow third party cookies, etc.
* Then you need to go back to the captcha page, where the token may or may not work. If it works, then for the next captcha you have to do this process again with the same registered account.

Alternatively, captcha implementors can request an alternate method to be available via their added captchas, which is text based. This works, but it is marketed as something recommended only in specific cases, because it weakens the secureness of the added captcha.

When asked about this, the company continues to advocate for accessibility as an aid for a minority, rather than ensuring that their captchas can be solved by everyone equally. For example they were planning to develop a hardware device which would generate tokens, which has the same issues as having to register via an email address, except with this they would also require a physical address to ship to.

A lot of people solve their captchas via AI, because it is a lot easier than having to go through all this just to get access to a page which sometimes can't be avoided (some governments use this captcha as well).

mouse-reeve:
Oof! Thank you so much for this. I’ve been looking for a ReCAPTCHA alternative but really striking out

Flameborn:
The only one that I have high hopes for is [mCaptcha](https://github.com/mCaptcha/mCaptcha), which is a one click solution that works really well. It's unfortunately not trivial to host and implement yet, but it's slowly getting there.

mouse-reeve:
That looks promising! I’ll take a look and if it’s too much to get using now I’ll use recaptcha for the time being and try to keep this in the pipeline

Flameborn:
Sorry for somewhat resurrecting an unmerged PR. A bit of a shameless plug, but I've recently wrote an open source server for [Altcha](https://altcha.org), which is something similar to MCaptcha and it's super easy to implement. To my knowledge, there are Python implementations of the library as well.

The server can be found [here](https://github.com/Flameborn/verity). Alternatively, Altcha provides a free API as well, so hosting a server is not required.

PR #3471, opened 2024-11-16
Fix imports creating duplicate Editions

### LINKED ISSUE #3019: Duplicate book after manual addition
**Describe the bug**
I think this is a fairly significant bug. If I add a book manually, I've noticed that sometimes it ends up duplicated on some instances. But not all, and not always.

**To Reproduce**
1. I've created a new book on my instance. A book that does not yet exist on external sources.
https://bw.heraut.eu/book/26362/s/chroniques-poetiques-dun-voyage-a-montreal

2. I look at bookwyrm social the day after tomorrow and I see that it's a duplicate (even though it's exactly the same book):
https://bookwyrm.social/book/1427616/editions

3. I look on another instance and see that it's a single copy:
https://bouquins.zbeul.fr/book/29947/s/chroniques-poetiques-dun-voyage-a-montreal

I've noticed this bug several times and it seems to be a real nuisance.
For example (I have two accounts, on two instances) that follow each other.
I see that sometimes the book is duplicated on my own timeline. For a book that should be the same, I have two publications that lead to two different books.
- the publication that announces the start of reading leads to a book
- the publication that announces (later) a quotation leads to a duplicate of the first book.

I'm using a translator and I hope my message is understandable.

### LINKED ISSUE #3357: ValueError('Invalid related remote id: ***')
In Bookwyrm 0.7.3

I often get this error (via Flower) regarding the propagation of book changes between instances.
When I look at the link I think there is confusion in the id. Here for example :

`('Edition', 'Work', 'parent_work', 'https://bw.heraut.eu/book/52087', 'https://bouquins.zbeul.fr/book/42398')`

52087 corresponds [to an edition](https://bw.heraut.eu/book/52087/editions) (on an instance) whereas 42398 corresponds [to a book](https://bouquins.zbeul.fr/book/42398/s/citadins-de-demain) on the other instance.

It's a shame because changes are not propagated.

  File "/opt/bookwyrm/venv/lib/python3.11/site-packages/celery/app/trace.py", line 477, in trace_task
    R = retval = fun(*args, **kwargs)
                 ^^^^^^^^^^^^^^^^^^^^
  File "/opt/bookwyrm/venv/lib/python3.11/site-packages/celery/app/trace.py", line 760, in __protected_call__
    return self.run(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/contextlib.py", line 81, in inner
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/bookwyrm/bookwyrm/activitypub/base_activity.py", line 277, in set_related_field
    raise ValueError(f"Invalid related remote id: {related_remote_id}")
ValueError: Invalid related remote id: https://bw.heraut.eu/book/52087

### LINKED ISSUE #3381: Account import fails with error: 'attributedTo'

hughrun:
This fixes bugs at both the sending and receiving instances: As the primary fix is at the sending end, a full fix will need to be collaborative at the network level.

**Fixing at the sending end: don't send duplicate GeneratedNotes**

The primary fix is to `handle_reading_status`: it calls `create_generated_note` which saves and broadcasts the GeneratedNote. Before this fix, `handle_reading_status` then saved and broadcast the note again, creating two incoming GeneratedNotes for all federated BookWyrm servers, and thus triggering simultaneous or near-simultaneous imports of the associated book within the `tag` reference. This is then dereferenced, creating two identical Works, the first of which ends up empty and the second of which ends up with two identical Editions.

**Fixing at the sending end: delay BookStatus broadcasts**

My testing doesn't indicate one way or the other whether this happens, but there's a theoretical race condition for incoming editions that are ListItem or ShelfBook items. The Add activity needs to create the Edition and Work if they are not already on the receiving instance, and so does the Comment or GeneratedNote. The fix in `ActivitypubMixin::broadcast` sets a delay on these statuses to ensure that the Add activity is processed first, and thus the book will already exist when the status is processed.

**Fixing at the receiving end: prevent dereferencing the incoming Edition**

The book attached to a ListItem or ShelfBook is an Edition. This means that when a receiving instance dereferences it, we end up with a recursion: The Edition dereferences the parent Work, which dereferences all the child editions in celery tasks. This includes the Edition that triggered the whole process in the first place.

I couldn't create a reproducable test of this, but I've seen evidence that this may cause a race condition where the incoming Edition ends up being duplicated - the task to create child Editions checks for existence before the original triggering Edition has saved, and in combination with the above we can then end up with three duplicates.

The changes to Note, Fields, and `ActivityObject::to_model` are aimed at preventing this by allowing us to compare related field ids to the original triggering remote_id before running `set_related_field` and excluding them if there is a match.

## Description

- Closes #3381
- Closes #3357 
- Closes #3019

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -120,6 +120,7 @@ class ActivityObject:
         save: bool = True,
         overwrite: bool = True,
         allow_external_connections: bool = True,
+        trigger=None,
     ) -> Optional[TBookWyrmModel]:
         """convert from an activity to a model instance. Args:
         model: the django model that this object is being converted to
@@ -133,6 +134,9 @@ class ActivityObject:
             only update blank fields if false
         allow_external_connections: look up missing data if true,
             throw an exception if false and an external connection is needed
+        trigger: the object that originally triggered this
+            self.to_model. e.g. if this is a Work being dereferenced from
+            an incoming Edition
         """
         model = model or get_model_from_type(self.type)
 
@@ -223,6 +227,8 @@ class ActivityObject:
             related_field_name = model_field.field.name
 
             for item in values:
+                if trigger and item == trigger.remote_id:
+                    continue
                 set_related_field.delay(
                     related_model.__name__,
                     instance.__class__.__name__,
--- a/bookwyrm/activitypub/note.py
+++ b/bookwyrm/activitypub/note.py
@@ -50,6 +50,7 @@ class Note(ActivityObject):
         save=True,
         overwrite=True,
         allow_external_connections=True,
+        trigger=None,
     ):
         instance = super().to_model(
             model, instance, allow_create, save, overwrite, allow_external_connections
--- a/bookwyrm/models/activitypub_mixin.py
+++ b/bookwyrm/models/activitypub_mixin.py
@@ -129,7 +129,20 @@ class ActivitypubMixin:
 
     def broadcast(self, activity, sender, software=None, queue=BROADCAST):
         """send out an activity"""
+
+        # if we're posting about ShelfBooks, set a delay to give the base activity
+        # time to add the book on remote servers first to avoid race conditions
+        countdown = (
+            10
+            if (
+                isinstance(activity, object)
+                and not isinstance(activity["object"], str)
+                and activity["object"].get("type", None) in ["GeneratedNote", "Comment"]
+            )
+            else 0
+        )
         broadcast_task.apply_async(
+            countdown=countdown,
             args=(
                 sender.id,
                 json.dumps(activity, cls=activitypub.ActivityEncoder),
@@ -227,6 +240,7 @@ class ObjectMixin(ActivitypubMixin):
                 return
 
             try:
+                # TODO: here is where we might use an ActivityPub extension instead
                 # do we have a "pure" activitypub version of this for mastodon?
                 if software != "bookwyrm" and hasattr(self, "pure_content"):
                     pure_activity = self.to_create_activity(user, pure=True)
--- a/bookwyrm/models/fields.py
+++ b/bookwyrm/models/fields.py
@@ -86,7 +86,9 @@ class ActivitypubFieldMixin:
                 raise
             value = getattr(data, "actor")
         formatted = self.field_from_activity(
-            value, allow_external_connections=allow_external_connections
+            value,
+            allow_external_connections=allow_external_connections,
+            trigger=instance,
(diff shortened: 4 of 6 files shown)


nesges:
Can we have this as a xmas gift, please? 😎 I'm at about 6500 manual merges an counting, and this one looks like it could mitigate these issues greatly

hughrun:
@bookwyrm-social/code-review if anyone has capacity to test and review this that would be great :)

mouse-reeve:
Wow, thank you! I was completely stumped as to why these duplications were happening

nesges:
just some feedback: we're still seeing influx of duplicate origin IDs. Here's a recent example: https://buecher.pnpde.social/book/100289/s/neogenesis https://buecher.pnpde.social/book/100290/s/neogenesis

hughrun:
@nesges unfortunately for this fix to work across the network, all instances need to upgrade, because some fixes are required at the source and others are at the importing instance. I can't see the origin ids in those books but it's likely that the source instance hadn't been upgraded at the time the book(s) were imported. It will likely take a while to roll out through the network.

nesges:
Thanks @hughrun! I got curious and looked up the versions of connected instances to ours and among others bookwyrm.social seems to be running on v0.7.4 still

PR #3496, opened 2025-01-28
Fix isbn search, matching the regular search

renatolond:
## Description

I noticed on my instance that looking for ISBN directly was always resulting in an "Oops" page. This only happens if the isbn-like search does not have local results (otherwise the local results are displayed).

I tried matching the ISBN results to the regular search results, I guess one method was changed and the other wasn't.

I haven't written any tests, though probably some should be written to avoid this method to get out-of-sync again in the future.

- Related Issue #
- Closes #

- [X] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/connectors/openlibrary.py
+++ b/bookwyrm/connectors/openlibrary.py
@@ -181,17 +181,19 @@ class Connector(AbstractConnector):
             )
 
     def parse_isbn_search_data(self, data: JsonDict) -> Iterator[SearchResult]:
-        for search_result in list(data.values()):
+        for search_result in list(data.get("docs", [])):
             # build the remote id from the openlibrary key
             key = self.books_url + search_result["key"]
-            authors = search_result.get("authors") or [{"name": "Unknown"}]
-            author_names = [author.get("name") for author in authors]
+            authors = search_result.get("author_name") or ["Unknown"]
+            cover_blob = search_result.get("cover_i")
+            cover = self.get_cover_url([cover_blob], size="M") if cover_blob else None
             yield SearchResult(
                 title=search_result.get("title"),
                 key=key,
-                author=", ".join(author_names),
+                author=", ".join(authors),
                 connector=self,
-                year=search_result.get("publish_date"),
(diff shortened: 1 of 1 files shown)


hughrun:
Hi @renatolond 

Thanks for this.

I can't replicate the problem of "returns nothing at all" if there is an edition with that ISBN in OpenLibrary but not locally. e.g. https://bookwyrm.social/search?q=9780843963632&type=book works.

However, I have noticed that we don't currently get any cover images from Open Library. Looking into that more, I think it's because `get_cover_url` uses a bad signature for the API we currently use for ISBNs.

For ISBN search, BookWyrm is currently using the old deprecated "Legacy Book API" rather than [the newer "Work and Edition" one for books and ISBN searches](https://openlibrary.org/dev/docs/api/books). Based on your code, it looks like you're trying to use the general "book search" API. This is why the test data format doesn't match what you're getting.

It's hardly surprising you're confused because it seems we're using a legacy API call and then accidentally referencing the covers incorrectly, using a value that appears in the newer general book search API result.

* "Legacy Book API" (currently used, as set in `initdb.py`): https://openlibrary.org/api/books?jscmd=data&format=json&bibkeys=ISBN:9780843963632
* "ISBN API" (special case of "editions" API): https://openlibrary.org/ISBN/9780843963632.json
* "Book Search API" querying on ISBN: https://openlibrary.org/search.json?isbn=9780843963632

I've done some poking around and I think you probably have the right solution to the problem of covers since it removes the legacy API and aligns with the expected data format for `get_cover_url`, but we need to change the `isbn_search_url` for Open Library to get it to work properly, otherwise it just returns nothing.

In order to change the default `isbn_search_url` we need to update `initdb.py` but that will only change it for new instances. We either need a management command to update existing servers, or wait for something I am meaning to do anyway which is create a page in the admin web UI to allow admins to enable and disable connectors and change their order of priority. That would also be an interface that could be used to update them if any of the API urls change as in this case.

mouse-reeve:

PR #3498, opened 2025-01-29
add Finna API connector

### LINKED ISSUE #3147: Add a connector to the Finna API
I'm running an instance for Finnish people, who read a lot of Finnish books. The problem is that OpenLibrary and Inventaire databases are very lacking on Finnish books, which means that lots of those books must be entered manually. This can be very demotivating and lead to many Finnish people stop using Bookwyrm.

There's Finna, a public database of most public libraries, the national library and many museums in Finland, and this database contains almost every book published in Finland. The web UI can be accessed at https://finna.fi. There's also public API to the Finna database that everyone can use, and the documentation for the API can be found at https://api.finna.fi/swagger-ui/?url=%2Fapi%2Fv1%3Fswagger.

What I'd like to see is a new connector for the Finna API so that books in the Finna database could be found via Bookwyrm search, just like books from OpenLibrary and Inventaire.

Another public database with open data is Helsinki Metropolitan Library's database (https://haku.helmet.fi/iii/encore/?lang=fin&ivts=jzkddmLlhpX8J1aYVzgJBA%3D%3D&casts=mtJgsueW4%2BgQF9%2Bfmt5sIA%3D%3D) that also has an open API. However, it doesn't contain as much data as Finna, and so far I haven't found the API documentation.

ilkka-ollakka:
## Description

Finna connector

Finna is finnish open API that contains among other things books released in finland

terms of use this API is located in https://www.kiwi.fi/display/Finna/Finna+API+Terms+of+Use and is CC0

more info https://www.finna.fi/Content/about_finnafi

Connector is heavily borrowed from openlibrary and inventaire codebase, this is my first PR toward bookwyrm codebase, so there might be somethings that I have fully missed.

PR adds separate management command to enable this connector, so it is not enabled by default even on clean install. Command also allows to disable connector.

kuva
kuva

- closes #3147 

Original thread where the idea started and people pitched ideas and approaches https://mementomori.social/@anotherdream/113881912359314786

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/connectors/finna.py
```diff
@@ -0,0 +1,298 @@
...
+        ]
+
+        self.author_mappings = [
+            Mapping("id", remote_field="authors", formatter=self.get_remote_author_id),
+            Mapping("name", remote_field="authors", formatter=self.get_first_author),
+        ]
+
+    def guess_page_numbers(self, data: JsonDict) -> str | None:
+        for row in data:
+            page_search  = re.search(r"(\d+) (sivua|s\.)", row)
```
osma, line 63: This regex looks a bit fragile. The number of pages may be expressed in different ways and in different languages depending on the source database. For example it could be `"266 pages"`, as in the first example in the included file `finna_author_search.json`. Maybe just take the first part that looks like a number and ignore the rest?

bookwyrm/connectors/finna.py
```diff
@@ -0,0 +1,298 @@
...
+            result_languages.append(languages.get(f"/languages/{language_code}", language_code))
+        return result_languages
+
+    def join_subject_list(self, data: JsonDict) -> list[str]:
+        return [" ".join(info) for info in data]
+
+    def describe_physical_format(self, formats: JsonDict) -> str:
+        format = "Hardcover"
+        # Map finnish finna formats to bookwyrm codes
+        format_mapping = {
```
osma, line 81: Instead of mapping the Finnish translations (which could change over time), I think it would be better to map the codes that Finna uses for formats. For example `0/Book` and `1/Book/Audiobook`. These are more stable.

bookwyrm/connectors/finna.py
ilkka-ollakka: First part could be just size, so '26 cm'. But yes, it is somewhat fragile, but I would prefer maybe to add 'pages' also matchable strings instead of making it wider to match first number.

bookwyrm/connectors/finna.py
ilkka-ollakka: Ah yes, good idea. I'll change that

bookwyrm/connectors/finna.py
ilkka-ollakka: I started checking this, seems that there are some books (first one that I spotted was 'Garpin maailma' that has format-description like this: '1 nide (559 sivua, 1 numeroimaton sivu) ; 21 cm'

ilkka-ollakka: There seems to be also examples when this don't match the pagecount but first number would match it. I don't really have strong opinion on which way this should be.

ilkka-ollakka: I ended up in flow where I try to match those sivua/pages/sidor, and if no fish, then try to match starting number.

ilkka-ollakka:
Fixed editions loading and picking edition from work so it prefers books over audiobook/ebook and finnish/swedish language ones if any found.

ilkka-ollakka:
Fixed the issues that pylint/mypy/black complained, seems that I needed to add finna_key, so added it as separated first commit.

ilkka-ollakka:
Cleaned out locale changes that jumped in by accident in some rebase

bookwyrm/management/commands/initdb.py
```diff
@@ -114,6 +114,24 @@ def init_connectors():
...
+        search_url="https://api.finna.fi/api/v1/search?limit=20"
+        "&filter[]=format%3a%220%2fBook%2f%22"
+        "&field[]=title&field[]=recordPage&field[]=authors&field[]=year&field[]=id&field[]=formats"
+        "&lookfor=",
+        isbn_search_url="https://api.finna.fi/api/v1/search?limit=1"
+        "&filter[]=format%3a%220%2fBook%2f%22"
+        "&field[]=title&field[]=recordPage&field[]=authors&field[]=year&field[]=id&field[]=formats"
+        "&lookfor=isbn:",
+        priority=1,
+    )
```
ilkka-ollakka, line 133: This is one thing that I'm unsure if it needs to be in migrations also? This works with clean setup that runs initdb, but I'll need to check how does it work with existing database.

ilkka-ollakka: Changed this to be migrations so it works as expected from current version updating to this one.

bookwyrm/migrations/0212_configure_finna_connector.py
```diff
@@ -0,0 +1,43 @@
...
+    app_registry.get_model("bookwyrm", "Connector").objects.using(db_alias).filter(
+        connector_file="finna"
+    ).delete()
+
+
+def add_finna_connector(app_registry, schema_editor):
+    """doesn't need to do anything"""
+    db_alias = schema_editor.connection.alias
+    model = app_registry.get_model("bookwyrm", "Connector")
+    model.objects.using(db_alias).create(
```
ilkka-ollakka, line 16: This seems to break lot of connector tests, indicating that it is not actually the preferred flow to add new connector. So should I add it back to initdb with different limit or some other way what is intended flow?

bookwyrm/models/book.py
```diff
@@ -41,6 +41,9 @@ class BookDataModel(ObjectMixin, BookWyrmModel):
     openlibrary_key = fields.CharField(
         max_length=255, blank=True, null=True, deduplication_field=True
     )
+    finna_key = fields.CharField(
```
hughrun, line 44: Ideally this would be visible in `templates/book/book_identifiers.html` and perhaps also `templates/book/edit/edit_book_form.html`.

hughrun (changes requested):
Overall this looks like a really well-considered PR - thankyou!

I have added a few comments, hopefully they are clear but please let me know if you have any questions or need help.

bookwyrm/connectors/finna.py
```diff
@@ -0,0 +1,386 @@
...
+    """instantiate a connector for finna"""
+
+    generated_remote_link_field = "id"
+
+    def __init__(self, identifier: str):
+        super().__init__(identifier)
+
+        get_first = lambda x, *args: x[0] if x else None
+        format_remote_id = lambda x: f"{self.books_url}{x}"
+        self.book_mappings = [
```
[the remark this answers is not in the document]
ilkka-ollakka: There is copyright flow that if the image url contains data, it is under finna cc and can be used, but like you noticed, most of the images are not under that.

I'll add the image parsing back to keep it consistent. Linking to Solr-urls directly was not seem as good idea from Finna side.

bookwyrm/connectors/finna.py
```diff
@@ -0,0 +1,386 @@
...
+            Mapping("id", remote_field="id", formatter=format_remote_id),
+            Mapping("finna_key", remote_field="id"),
+            Mapping("title", remote_field="title"),
```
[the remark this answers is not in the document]
ilkka-ollakka: Thanks, I'll update the mapping accordingly.

ilkka-ollakka:
Thanks for the review, I'll do the small changes and update the PR according to comments.

hughrun (approved):
This is really cool! Very thorough job as far as I can see.

There are two remaining issues I've noticed, but I am going to pull this in anyway.

1. When adding a new author, `get_authors_from_data` will pull the name from an author search in the finna API even though we already have the name. This is because of how `get_or_create_author` works in `abstract_connector` (it doubles as the way to get author data from other BookWyrm instances). The only reason I noticed this is that like a lot of other library catalogues, the Finna collection seems to have some older records that capitalise author names. This means sometimes the author name will appear initially like "Minna Canth" but then when the book is imported it will be changed to "MINNA CANTH" because the first "most relevant" record was like that. I couldn't work out a simple and safe way to adjust the query so we will just have to put up with this for now since it is more about the data source than our code.
2. If an admin adds Finna, then deactivates it, there is no way to _re_-activate it. However I'm planning to resolve this by adding my own PR to allow admins to activate and enable/disable connnectors from the web UI.

PR #3519, opened 2025-03-03
Replace imghdr with use of pil.Image.format

### LINKED ISSUE #3304: Replace now-deprecated imghdr module
**Describe the bug**
The imghdr module is deprecated as of Python 3.11, and will be removed in Python 3.13. The deprecation notice is here: https://docs.python.org/3/library/imghdr.html but the PEP referenced as featuring details and alternatives has only details, no specific alternative. 

We'll need to find an alternative to imghdr by the time we move to 3.13, and when we do we can remove the linter exception introduced into bookwyrm/connectors/abstract_connector.py with PR #3303 as part of the Python 3.11 upgrade.

**To Reproduce**
See https://docs.python.org/3/library/imghdr.html  and https://peps.python.org/pep-0594/#imghdr for details of the deprecation

**Expected behavior**
We should find a replacement! <3 

**Screenshots**
N/A

**Instance**
All

**Additional context**
N/A

ilkka-ollakka:
imghdr is deprecated, so replace it with something else

image content is actually already bitstream in memory, so we can just do read() on it without huge overhead, as was done with imghdr.

## Description

Remove use of deprecated library imghdr. As image is bitstream from loaded url in get_image, we just provide it to pillow to check bitstream and give us format.

- Related Issue #
- Closes #3304

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/connectors/abstract_connector.py
+++ b/bookwyrm/connectors/abstract_connector.py
@@ -5,7 +5,7 @@ from typing import Optional, TypedDict, Any, Callable, Union, Iterator
 from urllib.parse import quote_plus
 
 # pylint: disable-next=deprecated-module
-import imghdr  # Deprecated in 3.11 for removal in 3.13; no good alternative yet
+from PIL import Image, UnidentifiedImageError
 import logging
 import re
 import asyncio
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you so much for this! It's nice to be able to remove a deprecation comment from the code :)

PR #3527, opened 2025-03-26
Import tombstone migration

mouse-reeve:
## Description

This merges #3431 along with a merge migration

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
That's simple, @mouse-reeve — the problem is that False is not True 🫠

Looks like `bookwyrm.tests.importers.test_importer.GenericImporter::test_complete_job` is failing.

I can take a look after work tonight my time.

hughrun:
I haven't been able to confirm _why_ this is failing, but it looks like it's to do with a709121846349e3ee7ba9fe3b5bef70a13021a70, though it's mysterious that it didn't fail in that PR before it was merged. It's not just #3527 that is failing - if you run tests on `main` right now it will fail the same way.

* If I remove the code from a709121846349e3ee7ba9fe3b5bef70a13021a70, all tests pass.
* If I re-order the tests by changing the name of `test_goodreads_importer.py` to make it run last, all tests pass
* If I run `./bw-dev pytest -k "importer and not goodreads"`, all tests pass
* If I skip `GoodreadsImport::test_create_retry_job`, all tests pass 
* If I run `./bw-dev pytest -k "importer"`, I get the same error


My theory therefore is that the addition of `GoodreadsImporter::normalize_row` is making the goodreads `create_retry_job` test run a tiny bit slower, and we haven't managed to isolate our tests enough so that is causing a slight delay in finalising the import jobs on the generic test, hence the notification (that the import job is finished) hasn't triggered yet when we test the assertion that it exists.

However I'm still not completely across how the test environment works, so maybe I'm barking up the wrong tree.

hughrun:
Aaactually I think I had this issue with the BookWyrm csv importer. We're testing importing the same books in two different tests even though the csv files are technically different, and we run test in parallel. So somehow they are interfering with each other, I guess if the book already exists.

Probably the issue can be "fixed" by changing the test csv data but it feels a bit like it shouldn't be happening anyway.

hughrun:
@mouse-reeve I still can't find a root cause but I'm going to suggest simply removing `test_create_retry_job` from `test_goodreads_import`. 

 `GoodreadsImporter` inherits `create_retry_job` from `Importer` so it's not really adding anything useful to test it on both classes, and removing this test fixes our problem here.

Obviously it would be ideal to find and fix a root cause here but removing the test doesn't seem like it would lose any real test coverage. 

As far as I can tell the import job in the goodreads test is overwriting the test in the generic importer when they run in parallel, though I can't see any obvious reason why this would occur.

mouse-reeve:
That sounds like a good solution under the circumstances

hughrun:
@mouse-reeve wait wait, I have a better idea. 

I just realised that `test_complete_job` doesn't actually *need* to load the test file at all. The only thing we're testing is whether a notification is sent when the import is completed. So we can just created an in-memory "csv file" on the fly and the content doesn't even matter.

I'll make a patch for this tonight my time.

hughrun:
@mouse-reeve I updated the migrations so #3431 should be good to pull directly now, and will fix the problem with tests.

PR #3541, opened 2025-04-05
remove port 8000 from docker-compose as unnecessary for web container

ilkka-ollakka:
## Description

Nginx container doesn't need it to be able to connect to web container, so it just unnecessarily opens the port to outside connections

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


ilkka-ollakka:
Seems that bw-dev command contains portion to show in which that port is mapped, so it seems to have some use in some cases?

hughrun:
I think maybe we need `8000` open to use the Django development server? As things are currently set up, `main` is for development and `production` is for live production servers.

ilkka-ollakka:
The port needs to be opened only if you want to access that development server directly without going via ngixn container. Access via nginx container works in all modes (reverse-proxy, production https and development) fine without it.

Seems that it might be there for some historical reason as it has been there since the start of the docker-compose-file, and is not needed anymore?

ilkka-ollakka:
I'm not sure if there is some use case (debug etc) that needs direct connect to django run development server?

hughrun:
Ok now I see what you mean and have tested this successfully. I can't see any reason we need to directly access `8000`.

PR #3542, opened 2025-04-05
Rework dockerfile to do multi-stage build and run as non-root uid by default

ilkka-ollakka:
## Description

Change Dockerfile to do multi-stage build to get smaller container (from ~1.3G -> ~0.5G). Removing lot of not-needed stuff that is needed in build-phase but not when running things, things like gcc compilers and things. It builds/installs dependencies in full python image but installs things in virtualenv and copies that to -slim image that also contains non-root account that is used by default to run commands.

Bw-dev command is slightly modified to run pylint etc things as root-uid, making it smallest feasible change without needing to rework all the flows and commands.

I did test this by running `bw-dev pytest` and running bookwyrm in locally on development mode, running csv file import task and check that it didn't show up any new errors.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

If instance has local storage in use, command `bw-dev fix_volume_permissions` should be run during update to change local volume and asset permissions for exports/images/static files.

## Documentation

### Tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,29 @@
-FROM python:3.11
+FROM python:3.11 AS build-image
 
-ENV PYTHONUNBUFFERED 1
 
-RUN mkdir /app /app/static /app/images
-
-WORKDIR /app
 
 RUN apt-get update && apt-get install -y gettext libgettextpo-dev tidy && apt-get clean
 
+WORKDIR /app
+
 COPY requirements.txt /app/
-RUN pip install -r requirements.txt --no-cache-dir
+RUN python -mvenv /venv
+ENV PATH=/venv/bin:$PATH
+RUN pip install --compile -r requirements.txt --no-cache-dir
+
+
+FROM python:3.11-slim
+
+RUN apt-get update && apt-get install -y gettext libpq5 tidy && apt-get clean
+
+RUN mkdir -p /app/images /app/static
+
+
+
+RUN addgroup --system app && adduser --system --group bookwyrm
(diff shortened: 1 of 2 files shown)


ilkka-ollakka:
Just realized that in upgrade of production setup, this makes images and static directories inside container still be owned by root uid, which causes that new account can't write to those directories.

I'll check if there is some reasonable flow for that.

hughrun:
@ilkka-ollakka I'm a little unclear on the status of this PR - is it ready to check, or did you want to do some more work on it?

ilkka-ollakka:
@hughrun I had some changes in local that I had forgot to push into this PR. Now it is in state that it could be checked.

I was checking if using entrypoint in web-container could be used instead of separate bw-dev command for permissions, but that is ran under bookwyrm user, so it can't fix the permissions.

I'm not sure how typical bookwyrm instance is used, does it use s3-style storage or local storage, mainly how this should be informed on update flow that permissions needs to be fixed/changed with local volumes?

hughrun:
I think @mouse-reeve has a list of instance admins, or you could just ask in the main Matrix room. My guess would be most are using s3 but that's just a guess.

ilkka-ollakka:
Actually now that I've been looking this, after https://github.com/bookwyrm-social/bookwyrm/pull/3605 we don't need build phase and can use the slim-image directly as psycopg3 doesn't need to be compiled.

ilkka-ollakka:
I'll close this PR as I'm starting to prepare PR that has container build/entrypoint for migrations and container publishing flows included.

PR #3570, opened 2025-05-02
Add bw-dev create_secrets -command

ilkka-ollakka:
## Description

Command can be used to generate secrets for .env file that are not yet generated. It uses either python3 secrets module to generate urlsafe token of random 50 bytes or openssl to generate hex output of 50 random bytes.

50 bytes is than what django 4.2 generates by default, so it should be sane default for now. Owasp also recommends to have atleast 50 random characters as secret-key (https://cheatsheetseries.owasp.org/cheatsheets/Django_Security_Cheat_Sheet.html#key-management).

Do note that even if the output is alpha-numeric/hex, we do generate 50 random bytes that are encoded, so the generated password/key is longer than 50 characters.

Only credential from the list what admin should use is the `FLOWER_PASSWORD`, but that can be changed afterwards. Only password that is not trivial to change after initial setup is the `POSTGRES_PASSWORD`

Change defaults in settings for `SECRET_KEY` to be None and complain if `SECRET_KEY` is not defined. Comment out secrets from `.env.example` that should not have default values but unique per deployment

List of .env values that are checked and generated if not enabled in `.env`:
```
- SECRET_KEY
- POSTGRES_PASSWORD
- REDIS_ACTIVITY_PASSWORD
- REDIS_BROKER_PASSWORD
- FLOWER_PASSWORD
```

Alternative for openssl/secrets module could be to call django inside web-container for example to generate secrets (https://github.com/django/django/blob/stable/4.2.x/django/core/management/utils.py#L79).

- Related Issue #
- Closes #

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

Comments out default credentials in .env.example, so user needs to provide those credentials either manually editing .env file or running the `bw-dev create_secrets` command.

## Documentation

- [x] I have created a matching pull request in the Documentation repository

### Tests


ilkka-ollakka:
It might be useful to have flow to rotate secrets also, all expect database password should be trivial to rotate.

.env.example
```diff
@@ -1,5 +1,5 @@
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY="7(2w1sedok=aznpq)ta1mc4i%4h=xx@hxwx*o57ctsuml0x%fr"
+#SECRET_KEY="7(2w1sedok=aznpq)ta1mc4i%4h=xx@hxwx*o57ctsuml0x%fr"
```
hughrun: Standard practice to now has always been to leave a space when commenting out in `.env.` e.g.

```toml
# SECRET_KEY="like this"
#SECRET_KEY="not like this"
```

This is also the default behaviour in VSCode/SCodium. `create_secrets` in this case will say it is updating the secrets but doesn't actually do anything because it can't find the matching string.

I think the ideal scenario would be to be able to match on either style of commenting-out. Otherwise, it should use the style with the space.

bw-dev
```diff
@@ -69,7 +69,28 @@ function awscommand {
...
+        echo "you need either python or openssl executable to generate secrets with this script"
+        exit 1
+    fi
+    for env_variable in SECRET_KEY POSTGRES_PASSWORD REDIS_ACTIVITY_PASSWORD REDIS_BROKER_PASSWORD FLOWER_PASSWORD; do
+        if grep -q -qi "^$env_variable" .env; then
+            echo -e "${env_variable} already configured, skipping"
+        else
+            echo -e "generating random secret in ${env_variable}"
+            generated_secret=$(eval "${SECRETS_COMMAND}")
+            sed -i -e "s/^#${env_variable}=.*/${env_variable}=\"${generated_secret}\"/" .env
```
hughrun: This has surprising behaviour on MacOS (and possibly others) because it thinks `-e` is the backup file extension and creates a file confusingly named `.env-e`. Suggest you change it to:

```sh
sed -i .bak -e 
```
...or some other explicitly named extension.

The other thing I notice is that if the variable isn't in the file at all, it won't do anything but will say that it has. Potentially could fix with something like this:

```sh
if grep -q -qi "^$env_variable" .env; then
    echo -e "${env_variable} already configured, skipping"
elif grep -q -qi "^# $env_variable" .env; then
    echo -e "generating random secret in ${env_variable}"
    generated_secret=$(eval "${SECRETS_COMMAND}")
    sed -i .bak -e "s/^# ${env_variable}=.*/${env_variable}=\"${generated_secret}\"/" .env
else
    echo -e "generating random secret in ${env_variable}"
    generated_secret=$(eval "${SECRETS_COMMAND}")
    echo "${env_variable}=\"${generated_secret}\"" >> .env
fi
```

.env.example
ilkka-ollakka: Ah yes, I'll update the code to handle both cases

bw-dev
ilkka-ollakka: I updated the sed part, it should work with freebsd sed (same as os x has to my understanding), but I don't have such os currently to test.

I also extended the sed to append at the end of the file the variable if not found. I do see that the sed flow in there might be intimidating and could be better to replace it with simpler to follow logic, like the one you proposed.

ilkka-ollakka: I changed the sed to simpler flow to follow

hughrun (approved):

hughrun:
Looks good!

PR #3573, opened 2025-05-02
add isbn10 and isbn13 validators to Edition model

### LINKED ISSUE #3564: An invalid ISBN causes the book page to crash (quite problematic)
On bookwyrm 0.7.5

**Describe the bug**
An invalid ISBN causes the book page to crash.

**To Reproduce**
All you have to do is edit a book and enter this isbn: 97801X45 (please note that you then have to correct this directly in the database).

Following a server-side error on a book, I looked at the logs and saw a line like :

Apr 29 09:57:02 master gunicorn[1647]:   File "/DIR/bookwyrm/isbn/isbn.py", line 122, in __find_registrant
Apr 29 09:57:02 master gunicorn[1647]: if registrant_range[0] <= int(registrant) <= registrant_range[1]:
Apr 29 09:57:02 master gunicorn[1647]:                               ^^^^^^^^^^^^^^^
Apr 29 09:57:02 master gunicorn[1647]: ValueError: invalid literal for int() with base 10: '1X'

For some reason the book's ISBN was wrong. The value 97801X45 caused the page to crash.

ilkka-ollakka:
## Description

Checks if isbns looks valid when manually editing book before it goes to database. This allows the form inform if isbn values look something else than what bookwyrm expects them to be.

- Related Issue #
- Closes #3564

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/models/book.py
```diff
@@ -448,15 +449,59 @@ def to_edition_list(self, **kwargs):
...
+    if not maybe_isbn.upper()[:-1].isnumeric():
+        raise ValidationError(
+            _("%(value)s doesn't look like isbn"), params={"value": maybe_isbn}
+        )
+
+
+def validate_isbn13(maybe_isbn: str) -> None:
+    """Check if isbn13 mathes some expectations"""
+
+    if maybe_isbn[:3] != "978":
```
timothyjrogers, line 471: ISBN 13s can also begin with `979`, should this check allow both?

ilkka-ollakka: ah yes, it should, I'll add it.

dato:
Hello!, and thank you for your PR.

Regarding unit tests:

> - [x] I have not written tests and need help to write them

I’ve pushed a branch with unit tests for your changes (diff can be seen [here](https://github.com/ilkka-ollakka/bookwyrm/compare/tweak/validate_isbn_in_model...dato:bookwyrm:isbn_tests?expand=1); I could create a PR if you’d like).

Some of the tests are commented out, since they would fail with the current implementation. Starting with the least complex, these are:

1.  `validate_isbn13` properly strips dashes from the value, but `validate_isbn10` doesn’t.

2.  `validate_isbn10` doesn’t detect invalid checksum characters, e.g. `012345678Z` (very minor; could be ignored).

3. ISBNs include a “check digit” that guards against typos. It would be nice if form validation detected this as well, in addition to malformed values only.
    Wikipedia [describes](https://en.wikipedia.org/wiki/Check_digit#ISBN_10) check digits in ISBNs, and our `isbn_10_to_13` and `isbn_13_to_10` functions include code to compute it.

    - NOTE: fixing this item involves a fair amount of work, so please do not feel obligated to implement it.

ilkka-ollakka:
Hi,

Thanks for the test-examples and comments, those help a lot! Actually looking the models/book.py file, I could validate isbn10 by converting it to isbn13 and back and check result to match. For isbn13 similar should work if I just extend isbn13_to_10 also accept '979' prefix.

But please open PR for those tests and I'll take them in and work on those failing cases you spotted.

ilkka-ollakka:
Realised that I can make the PR with your link directly, so took it in, thanks!

bookwyrm/models/book.py
```diff
@@ -448,15 +449,98 @@ def to_edition_list(self, **kwargs):
...
+        )
+
+    if (checksum_version := isbn_13_to_10(isbn13_version)) is None:
+        raise ValidationError(
+            _("%(value)s doesn't look like isbn"), params={"value": maybe_isbn}
+        )
+
+    if checksum_version != normalized_isbn:
+        raise ValidationError(
+            _("%(value)s doesn't look like isbn, we expected %(check_version)s"),
```
mouse-reeve: I tried an invalid ISBN 13 for this test case, and got this result, which I don't think makes sense:
Screenshot 2025-06-07 at 5 42 07 PM

For the intended use case on ISBN 10, I think it would be clearer to say `"%(value)s doesn't match the ISBN 13 provided, we expected %(check_version)s"`

bookwyrm/models/book.py
```diff
@@ -448,15 +449,98 @@ def to_edition_list(self, **kwargs):
 ]
 
 
+def validate_isbn10(maybe_isbn: str) -> None:
+    """Check if isbn10 mathes some expectations"""
+
+    if not (normalized_isbn := normalize_isbn(maybe_isbn)):
+        raise ValidationError(
+            _("%(value)s doesn't look like isbn"), params={"value": maybe_isbn}
```
mouse-reeve: I think this should be phrased ever so slightly differently:

`_("%(value)s doesn't look like an ISBN"), params={"value": maybe_isbn}`

mouse-reeve (changes requested):
If the ISBN 13 value showing the ISBN 10 message issue is complex to resolve, I think I reasonable solution would be just to always give the same generic error message. Otherwise this looks good!

bookwyrm/models/book.py
ilkka-ollakka: Ah yes, you are right, I'll check those out.

ilkka-ollakka: validator config doesn't currently cross-check between isbn values if they match correctly. It only checks that isbn provided itself looks valid in given form.

I updates the error message to give provided prefix also, to make it more clearer that we expected different value internally in isbn-13 as checksum didn't match.

mouse-reeve (approved):
Works great, thank you!

PR #2809, opened 2023-04-08
When broadcasting, send initial small batch to improve caching

WesleyAC:
Related: #2792

Marked as draft since I've yet to test this.

Issue #2792, opened 2023-04-07
Investigate performance of user pages


WesleyAC:
Looking at Honeycomb data for HTTP latency, the pages that stand out the most are `/inbox` (expected due to #2717, although we should improve that) and `/user/`. Specifically:

```
^user/(?P<username>@?[a-zA-Z_\-\.0-9]+(@[\w_\-\.]+\.[a-z\-]{2,})?)/(status|review|reviewrating|comment|quotation|boost|generatednote)/(?P<status_id>\d+)(.json)?/?$
^user/(?P<username>@?[a-zA-Z_\-\.0-9]+(@[\w_\-\.]+\.[a-z\-]{2,})?)/?$
```

We should figure out what queries are responsible for this and improve them if possible.

![2023-04-06_23 14 20_1318x863](https://user-images.githubusercontent.com/5001092/230533864-b204e5bc-bd32-4869-8853-dd1cedf935d3.png)
![2023-04-06_23 16 16_560x189](https://user-images.githubusercontent.com/5001092/230533984-833730df-4691-470a-adc6-664f3e499898.png)
![2023-04-06_23 17 22_262x165](https://user-images.githubusercontent.com/5001092/230534089-e1c791f3-b1a8-4874-819b-3f926c337f87.png)

WesleyAC:
SQL queries involved in making a request to `/user/b` (user ID 2) as user "a" (user ID 1):

<details>
<summary>Queries</summary>

```
[...]

PR #3604, opened 2025-06-08
nginx: add proxy_cache_lock, so we wait one request to populate cache if there is one ongoing for that cache-key

ilkka-ollakka:
## Description

Add lock in nginx level that only one request gets in if cache is miss and others wait that one request at max of 10s to get response and populate cache, which is served to other requests hitting the same cache-key.

Alternative approach to https://github.com/bookwyrm-social/bookwyrm/pull/2809

- Related Issue #2792
- Closes #

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

Nginx caching rules are changed, to prevent huge amount of identical requests to cause load on instance

## Documentation

### Tests

### DIFF
--- a/nginx/server_config
+++ b/nginx/server_config
@@ -20,3 +20,7 @@ proxy_cache_path
 # since activitypub endpoints have both HTML and JSON
 # on the same URI.
 proxy_cache_key $scheme$proxy_host$uri$is_args$args$http_accept;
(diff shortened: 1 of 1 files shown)


hughrun:
This looks like it should work to me. @WesleyAC @mouse-reeve thoughts?

Issue #3568, opened 2025-05-02
Upgrade postgresql before django 4.2 LTS turns to EOL


ilkka-ollakka:
**Is your feature request related to a problem? Please describe.**

Django released new 5.2 LTS and it doesn't anymore support postgresql 13 (https://docs.djangoproject.com/en/5.2/releases/5.2/#dropped-support-for-postgresql-13).

Django 4.2 is still supported until mid 2026 (https://www.djangoproject.com/download/#supported-versions) so there is time to start checking feasible migration path.

**Describe the solution you'd like**
Postgresql image in docker-compose updated to more recent version with automatically migration current data.

**Describe alternatives you've considered**
Manual database-upgrade description with database dump and importing to new database engine version.

**Additional context**

ilkka-ollakka:
I did quick check by running `podman run --rm -it -v bookwyrm_pgdata:/var/lib/postgresql/data --env-file ./.env -e PGAUTO_ONESHOT=yes docker.io/pgautoupgrade/pgautoupgrade:17-bookworm` after taking backup.

After migration was ran updated docker-compose postgresql version to :17, pytest seems to pass and normal things seems to work ok.

ilkka-ollakka:
https://github.com/pgautoupgrade/docker-pgautoupgrade is the docker container image I used

PR #3615, opened 2025-06-14
update docker-compose to postgresql17 and add bw-dev command to upgrade current database volume

ilkka-ollakka:
## Description

Change postgresql container to version 17 and add bw-dev command and separate docker-compose file to upgrade current database volume in-place.

I'm not sure how instance admins should be informed for the need to upgrade db. For that I'm open for comments from other people who have more insight on that and how these style of things have been communicated previously.

Opening this as draft as it might not be desired approach to use 3rd party container for migration.

- Related Issue #3568
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

Postgresql container version changes and it doesn't get up with version 13 volume without migration. Also if database-volume is migrated to version 17, it doesn't start up again with version 13 container.

## Documentation

### Tests

### DIFF
--- a/docker-compose-upgrade-db.yml
+++ b/docker-compose-upgrade-db.yml
@@ -1,0 +1,14 @@
+services:
+  db_update:
+    image: docker.io/pgautoupgrade/pgautoupgrade:17-bookworm@sha256:ead773b5ace405c8ea92375c63216d7677a618f6bef8708f8c52ac92427b23cf
+    env_file: .env
+    environment:
+      PGAUTO_ONESHOT=yes
+    volumes:
+      - pgdata:/var/lib/postgresql/data
+    networks:
+      - main
+volumes:
+  pgdata:
+networks:
+  main:
--- a/bw-dev
+++ b/bw-dev
@@ -96,6 +96,9 @@ case "$CMD" in
     init_ssl)
         $DOCKER_COMPOSE --file ./docker-compose-init_letsencrypt.yml up --exit-code-from=certbot
         ;;
+    upgrade_db_version)
+        $DOCKER_COMPOSE --file docker-compose-upgrade-db.yml up --exit-code-from=db_update
+        ;;
     resetdb)
         prod_error
         $DOCKER_COMPOSE rm -svf
@@ -286,6 +289,7 @@ case "$CMD" in
         echo "    up [container]"
         echo "    down"
         echo "    service_ports_web"
+        echo "    upgrade_db_version"
         echo "    initdb"
         echo "    resetdb"
         echo "    makemigrations [migration]"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       - ./certbot/logs:/var/log/letsencrypt
       - ./certbot/data:/var/www/certbot
   db:
-    image: postgres:13
+    image: postgres:17
     env_file: .env
     healthcheck:
       test: ["CMD-SHELL", "pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}"]


docker-compose-upgrade-db.yml
```diff
@@ -0,0 +1,14 @@
+services:
+  db_update:
+    image: docker.io/pgautoupgrade/pgautoupgrade:17-bookworm@sha256:ead773b5ace405c8ea92375c63216d7677a618f6bef8708f8c52ac92427b23cf
+    env_file: .env
+    environment:
+      PGAUTO_ONESHOT=yes
```
hughrun: This causes an error: `services.db_update.environment must be a mapping`.

It should be:

```yaml
environment:
  PGAUTO_ONESHOT: yes
```

ilkka-ollakka: Ah, good catch, it should be either `- PGAUTO_ONESHOT=yes` or `PGAUTO_ONESHOT: yes`. I'll fix it.

ilkka-ollakka: I'll also check if I can get those `celery_beat` warnings visible and if there is something that can be done for those.

ilkka-ollakka: Only celery_beat warning that I spotted was:

```
WARNING/MainProcess] /usr/local/lib/python3.11/site-packages/celery/worker/consumer/consumer.py:507: CPendingDeprecationWarning: The broker_connection_retry configuration setting will no longer determine
[celery_worker]  | whether broker connection retries are made during startup in Celery 6.0 and above.
[celery_worker]  | If you wish to retain the existing behavior for retrying connections on startup,
[celery_worker]  | you should set broker_connection_retry_on_startup to True.
[celery_worker]  |   warnings.warn(
```

and I don't think it relates to database-version, but more celery-configurations?

hughrun: Ok I tried this again and it seems to work, but I still get hundreds of lines of logs that are variations on this:

```sh
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] ^
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] ^
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] ^
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] ^
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] ^
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]   File "/usr/local/lib/python3.11/site-packages/django_celery_beat/schedulers.py", line 252, in __init__
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] Scheduler.__init__(self, *args, **kwargs)
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]   File "/usr/local/lib/python3.11/site-packages/celery/beat.py", line 264, in __init__
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] self.setup_schedule()
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]   File "/usr/local/lib/python3.11/site-packages/django_celery_beat/schedulers.py", line 260, in setup_schedule
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess] self.install_default_entries(self.schedule)
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
celery_beat-1     | [2025-07-14 04:34:05,674: WARNING/MainProcess]
```

So most lines are blank or just read `^`, but even when there is a line with a reference to a file, it's not particularly clear what it means and whether admins should be concerned.

I did notice some kind of message about different operating systems when the upgrade ran, but I didn't manage to capture the message, sorry.

Here I am upgrading from the latest `main`.

This doesn't _necessarily_ need a fix since they are just warning logs, but if we decide not to worry about it, I would at least want to understand why there are warnings and document it for people upgrading so it is not a surprise.

ilkka-ollakka: If you happen to have volume export of that database and you can share it, I could try to take a look what happens there?

I'll also try to setup new instance, start some import and stop/upgrade db so there are tasks in old database when it migrates.

ilkka-ollakka: with clean instance and importing goodreads csv, upgrading db before celery-worker jobs started, didn't seem to reproduce the messages you showed :/

hughrun: @ilkka-ollakka Ok I think I have worked this out, and it's probably going to irrelevant to nearly everyone. I'm running this on an Apple M2 (Silicon) machine and the pgautoupgrade image is, obviously, not built on this architecture (It appears to be AMD64). So I think it's just behaving weirdly for that reason. If I add `platform: linux/amd64` to your `docker-compose`, then it runs cleanly:

```yaml
    env_file: .env
    platform: linux/amd64
    environment:
      - PGAUTO_ONESHOT=yes
```

See [the docker-compose docs](https://docs.docker.com/reference/compose-file/services/#platform) for `platform`. Do you think it would be helpful to specify this, or will it cause issues on other hosts?

ilkka-ollakka: Great finding @hughrun! Issue might be that the sha-digest is amd64 image one. Pgautoupgrade has arm64 container too.

Does it work if you just remote the `@sha256:ead773b5ace405c8ea92375...` part from image without platform? Most likely you need to remote/redownload that image so it picks up native platform for your M2.

ilkka-ollakka: I rebased the PR and replaced digest with index digest instead of platform specific image digest. Tested locally and it seemed to be able to find both linux/arm64 and linux/amd64 platform images with the index digest.

hughrun: Cool, that works well!

Ok (sorry!) one last thing: I notice that the `db_update` container hangs around and is still connected to our database volume. It's probably only a problem right now for development work (`resetdb` throws because you can't delete a volume that's still in use), but I imagine that at some point this might cause confusion in production settings. Should the upgrade command in `bw-dev` clean up after itself by removing the container?

ilkka-ollakka: Just looked that and with upgrade-db we could just use `run --rm` instead of `up ` with docker compose call, same that is used with black etc tooling.

I'll update the call to use run, so it doesn't leave the container hanging around.

ilkka-ollakka: Adn thanks for testing and pointing out things that don't just work. I appreciate it a lot :)

hughrun (approved):

hughrun:
@ilkka-ollakka this looks good to go. Just wondering if we can provide upgrade instructions for non-Docker installs? We'll need to advise administrators in the release docs.

ilkka-ollakka:
> @ilkka-ollakka this looks good to go. Just wondering if we can provide upgrade instructions for non-Docker installs? We'll need to advise administrators in the release docs.

I think we could at least link to https://www.postgresql.org/docs/17/upgrading.html for options how to upgrade. 

I'm not sure is there any more instructions non-docker users. Most of the distros have some automatic upgrade via packaging, but simplest would be to take backup and restore to new database.

ilkka-ollakka:
I could try how does the django handles different postgresql versions, meaning that non-docker users could be fine if the postgresql is just more recent than 13. That could give non-docker users some option of non-needing to upgrade if they are already in new enough postgresql for django 5.2 LTS.

ilkka-ollakka:
I tested now with postgresql 14, 15 and 16, removing container volumes between. Build, tests and makemigrations with --check. All went without issues, so I don't see reason why non-docker people could stay in their current postgresql if it is newer than 13.

Issue #3373, opened 2024-06-07
I would like the max upload size to be configurable


ruudschilders:
**Is your feature request related to a problem? Please describe.**
A user wanted to import their .tar.gz file from another server, but the file was 40MB. The 'Import' button wouldn't work

**Describe the solution you'd like**
The `MAX_FILE_SIZE_BYTES` is currently hard-coded in bookwyrm.js, it would be nice if it's configurable in the .env file

**Describe alternatives you've considered**
I changed the value in bookwyrm,js but that will be overwritten at each update

**Additional context**
I am running Bookwyrm (https://bookwyrm.world) in docker.

hughrun:
Max upload size should be controlled by `DATA_UPLOAD_MAX_MEMORY_SIZE` in Django - which as of a recent change is controlled by an `env` setting of `DATA_UPLOAD_MAX_MEMORY_MiB` in BookWyrm to make it easier for admins.

I didn't realise we had another totally different custom setting in the JS file. This seems unnecessary - Django will reject an upload that's too big with an appropriate error (see #3121).

@mouse-reeve looks like you put this in with #1165 - do you remember why?

hughrun:
fixed in #3627

PR #3627, opened 2025-06-25
Make file-upload limit configurable and inform when import-file is too big

### LINKED ISSUE #3623: Import button gets disabled when select import file
**Describe the bug**
A user of our instance told us that she is trying to import her file (52 MB from .social) and, when she selects the file, the import button gets "unselected" (that is the word she used, I guess the button got fainted).

Can file size be the reason? Our instance (v 0.7.4) has no import limit (we have the "0" value on that setting).

**To Reproduce**
1. Go to https://BWinstance.org/import
2. Click on 'Browse' and select the 52 MB file
3. See error: "Import" button gets disabled

**Instance**
Our instance is v 0.7.4 now

ilkka-ollakka:
## Description

Instead of static 10MB limit on upload-file-size, take the limit from `DATA_UPLOAD_MAX_MEMORY_MiB` `.env` setting and use it in both nginx and django-side.

In Django-side, set data-attribute for file-form to tell what the max limit is, both MB and bytes, to make it easier for widget to tell error limit and javascript to check the size-limit on upload.

Use the `ClearableFileInputWithWarning` also in import-file-uploads, as it has already warning in place in case the file is too big.

in `bookwyrm.js` read the `data-max-upload` from file-input attributes and use that as limit on checking instead of static 10MB limit.

- Related Issue #
- Closes #3623, #3373, 

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests

- [X] I have not written tests and need help to write them

I'm not sure what is the best way to test those widget-form and javascript changes. Manually testing them seems to indicate it works as expected.

### DIFF
--- a/bookwyrm/templates/widgets/clearable_file_input_with_warning.html
+++ b/bookwyrm/templates/widgets/clearable_file_input_with_warning.html
@@ -25,7 +25,7 @@
 
     <input type="{{ widget.type }}" name="{{ widget.name }}"{% include "django/forms/widgets/attrs.html" %}>
     <span class="help file-cta is-hidden file-too-big">
-        {% trans "File exceeds maximum size: 10MB" %}
+        {% blocktrans with max_size=widget.attrs.max_mb %}File exceeds maximum size: {{ max_size }}MB{% endblocktrans %}
     </span>
 </p>
 
--- a/bookwyrm/forms/forms.py
+++ b/bookwyrm/forms/forms.py
@@ -6,6 +6,7 @@ from django.utils.translation import gettext_lazy as _
 
 from bookwyrm import models
 from bookwyrm.models.user import FeedFilterChoices
+from bookwyrm.models.fields import ClearableFileInputWithWarning
 from .custom_form import CustomForm
 
 # pylint: disable=missing-class-docstring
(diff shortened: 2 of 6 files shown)


bookwyrm/templates/widgets/clearable_file_input_with_warning.html
```diff
@@ -25,7 +25,7 @@
 
     <input type="{{ widget.type }}" name="{{ widget.name }}"{% include "django/forms/widgets/attrs.html" %}>
     <span class="help file-cta is-hidden file-too-big">
```
hughrun: Might be better for this to also have class `has-background-danger-light` ?

hughrun (commented):
This works well. A couple of comments:

1. Given this prevents submitting the form, I think it would be better to make the alert more obviously an error message.
2. I don't want to make this PR more complicated, so this probably should be looked at separately, but another place people have experienced confusing filesize upload errors is with cover image uploads. This will be more complex because it uses a custom `ResubmitImageWidget` to enable us to keep the file in memory when checking for existing authors.

bookwyrm/templates/widgets/clearable_file_input_with_warning.html
ilkka-ollakka: I'll add it. I didn't look that much on ResubmitImagewidget yet, but I can check if there is more than just defining same template for it needed.

ilkka-ollakka: Seems that ResubmitImageWidget also started working with just template/context updates. Added commit for that in this PR.

ilkka-ollakka:
rebased the PR on top of main

hughrun:
This looks good to me! Thanks for that update @ilkka-ollakka.

@mouse-reeve FYI this effectively changes the default update limit to 100MB if it's not set, which I think is a good thing but probably should be mentioned in release notes.

PR #2467, opened 2022-12-03
docker: move initial setup into docker startup to simplify deployment

osresearch:
This patch makes it easier to perform non-development installations with `docker-compose` by performing the initial setup in the `web` container on first startup (while the `celery` and `flower` containers wait for it to complete).

The production `docker-compose.yml` also needs an update to pull from an official image (I'm testing it with my OIDC patches in #2464 here):
```
version: '3'

services:
  nginx:
    image: nginx:latest
    restart: unless-stopped
    ports:
      - "1333:80"
    depends_on:
      - web
    networks:
      - main
    volumes:
      - ./nginx:/etc/nginx/conf.d:ro
      - ./data/static_volume:/app/static:ro
      - ./data/media_volume:/app/images:ro
      - ./data/cache:/cache

  db:
    image: postgres
    env_file: .env
    volumes:
      - ./data/pgdata:/var/lib/postgresql/data
    networks:
      - main

  redis_activity:
    image: redis
    command: redis-server --requirepass ${REDIS_ACTIVITY_PASSWORD} --appendonly yes --port ${REDIS_ACTIVITY_PORT}
    volumes:
      - ./redis.conf:/etc/redis/redis.conf
      - ./data/redis_activity_data:/data
    env_file: .env
    networks:
      - main
    restart: on-failure

  redis_broker:
    image: redis
    command: redis-server --requirepass ${REDIS_BROKER_PASSWORD} --appendonly yes --port ${REDIS_BROKER_PORT}
    volumes:
      - ./redis.conf:/etc/redis/redis.conf
      - ./data/redis_broker_data:/data
    env_file: .env
    networks:
      - main
    restart: on-failure

  web:
    image: osresearch/bookwyrm:oidc
    command: python manage.py runserver 0.0.0.0:8000
    env_file: .env
    volumes:
      - ./data/static_volume:/app/static
      - ./data/media_volume:/app/images
    environment:
      - DB_INIT=True
    depends_on:
      - db
      - celery_worker
      - redis_activity
    networks:
      - main
    ports:
      - "8000"

  celery_worker:
    image: osresearch/bookwyrm:oidc
    command: celery -A celerywyrm worker -l info -Q high_priority,medium_priority,low_priority
    env_file: .env
    volumes:
      - ./data/static_volume:/app/static
      - ./data/media_volume:/app/images
    depends_on:
      - db
      - redis_broker
    restart: on-failure
    networks:
      - main

  celery_beat:
    image: osresearch/bookwyrm:oidc
[...]

Issue #2095, opened 2022-05-03
building / publishing generic Docker images


magnus919:
**Is your feature request related to a problem? Please describe.**
This seems a convention with a few Fediverse apps, but I'm noticing the Docker instructions include setting up environment variables and then having the user build their own container images rather than pulling generic images and externalizing environment. This could really boost participation and hosting of more instances!

**Describe the solution you'd like**

* Publish regular tagged container images as part of build process to Docker Hub.
* Amend documentation to avoid end users having to build, and instead shift toward consuming these images (I think I can help figure out this part).

**Describe alternatives you've considered**
I could follow the Docker deployment instructions as-provided, but it would be super weird in my setup. Also the way these are (not) provided now means it can't be deployed to K8s without extra work.

**Additional context**
Not sure how much I can help with sorting out a generically consumable image build (I'd like to help). But I am pretty sure I can help out with consumption doc.

magnus919:
Just expounding a bit...

Generic Docker images would free end users up to easily deploy to Docker, Docker Swarm, Kubernetes, CloudFoundry, OpenShift, and probably other container platforms I'm not considering.

ChaosExAnima:
💯 to this! I was actually looking to spin up an instance but dismayed to find prebuilt packages aren't available.

I've done this for a few of my own repos, so for automated docker builds I'd suggest leveraging your existing GitHub Actions to incorporate this :octocat:-made one: https://github.com/marketplace/actions/build-and-push-docker-images

You can probably host the prebuilt images here, but the defacto seems to be hosting with Docker Hub. It should be easy to create an official BookWyrm org there and have each image be it's own repo. If you have access to a paid Docker account, you can skip GitHub entirely and just use their automated builds.

I'll also add as I see this a lot with OSS Docker images: [please support Docker+K8S secrets](https://docs.docker.com/engine/swarm/secrets/#build-support-for-docker-secrets-into-your-images)! It's easy to pass in env vars for sensitive data, but storing that info in either a `.env` or a `docker-compose` file isn't as secure. It may not make sense for a standalone setup, but if you have a k8s or Swarm setup already it would be nice.

skobkin:
Came here to post issue about that.

[...]

PR #3636, opened 2025-07-06
Build container and release it to github, merge files to pyproject.toml and handle basic setup/migrations automatically in container entrypoint-script

### LINKED ISSUE #3551: Consider packaging releases as containers
There have been a few suggestions in the Matrix chat about releasing production versions as pre-built containers rather than pulling the full repository and then running `docker compose build` to build the latest version.

This Issue is to give that discussion a presence here and flesh out what it might potentially look like, potential ramifications etc.

## The basic idea

As I understand it, the principle is that instead of `pull`ing the whole git repository, potentially adjusting the `docker-compose.yml` file to personal needs, and then `build`ing the docker images locally, a BookWyrm administrator would reference a prebuilt, hosted docker container in a local `compose` file and pull the container from the container repository. This gives administrators a cleaner environment and reduces the likelihood of merge conflicts where admins have made local adjustments to deployment files to suit their own needs.

## An example of how it might work

One example of how this _might_ work is the Plausible Analytics project. Plausible runs a commercial business hosting less-invasive website analytics, and the software itself is open source.

They run **development** in [the main repository](https://github.com/plausible/analytics).

When there is a **release**, [a new versioned container](https://github.com/plausible/analytics/pkgs/container/community-edition) is published using [GitHub Packages](https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages). 

This container is referenced in [a separate plausible/community-edition](https://github.com/plausible/community-edition) **repository with just a compose file, license and readme**. Admins simply clone this lightweight repository and the production container is pulled down when they deploy.

## Considerations

We have a little bit of tooling for both devs and admins, primarily as the `bw-dev` script. This would probably need to be broken up into dev and prod versions and included in the "production repository" with the compose file. There would also need to be some thinking around what appropriately goes in an `.env` file (e.g. ports) and various needs around web server setup (reverse-proxies on the host machine for example).

At a practical level, anyone responsible for creating a new release would need to have appropriate permissions to publish the container.

## Possible advantages for version releases

Although in some ways this is more complex from a dev and release perspective, in other ways it has the potential to simplify things. Currently we're running both `main` (for development) and `production` (for production deployments), and @mouse-reeve has to do a bit of a song and dance to merge into `production` but retain some differences. If production config was all in a completely separate `compose` file potentially even in a different repository like Plausible community edition, version release could be as simple as building a new container, loading it as a new package, and adjusting the referenced container version in the production repository.

## Related issues and discussions

This could potentially resolve or influence a number of existing issues:

#793 
#1786 
#2095 
#3279 
#3289 
#3377

ilkka-ollakka:
## Description

This PR contains changes to build bookwyrm container and release it to be used without closing the source repository. Currently it builds arm64 and amd64 platforms.

Container builds are done when tag is pushed to use tag as version, for example `bookwyrm:0.8.0`. Main repo push build container with name `bookwyrm:main`.

It also combines requirements.txt, pylint/mypy/coverage configs to pyproject.toml config.

Container entrypoint script runs migrations,initdb and permissions checks before changing to `bookwyrm`-user to start provided command. So Celery/bookwyrm services are running non-root user. I used the https://github.com/bookwyrm-social/bookwyrm/pull/2467 as initial entrypoint script and modified it littlebit.

`bw-dev` command now has command `dev` to start local code in container instead of build container. It should allow local development easily. `dev`-config also sets `DEBUG=1` environment variable, making heavy assumption that you want to run it with debug-mode.

> One flow is not yet handled properly, that is that celery doesn't wait migrations after initial migrations, so new migrations are done, but celery/others doesn't know to wait them to finish before starting.

I modified Dockerfile and entrypoint for this, to take build-timestamp as marker what to look for in file when waiting migration checks, still might not be fully working, but should handle new builds/migrations to be waited correctly.

Changes include using dev-tools container changes, build it on top of bookwyrm container, install all dev-dependencies there instead of main container and run all pytest/mypy etc using that container.

- Related Issue #2095 
- Closes #3551 

- [ ] Bug Fix
- [X] Enhancement
- [X] Refactor

We don't anymore mount local directory to container, but build container. We also run migrations and database-initdb autotically on each start.

We run pytest/mypy/etc in derived container that mounts source-code in.

Dev flow now uses `bw-dev dev` instead of `bw-dev up` to start local code.

`requirements.txt`, `.pylintrc`, `.coveragerc` and `mypy.ini` are all merged to `pyproject.toml`

## Documentation

### Tests

I think this needs some manual testing, as it is big change and it is probable that I have missed some flow/use-case that is broken.


ilkka-ollakka:
rebased the PR on top of current main

ilkka-ollakka:
If there is something that you are not sure/just curious, please do ask and I'll try to clarify what it does.

Dockerfile
```diff
@@ -1,12 +1,43 @@
...
+
+COPY --from=build /app /app
+COPY --from=build /venv /venv
+COPY README.md LICENSE.md VERSION /app/
+COPY manage.py gunicorn.conf.py /app/
+COPY celerywyrm /app/celerywyrm
+COPY locale /app/locale
+COPY bookwyrm /app/bookwyrm
+COPY entrypoint.sh /entrypoint.sh
+RUN date +%s > /build_timestamp
```
ilkka-ollakka: Instead of plain timestamp, we could run for example hash of migration-files hashes to have identifier. Biggest difference would be that it wouldn't recheck the migrations if there is no changes, so in release flow not huge difference.

ilkka-ollakka:
I'll add commits to this PR to have development flow working with local source-files.

ilkka-ollakka:
I moved dev-tools to dev-compose file and point docker-compose bookwyrm images to ghcr.io, so release-wise the docker-compose modifications would be just adding some production related db-changes and change bookwyrm:main to bookwyrm:release-number.

marcusyoung:
@ilkka-ollakka Can I check that consideration has been given to any potential impact of these changes on dockerless installs. For example, I noticed that requirements.txt has been removed, But we need to run `./venv/bin/pip3 install -r requirements.txt` on a dockerless install during upgrade. There may be other things, but that's the one that stood out.

ilkka-ollakka:
> @ilkka-ollakka Can I check that consideration has been given to any potential impact of these changes on dockerless installs. For example, I noticed that requirements.txt has been removed, But we need to run `./venv/bin/pip3 install -r requirements.txt` on a dockerless install during upgrade. There may be other things, but that's the one that stood out.

Ah yes, good point. For now it should be something like `pip install "pip>=25.1.0" && pip install --group main . ` so to have new enough pip to support group and it should be able to use the `pyprojet.toml`-file just fine.

ilkka-ollakka:
Rebased on top of current main and update pyproject.toml with new Django-version from updated requirements.txt.

Dockerfile
ilkka-ollakka: I changed the entrypoint/Dockerfile for hash-flow instead of build-timestamp.

ilkka-ollakka:
rebased on top of current main with updated requirements.txt changes.

ilkka-ollakka:
rebased on top of main to update requirements changes

ilkka-ollakka:
I'll split separate PR's out from this, separatable changes so those are easier to review/merge separately after current release is in order.

hughrun:
@ilkka-ollakka I'm trying to understand this. Can you explain how `entrypoint.sh` gets called? I can't see it referenced anywhere. I'm also wondering if we can use `depends_on` with a (maybe simpler?) health check instead?

ilkka-ollakka:
> @ilkka-ollakka I'm trying to understand this. Can you explain how `entrypoint.sh` gets called? I can't see it referenced anywhere. I'm also wondering if we can use `depends_on` with a (maybe simpler?) health check instead?

Ah yes, in Dockerfile we define 'ENTRYPOINT' pointing to that script in line 42. So all commands get run as parameter to entrypoint script.

Documentation on the difference of ENTRYPOINT and CMD can be found in https://docs.docker.com/reference/dockerfile/#understand-how-cmd-and-entrypoint-interact so docker/podman itself handles the calling

I think it is good idea use health checks to make entrypoint simpler.

Or did you have some idead to use healtchecks instead entrypoint script? That would be also nice approach, and if you have something started with that flow, I would be happy to take a look and adjust the PR?

ilkka-ollakka:
@hughrun I think the initial idea with entrypoint instead of depends/healtcheck/separate container was that with entrypoint it doesn't necessary need to configure separate containers if someone is not using bookwyrms docker-compose file.

But now saying it out loud, I think those cases can easily accomodate their setups to health-checks or other approaches.

hughrun:
> Ah yes, in Dockerfile we define 'ENTRYPOINT' pointing to that script in line 42. So all commands get run as parameter to entrypoint script.

I completely missed this reference! That makes sense now.

I don't have any particular suggestions or better ideas – I just figured this looks like the sort of thing that `depends`/`healthcheck` is designed for, and I feel that would be a little more intuitive to understand what it is doing.

ilkka-ollakka:
> I don't have any particular suggestions or better ideas – I just figured this looks like the sort of thing that `depends`/`healthcheck` is designed for, and I feel that would be a little more intuitive to understand what it is doing.

I think depends is more of start-order of containers and healtcheck to know they are operating correctly or needing restart. I agree that depeds is used to do some one-off things before main container starts. Entrypoint is used more internally in container to make sure things are correct when starting up, reducing the need of external container/command runs.

I agree that entrypoint in current form could do some clarification on what it actually does and what the flow of things is. I'll atleast add some comments and and try to make it more clear how it is used.

ilkka-ollakka:
I modified the docker-compose to have depends_on to web-container being healthy on bookwyrm containers and defined healthcheck to web-container to check that django is up. Also cleaned out entrypoint to run migrations and not to do polling in other containers.

ilkka-ollakka:
also rebased on top of curren main

SensibleSalmon:
Hi all, very happy to see this as someone looking to deploy this on a swarm. Is there any additional testing or development that I could help with on this? I see #3759 is approved and passing build; I'm happy to jump in wherever is useful. I think a canonical docker image is a fantastic addition to the project.

ilkka-ollakka:
> Hi all, very happy to see this as someone looking to deploy this on a swarm. Is there any additional testing or development that I could help with on this? I see #3759 is approved and passing build; I'm happy to jump in wherever is useful. I think a canonical docker image is a fantastic addition to the project.

Thats nice to hear! I've been splitting this PR to more bite-size PRs and I think only things left currently are:
1) build full container image and publish it, most likely I'll split this into 2 separate PR
2) run container as non-root user
3) dev-tools/etc fiddling

I most likely try to split those remaining things during this or next week to new PRs, and if you can take a look and test/comment, it would be great. Ofcourse if you have some questions/comments already now in things, I would love to hear those.

SensibleSalmon:
> Ofcourse if you have some questions/comments already now in things, I would love to hear those.

Sure thing! My docker compose experience is mildly limited; I have ~10 stacks running across either my NAS or my swarm and while my employer uses containers heavily I'm a step removed from the infra teams. I put myself firmly in the "know enough to be dangerous" category :sweat_smile: so just be aware that my questions are just as likely to come from a lack of (beginner) knowledge as being a structural concern, and I'll try to flag them accordingly.

I'll add what questions I have against the PR and I'll see if I can get an instance running on my swarm with the compose file in this PR / the `ghcr.io/bookwyrm/bookwyrm:main` image, though it might be the weekend before I have the time to do that.

docker-compose.yml
```diff
@@ -45,20 +46,23 @@ services:
     networks:
       - main
   web:
+    image: ghcr.io/bookwyrm/bookwyrm:main
     build: .
```
SensibleSalmon, line 100: curiosity: for the other services in the compose file you removed the `build` directive in favor of the `image: ghcr.io/bookwyrm/bookwyrm:main`; why keep the `build: .` here?

pyproject.toml
```diff
@@ -1,2 +1,151 @@
...
+env = [
+    "LANGUAGE_CODE = en-US",
+    "SECRET_KEY = beepbeep",
+    "DEBUG = false",
+    "DOMAIN = your.domain.here",
+    "PORT = 4242",
+    "ALLOWED_HOSTS = your.domain.here",
+    "BOOKWYRM_DATABASE_BACKEND = postgres",
+    "MEDIA_ROOT = images/",
+    "CELERY_BROKER_URL = memory://",
```
SensibleSalmon, line 90: curiosity: is this meant to be a default? in the docker-compose case, would `memory://` ever work when the broker is a separate container?

entrypoint.sh
```diff
@@ -0,0 +1,59 @@
...
+    # unnecessarily
+    if [ ! -r "$CANARY" ]; then
+        info "**** Doing initial database population"
+        python manage.py initdb || die "failed to initdb"
+    fi
+    # We are running container as non-root id, so check that permissions are correct
+    info "**** Checking images and exports directory permissions to be correct"
+    chown -c -R bookwyrm /app/exports
+    chown -c -R bookwyrm /app/images
+    chown -c -R bookwyrm /app/static/css/themes
```
SensibleSalmon: curiosity: what happens when the `bookwyrm` user doesn't exist on the host system? Docker user/group mapping is not something I understand well, and I feel like I often run into filesystem permissions errors.

ilkka-ollakka: it doesnt need to exist in host system. bookwyrm account is created inside container in build phase. Entrypoint is also runned inside the container in container start.

docker-compose.yml
ilkka-ollakka: allowing development and local build. Also as all have same image it only builds it once, atleast podman compose builds currently all images separately.

pyproject.toml
ilkka-ollakka: this is oblyfor pytest run variables, so not default otherwise. It is in tool pytest.ini_options block

docker-compose.yml
```diff
@@ -84,55 +88,70 @@ services:
...
     depends_on:
-      - db
-      - redis_broker
+      web:
+        condition: service_healthy
+      db:
+        condition: service_healthy
+      redis_broker:
+        condition: service_started
     restart: on-failure
```
SensibleSalmon: this conflicts with the `restart: always` directive on line 105, above

docker-compose.yml
```diff
@@ -84,55 +88,70 @@ services:
...
+      retries: 6
     volumes:
-      - .:/app
       - static_volume:/app/static
       - media_volume:/app/images
       - exports_volume:/app/exports
+    restart: always
     depends_on:
       - celery_worker
     restart: on-failure
```
SensibleSalmon: another pair of conflicting restart directives

docker-compose.yml
```diff
@@ -84,55 +88,70 @@ services:
...
     depends_on:
-      - db
-      - redis_broker
+      db:
+        condition: service_healthy
+      redis_broker:
+        condition: service_started
+      web:
+        condition: service_healthy
     restart: on-failure
```
SensibleSalmon: and another

ilkka-ollakka: good catches, I'll try to remember to sort those out when I rebase/split remaining things out.

hughrun:
@ilkka-ollakka is the idea that we won't actually need this PR once we merge #3803 #3809 plus maybe some small cleanup after that?

ilkka-ollakka:
> @ilkka-ollakka is the idea that we won't actually need this PR once we merge #3803 #3809 plus maybe some small cleanup after that?

yes, idea is to split things to smaller PRs and then this can be closed.

ilkka-ollakka:
@hughrun just making sure, there are still few PRs that are needed beside those that are already open. Mainly the publish change is one and the non-root/dev-tools thing is optional but nice to have.

ilkka-ollakka:
closing this PR as relevant changes are merged from PR splitted out from this.

Issue #3644, opened 2025-07-19
User import fails


marcusyoung:
**Describe the bug**
Importing a user fails.

**To Reproduce**
Import a user

**Expected behavior**
The selected objects should be imported for the user

**Instance**
- ramblingreaders.org
- Using S3-comaptible storage (Cloudflare R2)
- Dockerless install

**Additional context**
- We are running version 0.7.5 though that version isn't checked out as I would have pulled latest production branch when I upgraded. We are at commit e217a17ba with manual fix from commit 9e544e86f applied. 
- We are using S3-compatible storage.
- User export is not enabled on this instance
- When a user import is attempted (using an export from another instance), the tar file is uploaded and saved in the /images folder of the S3 bucket. 
- In the database entries are created in `bookwyrm_bookwyrmimportjob` and `bookwyrm_parentjob` 
- The status column in `bookwyrm_parentjob` switches to fail immediately (or as fast as I can refresh the table!)
- I have examined journal and system logs and can't find any relevant errors. If you can provide a likely candidate string for a grep search across the logs I can try that.

hughrun:
@marcusyoung I'm looking into this and there's a few things we need to tidy up with user import, but I'm wondering whether you know which version of BookWyrm the user export file came from?

marcusyoung:
@hughrun Unfortunately not, I don't think that information is recorded with the export? One was from bookwyrm.social and the other theunseen.city. Both within the last couple of weeks.

hughrun:
"within the last couple of weeks" is good enough 😄 bw.s exports will definitely be `0.7.5`, so we can now rule out incompatible export file formats.

hughrun:
Ok I've done some preliminary testing with two basic export files and I can't replicate this yet. I have noticed however a few things that need to be fixed regardless:

1. we should be uploading to an `imports` directory rather than the `images` directory (which is the default), for the same reason we don't want to send exports to `images`.
2.  we should be deleting the file after the import completes. I considered uploading as a temporary file, but some exports could get pretty big, so that is probably not a great idea as it will choke the instance.
3. we don't send a notification when the import finishes, so users have to check the user import page which is fine I guess but a notification would be nice, like we do for exports

@marcusyoung I'm wondering if one of your users experiencing this problem would be willing to share their export file with me to assist with troubleshooting? If so, let me know on Matrix and I'll share my email address.

marcusyoung:
@hughrun will make contact via Matrix.

[...]

PR #3653, opened 2025-08-03
Prefer existing book data from the fedi when importing user accounts

hughrun:
This is the first of four related PRs intended to improve user import and export handling.

This first one addresses a number of issues discussed in #3644. Apologies for the extra-large PR, I had trouble working out how to split this up in a way that made sense. This PR should:

- fix notifications not working for completed user imports
- fix import file erroneously saving in `/images` (`default` storage), changing it to save in the `exports` storage
- fix import and export files potentially being saved in public S3 buckets
- change to a default of saving import and export files in local storage even when S3 storage is in use
- delete import files after import is finished
- import books and covers from external instances by default instead of needing export files with many cover images
- import books (without their covers) from the JSON import as a last resort
- improve logging for production instances
- rename `UserRelationshipImport` for more consistent names

## Storage configuration changes

This PR will change the default for export/import files to be saved in the local file system, even when `USE_S3` is `True`. This can be overridden with a new ENV variable `USE_S3_FOR_EXPORTS` if admins really want to save export files in S3, however if they do so they will need to set S3 values for a separate bucket for exports. This is so that instances using S3 services without ACL (e.g. Backblaze B2 or Cloudflare R2) can set the exports bucket to private.

This may seem convoluted, but the intention is that with the associated PRs, export files will be periodically deleted from storage, and will also be significantly smaller since they won't include cover images. Therefore the amount of file storage required will be quite minimal, meaning S3 is unnecessary. With standard configuration, local `/exports` directory should be secure, whereas S3 brings in complication. This also possibly means we could add the ability to enable user exports for Azure, which to date is not possible.

## Description

- Related Issue #3644
- Related PRs: #3654, #3655, future PR for automatically scanning for covers from connectors

- [x] Bug Fix
- [x] Enhancement
- [x] Refactor

## Documentation

### Tests


ilkka-ollakka (approved):

ilkka-ollakka (approved):

hughrun:
I meant to explain a little more about why we didn't do this originally, and why I think we should have.

The initial thinking for user imports was that users should be able to completely recreate their books etc even if the instance they were coming from had already disappeared. At the time there were many Mastodon instances disappearing virtually overnight, and `bookwyrm.social` was experiencing significant performance issues. The idea was that the "source" instance should not be hit with a heap of requests every time someone migrates to a new instance, and that we shouldn't assume it's still available.

I have realised, however, that this is likely to happen anyway because when you import a book, BookWyrm looks for relevant reviews, and the authors of those reviews etc. So all we're doing is making gigantic export files which solve the fairly niche problem of "many of my books are only on my source instance and it has already gone offline". It seems to me that the trade-off isn't worth it. This rethink was prompted by a very reasonable question from @marcusyoung asking why we don't just do it like the csv book import.

bookwyrm/models/bookwyrm_import_job.py
```diff
@@ -174,84 +264,105 @@ def start_import_task(**kwargs):
...
 
-            for item in UserRelationshipImport.objects.filter(parent_job=job).all():
+            for item in UserImportRelationship.objects.filter(parent_job=job).all():
                 item.start_job()
 
+            url_parts = urlparse(job.import_data.get("id"))
+            url = f"{url_parts.scheme}://{url_parts.netloc}"
+            # Check https://example.com to see if the instance is still online
+            # If not, we don't bother trying to pull book data from it.
+            resp = requests.get(
```
mouse-reeve: Would this be more efficient as `requests.head`?

bookwyrm/models/bookwyrm_import_job.py
```diff
@@ -174,84 +264,105 @@ def start_import_task(**kwargs):
...
+                url,
+                headers={
+                    "User-Agent": settings.USER_AGENT,
+                },
+                timeout=10,
```
mouse-reeve: Does it make sense to use the [system timeout variable](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/settings.py#L247)?

bookwyrm/models/bookwyrm_import_job.py
```diff
@@ -174,84 +264,105 @@ def start_import_task(**kwargs):
...
+            )
```
mouse-reeve: I'm running into a connection refused error on this block, albeit running in localhost without ngrok and not configured to communicate with the outside world. While that's not how a live instance would be operating, should this block be resilient to errors like these? Would this behave correctly for a garbled url?

hughrun: Good point, its' a bit flaky like this. I was assuming `resp.ok` would take care of this but if there's an issue at our end it probably won't. I guess we should allow it to complete the import however it can.

hughrun:
Thanks for the reviews @ilkka-ollakka and @mouse-reeve !

I'll make some updates in line with Mouse's comments and then ask for a final check/pull.

PR #3655, opened 2025-08-03
Delete old export and import files with a scheduled job

hughrun:
Adds a scheduled job to delete old export and import jobs

- Admins can determine how often the job runs, and how old the files should be before they are deleted, from a new page in the Admin settings.

- Defaults to 72 hours old, with no job scheduled.

- **Note** that this will only delete files associated with a known BookwyrmImportJob or BookwyrmExportJob, where the file is using the storage currently set in ENV. We will need release notes explaining the need to delete all other files.

- Old export and import files in /images and using old /exports storage (e.g. S3 when exports has been changed to local storage) will need to be deleted manually.

Screenshot 2025-08-16 at 3 15 58 PM

- Related Issue: #3644
- Related PR: #3653

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -321,6 +321,7 @@ case "$CMD" in
         echo "    upgrade_db_version"
         echo "    initdb"
         echo "    resetdb"
+        echo "    init_ssl"
         echo "    makemigrations [migration]"
         echo "    migrate [migration]"
         echo "    create_secrets"


bookwyrm/templates/preferences/export-user.html
```diff
@@ -96,6 +96,11 @@ <h2 class="title">{% trans "Recent Exports" %}</h2>
     <p class="content">
     {% trans "User export files will show 'complete' once ready. This may take a little while. Click the link to download your file." %}
     </p>
+    <p class="content">
+        {% blocktrans %}
```
mouse-reeve, line 100: This needs to be `{% blocktrans trimmed %}` or the whitespace will be added to the translate file and mess up it's formatting

bookwyrm/templates/settings/files.html
```diff
@@ -0,0 +1,212 @@
...
+                {{ task_form.period.label }}
+                </label>
+                <div class="select">
+                    {{ task_form.period }}
+                </div>
+                <p class="help" id="desc_period">
+                    {{ task_form.period.help_text }}
+                </p>
+            </div>
+            <button class="button is-warning">{% trans "Schedule scan" %}</button>
```
mouse-reeve, line 95: I didn't see an obvious reason why this would be warning colored. I think it would be better to either change it to `is-primary` or add text explaining it.

bookwyrm/templates/settings/files.html
```diff
@@ -0,0 +1,212 @@
...
+{% trans "Files maintenance" %}
+{% endblock %}
+
+{% block header %}
+{% trans "Files maintenance" %}
+{% endblock %}
+
+{% block panel %}
+
+<div class="box block">
```
mouse-reeve, line 16: I found it a little confusing to have both these forms in the same "box" element

bookwyrm/templates/preferences/export-user.html
```diff
@@ -96,6 +96,11 @@ <h2 class="title">{% trans "Recent Exports" %}</h2>
...
+        Export files will be deleted after {{ expiry_hours }} hour{{ expiry_hours|pluralize}}.
```
mouse-reeve, line 101: This filter didn't work for me (it displays "72 hour" instead of "72 hours"). It would be better for the translation files to format it using the verbose method:

```
{% blocktrans trimmed count counter=expiry_hours %}
Export files will be deleted after {{ expiry_hours }} hour.
{% plural %}
Export files will be deleted after {{ expiry_hours }} hours.
{% endblocktrans %}
```

hughrun:
Thankyou! I still get confused about blocktrans 🙁

PR #3657, opened 2025-08-06
Automatically parse series from title in openlibrary connector and try to retrieve series info in inventaire connector

### LINKED ISSUE #3255: Series number ends up in series name
**Describe the bug**
Many books I checked have "Series Name #42" as their series name and an empty series number.

**To Reproduce**
1. Go to https://bookwyrm.social/book/194070/s/mutineers-moon
2. Click the edit button

**Expected behavior**
Series should be "Mutineer's Moon" and series number should be "1".

**Screenshots**
image

**Instance**
bookwyrm.social

**Additional context**
n/a

ilkka-ollakka:
## Description

Modify openlibrary connector to split series-field to series-name and series-number.

Add identifiers for inventaire to parse series-name and series-number data if found.

- Related Issue #
- Closes #3255

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
I don't want to block this PR, but this implementation is fairly fragile because of how Open Library deals with series (badly!).

For example, Terry Pratchett's _Small Gods_ has the following unique entries in `series`:

* A Discworld novel
* The Discworld series
* Discworld -- 13
* Discworld, Book 13
* Discworld
* Ploskiǐ mir
* Discworld (13)
* Les annales du Disque-Monde (13)
* Harper fantasy

So just for this one book we have to manage series numbering alternatives like:

* Title -- {number}
* Title, Book {number}
* Title (number)

Also for the same book I can't get any series information at all when importing from Inventaire, even though it's definitely in wikidata and Inventaire - there appears to be some issue parsing the series name into a `Work`, but I don't want to scope-creep this too much so I'll just note that. I was starting to think we should maybe preference WikiData/Inventaire as a data source because it's so much better structured, but if we're not actually getting any data at all then that's a bigger problem.

@ilkka-ollakka I'm interested in your thoughts on next steps on this. I'd be ok to pull something with a few more variations since it will be better than what is happening now – as long as everyone understands that this is a temporary and fragile "fix". Ultimately we probably should treat series as its own model, but incoming data quality is always going to be a problem.

ilkka-ollakka:
Thanks for the comment @hughrun, I'll take a look for the small gods on inventaire side, could be that I have some mistake in parsing or missed something.

Also on openlibrary for same book, I'll check it out. But I agree that it is 'hopen for the best' level at best from there to get the series info.

hughrun:
Hmm I was so busy trying to troubleshoot I missed the fact you added custom mapping for series! Series needs to be a string, and your change turns it into a list. But even if I adjust it to choose the first item in the list it still doesn't parse into a `Work` properly so I'm not clear what's happening.

hughrun:
@ilkka-ollakka Sorry, just to clarify: as far as I can tell, we are not actually pulling in series names from Inventaire _using the current `main` code_ – this is not because of the changes you made in this PR, it's a pre-existing problem. I've also noticed we're not pulling in cover images within the search screen.

I think if you focus the scope of this change to OpenLibrary and cover a few extra variations (at "best guesss" level, as you suggested) that would be a good improvement.

We then can worry about the Inventaire problems separately rather than muddy things up here.

ilkka-ollakka:
I'll split out the Inventaire bits to separate PR, so it can be also cleaned up nicer if needed and I'll keep the openlibrary ones in here and add some guesses on series things.

ilkka-ollakka:
I took the inventaire bits out from this and added some regex-guesses for series-name/number extraction.

PR #3664, opened 2025-08-11
Add ability to view and terminate user sessions

### LINKED ISSUE #2278: Add an option to view and terminate activity sessions

### LINKED ISSUE #3082: Make sessions last longer

hughrun:
Adds a dashboard in user preferences to enable users to see all logged-in sessions under their account, and terminate any if they choose to do so.

Also renames 2FA page to "Security Settings" since it seemed reasonable to put these two functionalities together.

Because we use the cache engine for sessions, we can't efficiently check directly for all a user's sessions. Instead we add an object to a new UserSession model the first time they log in to a session. On logout, the actual session is terminated, and we delete the model object. Because sessions can expire, we check all associated sessions for a user's UserSession objects when they load the security page and remove any that no longer have active sessions.

Existing sessions when this is first rolled out will not appear in the dashboard since they won't have a UserSession object. Once they expire, this problem will resolve itself.

Users can't accidentally log out their current session ;)

This also [changes the default session age to 365 days](https://github.com/bookwyrm-social/bookwyrm/issues/3082).

light_sessions

dark_sessions

- Closes #2278 
- Closes #3082

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/models/session.py
```diff
@@ -0,0 +1,49 @@
...
+
+from bookwyrm.models import User
+
+
+class UserSession(models.Model):
+    """A session for a logged-in user. We only use this model to save info about
+    a logged-in session when the user first logs in, so users can remove sessions
+    e.g. for devices they have lost."""
+
+    user = models.ForeignKey("User", on_delete=models.PROTECT, related_name="sessions")
```
ilkka-ollakka: Should this be Cascade, so if user is deleted, sessions are also deleted?

hughrun: Uh, yes!

bookwyrm/templates/preferences/security.html
```diff
@@ -98,4 +101,65 @@ <h3>Backup codes</h3>
...
     {% endif %}
 </div>
+
+<div class="block">
+
+</div>
+<div class="block">
+    <h2 class="title is-4">
+        {% trans "Sessions" %}
+    </h2>
```
dato, line 112: The only glitch I noticed was that if I now go _and actually set up 2FA_, this section is shown (with no sessions) together with the QR and confirmation code and whatnot.

Should this block perhaps be hidden (or absent) if, say,`this_session` is not passed?

bookwyrm/templates/preferences/security.html
```diff
@@ -98,4 +101,65 @@ <h3>Backup codes</h3>
...
+        </p>
+    </div>
+    <div class="block">
+        <table class="table">
+            <thead>
+                <tr>
+                    <th><abbr title="{% trans 'Date first logged in' %}"></abbr>{% trans "Date" %}</th>
+                    <th><abbr title="{% trans 'IP address' %}"></abbr>{% trans "IP" %}</th>
+                    <th><abbr title="{% trans 'Operating System' %}"></abbr>{% trans "OS" %}</th>
+                    <th><abbr title="{% trans 'Web Browser' %}"></abbr>{% trans "Browser" %}</th>
```
dato: Here the `<abbr>` should include the abbreviation itself.

```suggestion
                    <th><abbr title="{% trans 'Date first logged in' %}">{% trans "Date" %}</abbr></th>
                    <th><abbr title="{% trans 'IP address' %}">{% trans "IP" %}</abbr></th>
                    <th><abbr title="{% trans 'Operating System' %}">{% trans "OS" %}</abbr></th>
                    <th><abbr title="{% trans 'Web Browser' %}">{% trans "Browser" %}</abbr></th>
```

bookwyrm/models/session.py
```diff
@@ -0,0 +1,49 @@
+""" functions for managing user sessions """
+import ua_parser
+
+from django.contrib.sessions.backends.db import SessionStore
```
dato: Perhaps do the `SessionStore = import_module(settings.SESSION_ENGINE).SessionStore` dance here too, as elsewhere in the code, just to be sure?

bookwyrm/models/user.py
```diff
@@ -509,6 +512,15 @@ def raise_not_editable(self, viewer):
             return
         raise PermissionDenied()
 
+    def refresh_user_sessions(self):
+        """Check sessions still exist
+        We delete them on logout but not when sessions expire"""
+
+        for sess in self.sessions.all():
+            cache_session = SessionStore()
```
dato: Unless I'm missing something, this object creation can safely be moved right before the loop?

bookwyrm/models/session.py
```diff
@@ -0,0 +1,49 @@
...
+def create_user_session(
+    user_id: int, session_key: str, ip_address: str, agent_string: str = ""
+):
+    """create a session object"""
+
+    user = User.objects.get(id=user_id)
+    os = ua_parser.parse_os(agent_string)
+    user_agent = ua_parser.parse_user_agent(agent_string)
+    system = getattr(os, "family", "Unknown")
+    browser = getattr(user_agent, "family", "Unknown")
```
dato: nit: a single call to `ua_parse.parse()` would be sufficient:


```suggestion
    parsed = ua_parser.parse(agent_string)
    system = getattr(parsed.os, "family", "Unknown")
    browser = getattr(parsed.user_agent, "family", "Unknown")
```

dato: Also should `"Unknown"` be translated?

bookwyrm/tests/models/test_session.py
```diff
@@ -0,0 +1,98 @@
...
+    """test session tracking functions"""
+
+    def setUp(self):
+        """need a user and some pre-existing sessions"""
+
+        with (
+            patch("bookwyrm.suggested_users.rerank_suggestions_task.delay"),
+            patch("bookwyrm.activitystreams.populate_stream_task.delay"),
+            patch("bookwyrm.lists_stream.populate_lists_task.delay"),
+        ):
```
dato: (completely optional) 

FWIW patching out `.delay` methods is no longer needed after #3151.

I have an old branch that removes the `patch()` calls from the test suite, but this is new code so you may as well make use of the feature already.

bookwyrm/templates/preferences/security.html
hughrun: TBH I didn't quite understand how this works so thanks for the tip!

bookwyrm/tests/models/test_session.py
hughrun: @dato I think I missed this change. Does that mean we just don't need to patch celery tasks at all?

dato: Yeah, that's correct. When created during tests, they now go into an in-memory queue which nobody reads.

PR #3669, opened 2025-08-16
Add background jobs to find and add missing covers

### LINKED ISSUE #3668: Remove all removed images
I recently found out that the s3 storage I used with my Bookwyrm instance somehow vanished. After finding another s3 provider I could only sync those assets that were on the server, meaning that a lot of image files have been lost. So now I see everywhere alt text instead of images, because apparently Bookwyrm does not remove image existances from its database when a file is not found.

I would like to have a way to remove from the database all image file paths that lead to files that are no longer present, because of data loss. (I know that I should have made proper back-ups, but I usually do not do this, or I remove the backup after some time thinking I no longer need it, needing the free space)
It would be great if vanished media would be re-synchronised or something with other Bookwyrm servers. So that if a book cover file turns out to be gone, that Bookwyrm then takes the cover from another Bookwyrm instance.

Yes, I have considered manually removing all the book covers and re-uploading covers. This would take way too much time even if I had kind of enough time to do things. Since I have way too much things to do already, always, it is not feasible to do this by hand.

**Additional context**
Although I spoke mainly of book covers, avatars are affected as well.
Books that are newly discovered (I think, or else they are newly bookwyrmed about by someone I follow) *do* display book covers.
Here is a screenshot.
Screenshot of my Bookwyrm instance, showing zero book covers, also not showing avatars.

hughrun:
Builds on previous work in `scheduled-delete-exports` (#3655) – this PR includes code from that one so #3655 should be pulled first.

Adds new background jobs:

* schedulable job to scan connectors for book covers for editions without them
* non-schedulable job to scan connectors for book covers for editions where they have a cover filepath but it does not resolve to an actual file

I also had to move `set_cover_from_url()` to utils.images because trying to use it from a model was causing circular import problems.

I made some choices that others may not agree with (e.g. don't allow regular schedule for finding incorrect cover filepaths - mostly because it seems like it would take a lot of resources to check but is a niche problem).  Also there's a nested series of if statements I'm not thrilled about - so I'm happy to take feedback for improving it!

- Related Issue #
- Closes #3668

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
Sorry about all the commits 🙃

bookwyrm/models/housekeeping.py
```diff
@@ -95,3 +105,141 @@ def start_export_deletions(**kwargs):
...
+
+                if data and data.get("cover"):
+                    if data.get("cover").get("url") in ["", None]:
+                        continue
+
+                    url = data["cover"]["url"]
+                    image = set_cover_from_url(url)
+                    if image:
+                        edition.cover.save(*image)
+                        return edition.id
```
ilkka-ollakka, line 213: One option for this could be rearrange this in something like this:
```
if not(data and data.get("cover")):
   continue

if not (url := data.get("cover").get("url")):
   continue

if (image := set_cover_from_url(url)) is not None:
  edition.cover.save(*image)
  return edition.id
```

bookwyrm/utils/images.py
```diff
@@ -41,3 +47,15 @@ def remove_uploaded_image_exif(source: UploadedFile) -> UploadedFile:
             "Could not open image file for EXIF removal, saving unmodified image"
         )
     return source
+
+
+def set_cover_from_url(url: str) -> None | list[Any]:
+    """load cover image from a url"""
+    try:
+        image_content, extension = get_image(url)
+    except:  # pylint: disable=bare-except
```
ilkka-ollakka, line 56: I think in this it should be enough to just get 'except Exception'. Bare exceptions shouldn't bring any benefit in here to my understanding.

It is previous code that is moved here, but as it is moving, could be good change to reduce one pylint point away.

ilkka-ollakka (approved):

ilkka-ollakka:
I think overall this looks good. 2 small things that I spotted, but nothing that prevents going forward.

PR #3672, opened 2025-08-18
Force https on embed urls

### LINKED ISSUE #3648: Make embed urls secure by using https
I cannot embed my bookwyrm urls (http://bookwyrm.social/list/4242/embed/7576ada4831f44ecacab0bff4c12e5f6) on my Github Pages website because the URL is insecure which Github Pages blocks. It uses http where only https is allowed.

Please use HTTPS only for url embedding.

The alternative is using another book website as embedded is only of the main features I am after.

**Additional context**
See error message from my browser console:

`Mixed Content: The page [...] was loaded over HTTPS, but requested an insecure frame 'http://bookwyrm.social/list/4242/embed/7576ada4831f44ecacab0bff4c12e5f6'. This request has been blocked; the content must be served over HTTPS.`

hughrun:
As per #3648, embed code URLs are http if a proxy server is being used. This fix forces them to always use https.

- Closes #3648

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/views/list/list.py
+++ b/bookwyrm/views/list/list.py
@@ -1,5 +1,6 @@
 """ book list views"""
 from typing import Optional
+from urllib.parse import urlparse, urlunparse
 
 from django.contrib.auth.decorators import login_required
 from django.core.exceptions import PermissionDenied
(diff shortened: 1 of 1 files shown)


ilkka-ollakka:
I think it could be wiser to modify default settings and nginx config to provide that X-Proxy-protocol and enable it by default, similar what we do for assuming we have https unless we are localhost-mode?

Even if django documentation is talking like the header use would need some assesment, I think it is quite safe unless you don't do some odd configuration on purpose (meaning nginx doesn't forward those headers by default, so it is safe to define them to be passed)

I can provide PR to show the changes in nginx and django-settings for that?

PR #3673, opened 2025-08-20
https: send x-forwarded-proto from nginx and use it in django

ilkka-ollakka:
Django uses it to check if connection is secure (https). Enable it with same logic that is used to enable other https settings.

## Description

Forward connection scheme from nginx to django and allow django to use the header to figure out if connection is via https. Alternative to https://github.com/bookwyrm-social/bookwyrm/pull/3672

- Related Issue #
- Closes #

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

Remove non-working config from `.env.example` and change settings to use X-forwarded-proto to check if site is accessed via https. Also modify nginx locations to send the client access scheme as X-Forwarded-Proto header.

## Documentation

### Tests


ilkka-ollakka:
hmm.. it should work with ngrok too, maybe it went to codepath that disables it, same what we do with https, so if domain is localhost?

ilkka-ollakka:
Just curious, do you run ngrok in front of nginx or how the setup goes? Mainly that most likely it is somewhat trivial to check how ngrok is used and make it working out of the box better in dev-flow.

hughrun:
I run it completely outside of docker (i.e. in another shell session) and just add the URL to the `.env` file:

```yaml
# .env
DEBUG=true
DOMAIN=ba222577xxxx.ngrok-free.app
EMAIL=your@email.here
# docker compose NGINX setup: https, reverse_proxy
NGINX_SETUP=reverse_proxy
```

So I expected it to just work like a normal reverse-proxy.

ilkka-ollakka:
ah yes, in reverse-proxy the scheme is http, totally missed that one. I'll update the change to set it https to both https and reverse-proxy setup.

ilkka-ollakka:
nginx doesn't seem to have similar `proxy_add_x_forwarded_proto` env yet in that version of nginx that it has for `proxy_add_x_forwarded_for` headers. So I changed the flow to work with assumption that we are always behind https unless it is localhost.

PR #3710, opened 2025-10-15
book: fix validation for 979 prefixed isbn-13 and isbn_13_to_10 function

### LINKED ISSUE #3709: Important ISBN problem (in 0.8.0 version)
Bookwyrm 0.8.0

I think there is a problem with ISBN validation.
When adding a book with the ISBN: 9791038704022 

ex. https://bookwyrm.social/book/2130964/s/datura

I get the error : `9791038704022 doesn't have correct ISBN checksum, we expected 9791038704023` 

I think Bookwyrm is trying to convert it to ISBN 10, but that's not possible with this ISBN (979).

I checked, and the ISBN in question is valid and appears on all websites selling the book.

It therefore appears that all ISBN numbers beginning with 979 are rejected.

Have a good day.

### LINKED ISSUE #3711: Important ISBN problem with ISBN10 (in 0.8.0 version)
When trying to enter the ISBN numbers for this book 
ISBN-10 2253002690
ISBN-13 978-2253002697

e.g.: https://bookwyrm.social/book/569546/s/tonio-kr%C3%B6ger

the interface tells me that the ISBN 2253002690 is incorrect : 

`2253002690 doesn't have correct ISBN checksum, we expected 22530026911 `

So, to test it, I only enter the ISBN-13: 978-2253002697
but then I see that when I return to the book, BookWyrm has automatically inserted an incorrect ISBN 10:
22530026911 (an 11-digit ISBN??)

Sorry @ilkka-ollakka  but this is likely to cause chaos in the database.

Thanks again, keep up the good work, and have a great day !

Anthony

ilkka-ollakka:
## Description

Check ISBN-13 version directly with checksum instead converting to isbn-10 and back. 979 prefixed isbn-13 doesn't seem to have way to convert to isbn-10.

- Related Issue #
- Closes #3709, Closes #3711

- [X] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/management/commands/fix_isbn10_entries.py
+++ b/bookwyrm/management/commands/fix_isbn10_entries.py
@@ -1,0 +1,28 @@
+from django.core.management.base import BaseCommand
+from django.db.models.functions import Length
+from bookwyrm import models
+
+
+def find_long_isbn10_editions():
+    """Fix isbn-10 entries that have incorrect checksum"""
+    entries_to_fix = models.Edition.objects.annotate(
+        isbn10_length=Length("isbn_10")
+    ).filter(isbn10_length__gt=10)
+
+    print(f"{entries_to_fix.count()} editions to find")
+
+    for edition_to_fix in entries_to_fix:
+        if edition_to_fix.isbn_10.endswith("11"):
+            edition_to_fix.isbn_10 = edition_to_fix.isbn_10[:9] + "0"
+            edition_to_fix.save(broadcast=True, update_fields=["isbn_10"])
+
+
+class Command(BaseCommand):
+    """Fix isbn-10 entries that have incorrect checksum"""
+
+    help = "Find and fix isbn-10 entries that have incorrect checksum"
+
+    # pylint: disable=no-self-use,unused-argument
+    def handle(self, *args, **options):
+        """run fix"""
+        find_long_isbn10_editions()


hughrun:
~LGTM~, nice work.

@mouse-reeve this will need to be added to a 0.8.1 patch since it is a bugfix on the `v0.8.0` pre-release.

hughrun:
hmm, actually #3711 still applies with this patch. Consider a check specifically for ISBN10 rather than converting to ISBN13 to check - they [have different checksums](https://freeisbn.com/check-digit)

ilkka-ollakka:
> hmm, actually #3711 still applies with this patch. Consider a check specifically for ISBN10 rather than converting to ISBN13 to check - they [have different checksums](https://freeisbn.com/check-digit)

hmm.. seems that there was other bug also lingering in isbn_13_to_10. I'll add commit for that fix too and add those isbns to test-cases that have raised the issue.

ilkka-ollakka:
Looking at the second commit, there is a change that there are 11 digit isbn10 values in databases in bookwyrm instances if it has converter incorrectly from isbn-13 to isbn-10.

It might cause some odd issues if you edit such book or try to import one. I'm not sure what would be the best flow to handle those, data-migration maybe to check if isbn_10 is 11 digits and last are '11' change it to '0'?

hughrun:
Those records might already be federating so I don't think a once-off update will be robust enough? I would suggest something in the ISBN check code itself, something like:

If ISBN has 11 characters:
- are the last 2 digits "11"?
- if so change to "0"
- run the checksum check
- if it passes, change the ISBN and it has passed the check
- if it doesn't pass, it's wrong and return an error

Essentially, we are silently changing the data which I normally wouldn't recommend but in this case we can be confident that it's either our own data error or somehow a typo that we can confidently fix.

I guess the other thing to consider is that if we have an ISBN13 I think it would be reasonable to simply check that, and then if it passes and starts with `978`, derive an ISBN10 from it and just overwrite whatever is already there (remembering that `979` ISBNs don't have an ISBN10 equivalent so we can't do this for them).

hughrun:
Although thinking about this again, with ISBNs that are wrong we may be causing duplicate records when things can't be matched so perhaps we need a data migration _and_ a double check in the code?

ilkka-ollakka:
Would it be possible to check from bookwyrm.social db, what is the amount of isbn10 entries that have length of 11 and ending at 11 (well, we should check isbn10 entries length of 11 anycase)?

I think it could give us some info who big the fixable amount is.

mouse-reeve:
It looks like there are 31,328 isbn 10s that are 11 characters long and end in `11`

prolibre:
Oops. I’m afraid my report may have (as we say in French) “soulevé un lièvre.” I see you say “to open a can of worms” or “to raise a red flag” your side. Indeed, I’ve found several thousand 11-character ISBNs in my own instance.

ilkka-ollakka:
I think management command could be good option for fixing those, as changes federate to other instances when edition isbns are fixed?

I think biggest issue is that we can't mark management command to be run as migration-flow and it needs to be documented in release-notes/somewhere that admins will run it. Or atleast most of the admins should run it, assuming that updates are federated to other instances.

But I don't think 31k is that huge amount, database migrations should also be able to handle it fine?

hughrun:
If possible, I think ideally a Django [data migration](https://docs.djangoproject.com/en/5.2/topics/migrations/#data-migrations) file might be the best way to handle it. That would mean devs don't have to do anything unusual, though I think it's always good to be explicit in release notes about what steps they should take.

However, I'm not sure what would happen if we save an `Edition` whilst the instance is offline – would it still queue up a task to send an external update request? I would expect not??? So maybe running a management command whilst the instance is live would be best.

I expect postgres would handle 31k updates easily but `bookwyrm.social` would then queue 31k `Update` activities fanning out to many federated instances per update which ...might cause a fairly gnarly Celery backlog.

@mouse-reeve any thoughts on this? It's likely to impact b.s much more than any other instance.

ilkka-ollakka:
I added management command to check/fix issue, I'm all for making it migration-run, but like you mentioned @hughrun it might not work from migrations and admin command can cause lot of activities out from bookwyrm.social, so I don't really have strong opinion on the approach.

mouse-reeve:
I don’t think it’s a good idea to have a migration trigger activitypub celery tasks, but the migration could update those rows without triggering broadcast, if sending the Update activity is not necessary

ilkka-ollakka:
I'll change the commit to be migration

ilkka-ollakka:
Changed the commit to be migration instead of management command.

prolibre:
Hello @hughrun 
Is there a way to run this command from time to time (even as a command line via cron) to do a little cleanup? There are instances that are rarely updated and could continue to transmit incorrect ISBNs to the entire network.

hughrun:
potentially we could ask @ilkka-ollakka to create a command (that also doesn't broadcast) in addition to the migration. But I am not particularly optimistic that an admin who does not keep their instance regularly updated would nevertheless regularly run the cleanup command.

I guess a possible other addition would be a scheduled task that could just automatically run say, weekly.

prolibre:
@hughrun 

I was thinking more about the fact that outdated instances would continue to distribute incorrect ISBNs, which we would receive on our own updated instances. And I know that when new books are distributed via the network, there is no ISBN check. As a result, we would still receive incorrect ISBNs on updated instances :-(
I hope my comments are clear.
Have a good day.

ilkka-ollakka:
> potentially we could ask @ilkka-ollakka to create a command (that also doesn't broadcast) in addition to the migration. But I am not particularly optimistic that an admin who does not keep their instance regularly updated would nevertheless regularly run the cleanup command.
> 
> I guess a possible other addition would be a scheduled task that could just automatically run say, weekly.

I can add management command to do the cleanup, but I think that should then broadcast. As we do the heavy change in migration, cleanup after that should broadcast the changes to reduce the need to fix same edition multiple times and there shouldn't be that large amount of updates to fanout in that case?

ilkka-ollakka:
added management command to check items after migrations have ran

PR #3743, opened 2025-11-17
Fix Safari dropdown arrow

### LINKED ISSUE #3735: profile dropdown has additional arrow in Safari
**Describe the bug**

**To Reproduce**
1. Load BookWyrm using Safari
2. Notice there is an additional side-facing arrow to the left of the user avatar

**Expected behavior**
No additional arrow

**Screenshots**

Image

**Instance**

**Additional context**
This [can be fixed](https://www.sitepoint.com/style-html-details-element/#removingthecustommarker) by setting `list-style-type` to `none` for the `summary` element. Might need some care to ensure that doesn't cause cascading problems elsewhere, especially since I'm not quite sure why this is already hidden in other browsers.

---

    - OS: MacOS
    - Browser Safari

cnaples79:
## Summary
- Fixes additional side-facing arrow displayed in Safari on profile dropdown
- Adds `list-style-type: none` to `details summary` selector

## Problem
Safari displays an extra arrow to the left of the user avatar in the profile dropdown, despite existing marker hiding rules (`::marker` and `::-webkit-details-marker`).

## Solution
Added `list-style-type: none` to the `details summary` CSS rule in `_details.scss`. This is a known Safari quirk where the browser requires this additional property to fully suppress the default list marker on `<summary>` elements.

## Changes
- **File modified:** `bookwyrm/static/css/bookwyrm/components/_details.scss`
- **Line 6:** Added `list-style-type: none;` to the `details summary` selector

## Testing
This is a Safari-specific CSS fix. The change should not affect other browsers since they already hide the marker with existing rules. Safari users will see the extra arrow removed.

Fixes #3735

### DIFF
--- a/bookwyrm/static/css/bookwyrm/components/_details.scss
+++ b/bookwyrm/static/css/bookwyrm/components/_details.scss
@@ -3,6 +3,7 @@
 
 details summary {
     cursor: pointer;
+    list-style-type: none;
 }
 
 summary::-webkit-details-marker {


hughrun:
LGTM! Thanks 🙂

PR #3745, opened 2025-11-18
Striptags on book description in <meta> tags

### LINKED ISSUE #3695: opengraph and twitter descriptions should strip tags
**Describe the bug**

**To Reproduce**
1. Go to any book page
2. Check `og:description` and `twitter:description` meta tags
3. Notice that they include html tags but probably shouldn't
4. Add to Zotero using browser plugin
5. Also notice that the abstract has `<p>` tags

**Expected behavior**
I think we probably should be using `striptags` to remove the HTML markup.

kasiarog:
## Description

Fixes showing html tags in <meta og:description> and <meta twitter:description>.

- Related Issue #3695 
- Closes #3695

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/views/helpers.py
+++ b/bookwyrm/views/helpers.py
@@ -49,7 +49,7 @@ def get_user_from_username(viewer, username):
 def is_api_request(request):
     """check whether a request is asking for html or data"""
     is_api = "json" in request.headers.get("Accept", "") or re.match(
-        r".*\.json/?$", request.path
+        r"\S{1,100}\.json/?$", request.path
     )
 
     if is_api:


kasiarog:
Hey, could you check on this? @mouse-reeve @hughrun

bookwyrm/templates/snippets/opengraph.html
```diff
@@ -20,6 +20,5 @@
 <meta name="twitter:title" content="{% if title %}{{ title }} - {% endif %}{{ site.name }}">
 <meta name="og:title" content="{% if title %}{{ title }} - {% endif %}{{ site.name }}">
 
-{% firstof description site.instance_tagline as description %}
-<meta name="twitter:description" content="{{ description }}">
-<meta name="og:description" content="{{ description }}">
+<meta name="twitter:description" content="{% if description %}{{ description|striptags }}{% else %}{{ site.instance_tagline }}{% endif %}">
```
hughrun: It's a good instinct to do a check, but the `firstof` filter above already assigns `description` to `site.instance_tagline` if `description` doesn't exist, so there's no need for the `if/else` blocks here. You can just do:
```html
<meta name="twitter:description" content="{{ description|striptags }}">
```

Because `instance_tagline` [has a default value](https://github.com/bookwyrm-social/bookwyrm/blob/c424d01d651a135146c2d7d789855d7777f64e88/bookwyrm/models/site.py#L46) this is safe to do.

hughrun (changes requested):
Thanks for this @kasiarog. It will work as it is now, but the if/else checks aren't really needed so it is better to keep this as clean as we can.

bookwyrm/templates/snippets/opengraph.html
kasiarog: This is what I did at first, but it didn't give me the expected results. Magically the tags were still there after doing a simple filter `description|striptags`. That's why I tried doing it this way

hughrun: hmm odd. I'll take another look later.

hughrun: I've had a look and I can't replicate this:

`Magically the tags were still there after doing a simple filter description|striptags`

@kasiarog do you have an example of a book you were testing with?

kasiarog: Yes, for example "Sense and Sensibility" @hughrun 

image

hughrun: My apologies, @kasiarog, I was not testing this properly.

I think the problem you're experiencing is because [`firstof` autoescapes variable values](https://docs.djangoproject.com/en/5.2/ref/templates/builtins/#firstof)

You should be able to get around this by stripping tags first:

```py
{% firstof description|striptags site.instance_tagline|striptags as description %}
<meta name="twitter:description" content="{{ description }}">
<meta name="og:description" content="{{ description }}">
```

This seems to work for me.

kasiarog: Right, thank you! I corrected it

Issue #3681, opened 2025-08-29
Proposal: Authority control and better modelling for contributors and series


hughrun:
This issue is to outline changes I am planning to make via a small number of PRs. This is to explain my thinking, and invite any feedback on this proposal.

## Contributors

This change would introduce new models `Contributor` and `Contribution`, and turn `Author` into a subclass of `Contributor`.

This would address the following issues:

* #1120
* #2109 (potentially)
* #3160

### Why we need it

Currently the only creators recognised in BookWyrm are authors. This prevents us from properly acknowledging contributors such as illustrators, translators, and editors. It also causes less obvious problems like the inability to attach a series editor to anything.

Currently non-author contributors are either not listed at all, or are incorrectly listed as authors, mostly depending on the original data source.

#### `Contributor`

This would essentially be exactly the same as the current `Author`.

#### `Author`

This would become a subclassed `Contributor` - required for backwards compatibility

#### `Contribution`

Would look something like this:

```py
class Contribution()
  contributor = ForeignKey("Contributor", related_name="contributions")
  book = ForeignKey("Book", related_name="contributors") # contributions can be at the Work or the Edition level
  contribution_type = Charfield(choices=ContributionChoices) # default "author"
  contribution_desc = Charfield() # string to be translated in UI. Would be free text if contribution_type is "other", otherwise we use a default value based on contribution_type
  onix_code = Charfield(choices=OnixCodes) # relevant values from https://ns.editeur.org/onix/en/17
  wikidata_key = fields.CharField(choices=WikiContributionCodes) # aligned to onix types - we can then pull in the correct contribution_type from Inventaire, for example
```

### Why `Contributor`?

There are many different types of contribution, but we always want to connect all of them to a given work or edition - so it's neater to call all of these people one thing (Contributor). This also means if we decide to create a new kind of contribution in future, we only have to change the choices available for `contribution_type`.

### Why `Contribution`?

The same person could make a different contribution to different books - they could be an author on one, a translator of another, and potentially both an author and illustrator on something else. Separating out the _particular relationship_ between each contributor and book allows us to account for all of these possibilities, and also allows us to show on a Contributor's page what kind of contribution they made to different books, much more efficiently than if all that data was stored on the books directly.

### Why limit it to certain ONIX values?

[...]

Issue #2959, opened 2023-08-19
[Feature Request] automatic Book-series-list to keep track of your book series


fabiscafe:
Hello :wave: 

**Is your feature request related to a problem? Please describe.**
I do use bookwyrm in part to keep track of my manga-list. Currently I do this by adding each Manga to an own shelf for the series. I think this is kind of silly, since there is metadata of what series a book belongs to, it would be possible to just add this to the users page, without additional work or shelf building.

**Describe the solution you'd like**
It would be nice if the user would have a category on their users page "book series list" that contains a list of your currently read, and already finishes books that belong into a series.
Lets say it's it's a simple list with clickable items
* [One Week Friends (finished)](https://bookwyrm.social/user/fabiscafe/books/oneweekfriends-62881)
* [Everyday Escape](https://bookwyrm.social/user/fabiscafe/books/everydayescape-62882)
* [Ich habe mein Leben für 10.000 Yen pro Jahr verkauft(finished)](https://bookwyrm.social/user/fabiscafe/books/ichhabemeinlebenf%C3%BCr10000yenprojahrverkauft-62885)
* [I Dad That Same Dream Again(finished)](https://bookwyrm.social/user/fabiscafe/books/ihadthatsamedreamagain-62890)

Nothing complicated, just as simple as that.

**Describe alternatives you've considered**
The alternative is to group them manually into shelfs, but this is troublesome, since you have to create an own shelf for every series, you need to group them to the read list and to the series shelf. It's just a lot to do.
Another alternative are lists and they face the same problem as shelfs.

**Additional context**
Additional it could be possible to rate the finished series in the end and federate this.

PR #3747, opened 2025-11-20
Make series a separate model

### LINKED ISSUE #2991: Series are tied to the first-named author in each edition
**Describe the bug**
As best I can tell, when a series title is added to a book, the matching mechanism to find the rest of the books in the series (i.e. from the book home page) assumes the first-listed author is the author of all books in the series.

Looks like this comes from `/templates/book/book.html`:

```django
{% if book.series %}
    <meta itemprop="isPartOf" content="{{ book.series | escape }}">
    <meta itemprop="volumeNumber" content="{{ book.series_number }}">

    {% if book.authors.exists %} <!-- BELOW WE SIMPLY USE THE FIRST LISTED AUTHOR -->
        <a href="{% url 'book-series-by' book.authors.first.id %}?series_name={{ book.series }}">
    {% endif %}
    {{ book.series }}{% if book.series_number %} #{{ book.series_number }}{% endif %}
    {% if book.authors.exists %}
        </a>
    {% endif %}
{% endif %}
```
And then `BookSeriesBy`.

Unfortunately it's not always this simple. e.g. the _First Knowledges_ series is edited by Margo Neale but she is only an author on the first book in the series. So we need some way of identifying the _series author_ in order to bundle them all up.

**To Reproduce**
1. Go to the page of a book with a series where each book in the series has different authors,
2. Click on the series name.
3. Notice that the series page only lists the book you were just looking at, and claims the first listed author is the author of the entire series.

**Expected behavior**
Clicking on the series name should list all known editions within the series.

**Instance**

**Additional context**

---

    - OS: all
    - Browser all
    - Version 0.6.5

### LINKED ISSUE #3306: Move series membership from Edition to Work
The Wikidata property `part of the series` ([P179](https://www.wikidata.org/wiki/Property:P179)) links a `work` ([Q386724](https://www.wikidata.org/wiki/Q386724)) to a `series of creative works` ([Q7725310](https://www.wikidata.org/wiki/Q7725310)).

`work` is equivalent to our `Work` model. We currently store series information in the `Book` model, which is the parent class of both `Work` and `Edition`, but we don't expose `Work` to the end user: editing a "book" always edits an `Edition`.

I propose moving series membership from `Book` to `Work`, and making `Work` editable by the end user.

This would be a good time to introduce a `Series` model so the `series of creative works` Wikidata key can also be stored.

### LINKED ISSUE #3579: Multiple series to a single book
Please consider adding multiple series to a single book.
For example:
- Avalon (Chronological Order)
- Avalon  (Publication Order)

hughrun:
## Description

This PR moves book series to a pair of new models: `Series` and `SeriesBook`. This will allow more consistency around series, and for books to be matched to series regardless of whether they share an author. Basically it makes series "things, not strings".

A `Series` has basic `BookData` to better enable us to de-duplicate them, although only series imported from Inventaire are likely to have anything beyond a name. A `SeriesBook` connects a Series to a `Book`, including a `series_number`: it's the instantiation of a Book being part of a series.

When saving a book, users will be prompted to confirm whether it is part of an existing series, in the same way we check for existing authors.

All importers will now convert series information to a `SeriesBook` and `Series` where possible. If there is a matching series but no other book in the series with a matching author, the series information is left on the incoming book and users will be prompted to confirm the series information manually on the book view page.

Series can be edited, including updating the `seriesnumber` on one or more `seriesbooks` - users need to have permission to edit books to do this.

Creating a data migration proved to be challenging, so I have instead created an admin command to convert "legacy" series to Series and SeriesBooks where possible. This will need to be run by admins manually.

I have tried to make this backwards-compatible but it's entirely possibly that I've missed something.

I also added an atomic block to the user export model which is entirely unrelated except that I made a mistake when I was testing, and without this block I got a bunch of obscure errors.

This was a bit of a beast to put together so there may well be parts that could be more efficient or that I've missed - happy for any feedback!

- Related Issue #3681 #2959
- Closes #3579 
- Closes #3306 
- Closes #2991 

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/views/books/edit_book.py
```diff
@@ -220,6 +235,40 @@ def add_authors(request, data):
...
+    # check for existing series
+    vector = SearchVector("name", weight="A") + SearchVector(
+        "alternative_names", weight="B"
+    )
+    series = data["form"]["series"].value()
+
+    matches = (
+        models.Series.objects.annotate(search=vector)
+        .annotate(rank=SearchRank(vector, series, normalization=32))
+        .filter(rank__gt=0.019)
```
ilkka-ollakka: I think this should be 0.19 instead of 0.019 ?

hughrun: oops

hughrun:
hmm. Not sure why this `NotifyInviteRequest` test is failing, I didn't do anything to notifications. I suspect something about ordering/concurrency in the test suite?

ilkka-ollakka:
> hmm. Not sure why this `NotifyInviteRequest` test is failing, I didn't do anything to notifications. I suspect something about ordering/concurrency in the test suite?

I didn't manage to reproduce the failure locally either.

ilkka-ollakka (approved):
I think this looks good. The failing test is something that could be tried to reproduce but shouldn't block merging in my opinion.

hughrun:
I can replicate this test failure locally if I run `pytest` with `-n 3` (as per the GitHub action) but not without concurrent tests. I'm guessing there's an underlying problem and this just happens to have surfaced it.

bookwyrm/templates/book/book.html
```diff
@@ -33,45 +33,61 @@
...
 
-                        <span class="has-text-weight-bold" dir="auto">
-                            {{ book.subtitle }}
-                        </span>
+                {% if book.parent_work.seriesbooks %}
+                {% for sbook in book.parent_work.seriesbooks.all %}
+                <div>
+                    {% spaceless %}
+                    {% if sbook.series_number %}
+                    {% blocktrans with title=sbook.series.name path=sbook.series.local_path number=sbook.series_number %}
```
mouse-reeve: You want to include as little HTML as possible within a `blocktrans`, so this should be inside the following `span`, rather than wrapping it. It also needs `trimmed` added to remove the line breaks.

There must be translation systems that can tolerate line breaks, but in our case they cause syntax errors in the translation files.

bookwyrm/models/book.py
```diff
@@ -777,3 +813,64 @@ def preview_image(instance, *args, **kwargs):
...
+    series = fields.ForeignKey(
+        "Series", on_delete=models.CASCADE, related_name="seriesbooks"
+    )
+    book = fields.ForeignKey(
+        "Book", on_delete=models.CASCADE, related_name="seriesbooks"
+    )
+    series_number = fields.CharField(max_length=255, blank=True, null=True)
+    user = fields.ForeignKey(
+        "User", on_delete=models.PROTECT, activitypub_field="actor", related_name="+"
+    )  # for broadcast, should always be instance user but we can't set that here
```
mouse-reeve: This model will default sort to created date, rather than numerical order, which looks odd in the UI

bookwyrm/templates/book/series.html
```diff
@@ -2,35 +2,79 @@
...
+		</div>
+	</div>
+
+	<div class="columns is-multiline is-mobile">
+	{% for seriesbook in series.seriesbooks.all %}
+		<div class="column is-one-fifth-tablet is-half-mobile is-flex is-flex-direction-column" itemscope itemtype="https://schema.org/CreativeWork">
+			<div class="is-flex-grow-1 mb-3">
+				<meta itemprop="isPartOf" content="{{series.name}}">
+				<meta itemprop="name" content="{{book.title}}">
+				<span class="subtitle">{% if seriesbook.series_number %}{% blocktrans with series_number=seriesbook.series_number %}Book #{{ series_number }}{% endblocktrans %}{% endif %}</span>
```
mouse-reeve: Since series number doesn't have to be a number, should the pound sign be here? It looks odd if the value of `series_number` is a word. I'm trying bolding the series number field and removing the #

bookwyrm/templates/book/series.html
```diff
@@ -2,35 +2,79 @@
...
+						<span class="mx-4" itemprop="alternativeName">{{title}}</span>
+						{% endfor %}
+					</div>
+				</div>
+				{% endif %}
+			</div>
+			<div class="column is-one-quarter">
+				{% if series.inventaire_id or series.wikidata or series.isfdb %}
+				<div class="box">
+					<h4 class="title is-4">{% trans "Identifiers" %}</h4>
```
mouse-reeve: I changed this to "External links" to be consistent with the similar box on the author page

mouse-reeve:
I have some small tweaks, but I ran into two bugs that weren't obvious to me:

**No. 1:** If I add a new series with the same name as an existing series, I get this error:
```
FieldError at /book/18664/confirm

Cannot resolve keyword 'title' into field. Choices are: aasin, absorbed, alternative_names, asin, bnf_id, created_date, finna_key, goodreads_key, id, inventaire_id, isfdb, last_edited_by, last_edited_by_id, librarything_key, libris_key, name, openlibrary_key, origin_id, remote_id, search_vector, seriesbooks, updated_date, user, user_id, viaf, wikidata
```
Which traces to:
 /app/bookwyrm/views/books/edit_book.py, line 378

**No. 2:** If I try to delete a series using the checkbox on the edit book page, I get this error:
```
DoesNotExist at /book/18664/edit

SeriesBook matching query does not exist.
```
Which traces to:
/app/bookwyrm/views/books/edit_book.py, line 74

mouse-reeve (changes requested):
I submitted fixes for all these comments on your repo, hopefully that helps??? But the two errors I found unfortunately remain

bookwyrm/management/commands/upgrade_series.py
```diff
@@ -0,0 +1,77 @@
...
+    print("-------")
+    print(f"Created {net_series} new Series and {net_books} new SeriesBooks")
+
+
+class Command(BaseCommand):
+    """Turn legacy series data into Series and SeriesBook objects"""
+
+    help = "Turn legacy series data into Series and SeriesBook objects"
+
+    # pylint: disable=no-self-use,unused-argument
```
dato:
```suggestion
```

bookwyrm/connectors/abstract_connector.py
```diff
@@ -111,6 +113,74 @@ async def get_results(
             logger.info(err)
         return None
 
+    def get_or_create_seriesbook_from_data(  # pylint: disable=no-self-use
```
dato, line 116:
```suggestion
    def get_or_create_seriesbook_from_data(
```

bookwyrm/models/book.py
```diff
@@ -891,3 +927,67 @@ def preview_image(instance, *args, **kwargs):
...
+
+    @property
+    def collection_queryset(self):
+        """list of books for this series, overrides OrderedCollectionMixin"""
+        return SeriesBook.objects.filter(series=self).order_by("-series_number")
+
+    def raise_not_editable(self, viewer):
+        if viewer.has_perm("bookwyrm.edit_book"):
+            return
+        raise PermissionDenied()
```
dato:
```suggestion
        if not viewer.has_perm("bookwyrm.edit_book"):
            raise PermissionDenied()
```

bookwyrm/models/book.py
```diff
@@ -891,3 +927,67 @@ def preview_image(instance, *args, **kwargs):
...
+    activity_serializer = activitypub.SeriesBook
+
+    def get_remote_id(self):
+        """need a remote id to provide the URI for series"""
+        return f"{BASE_URL}/seriesbook/{self.id}"
+
+    def raise_not_editable(self, viewer):
+        if viewer.has_perm("bookwyrm.edit_book"):
+            return
+        raise PermissionDenied()
```
dato:
```suggestion
        if not viewer.has_perm("bookwyrm.edit_book"):
            raise PermissionDenied()
```

bookwyrm/models/book.py
```diff
@@ -891,3 +927,67 @@ def preview_image(instance, *args, **kwargs):
...
+    deserialize_reverse_fields = [("seriesbooks", "seriesBooks")]
+
+    def get_remote_id(self):
+        """series need a remote id"""
+        return f"{BASE_URL}/series/{self.id}"
+
+    @property
+    def collection_queryset(self):
+        """list of books for this series, overrides OrderedCollectionMixin"""
+        return SeriesBook.objects.filter(series=self).order_by("-series_number")
```
dato: Why is this descending order? Wouldn't it make more sense ascending?

bookwyrm/models/book.py
```diff
@@ -891,3 +927,67 @@ def preview_image(instance, *args, **kwargs):
...
+class SeriesBook(ObjectMixin, BookWyrmModel):
+    """connect a book to a series with a series number"""
+
+    series = fields.ForeignKey(
+        "Series", on_delete=models.CASCADE, related_name="seriesbooks"
+    )
+    book = fields.ForeignKey(
+        "Book", on_delete=models.CASCADE, related_name="seriesbooks"
+    )
+    series_number = fields.CharField(max_length=255, blank=True, null=True)
```
dato, line 992: This is `CharField` here, but `int` in activitypub.SeriesBook. Should it be the same in both places?

bookwyrm/tests/activitypub/test_series.py
```diff
@@ -0,0 +1,262 @@
...
+            new_callable=set_related_field(
+                "SeriesBook",
+                "Work",
+                "book",
+                "https://example.com/book/2",
+                book_data.seriesBooks[0],
+            ),
+        ) as mocked:
+            book_data.to_model()  # run it again to trigger the mock
+            mocked.assert_called()
```
dato: `new_callable` in this block is not being used properly. `set_related_field` is invoked at the time it gets passed as `new_callable`, returning `None`, resulting in `new_callable=None` (which is the default).

The test is okay-ish because of the `mocked.assert_called()`, which verifies somebody called set_related_field().

An alternative would be to use a lambda to ensure that it's `to_model()` who makes the call to set_related_field():

```suggestion
        with patch(
            "bookwyrm.activitypub.base_activity.set_related_field.delay",
            new=lambda *args: set_related_field(*args),
        ) as mocked:
            book_data.to_model()  # run it again to set the related field
```

bookwyrm/tests/activitypub/test_series.py
```diff
@@ -0,0 +1,262 @@
...
+
+        with patch(
+            "bookwyrm.activitypub.base_activity.set_related_field.delay",
+            new_callable=set_related_field(
+                "SeriesBook",
+                "Series",
+                "series",
+                "https://example.com/series/2",
+                series_data.seriesBooks[0],
+            ),
```
dato: However if I attempt the same here, I get an error below because SeriesBook.count() is 3 instead of 2, and I don't know why it happens. :(

bookwyrm/tests/activitypub/test_series.py
```diff
@@ -0,0 +1,262 @@
...
+            "seriesNumber": "99",
+            "@context": [
+                "https://www.w3.org/ns/activitystreams",
+                {"Hashtag": "as:Hashtag"},
+            ],
+        }
+
+    def test_serialize_series(self):
+        """check presence of series fields"""
+        activity = self.series.to_activity()
```
dato, line 102: I get something weird with the serialization of actor. If I put here:

```python
        self.assertEqual(activity["actor"], "https://example.com/user/instance")
```

it fails because

```
('https://example.com/user/instance',) != 'https://example.com/user/instance'
```

i.e., it seems as if the actor is a tuple in the Series object.

Can you reproduce this behavior? Is it correct?

dato (commented):
I took a look at this PR, I have some minor comments and questions (all suggestions are optional).

Thanks!!

bookwyrm/models/book.py
hughrun: Yes, it should be CharField, thanks for the pickup.

bookwyrm/models/book.py
hughrun: Yes! But it actually wasn't working anyway because `OrderedCollectionMixin` was missing, so thanks for pointing this out.

bookwyrm/tests/activitypub/test_series.py
hughrun: Series doesn't need an actor, and also can't serialise back to books and seriesbooks, at least for now. So I've removed `actor`/user from the Series model, and also removed this test.

bookwyrm/tests/activitypub/test_series.py
hughrun: Thanks for checking this. I definitely struggled with this bit, thanks for explaining how to use it.

bookwyrm/connectors/abstract_connector.py
```diff
@@ -443,3 +516,41 @@ def maybe_isbn(query: str) -> bool:
...
+        new.user = user
+        series = new
+        series.save()
+
+    if not models.SeriesBook.objects.filter(book=work, series=series).exists():
+        # using the work.series_number for every series is safe because
+        # Inventaire doesn't supply series ordinal when more than one series
+        models.SeriesBook.objects.create(
+            user=user, book=work, series=series, series_number=work.series_number
+        )
```
ilkka-ollakka: You could make this atomic call with following:
```

models.SeriesBook.objects.get_or_create(book=work,
    series=series,
    defaults={"user": user, "series_number": work.series_number})

```
I learned about this when fixing initdb-things just recently. https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create

bookwyrm/views/books/edit_book.py
```diff
@@ -306,6 +355,83 @@ def post(self, request, book_id=None):
             for author_id in request.POST.getlist("remove_authors"):
                 book.authors.remove(author_id)
 
+            user = models.User.objects.get(localname=INSTANCE_ACTOR_USERNAME)
+
+            if series_match := request.POST.get("series_match"):
+                if series_match != "0":  # add known series
+                    series = models.Series.objects.get(id=int(series_match))
+
+                    if not models.SeriesBook.objects.filter(
```
ilkka-ollakka: same here, you could use get_or_create

bookwyrm/management/commands/upgrade_series.py
```diff
@@ -0,0 +1,86 @@
...
+
+        if not SeriesBook.objects.filter(
+            book=book.parent_work, series=series
+        ).exists():  # the series might match because the parent work is already attached to the series. Don't try to duplicate it.
+            SeriesBook.objects.create(
+                user=user,
+                series=series,
+                book=book.parent_work,
+                series_number=book.series_number,
+            )
```
ilkka-ollakka: and here also.

mouse-reeve (changes requested):
I'm sorry to be the bearer of this news, but now when I import a book _without_ a series from Inventaire, it creates a blank string series that has a url linked from the book page, but the link leads to a 404

hughrun:
hmm, that seems suboptimal! I must have broken the logic somewhere.

hughrun:
@mouse-reeve @ilkka-ollakka this is ready to look at again.

ilkka-ollakka:
I tested out this as dev instance and importing somewhat large goodreads export file in.

For me it seems to give exception when I manually try to move one book from 'Wayfarer, #3' series to 'Wayfarer' series, so misparsing on some data. Adding book into existing series works, but removing the 'Wayfarer, #3' series from the book seems to give followin:

```
Exception Value: SeriesBook matching query does not exist.
Raised during: 	bookwyrm.views.books.edit_book.EditBook
```
image


I think it happens because I already selected to remove the existing series when added new one, but it somehow stayed with the book and next removal try gives error?

hughrun:
@ilkka-ollakka I'll add some error handling here but can you give me any more information about your workflow to produce this error? I can't replicate it, unless I have two edit pages for the same book open in different windows/tabs.

bookwyrm/templates/book/edit/edit_book_form.html
```diff
@@ -147,6 +124,57 @@ <h2 class="title is-4">
...
+                        <label class="label">
+                            {% trans "Remove Series:" %}
+                        </label>
+                        <fieldset>
+                            {% for sb in seriesbooks.all %}
+                            <div class="">
+                                <label class="label mb-2" for="remove_series">
+                                    <span class="tag is-link is-light is-medium">
+                                        <a href="{{ sb.series.local_path }}">{{ sb.series.name }}</a>
+                                        <input class="ml-4" type="checkbox" name="remove_series" id="id_remove_series_{{sb.series.id}}" value="{{ sb.series.id }}" {% if sb.series.id|stringformat:"i" in remove_series %}checked{% endif %} aria-describedby="desc_remove_series_{{sb.series.id}}">
```
ilkka-ollakka, line 146: should the value be `sb.id` instead of `sb.series.id` in here? In the code it seems to seek SeriesBook.id=this-value in remove_series list.

hughrun: omg.

You are absolutely correct. I changed the way the models worked a couple of times and I must have missed this when I was cleaning up. I guess when I was testing, the seriesbook and the series had the same ID so it seemed to be working. Thanks for your attention to detail, @ilkka-ollakka, this would have been pretty disastrous to release!

ilkka-ollakka:
Adding and removing book from series now works for me, I seemed to hit some issue after I ran 'deduplicate_book_data', after running that upgrade_series gives me integrityerror:
```
django.db.utils.IntegrityError: null value in column "book_id" of relation "bookwyrm_seriesbook" violates not-null constraint
DETAIL:  Failing row contains (1356, 2026-04-25 10:53:03.270295+00, 2026-04-25 10:53:03.270303+00, null, , null, 1332, 2)
```

I'll give it another run later to see if I can reproduce the issue and pin it down more precisely.

hughrun:
@ilkka-ollakka Is it possible you have editions with no `parent_work`? I think that might be what causes this. I will add a check for this, but it would be good to have confirmation whether this is the cause.

ilkka-ollakka:
I'm unable to reproduce the issue with recent commit.

One thing that I noticed, after running 'upgrade_series' command, books that don't have series, get found edition amount of empty series linked to.

kuva

In edit book view those are all listed as empty series as

kuva


Empty series amounts in all the books that I checked seems to match edition count of the book.

hughrun:
@ilkka-ollakka can you show me a relevant sample of the goodreads file? I'm wondering if this might be happening if series is a blank string (`""`) rather than null.

bookwyrm/management/commands/upgrade_series.py
```diff
@@ -14,7 +14,7 @@ def upgrade_series_data():
     series_count = Series.objects.count()
     seriesbook_count = SeriesBook.objects.count()
 
-    for book in Edition.objects.exclude(series=None):
+    for book in Edition.objects.exclude(parent_work=None).exclude(series=None):
```
ilkka-ollakka: After I locally added also `.exclude(series="")` it stopped mapping to empty series. Seems that most of the books have blank series in db after import.

ilkka-ollakka:
> @ilkka-ollakka can you show me a relevant sample of the goodreads file? I'm wondering if this might be happening if series is a blank string (`""`) rather than null.

I can provide the full goodreads export if you like? But seems to be that series is blank instead of None, as you suspected.

hughrun:
Thanks @ilkka-ollakka! If you can't see any more issues then I think that is solved so I don't need the file.

@mouse-reeve I know you've already looked at this but do you mind doing a final check? It's a pretty big change so it would be good to have another set of eyes on it, especially given my silly mistakes @ilkka-ollakka identified.

mouse-reeve:
I don't know if this is a blocker, but I found this behavior a little unexpected from a user perspective:
When a book has a series , the edit page shows the option to remove the series, or add a new series. It doesn't show the option to change the existing series number. I can add a series number and it's saved but not used (because it's not associated with the existing series entry). 

The only way to change the series number, as far as I can tell, is to remove the series, and then re-add it with the number?

hughrun:
hmm. The expected way to change an existing series number is to go to the series page and do it there.

I agree that is not very intuitive. The problem is that a book can be in more than one series simultaneously, and in that case we don't know which series number you are changing.

This _should_ be a blocker because we already have too many obscure things like this. I'll consider some alternatives.

mouse-reeve:
I see! Would the simplest fix be to have a link to "edit series order" or something in the edit book page? Ideally you could edit the number on the existing series but I don't know how much that complicates things.

hughrun:
i'll investigate both options today. keen to get this merged, but it needs to work properly.

hughrun:
@mouse-reeve is this too busy?

Screenshot 2026-05-16 at 8 42 37 AM

mouse-reeve:
That works! Maybe if the individual series were in a `box` element and the series number on the same line as the label element?

hughrun:
Ok @mouse-reeve that was a little more complicated than I expected, but it did help me to notice that there was an unrelated bug with removing authors, and I hadn't put a unique-together constraint on series+series_number.

When you have time, I think this is ready for a hopefully-last check. I've tested every combination I can think of, but you might find something I missed.

mouse-reeve (approved):
Looks good! How exciting!! I added a couple missing translation tags but that's it.

joric3:
I just want to say thanks for the work you all put in. You're awesome!

PR #3748, opened 2025-11-20
fix is_api_request

hughrun:
## Description

This addresses a CodeQL error about unsafe regex.
'.*' matches anything, which leaves us open to DDoS attacks. This fix matches anything that isn't whitespace, up to 100 characters. As we're only matching the path after the domain, this should be plenty.

- Related Issue #
- Closes #

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/views/helpers.py
+++ b/bookwyrm/views/helpers.py
@@ -49,7 +49,7 @@ def get_user_from_username(viewer, username):
 def is_api_request(request):
     """check whether a request is asking for html or data"""
     is_api = "json" in request.headers.get("Accept", "") or re.match(
-        r".*\.json/?$", request.path
+        r"\S{1,100}\.json/?$", request.path
     )
 
     if is_api:


bookwyrm/views/helpers.py
```diff
@@ -49,7 +49,7 @@ def get_user_from_username(viewer, username):
 def is_api_request(request):
     """check whether a request is asking for html or data"""
     is_api = "json" in request.headers.get("Accept", "") or re.match(
-        r".*\.json/?$", request.path
+        r"\S{1,100}\.json/?$", request.path
```
ilkka-ollakka, line 52: this looks ok, but we could just do 'request.path.endswith(".json/?")' as we dont need to match midstring or get values out of the match?

ilkka-ollakka: nevermind, I missed that optional / in match.

PR #3750, opened 2025-12-02
Linter migration to Ruff

### LINKED ISSUE #3677: Consider replacing `black` and `pylint` with `ruff` and/or define more formatting rules
We're currently linting with `pylint` and formatting with `black`.

We seem to have quite a few pylint `ignore` messages in the codebase, suggesting either that we need to update the `.pylint` configuration to simply not check for certain things, or that perhaps we're not all that serious about the rules we have.

[ruff](https://docs.astral.sh/ruff) seems to be an increasingly popular alternative to pylint, and it can also be used as a drop-in replacement for `black`. That would allow us to reduce our dev dependencies and also to run `ruff` from the dev-tools container instead of from `web` as pylint seems to require.

So I am tentatively suggesting that maybe we should replace both black and pylint with `ruff`, and consider how we want to [configure it](https://docs.astral.sh/ruff/configuration) to enable consistent and easy-to-read clean code without nit-picking.

kasiarog:
## Description

Changes the linter for python from black and pylint to ruff.

A lot of files have changed here and these changes can be devided into few categories:
1. changes in linter configuration - this is the most important,
2. "# pylint:" statements removal - I got rid of unnecessary pylint ignore statements,
3. changes made by formatter - in the vast majority it is something like adding or deleting a blank row or space,
 
While checking the tones of changed files, the most important is to check these couple of files:
- `pyproject.toml`
- `bw-dev`
- `complete_bwdev.fish`
- `complete_bwdev.sh`
- `complete_bwdev.zsh`
- `README.md`
- `dev-tools/requirements.txt`
- `requirements.txt`
- `.github/workflows/python.yml`
- `.github/pull_request_template.md`

- Related Issue #3677
- Closes #3677 

- [ ] Bug Fix
- [ ] Enhancement
- [x] Refactor

I don't know whether I should check any of these, but there is some configurational code, but just for the development process.

## Documentation

### Tests

### DIFF
--- a/bookwyrm/static/css/bookwyrm/components/_details.scss
+++ b/bookwyrm/static/css/bookwyrm/components/_details.scss
@@ -3,6 +3,7 @@
 
 details summary {
     cursor: pointer;
+    list-style-type: none;
 }
 
 summary::-webkit-details-marker {
--- a/bookwyrm/tests/views/preferences/test_export.py
+++ b/bookwyrm/tests/views/preferences/test_export.py
@@ -39,6 +39,7 @@ class ExportViews(TestCase):
             remote_id="https://example.com/book/1",
             parent_work=cls.work,
             isbn_13="9781234567890",
+            pages=123,
             bnf_id="beep",
         )
 
@@ -69,7 +70,7 @@ class ExportViews(TestCase):
 
         self.assertEqual(
             export.content,
-            b"title,author_text,remote_id,openlibrary_key,finna_key,inventaire_id,librarything_key,goodreads_key,bnf_id,viaf,wikidata,asin,aasin,isfdb,isbn_10,isbn_13,oclc_number,start_date,finish_date,stopped_date,rating,review_name,review_cw,review_content,review_published,shelf,shelf_name,shelf_date\r\n"
-            + b"Test Book,,%b,,,,,,beep,,,,,,123456789X,9781234567890,,,,,,,,,,to-read,To Read,%b\r\n"
+            b"title,author_text,remote_id,openlibrary_key,finna_key,inventaire_id,librarything_key,goodreads_key,bnf_id,viaf,wikidata,asin,aasin,isfdb,isbn_10,isbn_13,oclc_number,pages,start_date,finish_date,stopped_date,rating,review_name,review_cw,review_content,review_published,shelf,shelf_name,shelf_date\r\n"
+            + b"Test Book,,%b,,,,,,beep,,,,,,123456789X,9781234567890,,123,,,,,,,,,to-read,To Read,%b\r\n"
             % (self.book.remote_id.encode("utf-8"), book_date),
         )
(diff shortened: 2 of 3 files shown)


requirements.txt
```diff
@@ -47,11 +47,10 @@ setuptools>=65.5.1
 tornado>=6.3.3
 
 # Dev
-black==22.*
+ruff>=0.1.0
```
ilkka-ollakka: I think overall this looks great!

For ruff version, could be feasible to stick with somewhat recent version (current is 0.14.7)? I haven't checked if ruff 0.1.0 supports all those options/gives same formating? Unless you used that version of ruff ofcource, then it should be fine.

pyproject.toml
```diff
@@ -1,2 +1,32 @@
...
+    "E",   # pycodestyle errors
+    "W",   # pycodestyle warnings  
+    "F",   # pyflakes (critical errors only)
+]
+
+lint.ignore = [
+    "E501",   # line too long
+    "E722",   # bare except
+    "E731",   # lambda assignment
+    "F401",   # unused imports
```
mouse-reeve: I don't think we want to ignore unused imports

pyproject.toml
```diff
@@ -1,2 +1,32 @@
...
+    "F403",   # star import (unable to detect undefined names from star imports)
+    "F405",   # `env` may be undefined, or defined from star imports
+    "F841",   # unused variable
```
mouse-reeve: Is this to avoid the no-self-use pylint lines? I do think it's helpful to flag unused variables left in by mistakes

requirements.txt
kasiarog: Right, I missed that. Here is the updated version

ilkka-ollakka (approved):
I think this looks good, I don't posses permissions to merge it currently.

kasiarog:
Thanks, so I will wait for @hughrun or @mouse-reeve :)

hughrun:
Thanks for doing this @kasiarog - I should be able to look at it in the next couple of days.

pyproject.toml
```diff
@@ -1,2 +1,32 @@
...
+[tool.ruff]
+exclude = [
+    "migrations",
+]
+
+line-length = 88
+
+target-version = "py311"
+
+lint.select = [
```
hughrun: These should both be under `[tool.ruff.lint]` like this:

```toml
[tool.ruff.lint]
select = [
    "E",   # pycodestyle errors
    "W",   # pycodestyle warnings  
    "F",   # pyflakes (critical errors only)
    ]
ignore = [
   "E501",   # line too long
   "etc",
    ]
```

kasiarog: Fair, I changed it.

hughrun:
I'm happy to merge this, I just want to give @ccamara a couple of days to update #3755 first to avoid merge conflicts.

PR #3759, opened 2025-12-13
Add container entrypoint to do migrations and static collections on start

ilkka-ollakka:
## Description

Add entrypoint script to run migrations and collections on start.

Split out from https://github.com/bookwyrm-social/bookwyrm/pull/3636 and modified not to include non-root running steps.

Also contains commits from https://github.com/bookwyrm-social/bookwyrm/pull/3757 as this depends on correct order, so other containers should wait for web container to run migrations and things.

This could contain also running init-db, as it is now safe to run in initialized database, but I left it out to keep this simpler.

Original entrypoint script, which this is heavily based on is from https://github.com/bookwyrm-social/bookwyrm/pull/2467

- Related Issue #
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,3 +10,7 @@ RUN apt-get update && apt-get install -y gettext libgettextpo-dev tidy && apt-ge
 
 COPY requirements.txt /app/
 RUN pip install -r requirements.txt --no-cache-dir
+COPY entrypoint.sh /entrypoint.sh
+
+# Entrypoint script is used to do database migrations and collectstatic
+ENTRYPOINT ["/entrypoint.sh"]
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+set -e
+
+info() { echo >&2 "$*"; }
+die() {
+    echo >&2 "$*"
+    exit 1
+}
+
+trap exit TERM
+
+# web container is running gunicorn, so we know to use that for migrations/init-db
+# other containers (celery, celery-beat, etc) just run their commands and are started
+# after web-container as depends_on in docker-compose.yml
+if [ "$1" = "gunicorn" ]; then
+    # we run all the migrations if needed, it is no-op if none needed
+    info "**** Checking needed migrations"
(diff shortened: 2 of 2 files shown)


ilkka-ollakka:
I think one case could be if the migration would be faulty, but in that flow you could run manage-command to skip that migration.

This would drop the need to run setup if we add initdb-run in the entrypoint-script. And after that just to edit documentation to tell instead of setup just run admin_code.

hughrun:
> in that flow you could run manage-command to skip that migration

Sorry, @ilkka-ollakka I don't understand what you mean. Could you give an example of what this would look like? I'm just trying to understand what advice and warnings we need to provide to admins.

ilkka-ollakka:
> Sorry, @ilkka-ollakka I don't understand what you mean. Could you give an example of what this would look like? I'm just trying to understand what advice and warnings we need to provide to admins.

I mean `./bw-dev runweb manage.py migrate --fake` (https://docs.djangoproject.com/en/5.2/ref/django-admin/#cmdoption-migrate-fake) command that marks migrations applied in DB without actually running those. Mainly in case migration would fail for some reason but wouldn't be fatal.

Same flow applies to skip migration without this PR. And I would advice in most cases not to run that command unless person is certain that migration can be skipped without issues.

hughrun:
To check I'm understanding correctly, if there's a problem with migration xxx, I would do:

> `./bw-dev runweb manage.py migrate --fake` (to skip migration xxx)
> `./bw-dev up` (spins up docker but no migrations run since I'm "up to date")

And then my local instance thinks it is up to migration xxx so doesn't actually run the migration in future?

ilkka-ollakka:
> To check I'm understanding correctly, if there's a problem with migration xxx, I would do:
> 
> > `./bw-dev runweb manage.py migrate --fake` (to skip migration xxx)
> > `./bw-dev up` (spins up docker but no migrations run since I'm "up to date")
> 
> And then my local instance thinks it is up to migration xxx so doesn't actually run the migration in future?

Yes, correct. It only places the row in django_migrations table for those migrations.

PR #3760, opened 2025-12-16
Fix shelf tabs redirecting to remote instance for remote users

### LINKED ISSUE #3685: [Bug] Shelf tabs when viewing a remote user redirect to their original instance
**Describe the bug**
Shelf tabs when viewing a remote user redirect to their original instance when clicked. 

I know it's intentional because sometimes an instance won't have the full book list of a remote user, however this is a horrible user experience. The user should be kept on their home instance.

**To Reproduce**
1. Find a remote profile from your home instance
2. Go to their user profile
3. Click on the Books tab
4. Click on one of the shelf tabs within that
5. See that you get redirected to their original instance.

It also occurs on the 'view more' link.

**Expected behavior**
It should remain on the user's home instance. That page does exist, so it's a matter of changing the url.

**Instance**
ramblingreaders.org when viewing a bookwyrm.social account.

nycterent:
## Summary
- Fixes shelf tab links redirecting to remote instance when viewing remote users' shelves
- Uses direct path construction instead of `local_path` which returns remote URLs for remote users
- Adds tests for remote user shelf viewing scenarios

## Problem
When viewing a remote user's shelf page, clicking on shelf tabs (e.g., "To Read", "Currently Reading") would redirect to the remote instance instead of showing the shelf on the local instance.

Root cause: `shelf.local_path` uses `get_remote_id().replace(BASE_URL, "")` which doesn't work for remote users since their `remote_id` doesn't contain the local `BASE_URL`.

## Solution
Replace `{{ shelf.local_path }}` and `{% url 'shelf' ... %}` with direct path construction `/user/{{ user|username }}/books/{{ shelf.identifier }}` which always generates local URLs.

The `{% url %}` tag approach was attempted first but failed due to the ambiguous `(shelf|books)` group in the URL pattern causing `NoReverseMatch` errors.

## Testing
- Added unit tests for remote user shelf viewing
- All existing tests pass
- Manually verified on staging instance

Fixes #3685

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bookwyrm/tests/views/test_user.py
+++ b/bookwyrm/tests/views/test_user.py
@@ -186,9 +186,13 @@ class UserViews(TestCase):
                 inbox="https://example.com/users/nutria/inbox",
                 outbox="https://example.com/users/nutria/outbox",
             )
-        # Create a shelf with a book for the remote user
+        # Create a shelf for the remote user (remote users don't get default shelves)
         with patch("bookwyrm.models.activitypub_mixin.broadcast_task.apply_async"):
-            shelf = remote_user.shelf_set.first()
+            shelf = models.Shelf.objects.create(
+                name="To Read",
+                identifier="to-read",
+                user=remote_user,
+            )
             models.ShelfBook.objects.create(
                 book=self.book,
                 user=remote_user,


ilkka-ollakka:
I think this PR contains way more than intended amount of commits?

PR #3767, opened 2025-12-21
Replace markdown library with mistune

iangreenleaf:
## Description
This replaces the library for parsing Markdown formatted text.

I ran into problems while working on a feature that uses media queries in the `sizes` attribute on `img`; markdown.py was tripping over symbols like `<=` and treating them like HTML even though they were in a string in an attribute and should have been passed through. In other words, I suspect the parser used by markdown.py isn't very robust.

There are a couple other libraries that could work, like [marko](https://github.com/frostming/marko?tab=readme-ov-file). To be honest I picked mistune without a ton of research because it seemed like it was modern, fast, and had wide adoption. Marko complains that mistune is not fully compliant with the CommonMark standard, but then mistune's own docs seem to suggest that it is. I didn't go all the way down that rabbit hole, so all that to say I'm open to using some other library instead, though I suspect whatever differences exist in adherence to spec are probably quite minor and not likely to be  big deal to Bookwyrm. Mistune appears to have excellent performance as well, which is helpful.

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

Changes one of the library dependencies.

## Documentation

### Tests


mouse-reeve (changes requested):
I have no objection to this change, and mistune passes the smell check of being actively maintained.

Unfortunately, you're a victim of timing on this one -- we just moved from `requirements.txt` to `pyproject.toml`. Would you be down to adjust accordingly? I'm not sure why mypy is unhappy but you should be able to run `./bw-dev ruff` to resolve that failing check

iangreenleaf:
Sure, I can merge and update! Might take me a little bit to do so--lot of other stuff going on at the moment.

iangreenleaf:
@mouse-reeve should be ready again. I had to cast a couple of the return values because mistune offers a weird return signature. I'm pretty sure that as called that method will only ever return a string, so hopefully casting is the appropriate solution there.

mouse-reeve:
Hello! I haven't forgotten about your PRs, just did some that required less brainpower. I expect I have created some side effects by merging #3868 before this one, which I don't have time to fix right this moment but will try to get to sooner rather than later. Once that's set, I think this is a pretty straightforward merge and I will look again at your bigger PR. I really appreciate your patience! I know it's been a long haul

PR #3768, opened 2025-12-21
Allow uploading of images in statuses

iangreenleaf:
## Description
Allows images to be uploaded and displayed inline in reviews. I wanted the ability to do this (I'm reviewing picture books) so I set out to make it happen. I tried to solve this in a clean and self-contained way while also making it flexible enough to be powerful, so this is what I came up with:

1. Allow `` elements and a few attributes through the HTML sanitizer. With this change alone, it's technically possible for people to insert images uploaded elsewhere into their statuses.
2. A pair of models and a view to allow users to upload images in a way that are not embedded in another model but rather are standalone records. Upon upload, an image is resized to any number of predefined sizes as determined by the server config.
3. A special `!image` markdown helper that integrates with the `UserUpload` class to display responsive images by making use of `srcset` to provide all the available sizes of image.
4. Some JavaScript that handles drag n drop events to automatically upload the dropped image and inserts the `!image` helper at the cursor.

End result:
image
image

A couple notes:

- I'm an experienced Rails dev but this is my first foray in Django, so there are probably places I've strayed from convention or simply missed easier ways to do stuff. Feel free to suggest improvements.
- The flaw to how I've set this up at the moment is that if you don't know about the feature, it's totally invisible and you'll never discover it. I could imagine providing some sort of UI to allow uploads outside of the drag n drop mechanic, but that seemed like a series of decisions that someone with more context should be making, and I wanted to ship the simplest version first before I tried to tackle any of that.
- This PR includes #3767 because it needs a new parser to successfully create the HTML in statuses for images.
- I probably need to do some linter cleanup before this will pass - I'll hopefully get to that soon.

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

New `UPLOAD_IMAGE_SIZES` .env value that controls what sizes of images will be generated.

## Documentation
I'm not sure if any documentation is needed here, let me know.

### Tests


bookwyrm/models/user_upload.py
```diff
@@ -0,0 +1,32 @@
+""" database schema for images uploaded by users """
+from django.db import models
+import time, os
+
+def user_upload_directory_path(instance, filename):
+    return "uploads/user_{0}/{1}_{2}".format(instance.user.id, time.time_ns(), filename)
+
+class UserUpload(models.Model):
```
ilkka-ollakka, line 12: I think UserUpload should be linked to status also, otherwise there might be lot of orpham uploads that are not removed while user is still in the system but nothing links to them.

bookwyrm/static/js/xhr_files.js
```diff
@@ -0,0 +1,70 @@
...
+        [...event.dataTransfer.items].forEach((item, i) => {
+            // If dropped items aren't files, reject them
+            if (item.kind === "file") {
+                const file = item.getAsFile();
+                this.uploadFile(file, event.target);
+            }
+        });
+    }
+
+    uploadFile(file, target) {
```
ilkka-ollakka, line 52: I think this could also check filesize. API/nginx will give error if file is too big, but not sure this codeflow gives reasonable explanation to the user in that case?

One example can be found in https://github.com/bookwyrm-social/bookwyrm/pull/3627/files how bookwyrm.js uses form dataset to check upload file-sizes.

bookwyrm/views/user_upload.py
```diff
@@ -0,0 +1,78 @@
...
+                    original_file=file
+                )
+        upload.save()
+
+        width, height = image.size
+        env = Env()
+        sizes = env.list("UPLOAD_IMAGE_SIZES", [400, 1200], subcast=int)
+
+        for size in sizes:
+            v = self.create_version(image, upload, size)
```
ilkka-ollakka: should this be background-jobs? I haven't checked lately how much memory/time/cpu takes with huge files.

bookwyrm/views/user_upload.py
```diff
@@ -0,0 +1,78 @@
...
+PREFERRED_EXTENSIONS.update({
+    "JPEG": ".jpg",
+    "TIFF": ".tif",
+})
+
+# pylint: disable= no-self-use
+@method_decorator(login_required, name="dispatch")
+class CreateUserUpload(View):
+    def post(self, request):
+        file = request.FILES['file']
```
ilkka-ollakka: I think this should strip the exif-data from file. 

```
from bookwyrm.utils.images import remove_uploaded_image_exif

file = remove_uploaded_image_exif(request.FILES["file"])
```

Something like that should handle it.

.env.example
```diff
@@ -35,6 +35,8 @@ DEFAULT_LANGUAGE="English"
 STATIC_ROOT=static/
 MEDIA_ROOT=images/
 
+UPLOAD_IMAGE_SIZES=400,1200,2000
```
ilkka-ollakka: Should this be named to something that tells that it is versions max-dimensions? Sizes could imply that it relates somehow to filesize as bytes. Or add comment to clarify how that variable is used?

ilkka-ollakka (commented):
I think overall it looks ok, few comments here and there.

Biggest issue I see, is that uploads are not linked to statuses where they are uploaded to, so I don't see clear flow how those are cleaned up if nothing links to those uploads anymore.

iangreenleaf:
Thanks @ilkka-ollakka, I will go through your comments in detail once the holidays are over and I'm back to my normal schedule, most likely early January.

Regarding not tying the uploads to statuses, yes, that's a design decision I made that's worth discussing. I've seen two schools of thought regarding how to treat uploads. One is to tie them closely to a database record, as you're suggesting, and manage them as a sort of additional piece of data on that record. The other approach is to treat them as independent and sort of idempotent: every uploaded file goes in a big bucket and gets a unique identifier that's the only link it has to anything in the database. Typically in that second way of doing things, every uploaded file lives forever, even if not needed—I suspect this approach really only came into practice once storage was cheap enough to not worry about a bit of "waste".

To align with the first approach instead of the second, I certainly could set up a join table to tie uploads to statuses. I think doing that might significantly increase the complexity of the front-end code, since we would need to figure out how to upload files at the same time as the status form is submitted, and tie each image to the correct place in the text of the status with some sort of placeholder for rendering. It's all a bit easier to handle when uploads happen prior to the creation of the status.

Another option that's in between the two poles would be to create an **optional** join table to statuses. That would allow me to keep the current flow for uploading immediately, but would give server admins the option to do some sort of clean-up action that deletes any images older than some threshold that aren't tied to a status.

hughrun:
Thanks for joining us @iangreenleaf! 

This is a cool idea. I'm wondering how attached you are to images uploads being a separate model, and if so can you explain a little about why?

The usual way we'd do this would be to add a file field to an existing model (in this case, probably [BookStatus](https://github.com/bookwyrm-social/bookwyrm/blob/af29b122e41edd5687b68840d4315289e74b3345/bookwyrm/models/status.py#L312)). Then you'd add that field to [ReviewForm](https://github.com/bookwyrm-social/bookwyrm/blob/af29b122e41edd5687b68840d4315289e74b3345/bookwyrm/forms/status.py#L13), `CommentForm` etc.

That partially solves the issue @ilkka-ollakka has noted with orphaned images (though Django has a few issues with cleaning up files generally).

The downside of that is that they're not "embedded" in the post with markdown, and would have to be added [like we do with book covers](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/books.py) which I suppose one might argue is a little clunky. But the upside is that the image/s are understood by the Django backend as being part of that status, and the code is easier to understand and maintain because it works in a "Django-like" way. When posts are pushed out as ActivityPub the images will be attachments rather than inline images.

iangreenleaf:
@hughrun I'm *quite* attached to the idea of the images being embedded in the markdown and thus being able to be located at specific points in the text. That is necessary to what I want out of this feature and fits with the perspective of treating book statuses more like a rich-text "article" and less like a microblogging "status". I'm *not* terribly attached to any of the particulars of how the code makes that functionality possible. If we can accomplish it while managing the files as part of the Django model, that would be fine with me - though as I mentioned in my previous comment I was a little apprehensive about complicating the code by going that route.

hughrun:
Ok there are a few options here. I think we can _probably_ get the best of both worlds with some care and not a huge amount of complication. 

The biggest issue I see at this point is that currently only `Review` is published to the rest of the fediverse as an `Article` - everything else is a `Note`. So to get the full value you're looking for, we would need to have a conversation about changing that. Mastodon refuses to accept the existence of `Article` as an object type, and turns them into a brief note with a link, so that has causes some discussions here about compatibility and whether to even publish things as `Activity` objects.

@mouse-reeve do you have any thoughts on any of this?

iangreenleaf:
I would be okay with having "inline" images limited to Reviews - that's the only place that it really feels like it makes sense to me. And maybe the other statuses actually *would* make sense with attached images; you could post a "started reading" status with a photo of you with your book and mug of hot cocoa or something.

I would want some advice on architecting it if we went that route, as I'm still new to the Bookwyrm codebase. I plugged it into all statuses just because that was the obvious and straightforward place to do it.

hughrun:
No problem at all @iangreenleaf. My own first PR for BookWyrm was my first time working with Django and ended up being a bit more complicated than I anticipated, so I empathise with the feeling of not quite knowing how everything fits together (also, BookWyrm is kinda complex because we have to consider ActivityPub as well).

If you give me a few days I'll think about what might be the most straightforward way to do this. I don't think you'll need to change too much, it's mostly how we handle saving images, and the ActivityPub aspect might need some additional changes. As @ilkka-ollakka noted, there are some methods already available for image handling that we can probably plug into as well.

bookwyrm/views/user_upload.py
hughrun: Probably better to ensure there's a limit on the size that can be uploaded. If it's a background job then the file won't be there when the status is published.

bookwyrm/models/user_upload.py
hughrun: I agree. Initially I was thinking about adjusting the `Status` or `Review` model but I think the best thing to do would be to add another `ForeignKey`. Currently we really only want this functionality for Reviews (because they turn into ActivityPub `Article`s, but everything else is a `Note`), but how we publish statuses might change in future so it's probably better to just use `Status` as the foreign key. 

i.e.

```py
status = models.ForeignKey(
    "Status",
    on_delete=models.CASCADE,
    null=True,
    blank=True,
    related_name="user_image_uploads"
)
```


Do you have any thoughts on that, @ilkka-ollakka?

hughrun (changes requested):
@iangreenleaf Now that I've had a proper look I can see what you're doing here. I haven't had time to do an actual test run of this yet but it generally looks good.

bookwyrm/models/user_upload.py
ilkka-ollakka: I agreen that adding foreing key to status should be enough in this.

iangreenleaf: @hughrun @ilkka-ollakka how's this look? Added a foreign key that is populated during status creation.

(I haven't addressed the other feedback from your reviews yet, it's on my radar still just need to find the time)

hughrun: Looks right to me. This should delete the images if the linked status is deleted, which is what we want.

bookwyrm/static/js/xhr_files.js
iangreenleaf: @ilkka-ollakka I think I've addressed all your feedback except this item. I got stuck here trying to figure out how to pass `DATA_UPLOAD_MAX_MEMORY_SIZE` through to the front end. To replicate how it's done in that other PR, I think I'd have to pass it through the status form, and I couldn't figure out how to do that - too many layers of stuff I'm unfamiliar with.

ilkka-ollakka: I'll take look if I can provide a way to solve that.

ilkka-ollakka: I think you could solve the issue by adding the upload-limits to text-area as dataset attributes, then you can access that data via event.currentTarget.dataset ?

So no need to go via the example PR on widgets as there is not really upload form like that. Instead you could for example add simple_tag to templatetags/utilities.py for adding data-set attributes with defined limits and use that tag in text-area where you defined the droppable-textfield ?

Not sure if it is the best way, but atleast it could be straightforward way to do it.

hughrun:
@iangreenleaf let us know when you're ready for this to be looked at again :)

bookwyrm/static/js/xhr_files.js
iangreenleaf: Works for me! Thanks for the help.

iangreenleaf:
@hughrun @ilkka-ollakka I believe I've addressed all the feedback and this is ready for another look!

bookwyrm/views/user_upload.py
```diff
@@ -0,0 +1,89 @@
...
+            user=request.user,
+            original_name=file.name,
+            original_content_type=file.content_type,
+            original_file=file,
+        )
+        upload.save()
+
+        width, height = image.size
+        env = Env()
+        sizes = env.list("UPLOAD_IMAGE_DIMENSIONS", [400, 1200], subcast=int)
```
ilkka-ollakka: I would move this parsing to settings.py and import from there.

ilkka-ollakka (commented):
I think this looks good, all my comments that I had have been resolved. Only comment is the location of env-parsing, but from my side it is not blocker.

I think I'll leave final review to @hughrun as this touches in things where I don't have that much experience currently on how those can break.

bookwyrm/views/user_upload.py
```diff
@@ -0,0 +1,87 @@
...
+    }
+)
+
+
+# pylint: disable= no-self-use
+@method_decorator(login_required, name="dispatch")
+class CreateUserUpload(View):
+    def post(self, request):
+        file = remove_uploaded_image_exif(request.FILES["file"])
+        image = Image.open(file)
```
hughrun: We need some error handling here. Currently if a user tries to upload an invalid file type, PIL throws an `UnidentifiedImageError` from `remove_uploaded_image_exif in our `utils.images.py` and there's no information for the end user about what (didn't) happen.

Ideally there would be some kind of user-facing error message - perhaps sent in the`JsonResponse`, but we also want to avoid uncaught errors generally.

bookwyrm/static/js/xhr_files.js
```diff
@@ -0,0 +1,87 @@
...
+        event.preventDefault();
+
+        // Use DataTransferItemList interface to access the file(s)
+        [...event.dataTransfer.items].forEach((item) => {
+            // If dropped items aren't files, reject them
+            if (item.kind === "file") {
+                const file = item.getAsFile();
+
+                if (file.size > event.currentTarget.dataset.maxUpload) {
+                    alert(
```
hughrun, line 38: @mouse-reeve do you have strong views on introducing `alert()` to the codebase? We don't use it anywhere else, but that may simply be that BookWyrm generally has a "only use JavaScript if you really have to" philosophy. In this case I think using JS is fine because there's not really any other way to do drag-and-drop image placement within the markdown, but I'm not sure if we want to maintain some kind of consistent UX in terms of modals/dialogues etc?

hughrun (changes requested):
This is really cool!

I have one blocker - there needs to be some error handling for filetypes that aren't recognised as images.

I also don't love using `alert()` for image size errors and would prefer something like a Bulma modal, but pending an opinion from @mouse-reeve I wouldn't block merging on that basis - we could add that shine later.

A couple of other comments, which also aren't blockers from my perspective but I do want to note for us to think about at some point:

1. Currently there is no way to delete a user uploaded image other than to delete the status or the user. If a user edits the status and removes the `!image(path/to/file)` then the image doesn't _appear_ in the status, but it still exists - it's effectively now orphaned. This has the potential to retain unnecessary image files in storage over time. I can think of a couple of (slightly hacky) ways we might handle this, but note that in that case there's no simple way for Django to know that the image is no longer attached to the status.
2. As @iangreenleaf has already noted, this functionality is effectively invisible to BookWyrm users - you have to know it exists. This is not really a problem specific to this feature so it's definitely _not_ a blocker. We need some kind of info or help interface for statuses that explains that you can use raw HTML, markdown, and drag-and-drop images. That should be a separate PR and I think it's ok to merge this functionality first.

bookwyrm/views/user_upload.py
iangreenleaf: Good call! Error handling added. I'm using `alert` again - inelegant but effective. I just have it reporting a generic message, though the server is returning error details in the JSON so in theory we could construct a more precise user error message if we wanted to.

iangreenleaf:
>  1. Currently there is no way to delete a user uploaded image other than to delete the status or the user. If a user edits the status and removes the `!image(path/to/file)` then the image doesn't _appear_ in the status, but it still exists - it's effectively now orphaned. This has the potential to retain unnecessary image files in storage over time. I can think of a couple of (slightly hacky) ways we might handle this, but note that in that case there's no simple way for Django to know that the image is no longer attached to the status.

Once the status is persisted, the foreign key will let us know the image is attached. One simple but relatively effective solution for this would be to have a periodic job of some sort that looks for all UserUploads that are over a certain age (created, say, more than 7 days ago) and not attached to a status -- those should be safe to delete.

>  2. As @iangreenleaf has already noted, this functionality is effectively invisible to BookWyrm users - you have to know it exists. This is not really a problem specific to this feature so it's definitely _not_ a blocker. We need some kind of info or help interface for statuses that explains that you can use raw HTML, markdown, and drag-and-drop images. That should be a separate PR and I think it's ok to merge this functionality first.

I think a small toolbar might be nice for making the formatting options more visible in general, and if that existed we could plug the image upload into it. I really like the way GitHub's editor handles this (and it was my inspiration for how I set up the image handler). It doesn't try to hide the markdown in the actual text box, but it gives you little buttons that help you add formatting if you don't know the markdown syntax. Unfortunately the JS packages that power this for GitHub, while open source, have a very complicated build chain and didn't seem suitable for use in Bookwyrm.

hughrun:
> Once the status is persisted, the foreign key will let us know the image is attached. One simple but relatively effective solution for this would be to have a periodic job of some sort that looks for all UserUploads that are over a certain age (created, say, more than 7 days ago) and not attached to a status -- those should be safe to delete.

We use a similar process for automatically clearing old user export files - it's a good idea and would definitely work if, for example, someone uploads an image to a draft post and then just never saves the post, or changes their mind about which image to use before saving the status. 

However, what I'm pointing to is that if you remove the markdown for an image from the status by editing the status after you originally saved/published it, the image will no longer appear in the status but neither the `Status` nor the `UserUpload` know that - Django sees it as still attached. i.e. we will know when an image is attached, but not when it is detached. 

I don't want to bog this PR down - especially since editing statuses to remove images probably won't happen a lot - so I think we're ok to merge without addressing this right now. But we'll need to look at it in future.

bookwyrm/static/js/xhr_files.js
mouse-reeve: Good question! I don't think I'm absolutely, it's-a-dealbreaker against it, but I would prefer to have the alert be part of the normal interface? One problem I see as well is that the string here isn't localized

mouse-reeve:
I'm sorry it's taken me so long to take a proper look at this. I just tried it out and it's very cool! I do think there needs to be some kind of hint in the UI about how to add images to reviews.

When I merged in `main`, it brought in a stricter set of `ruff` rules, so there are some linting issues surfaced -- sorry about that.

bookwyrm/static/js/xhr_files.js
iangreenleaf: I have a proof-of-concept for adding JS translations (thanks to Django, that turned out to not be too difficult). I put it in a separate PR to try to not complicate this one further: https://github.com/iangreenleaf/bookwyrm/pull/1

iangreenleaf:
@mouse-reeve thanks, and glad you like it! Are you okay with pushing some of those tweaks (i18n, UI hint for uploads, in-app alert message) to a later PR? I'm happy to work on making some of those changes, but I'll need guidance, since they touch on some bigger design philosophy questions I don't feel qualified to answer.

PR #3773, opened 2025-12-28
fix title_author search to find 'title author' style searches

### LINKED ISSUE #3769: Search for author and title not working
**Describe the bug**
When I searched for `thea riofrancos extraction` on bookwyrm.social I did not find any results, but later, by searching for just the author's name, and then going to an author page, I learned the book [is in fact there](https://bookwyrm.social/book/2056689/s/extraction-the-frontiers-of-green-capitalism).

**To Reproduce**
1. Go to bookwyrm.social
2. Type 'thea riofrancos extraction'
3. Click search
4. See no results

**Expected behavior**
[This book](https://bookwyrm.social/book/2056689/s/extraction-the-frontiers-of-green-capitalism) showing up

**Instance**
bookwyrm.social

---

**Desktop**
    - OS: Linux
    - Browser: Firefox
    - Version: 146

ilkka-ollakka:
## Description

Tune title-author search to allow finding books with 'title author' style searches

* Give book authors and titles same weight in searches
* Manually split search terms to separate SearchQuery

PR splits simple terms separately, keeping the english configuration as is.

- Related Issue #
- Closes #3769

- [X] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/book_search.py
+++ b/bookwyrm/book_search.py
@@ -137,10 +137,26 @@ def search_title_author(
 ) -> QuerySet[models.Edition]:
     """searches for title and author"""
     books = books or models.Edition.objects
-    query = SearchQuery(query, config="simple") | SearchQuery(query, config="english")
+    search_query = SearchQuery(query.strip(), config="english")
+    # We split up search terms for simple config so each term is checked individually against search vectors
+    # This helps matching author and title from query
+    for term in query.strip().split():
+        search_query |= SearchQuery(term, config="simple")
     results = (
-        books.filter(*filters, search_vector=query)
-        .annotate(rank=SearchRank(F("search_vector"), query, normalization=32))
+        books.filter(*filters, search_vector=search_query)
+        .annotate(
+            rank=SearchRank(
+                F("search_vector"),
+                search_query,
+                weights=[
+                    0.4,  # D, series name
+                    1.0,  # C, book author names and aliases
+                    0.4,  # B, book subtitle
+                    1.0,  # A, book title
+                ],  # give author and title same weight as we are searching both
+                normalization=32,
+            )
+        )
         .filter(rank__gt=min_confidence)
         .order_by("-rank")
     )
--- a/bookwyrm/tests/test_book_search.py
+++ b/bookwyrm/tests/test_book_search.py
@@ -22,8 +22,10 @@ class BookSearch(TestCase):
         cls.second_author = models.Author.objects.create(
             name="Author Two", aliases=["The Second"]
         )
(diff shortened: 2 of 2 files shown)


mouse-reeve:
I noticed that this leads to wildly over-matching book titles that have words like "a" and "the" in the title. Here's an example of searching for "a tree grows in brooklyn" on main:

Screenshot 2026-01-01 at 9 40 42 AM

 vs this branch:
Screenshot 2026-01-01 at 9 40 09 AM

On this branch, the desired result is on the second page.

bookwyrm/book_search.py
```diff
@@ -137,10 +137,34 @@ def search_title_author(
...
+    search_query = None
+    for search_term in query.split(" "):
+        if search_query is None:
+            search_query = SearchQuery(search_term, config="simple") | SearchQuery(
+                search_term, config="english"
+            )
+        else:
+            search_query |= SearchQuery(search_term, config="simple") | SearchQuery(
+                search_term, config="english"
+            )
```
mouse-reeve: I think the combination of the `or` here with splitting the string by word is causing over-matching. Words like "a" and "the" are being explicitly included in the query because they are used in the `simple` config, instead of being removed by the `english` config.

For a complete title, the `english` config changes `a tree grows in brooklyn` to `_ tree grows in brooklyn` (removing the `a` but keeping the length and order of the query string), but with this change, `a` is evaluated separately and since it produces a blank result, the `or` keeps it in place, producing `['a', 'tree', 'grows', 'in', 'brooklyn']` and then searching on `a`.

mouse-reeve (changes requested):
I think the impact on title-only searches is too high as-as, would it be possible to tweak how the query is being constructed to avoid this?

If you're interested in some context on the logic behind sometimes retaining common words: https://github.com/bookwyrm-social/bookwyrm/issues/1196

bookwyrm/book_search.py
ilkka-ollakka: ah good point. I'll check on that example and check if it can be easily fixed with anding configs and orring between config.

But anyway I'll check the examples you provided and figure out how to address that issue.

ilkka-ollakka: I reworked the PR and seems that only splitting the simple-part resolves the author finding and doesn't mess up the finding of 'a tree grows in brooklyn' style of books.

ilkka-ollakka:
I'll close this PR. I think trigram search could work better on searching book title/author names instead of FTS, but I haven't had time to pursue that direction lately to check if it is true.

PR #3798, opened 2026-01-16
add libris connector (Swedish libraries catalog)

peppercoffee:
## Description

This PR adds Libris, a book catalog of the Swedish libraries. It is heavily _(heavily)_ inspired by [the PR that added Finna](https://github.com/bookwyrm-social/bookwyrm/pull/3498). It will be great to have this for the new bookwyrm instance for Swedish-reading folks, https://bokdraken.se.

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

(i ran `./bw-dev formatters` and got 3905 errors reported on `bookwyrm/static/css/themes/bookwyrm-dark.css` which i don't think i touched; happy to fix anything that formatters (in CI?) will surface) 

### Tests

The tests are passing, and I've also tested manually to convince myself this is working.

The new connector shows up for the admin:
libris_connector_PR_1

There are some search results from it:
libris_connector_PR_2

Relevant search results, including for books which aren't available in the other catalogs!
libris_connector_PR_3

A book can be imported and then link to Libris works:
libris_connector_PR_4

You can also search by ISBN:
libris_connector_PR_5

If you import two books by the same author, no duplicate author is created:
libris_connector_PR_6

And it works for multiple authors:
image


ilkka-ollakka:
This looks great! thanks for the PR :)

bookwyrm/connectors/libris.py
```diff
@@ -0,0 +1,304 @@
...
+    parse a single author from Libris format.
+    format is typically: "Lastname, Firstname, year-" or "Lastname, Firstname"
+    sometimes includes trailing dot: "Lastname, Firstname, year- ."
+    returns: "Firstname Lastname" or None
+    """
+    if not author:
+        return None
+
+    # remove year information (e.g., "1976-", "1900-1980", or "1947- .")
+    author = re.sub(r",?\s*\d{4}-?\d*\s*\.?\s*$", "", author).strip()
```
hughrun: This regex is a little too generous and the code scanner is complaining. Something like this should work I think:

```sh
(\d\??){1,4}(\W\-?\??){0,3}((\d\?){1,4}$)?
```

hughrun (changes requested):

bookwyrm/connectors/libris.py
peppercoffee: hmm. the suggested regex makes 9 tests fail because it doesn't match the comma before the years (the full string at this point is something like `Lindgren, Astrid, 1907-2002`

```
E       AssertionError: 'Astrid, Lindgren' != 'Astrid Lindgren'
E       - Astrid, Lindgren
E       ?       -
E       + Astrid Lindgren
```

i _think_ the original regex is not as bad as the scanner says because it's anchored with `$` at the end, it's not arbitrary user input but rather a short string from the API, and afaik `\s*` can't cause some blown-up backtracking. but i might be wrong, so, happy to handle the commas a few lines lower like `return f"{parts[1].rstrip(',')} {parts[0].rstrip(',')}"` instead.

will work on this + the typing stuff from the failed CI job now

bookwyrm/migrations/0225_remove_book_bookwyrm_bo_remote__43009f_bloom_and_more.py
```diff
@@ -0,0 +1,38 @@
...
+        ),
+        migrations.AlterField(
+            model_name='connector',
+            name='connector_file',
+            field=models.CharField(choices=[('openlibrary', 'Openlibrary'), ('inventaire', 'Inventaire'), ('bookwyrm_connector', 'Bookwyrm Connector'), ('finna', 'Finna'), ('libris', 'Libris')], max_length=255),
+        ),
+        migrations.AddIndex(
+            model_name='book',
+            index=django.contrib.postgres.indexes.BloomIndex(fields=['remote_id', 'openlibrary_key', 'finna_key', 'libris_key', 'inventaire_id', 'librarything_key', 'goodreads_key', 'bnf_id', 'viaf', 'wikidata', 'asin', 'aasin', 'isfdb'], name='bookwyrm_bo_remote__36fe86_bloom'),
+        ),
```
ilkka-ollakka, line 21: There is change in bloomindex in main-branch, so most likely easiest for this to resolve the conflict is drop this migration-file and recreate migration.

peppercoffee: thanks, will do now

bookwyrm/connectors/libris.py
peppercoffee: done with both the comma and the typing stuff. very unsure about typing stuff since i never really got into types in Python, but there are no mypy errors anymore (locally at least).

bookwyrm/migrations/0225_remove_book_bookwyrm_bo_remote__43009f_bloom_and_more.py
peppercoffee: done

PR #3799, opened 2026-01-17
housekeeping: use first isbn values to search missing covers

ilkka-ollakka:
## Description

Most likely isbn values hits if any and others just generate unnecessary search requests so try first isbn values on search. We have quite a lot of deduplication_items=True fields in Book models.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/models/housekeeping.py
+++ b/bookwyrm/models/housekeeping.py
@@ -2,6 +2,7 @@
 
 import math
 from datetime import datetime, timedelta, timezone
+from itertools import chain
 
 from django.db.models import (
     CharField,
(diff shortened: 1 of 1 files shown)


hughrun:
Hmm good point.

Some books don't have ISBNs though. Should we have a second fallback set of unique IDs to check if ISBN doesn't find any records?

ilkka-ollakka:
> Hmm good point.
> 
> Some books don't have ISBNs though. Should we have a second fallback set of unique IDs to check if ISBN doesn't find any records?

That might be good idea, but not sure if the search checks for other than title/author/isbns ?

ilkka-ollakka:
Just checked and yes, search does check all deduplication identifiers with single search item, so maybe this should be changed that first the isbn is tested and then all the others just in case.

ilkka-ollakka:
updated the code to first go with isbn_10/isbn_13 and then rest of the fields.

bookwyrm/models/housekeeping.py
```diff
@@ -183,13 +184,19 @@ def get_cover_from_identifiers(edition):
...
-        if hasattr(f, "deduplication_field") and f.deduplication_field
+        if hasattr(f, "deduplication_field")
+        and f.deduplication_field
+        and f.name not in isbn_fields
     ]
 
-    for field in fields:
+    for field in chain.from_iterable([isbn_fields, fields]):
+        if not getattr(edition, field):
+            continue
```
hughrun, line 199: Oops, just this fix will help a lot!

hughrun (approved):
LGTM

PR #3809, opened 2026-01-22
Copies production-only backups code to main branch

mouse-reeve:
## Description

This moves some code that lived only on the production branch over to main. I actually think this is the last material difference between the two branches!

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/postgres-docker/backup.sh
+++ b/postgres-docker/backup.sh
@@ -1,11 +1,6 @@
 #!/bin/bash
 info() { echo >&2 "[$(date --iso-8601=seconds)] $*"; }
 
-if [ "${DEBUG}" != 'false' ]; then
-    info "backup: Skipping backups because DEBUG is true"
-    exit 0
-fi
-
 if [ -z "$POSTGRES_DB" ]; then
     info "backup: Database not specified, defaulting to bookwyrm"
 fi
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -3,6 +3,7 @@ services:
   web:
     volumes:
       - .:/app
+  db-backup-job: !reset null
   celery_worker:
     volumes:
       - .:/app


postgres-docker/cronfile
```diff
@@ -0,0 +1,5 @@
+0 0 * * * /usr/local/bin/bookwyrm-backup.sh
```
mouse-reeve: Should this be commented out be default to keep from filling people's dev environments with backup files?

ilkka-ollakka: I think it could be enough just to check if $DEBUG=true and skip the backup in such cases?

hughrun: I agree, I think it's better for defaults to be sensible for production rather than development, with as many dev-specific things as possible tied to `DEBUG=true`. We can also add this to the docs.

postgres-docker/backup.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+source /backups/.env
+
+if [ $DEBUG ]; then
```
ilkka-ollakka: I think you need to test this in bash like this: `if [ ${DEBUG} != 'false' ]`

postgres-docker/cronfile
```diff
@@ -0,0 +1,5 @@
...
+# If uncommented, this script will weed the backups directory. It will keep the 14
+# most-recent backups, then one backup/week for the next four backups, then one
+# backup/month after that.
+# 0 5 * * * /usr/local/bin/bookwyrm-weed.sh -d 14 -w 4 -m -1 /backups
```
ilkka-ollakka: I think this would be good to enable by default

ilkka-ollakka:
I think this is missing the use of that container in db-node in docker-compose.yml -file?

I might prefer to run this backup as separate container outside db, but that can be looked at later on if it is actually feasible.

mouse-reeve:
You're right, I should have looked more closely at whether things were actually working in production. I'm going through the history and sussing it out now

mouse-reeve:
@ilkka-ollakka  I've got it to what looks like the right state, but I'm not seeing backups generated, and when I go into the container and run the backup script, I'm getting this error:
```
pg_dump: error: connection to server on socket "/var/run/postgresql/.s.PGSQL.5432" failed: No such file or directory
        Is the server running locally and accepting connections on that socket?
```

I'm at an impasse on that for the moment.

ilkka-ollakka:
> impasse

Looking at the dockerfile, I think you should run that container instead of default postgresql, so it has cron running in same host.

If you run it as separate container, you need to provide hostname and password to pg_dump, you can use `-h db` to tell hostname and `PGPASSWORD` env-variable to tell password (PGPASSWORD=$POSTGRES_PASSWORD).

If you can show what docker-compose you had/how you tested it, I can retry to replicate the issue locally?

mouse-reeve:
The current state of the code in this PR is how I was testing

ilkka-ollakka:
ah, I'll take a look at this tomorrow to see most feasible flow to get it running.

ilkka-ollakka:
I did this PR against this branch: https://github.com/bookwyrm-social/bookwyrm/pull/3829 making the backup node to log output better from cron and run backups separate from database-container.

postgres-docker/backup.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+source /backups/.env
+
+if [ ${DEBUG} != 'false' ]; then
```
ilkka-ollakka: I think this fails as cron-jobs don't get environments directly into use, if you look the entrypoint.sh script, in line 18 it outputs postgres-related variables to /backups/.env which is used in this. So DEBUG should be added to that line.

postgres-docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# crontab
+RUN mkdir /backups
+COPY ./backup.sh /usr/local/bin/bookwyrm-backup.sh
+COPY ./weed.sh /usr/local/bin/bookwyrm-weed.sh
+COPY ./cronfile /etc/cron.d/cronfile
+RUN apt-get update && apt-get -y --no-install-recommends install cron
+RUN chmod 0644 /etc/cron.d/cronfile
+RUN crontab /etc/cron.d/cronfile
+RUN touch /var/log/cron.log
```
ilkka-ollakka: I think in these lines it install the same cronfile to systemd-wide crontab file and user-specific cron-jobs, /etc/cron.d/cronfile should have username (root) after time specifications, crontab installed file doesn't need it. Not sure which way it should be in this case.

hughrun:
Where are we on this @mouse-reeve? Ready for more testing?

mouse-reeve:
Yes! A third pair of eyes on this would be great

postgres-docker/backup.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+info() { echo >&2 "[$(date --iso-8601=seconds)] $*"; }
+
+if [ "${DEBUG}" != 'false' ]; then
```
hughrun: I think this probably makes sense as an explicit check. However I'm just noting here that we should be very clear in the documentation that if `DEBUG` is not set at all (e.g. it's commented out), then backups won't run. We don't want anyone to find this out the hard way.

ilkka-ollakka: I think we could leave this check out and just modify docker-compose.dev.yml not to have this container running in dev env. I can create PR for that tonight as an example.

ilkka-ollakka: I had some delays, but here is the example https://github.com/bookwyrm-social/bookwyrm/pull/3874

mouse-reeve:
@ilkka-ollakka I merged this too quickly by mistake -- I'm hitting `services.db-backup-job must be a mapping` when I try to use rundev. Is just removing the db-backup-job service line from the dev compose file a reasonable fix? I can also revert, I'm just hoping it's simpler to make a new commit

hughrun:
Does it work if you do this?

```
db-backup-job:
  - !reset none
```

mouse-reeve:
same error

ilkka-ollakka:
> @ilkka-ollakka I merged this too quickly by mistake -- I'm hitting `services.db-backup-job must be a mapping` when I try to use rundev. Is just removing the db-backup-job service line from the dev compose file a reasonable fix? I can also revert, I'm just hoping it's simpler to make a new commit

what docker version you are running? I'll check it out today what could be good alternative syntax or fix.

ilkka-ollakka:
for meanwhile, just removing the line from dev.yml should be fine. Just need to cleanup backups from dev volume if you run the dev instance over utc midnight

mouse-reeve:
Version 4.22.1 -- it sounds like other people aren't having the same problems I am, which is good overall!

mouse-reeve:
Upgrading docker fixed it! Thank you for the help (and no thank you to docker for not updating the normal way and having to be fully re-installed)

PR #3810, opened 2026-01-24
Re-enable stylelint disables after using them

dato:
This avoids errors like:

    bookwyrm/static/css/themes/bookwyrm-dark.css
     8442:3  ✖  "font-family-no-missing-generic-family-keyword"
                 has already been disabled       CssSyntaxError

which aborts the stylelint run and without showing the actual list of (other) issues.

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/static/css/bookwyrm/components/_shelving.scss
+++ b/bookwyrm/static/css/bookwyrm/components/_shelving.scss
@@ -4,7 +4,9 @@
 /** @todo Replace icons with SVG symbols.
     @see https://www.youtube.com/watch?v=9xXBYcWgCHA */
 .shelf-option:disabled > *::after {
-    font-family: icomoon; /* stylelint-disable font-family-no-missing-generic-family-keyword */
+    /* stylelint-disable font-family-no-missing-generic-family-keyword */
+    font-family: icomoon;
+    /* stylelint-enable font-family-no-missing-generic-family-keyword */
     content: "\e919"; /* icon-check */
     margin-left: 0.5em;
 }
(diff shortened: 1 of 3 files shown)


hughrun:
If I run `./bw-dev compile_themes` then it does indeed eliminate this message, but now I have 518 errors (mostly they seem innocuous). Does that mean it's ...working?

dato:
Yeah, I think what this change does is prevent a "syntax error" in the linter, so that it can produce the (long) list of actual lints. I'm unsure what should be done about those (the linter is disabled in Github Actions), but it seemed worthy of a fix. 

Thanks for taking a look!

mouse-reeve:
@hughrun, are you having the numerous errors when running `compile_themes` or when running `stylelint`? The former works fine for me on this branch, the latter produces 1218 errors (even more than 518, does that mean I'm winning?)

hughrun:
sorry I meant when running `stylelint` after updating the themes.

hughrun:
I would also recommend updating stylelint because it seems some of these checks have been deprecated.

mouse-reeve:
Having investigated a bit I have remembered that troubleshooting github workflows is one of the most irritating things to do, and also that I think the change in #3818 (at least the first commit that changes the ignore file) is a preferable fix, since I believe the underlying issue is that stylelint is checking css files when it should only be checking scss files

hughrun:
that makes sense.

and yes, I'm with you on GH actions!

The other issue I have had (could just be me...) is I don't seem to be able to force the dev tools service to update e.g. the node, stylelint version etc. Not sure why

PR #3825, opened 2026-02-04
Some tweaks to manifest.json

WKobes:
## Description

This PR fixes the description value in manifest.json, replaces a tab and ensures it does not start with an empty line.

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

none

## Documentation

### Tests

### DIFF
--- a/bookwyrm/templates/manifest.json
+++ b/bookwyrm/templates/manifest.json
@@ -1,6 +1,6 @@
 {% load static %}{
   "name": "{{ site.name }}",
-  "description": "{{ site.description }}",
+  "description": "{{ site.instance_short_description }}",
   "icons": [
     {
       "src": "{% if site.logo %}{{ media_full_url }}{{ site.logo }}{% else %}{% static 'images/logo.png' %}{% endif %}",


nginx/locations
```diff
@@ -70,6 +70,12 @@ location /robots.txt {
 
 #end of anubis related configuration
 
+# exempt manifest.json from bot checking
+location /manifest.json {
+    proxy_pass http://web;
+    proxy_set_header Host $host;
+}
+
```
ilkka-ollakka: Great spot to notice this didn't work.

I think it would be more usefull to add this exception in anubis config, mainly that if one point it is replaced, it is simpler to all exceptions related to it are in it's config.

For example this could be in anubis/botPolify.yaml in line 11 (so after the bookwyrm rule):
```
    - name: paths-skip-bot-checking
      action: ALLOW
      any:
        - 'path == "/manifest.json"'
```

And you could then add more paths that need skipping on anubis check at that list when those are discovered?

WKobes: I attempted something like this, both as:

```
 - name: paths-skip-bot-checking
   action: ALLOW
   expression:
      all:
        - 'path == "/manifest.json"'
```

as well as

```
 - name: paths-skip-bot-checking
      path_regex: ^/manifest\.json$
      action: ALLOW
```

But the browser error persists, since the request to /manifest.json is still redirected first through anubis. So I think this should be routed outside of anubis.

ilkka-ollakka: Yes you are right, there seems to be some issues with subrequest authentication and path handling. One thing is that we lack X-Original-URI header in nginx side, after that is added the path_regex seems to work, but path isnt.

I played littlebit more, and does following patch work in your test case?
```
diff --git a/anubis/botPolicy.yaml b/anubis/botPolicy.yaml
index 5b0939bf8..f914a8e7a 100644
--- a/anubis/botPolicy.yaml
+++ b/anubis/botPolicy.yaml
@@ -13,10 +13,7 @@ bots:
     expression: 'path.startsWith("confirm-email/")'
   - name: Paths to skip for checking
     action: ALLOW
-    expression:
-      any:
-        - 'path == "manifest.json"'
-        - 'path == "favicon.ico"'
+    path_regex: "^/manifest.json$"
 status_codes:
   CHALLENGE: 200
   DENY: 403
diff --git a/nginx/locations b/nginx/locations
index 7eb0a51d3..3a7089188 100644
--- a/nginx/locations
+++ b/nginx/locations
@@ -51,6 +51,8 @@ location /.within.website/ {
     proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     proxy_set_header X-Real-IP $remote_addr;
     proxy_set_header Host $host;
+    proxy_set_header X-Original-URI $request_uri;
+    proxy_set_header X-Forwarded-Host $http_host;
     proxy_pass_request_body off;
     proxy_set_header content-length "";
     auth_request off;
```

WKobes:
@ilkka-ollakka your changes to anubis paths fixed the routing issue. I force pushed this branch to just the fixes to manifest.json itself, I think it is ready to merge

ilkka-ollakka:
Looks good, thanks for the PR :)

PR #3841, opened 2026-02-10
Add github action to build and release containers

### LINKED ISSUE #3551: Consider packaging releases as containers

ilkka-ollakka:
## Description

Split from https://github.com/bookwyrm-social/bookwyrm/pull/3636 

This PR will enable building of bookwyrm container on main branch and tags. Tag-builds also get latest-tag.

So following container tags are present:
- v0.8.4 etc for tags
- latest for latest tagged build
- main for main branch build

Containers are stored in repository container registry but not used yet in docker-compose.yml. I'll open separate PR to take the containers into use, so this flow can be tested/tuned if needed before it is widely in use.

It also has manual dispatch, so you can start manually build for specific tag if needed.

Currently amd64 and arm64 containers are build, so for example 32bit raspberry pi can't use those.

Containers are also build from PR phase, but not published.

This doesn't yet change how containers are used, so the release flow is still that repo is checked out and containers are build locally by default.

- Related Issue #
- Closes #3551 #2095 

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -1,0 +1,62 @@
+name: Container building for release
+
+on:
+  pull_request:
+    branches: ["main"]
+  push:
+    branches: ["main"]
+    tags:
+      - v*
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Docker tag for build results"
+        type: string
+        default: manual
+        required: true
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # for checkout
+      packages: write # to upload build to GHCR
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Work out Docker tags
+        id: docker_meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ghcr.io/${{ github.repository }}
+          tags: |
+            type=raw,value=${{ inputs.tag }},event=workflow_dispatch
+            type=semver,pattern=v{{version}}
+            type=ref,event=branch
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Setup QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Setup Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Build and push
+        id: docker-build
+        uses: docker/build-push-action@v6
+        with:
+          platforms: linux/amd64,linux/arm64
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.docker_meta.outputs.tags }}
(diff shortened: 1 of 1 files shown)


dato (approved):
Looks great!

.github/workflows/container.yml
```diff
@@ -0,0 +1,62 @@
+name: Container building for release
+
+on:
+  pull_request:
+    branches: ["main"]
```
mouse-reeve, line 5: Should this be happening on pull requests, rather than just on pushes to `main`?

ilkka-ollakka: pull request state it just checks that container is buildable if it targets to main branch. So it is not pushing the container to ghcr from pr.

mouse-reeve: gotcha, thanks!

PR #3845, opened 2026-02-11
gunicorn: limit worker amount to between 1-12 and update version to 25.0.3

ilkka-ollakka:
## Description

https://gunicorn.org/design/?h=worker#how-many-workers mentions that 4-12 should be enough for heavy traffic, so limit worker amount default to be at max 12.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

- [X] This PR changes or adds dependencies

Pumps gunicorn version to 25.0.3

## Documentation

### Tests

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -5,6 +5,8 @@ env = Env()
 env.read_env()
 
 bind = "0.0.0.0:8000"
-workers = multiprocessing.cpu_count() * 2 + 1
+# clip worker count to be between 1 and 12
+#  https://gunicorn.org/design/?h=worker#how-many-workers
+workers = min([multiprocessing.cpu_count() * 2 + 1, 12])
 if env.bool("DEBUG", False):
     reload = True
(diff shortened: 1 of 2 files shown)


gunicorn.conf.py
```diff
@@ -5,6 +5,8 @@
 env.read_env()
 
 bind = "0.0.0.0:8000"
-workers = multiprocessing.cpu_count() * 2 + 1
+# clip worker count to be between 1 and 12
+#  https://gunicorn.org/design/?h=worker#how-many-workers
+workers = max([1, min([multiprocessing.cpu_count() * 2 + 1, 12])])
```
dato: Is the `max()` needed? Because of the `+ 1` in the formula, the second argument is going to be at least always 1.

dato (approved):

gunicorn.conf.py
ilkka-ollakka: it's not necessary, I think if someone runs with env-variable setting cpu-count to negative, they don't expect sane outcomes. I'll drop it out.

ilkka-ollakka:
Actually just noticed that we could run gunicorn with threads and use autoscaler with workers for better result. I'll close this PR and open another one.

PR #3846, opened 2026-02-11
Enable IPv6 for the main network so federation with IPv6-only instances is possible

WKobes:
## Description

This change enables IPv6 on the docker network, so (outbound) federation over IPv6 becomes possible. This is necessary, since even though the web server hosting an instance may support IPv6, docker will by default not enable IPv6 on the network.

This causes issues when federating with IPv6-only instances, like the Mastodon instance https://ipv6.camp

Federation to this instance is currently not possible from bookwyrm.social. My instance has this IPv6 setting enabled, making federation possible: https://wyrm.nl/user/alyx@ipv6.camp

- [X] Bug Fix
- [ ] Enhancement
- [ ] Refactor

This PR requires a one time restart of the network (docker-compose down and up --build). I could not test whether this breaks anything on IPv4-only instances, although I expect docker to fall-back to IPv4 in those cases.

## Documentation

### Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -185,3 +185,4 @@ volumes:
   redis_activity_data:
 networks:
   main:
+    enable_ipv6: true


ilkka-ollakka:
I think this requires running 'docker compose down' for whole stack once when it is taken into use?

WKobes:
> I think this requires running 'docker compose down' for whole stack once when it is taken into use?

True

ilkka-ollakka:
I think this looks good! I'll try to test this with ipv4-only host soon (Tm).

ilkka-ollakka:
This seemed to work fine for me when I did `sudo sysctl -w net.ipv6.conf.all.disable_ipv6=1`  and ./bw-dev down and up. And normally with ipv6 enabled host it works fine. I'm not sure if we should change gunicorn to bind to ipv6 port also?

Could you @hughrun or @mouse-reeve  check if this brings up something for your flow locally?

hughrun:
hmm, I have tried various things and I can't get this to work at all. On the other hand, it seems I can't even load `ipv6.camp` in Firefox, so it's probably something to do with my setup.

[gunicorn docs suggest we probably do need to explicitly bind to ipv6](https://gunicorn.org/reference/settings/#bind), but I'm not sure what would happen if the host doesn't have an ipv6 address assigned.

WKobes:
> hmm, I have tried various things and I can't get this to work at all. On the other hand, it seems I can't even load `ipv6.camp` in Firefox, so it's probably something to do with my setup.
> 
> [gunicorn docs suggest we probably do need to explicitly bind to ipv6](https://gunicorn.org/reference/settings/#bind), but I'm not sure what would happen if the host doesn't have an ipv6 address assigned.

You might not have an IPv6-connection, you can check this with http://conn.internet.nl/connection/. More interesting for us is whether this change breaks something for you? If not, we can assume that it will enable IPv6 federation where possible without breaking legacy IPv4-only set-ups

hughrun:
Ok not sure what's going on with my connection (Firefox can't see ipv6, Safari can). But I've done some quick testing on this and as far as I can tell it doesn't break anything.

PR #3849, opened 2026-02-15
Fix reading goal year check to respect user's timezone

### LINKED ISSUE #1743: Previous year's reading goal shows up at UTC midnight
It's NYE and I'm seeing the 2021 reading goal prompt -- I expect the timezone is being considered when the year is calculated, but not in the if statement that determines if the modal should be show. Low priority -- it'd be nice to fix this before 2023 :)

Screen Shot 2021-12-31 at 5 14 04 PM
.

LakshmiSravyaVedantham:
## Summary
- Uses `timezone.localtime().year` instead of `timezone.now().year` in feed and goal views
- Ensures the year check respects the user's preferred timezone (already activated by `TimezoneMiddleware`) rather than always using UTC
- Fixes the bug where users west of UTC see the new year's goal prompt before their local midnight

Fixes #1743

## Test plan
- Set user preferred timezone to a western timezone (e.g., America/Los_Angeles, UTC-8)
- At UTC midnight on January 1st (4pm Dec 31 Pacific), verify the feed does NOT show the new year's goal prompt
- Verify users in UTC+ timezones who have already entered the new year DO see the prompt

### DIFF
--- a/bookwyrm/views/feed.py
+++ b/bookwyrm/views/feed.py
@@ -240,7 +240,9 @@ def feed_page_data(user):
     if not user.is_authenticated:
         return {}
 
-    goal = models.AnnualGoal.objects.filter(user=user, year=timezone.now().year).first()
+    goal = models.AnnualGoal.objects.filter(
+        user=user, year=timezone.localtime().year
+    ).first()
     return {
         "goal": goal,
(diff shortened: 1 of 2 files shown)


mouse-reeve:
Thank you!

PR #3852, opened 2026-02-18
Reduce container default memory usage

ilkka-ollakka:
## Description

Replaces https://github.com/bookwyrm-social/bookwyrm/pull/3845

Use gthread worker in gunicorn instead of default sync worker. Most cases workers are quite idle and just allocating memory that could be better used by db container.

Use autoscaler in celery worker to scale worker amount between 1 and 10 depending on job amount. Similar approach to free memory to other use when there are no jobs to execute.

Difference of memory usage is from this:
Kuvakaappaus_20260218_175319-1

So web container using ~1G and celery worker ~800M memory to this:

Kuvakaappaus_20260218_175205

So web container using ~320M and celery worker ~170M when not under load. Of course when there are celery tasks running, memory usage increases.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/gunicorn.conf.py
+++ b/gunicorn.conf.py
@@ -5,6 +5,10 @@ env = Env()
 env.read_env()
 
 bind = "0.0.0.0:8000"
-workers = multiprocessing.cpu_count() * 2 + 1
+workers = 2
+threads = min([multiprocessing.cpu_count(), 12])
+max_requests = 1000
+max_requests_jitter = 100
+worker_class = "gthread"
 if env.bool("DEBUG", False):
     reload = True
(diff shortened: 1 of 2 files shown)


ilkka-ollakka:
one thing that might be useful to mention in documentation is to scale web nodes, so if someone runs out of performance from web node (and not from db), they can do 'docker compose up --scale web=2 -d' to add second web node. As nginx uses names, it should pass traffic to all web containers ~equally.

dato (approved):

Issue #3828, opened 2026-02-04
Development environment not picking up template changes


mouse-reeve:
**Describe the bug**
When I edit template files, I have to stop and then re-run and build docker to get the template changes to propagate. I've tried restarting, hard refresh, changing the `sendfile` to `off` in nginx, and prayer. Only re-running `docker-compose up --build` seems to do it

hughrun:
I think I might have found the solution - seems we're not the only ones experiencing problems.

See https://nickjanetakis.com/blog/django-4-1-html-templates-are-cached-by-default-with-debug-true

And an explanation of loaders settings here: https://docs.djangoproject.com/en/5.2/ref/templates/api/#template-loaders

tl;dr we need to adjust the settings.py file, probably like this:

```py
default_loaders = [
    "django.template.loaders.filesystem.Loader",
    "django.template.loaders.app_directories.Loader",
]

cached_loaders = [("django.template.loaders.cached.Loader", default_loaders)]

TEMPLATES = [
    {
        "BACKEND": "django.template.backends.django.DjangoTemplates",
        "DIRS": ["templates"],
        "APP_DIRS": True,
        "OPTIONS": {
            "context_processors": [
                "django.template.context_processors.debug",
                "django.template.context_processors.request",
                "django.contrib.auth.context_processors.auth",
                "django.contrib.messages.context_processors.messages",
                "bookwyrm.context_processors.site_settings",
            ],
            "loaders": default_loaders if DEBUG else cached_loaders,
        },
    },
]

```

mouse-reeve:
I tried that change (I also had to change `APP_DIRS` to false) and did not see any improvement. It does seem like this might be on the right track, because it does seem like `web` is the only container that needs to be rebuild to get changes to propagate

mouse-reeve:
I'm actually not even seeing python changes, so right now I'm having even bigger problems. I wonder if I messed up my local settings in some way 🤔

mouse-reeve:
Okay I ALSO notice that when I run makemigrations, the generated migration file is being sent into the void and not saved locally. (Is this just me??)

@ilkka-ollakka Do you think it's possible that one of the recent docker changes could have caused this? I've tried manually reverting a couple but haven't zeroed in on anything

mouse-reeve:
I do think this is caused by a PR from the last 3 days, and I think it might be https://github.com/bookwyrm-social/bookwyrm/pull/3815, but I am out of time for today for looking into this

[...]

PR #3855, opened 2026-02-22
bw-dev: dev workflow improvements

ilkka-ollakka:
## Description

Modify bw-dev and docker-compose files to allow following:
* bw-dev dev allows subcommands to build/up/down dev stack
* dev stack defaults to debug-mode for localhost-domain and port=1333 with reverse_proxy
* dev stacks run in separate name than normal bookwyrm, so allows in the future to remove the prod_error checks from bw-dev
* add bw-dev command to cleanup unused docker build-caches

So this allows development flow of following:
* `bw-dev up -d` to start dev-stack, access via http://localhost:1333, reading sensible defauls from .env.dev:
  * No need to edit .env to enable debug
  * No need to setup nginx-mode
  * No need to setup port
  * No need to edit domain in .env
* modify files and dev stack notices changes/reload files
* `bw-dev pytest -nauto` runs pytest against same changes/within same stack
* `bw-dev formatters` needs dev-tools container that can be build/rebuild with `./bw-dev dev build dev-tools`

To update dev-tools:
* `./bw-dev dev build`
* `./bw-dev dev build dev-tools`

Tool split-up is checked that if it needs some db rundev is used, otherwise runtools. As runtools doesn't depend on database, it should run faster.

It should work now littlebit faster as rundev/runtools don't stop the stack after run, so next runs should already have db/redis nodes up, or if you are running dev stack at the same time, all nodes are already up that are depended on.

I also removed prod_error check on calls that run in dev-stack, as they don't touch bookwyrm stack and should be safe to run even if you have bookwyrm containers present.

This might still contain some sharp edges, so add questions and comments with low threshold what doesn't work and what works and if you are missing something that you hope/expect to be in place.

- Related Issue #3828
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

- [x] I have created a matching pull request in the Documentation repository

Biggest thing is that this runs the development stack in dev-wyrm stack instead of bookwyrm stack, so if someone has some long-running development stack that they want to keep, they should backup database/rename db-volume and media-volume to new stack to continue using them.

### Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
       - media_volume:/app/images
   anubis:
     image: ghcr.io/techarohq/anubis:latest
+    restart: unless-stopped
     environment:
       BIND: ":8000"
       DIFFICULTY: "4"
@@ -48,6 +49,7 @@ services:
       - "./anubis/botPolicy.yaml:/data/cfg/botPolicy.yaml:ro"
   certbot:
     image: certbot/certbot:latest
+    restart: unless-stopped
     entrypoint: /bin/sh -c "trap exit TERM; while [ "${NGINX_SETUP:-https}" = "https" ];do certbot renew --webroot --webroot-path=/var/www/certbot; sleep 1d & wait $${!}; done"
     depends_on:
       - nginx
@@ -57,6 +59,7 @@ services:
       - ./certbot/data:/var/www/certbot
   db:
     image: postgres:17
+    restart: unless-stopped
     shm_size: ${POSTGRES_SHM_SIZE:-128mb}
     env_file: .env
     healthcheck:
@@ -71,6 +74,7 @@ services:
       - main
   web:
     build: .
+    restart: unless-stopped
     env_file: .env
     command: gunicorn bookwyrm.wsgi:application
     volumes:
@@ -99,7 +103,7 @@ services:
     env_file: .env
     networks:
       - main
-    restart: on-failure
+    restart: unless-stopped
   redis_broker:
     image: redis:7.2.1
     command: redis-server --requirepass ${REDIS_BROKER_PASSWORD} --appendonly yes --port ${REDIS_BROKER_PORT}
@@ -109,7 +113,7 @@ services:
     env_file: .env
     networks:
       - main
-    restart: on-failure
+    restart: unless-stopped
   celery_worker:
     env_file: .env
     build: .
@@ -133,7 +137,7 @@ services:
         condition: service_healthy
       redis_broker:
         condition: service_started
-    restart: on-failure
+    restart: unless-stopped
   celery_beat:
     env_file: .env
     build: .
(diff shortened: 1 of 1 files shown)


bw-dev
```diff
@@ -255,13 +268,7 @@ case "$CMD" in
     update)
         git pull
         $DOCKER_COMPOSE build
-        # ./update.sh
-        runweb python manage.py migrate
-        runweb python manage.py compile_themes
-        runweb python manage.py collectstatic --no-input
         $DOCKER_COMPOSE up -d
-        $DOCKER_COMPOSE restart web
-        $DOCKER_COMPOSE restart celery_worker
```
hughrun, line 264: Just wanting to double check: this is being removed because `up` now checks all of these, right?

ilkka-ollakka: yes, right.

ilkka-ollakka:
> This looks pretty good with the instructions you've drafted in the docs repository. However I get an error if I run `./bw-dev makemigrations`:
> 
...
> Running `/bw-dev rundev python manage.py makemigrations` works.
> 
> I don't understand why this error is being thrown, but the two commands seem to be identical except that `./bw-dev makemigrations` uses the `dev-tools` service. But that should import everything from `bookwyrm-web` so 🤷‍♂️

Ah good point, I'll check those makemigrations as no point actually have special flow for it anymore and it can be ran in web contains in dev-stack instead of dev-tools.

For dev-tools container, I'll check what it needs so it is propertly prebuild when needed, and add instructions in documentation side if it needs some steps.

ilkka-ollakka:
I updated makemigrations and rundev/runtools splitup. Let me know any other issues you manage to surface.

bw-dev
```diff
@@ -255,13 +257,7 @@ case "$CMD" in
     update)
         git pull
         $DOCKER_COMPOSE build
-        # ./update.sh
-        runweb python manage.py migrate
```
ilkka-ollakka, line 259: Now thinking this more, might be good idea to keep migrations here. As sometimes some big migrations might take longer time what the healtcheck is willing to give, causing some odd failures in starts.

ilkka-ollakka:
I tested this with the template-caching-pr. Seems taht with template-caching-pr if you are not logged in, nginx cache caches the output for default amount.

I think it can be fixed to not to cache with debug-mode, but I'm not sure if I want to include that tweaking in here. I'll open separate PR to allow better testing of that changeset.

hughrun:
@ilkka-ollakka ok this looks better. I've added a PR to your repository for consideration - it would add a change to `resetdb`. With this change, `resetdb` resets dev volumes. Without this, there's a chance to delete prod volumes (which we never want) and no way to reset dev volumes.

ilkka-ollakka:
Thanks for checking @hughrun and good catch on that resetdb.

hughrun:
@mouse-reeve did you want to check this with the draft docs explaining how to work in dev with these changes?

mouse-reeve:
This looks ready to go - @hughrun do you want to approve/merge? (This and the docs)

hughrun:
yep i will sort this out tonight. Thanks @ilkka-ollakka @mouse-reeve !

PR #3859, opened 2026-02-28
Restart services automatically, when server reboots.

dashohoxha:
## Description

When I reboot my server, I had to start up bookwyrm manually, because it does not start by itself. This patch fixes that.

Yes it is just a patch, not a big PR. Sorry to say it, but with all these complicated PR templates you are just increasing the barrier for submitting small patches.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,6 +27,7 @@ services:
       - media_volume:/app/images
   anubis:
     image: ghcr.io/techarohq/anubis:latest
+    restart: unless-stopped
     environment:
       BIND: ":8000"
       DIFFICULTY: "4"
@@ -48,6 +49,7 @@ services:
       - "./anubis/botPolicy.yaml:/data/cfg/botPolicy.yaml:ro"
   certbot:
     image: certbot/certbot:latest
+    restart: unless-stopped
     entrypoint: /bin/sh -c "trap exit TERM; while [ "${NGINX_SETUP:-https}" = "https" ];do certbot renew --webroot --webroot-path=/var/www/certbot; sleep 1d & wait $${!}; done"
     depends_on:
       - nginx
@@ -57,6 +59,7 @@ services:
       - ./certbot/data:/var/www/certbot
   db:
     image: postgres:17
+    restart: unless-stopped
     shm_size: ${POSTGRES_SHM_SIZE:-128mb}
     env_file: .env
     healthcheck:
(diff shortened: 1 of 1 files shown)


mouse-reeve:
Thank you for persevering! I appreciate the feedback on the template. It's an ongoing challenge to balance giving people enough information (like how to make checkboxes, and links to documentation resources) with not overwhelming people with a ton of text all at once. And you're right in observing that the template is tailored to larger-scale features rather than small patches like this one, where most of the template isn't relevant. 

Do you think it would be more approachable to have less lead-in to the description section of the template? Like going from what we have now:

```
<!--
Thanks for contributing! This template has some checkboxes that help keep track of what changes go into a release.

To check (tick) a list item, replace the space between square brackets with an x, like this:

- [x] I have checked the box

You can find more information and tips for BookWyrm contributors at https://docs.joinbookwyrm.com/contributing.html
-->
## Description
<!--
Describe what your pull request does here
-->
```

To something more like this, where the template starts out with the main body of the PR?
```
## Description


<!-- 
Thanks for contributing! This template has some checkboxes that help keep track of what changes go into a release. 

You can find more information and tips for BookWyrm contributors at https://docs.joinbookwyrm.com/contributing.html
-->
```

It would be really useful to get your honest feedback or ideas -- I'm so used to the systems we have that I don't have the same perspective as someone coming in fresh.

dashohoxha:
I am fresh here but I am not a student. I might be wrong and you don't have to follow my suggestions. But honestly I think that no template is needed.

A template is based on the assumption that someone knows how to program, how to use git and GitHub, how to make a PR, but somehow does not know how to describe it, or does not know how to use markdown.

When I make a PR I usually write an adequate description even if there is no template to request it. Each commit also has a description.

In case somebody fails to describe it properly, you can always ask questions and discuss it (like we are discussing here), and ask the contributor to modify/improve the PR.

mouse-reeve:
Thank you for your frankness! And to be clear I did mean "fresh" as is "hasn't opened a PR in this repo before" not as in "new to the concept of programming." We do get material benefits from the template (it is part of the release workflow), and I have found providing some structure has helped overall in getting contributions to be a little better documented. But it's always something I've been hoping to use flexibly, and I think you're right to point out that it's a lot!

Anyway I will think about it and let you get back to the actual PR at hand. Which I think would be in @ilkka-ollakka's wheelhouse

ilkka-ollakka:
Thanks for the patch, it looks good improvement.

Issue #3770, opened 2025-12-25
Visual refresh


camyrland:
The current visual style of Bookwyrm is quite dated and cluttered, especially on small screens such as phones. Much has happened to both Mastodon and Pixelfed in regard to visuals over the last few years, but BookWyrm hasn't changed much. 

I am by no means a designer or layout expert, so I don't have any definite ideas as to what is needed or what would work best, but some suggestions are:

* Drop the card layout, which only takes up space on screen. 
* Show book covers in full screen width, possibly crop height to 1:1 or 2:3 ratio. Look to Pixelfed..
* Rearrange the order of a post. There's too much going on between the cover of the book and the reader's thoughts about it. A suggestion to post layout could be something like post author at the top, then fullwidth book cover, showing replies, boosts and book rating (like "⭐4.5") just below the cover, followed by reader's thoughts. Clickable links to the book and author, and possibly a more detailed rating overview, should be hidden until the post is clicked/expanded. 
* Remove all the frame lines. There's no reason to put a box around every element in a post. It just looks messy.

There are probably many more design improvements to be made, so hopefully this post can spark a debate on which and how changes should be implemented.

A visual refresh could boost the attention BookWyrm is getting, and improve the impression new users get when signing up.

Another approach (if possible) could be to build a third party (web) app to test different design solutions without affecting the BookWyrm code base. Look to Phanpy or Elk..

mouse-reeve:
Thank you! These seem like sensible and actionable suggestions, and I'm constantly out of my depth when it comes to anything visual, so I really appreciate getting ideas.

I'm not sure I'm totally understanding your suggestion for rearranging order of a post. Would you be able to draw me a picture? Even something MSPaint level would be helpful.

In terms of the frame lines and card layout, is something like this what you're talking about?
Before:
<img width="1076" height="759" alt="Image" src="https://github.com/user-attachments/assets/a856435b-ea59-4752-970b-3348dbee3415" />

After:
<img width="1094" height="751" alt="Image" src="https://github.com/user-attachments/assets/e99599a4-937f-4672-a788-e8aade46c79d" />

dylanwe:
I think this is a great improvement for the UI. Feeling out of depth with design is pretty common.

Most designs don’t have to be revolutionary and aligning with existing applications often improves the navigation as users already expect things to be done in certain ways.

A book that really helped me out was https://refactoringui.com/. Another great resource is [dribbble](https://dribbble.com) to get inspiration.

[...]

PR #3861, opened 2026-03-01
Refresh UI display of statuses

mouse-reeve:
## Description

This removes a lot of lines, shadows, and whitespace from statuses in the feed.

Current feed:
Screenshot 2026-05-16 at 6 10 36 PM

Proposed change:
Screenshot 2026-05-17 at 2 49 38 PM
Screenshot 2026-05-17 at 2 49 57 PM

- Related Issue #3770 
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


mouse-reeve:
Updated to reflect feedback that the division between statuses was too hard to see:
Screenshot 2026-03-19 at 8 16 48 PM

j6k4m8:
I like the new look and definitely agree it needs the clearer delineation in the update!! 
I think also possibly left/right borders could help to ground it as well, I'm looking at the current bluesky and masto designs as inspiration: 

|--|--|
|--|--|
|  |  |

maxheadroom:
Mmh, I have to admit I like the old layout better. For this old layout is a clearer distinction between the two posts as one can see where the posts ends.

dylanwe:
Maybe this can help inspire the new look:
image

https://dribbble.com/shots/18396867-Goodreads-Redesign-In-Progress

This includes the full borders like @j6k4m8 suggests which makes it feel like cards but removes the shadow. All the shadows can make it feel to busy.

All credits to [Brittney Singleton](https://dribbble.com/brittneysingleton)

dylanwe:
The `hr` currently also is a thicker line than other lines which doesn’t help cohesion.

j6k4m8:
I love that design! Strong support for adopting flavors from that :D 

Let me know if you want a hand implementing :)

dylanwe:
Made some quick examples using the web inspector

1. Original version
image

---

2. Removed drop shadow and added full border

image

---

3. Changed the color to a dark green

image

---

4. A brown variation

image

---

5. Hints of serif font (feels more book like)

image

---

6. Add more rounded corners

image

I changed from the current colors because they don't feel like colors I associate with being calm and reading. I enjoy green and brown because it reminds me of nature and the warm tone reminds me of paper. I changed "Book Timeline" to just "Timeline" as the tab already shows which timeline is viewed.

I've added border radius around the cover containers and avatars. The sharp corners felt a bit off on them and this gives it a softer feel.

j6k4m8:
I do also REALLY like the inline "comment" text; it puts some of the whitespace in that bottom buttons panel to productive use by reducing the amount of interaction to write a reply, and also makes the bottom panel UI make way more sense.
image

dylanwe:
Something like this @j6k4m8? Could fit the current design.
 
image

In this case if the component becomes too wide you should put a max width on the input and align it to the right of the component whilst keeping the other buttons on the left.

dylanwe:
Putting in that comment action might be better suited for a different PR to not overload this with new functionality.

j6k4m8:
yeah that's a great point. @mouse-reeve let me know if I shoudl jump in here :)

mouse-reeve:
@j6k4m8 Please feel free to!

mouse-reeve:
I really like these suggestions! I do want to keep changes incremental for now, so I would prefer to change the borders and padding and not the colors in this PR. And I agree with @dylanwe, moving the comment box is a great idea for a separate PR. I realize it's hard to be super granular about design when it's the whole thing you actually deal with, but I'd like to do a larger-scale re-think in the future and just some cleanup right now.

I'm going to see if I can quickly make just the layout related changes, since I already have some work in place. It would be absolutely wonderful if someone here wanted to try their hand at moving the comment box!

mouse-reeve:
(I'd also like to be cautious about using GoodReads as a design reference, it's a little too close to home and I don't want to appear as if bookwyrm is a clone of it)

mouse-reeve:
Can I get a vibe check on this status display in the feed:
Screenshot 2026-05-15 at 10 50 42 AM

dylanwe:
@mouse-reeve very understandable about minimizing the changes and not wanting to look like Goodreads. I think your new design already looks a lot better than the previous iterations! Mind that the border above the actions reply, boost, like has a different border color on top than you used on the card border. I also think that the title should be next to the book:

image

The reason for that is that we group the action and the book overview. "User started reading" and below that we have the book details.

But I wonder what @camyrland thinks of this as well.

mouse-reeve:
I did set them to different colors, but it looks weird in dark mode so I'm changing them to the same color. Moving the text is out of scope for this issue (it's a little more complicated than it might seem on the surface).

Here is the current state of the PR for dark mode:
Screenshot 2026-05-16 at 6 02 34 PM

camyrland:
> @mouse-reeve very understandable about minimizing the changes and not wanting to look like Goodreads. I think your new design already looks a lot better than the previous iterations! Mind that the border above the actions reply, boost, like has a different border color on top than you used on the card border. I also think that the title should be next to the book:
> image
> 
> The reason for that is that we group the action and the book overview. "User started reading" and below that we have the book details.
> 
> But I wonder what @camyrland thinks of this as well.

I like it, but I'd drop the border between the main and footer part. I'd also prefer a single colour for the entire card, but hey, that's just me :)

mouse-reeve:
Thank you! I was skeptical but I tried it without the header color and border and I think it's more cohesive and clean. Updated screenshots:
Screenshot 2026-05-17 at 2 49 38 PM
Screenshot 2026-05-17 at 2 49 57 PM


At this point, it sounds like we're basically all agreed that this change is an improvement on the existing layout, whether or not it's 100% exactly how you would do it. I would like to call this PR complete and merge it. That doesn't mean this is set in stone, or the ideas suggested here for other changes are dropped, just that the specific changes in this PR feel like a step forward. 

Could I get an **emoji response** from people who are happy to see this merged, and a **reply** with your thoughts if you don't want to see it merged yet?

j6k4m8:
> @j6k4m8 Please feel free to!

omg I'm SO SORRY I missed this! I can jump in once this is merged to start working on some of the next features we're discussing :)

PR #3877, opened 2026-04-08
fix: use LibraryThing Collections column for shelf mapping

### LINKED ISSUE #3876: LibraryThing import puts everything onto to-read shelf
**Describe the bug**
When importing TSV from LibraryThing everything book is grouped as "to-read" regardless of collection set in LibraryThing and whether or not "create new shelves if they do not exist" is checked 

**To Reproduce**
1. Go to YOUR BOOKS
2. Click on IMPORT BOOKS
3. Select LibraryThing (TSV) from Data source. 
4. select File exporeted from LibraryThing
5. Select import

**Expected behavior**
After import all books should be on the same shelf as corresponds with their Librarything Collection

**Actual Results**
Every book is in "to-read" 

**Screenshots**

Image

<hr>
Image

**Instance**
tested on bookwyrm.social, bookwyrm.world and milleuilles.cloud

**Additional context**

On thing to be aware of is that LibraryThing doesn't actually have a field for "Shelf" but their "Collections" field is essential the same thing. Default choices include:  Currently reading, To read, Read but unowned, Your library, Wishlist, and Favorites.

I tried editing my TSV before import so that I only included books I already read and replaced of the contents of the Collections with "Read" and still everything was imported into to-read

One difference between LibraryThing Collections and Bookwyrm shelves is that a user can put a book in more than one Collection on LibraryThing and only one Shelf on Bookwyrm.

Long term I think it would be good to give a user a choice for each such instance, like "23 books in Collections 'Your Library & To read', would you like to (1.) place on [drop down menu with previously existing shelves] shelf (2) place on  new shelf [width editable field pre-populated with 'Your Library & To read'] (3.) do not import". 

Short term just defaulting to the first collection listed in the field should be sufficient. 

---

    - OS: Windows 11 Version 25H2 (Build 26200.8037)
    - Browser: Microsoft Edge 
    - Version: 146.0.3856.97 (Official build) (64-bit)

Bahtya:
## Problem

When importing books from LibraryThing TSV, every book is placed on the to-read shelf regardless of the Collection set in LibraryThing. The create new shelves option has no effect.

## Root Cause

LibrarythingImporter.get_shelf() only checks date_finished and date_started to determine shelf placement. It completely ignores the Collections column from the LibraryThing TSV, which is the primary indicator of reading status.

Additionally, LibrarythingImporter does not override row_mappings_guesses, so the Collections column is never mapped to the shelf field.

## Fix

1. Override row_mappings_guesses to include collections in the shelf mapping guesses
2. In get_shelf(), when no dates are available, fall back to the Collections/shelf field value and match it against known shelf names
3. Books with no dates and no recognizable collection still default to to-read

Fixes #3876


bookwyrm/importers/librarything_import.py
```diff
@@ -19,6 +19,28 @@ class LibrarythingImporter(Importer):
...
     encoding = "ISO-8859-1"
 
+    # LibraryThing uses "Collections" column instead of "shelf"
+    row_mappings_guesses = [
+        ("id", ["id", "book id"]),
+        ("title", ["title"]),
+        ("authors", ["author_text", "author", "authors", "primary author"]),
+        ("isbn_10", ["isbn_10", "isbn10", "isbn", "isbn/uid"]),
+        ("isbn_13", ["isbn_13", "isbn13", "isbn", "isbns", "isbn/uid"]),
+        ("shelf", ["shelf", "exclusive shelf", "read status", "bookshelf", "collections"]),
```
ilkka-ollakka: Instead of copying all the entries here to add the collections. You could add __init__() to this importer and just add that 'collections' to shels, so something like `self.row_mappings_guesses["shelf"].append('collections')`

bookwyrm/importers/librarything_import.py
```diff
@@ -32,8 +54,16 @@ def normalize_row(
...
             return Shelf.READ_FINISHED
         if normalized_row["date_started"]:
             return Shelf.READING
+        # Fall back to Collections/shelf field
+        shelf_name = normalized_row.get("shelf")
+        if shelf_name:
+            shelf_name_lower = shelf_name.lower().strip()
+            for shelf_key, guesses in self.shelf_mapping_guesses.items():
+                if any(g in shelf_name_lower for g in guesses):
+                    return shelf_key
```
ilkka-ollakka, line 47: Could you add tests to highlight the case where previously code failed and now works?

ilkka-ollakka (commented):
Thanks for the patch, few quick comments that I noticed. Mainly if you can add tests also for the case, it would be great!

bookwyrm/importers/librarything_import.py
mouse-reeve: I agree with this, but to make it even simpler you could just add `collections` to `importer.py`, since I don't think it would cause any problems to treat it as an alternative to `shelf` across the board

Bahtya:
Thanks both @ilkka-ollakka and @mouse-reeve for the helpful feedback!

I've pushed a new commit that simplifies the approach as suggested:

**Changes per @mouse-reeve's suggestion:**
- Added `"collections"` to the shelf `row_mappings_guesses` in the base `importer.py` instead of overriding `row_mappings_guesses` in the LibraryThing importer. This is much simpler and works because only LibraryThing exports have a Collections column.

**Simplification of `librarything_import.py`:**
- Removed the full `row_mappings_guesses` override entirely (no longer needed)
- Fixed `get_shelf()` to use exact matching (`shelf_name_lower in guesses`) instead of substring matching. The substring approach was buggy -- for example, `"to read"` would incorrectly match `"read"` in the read shelf guesses.

**Tests added per @ilkka-ollakka's request:**
- `test_get_shelf_date_finished`: dates take priority over Collections
- `test_get_shelf_date_started`: `date_started` without `date_finished` -> reading
- `test_get_shelf_from_collections_to_read`: `'To read'` -> to-read shelf
- `test_get_shelf_from_collections_reading`: `'Currently reading'` -> reading shelf
- `test_get_shelf_from_collections_read`: `'Read'` -> read shelf
- `test_get_shelf_unrecognized_collection_defaults_to_read`: `'Your library'` -> defaults to to-read
- `test_get_shelf_no_dates_no_collection`: no data at all -> defaults to to-read
- `test_create_job_with_collections`: end-to-end integration test with extended test TSV data covering all shelf types

Bahtya:
Apologies, I just noticed the AGENTS.md policy in this repository which states that AI agents must not make contributions. I will close this PR out of respect for the project's guidelines. Thank you for the feedback received so far.

PR #3880, opened 2026-04-11
celery: pump version to 5.6.2 and change to gevent pool

ilkka-ollakka:
## Description

5.4 version added some DjangoTask classes and documentation refers to those, also the 5.3 version is from year 2023. So update the version to make it easier for others to contribute by checking current celery documentation and django documentation.

Change to gevent pool, instead of prefork. This limits the cpu use to one core in worker, but limits also the memory use a lot less than with prefork default pool. One can use 'docker compose up --scale=n celery-worker' to spin up celery-worker container count to n workers if needed. 

- Related Issue #
- Closes #

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

- [X] This PR changes or adds dependencies

Updates celery dependency to 5.6.2

## Documentation

### Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -144,7 +144,7 @@ services:
     build: .
     networks:
       - main
-    command: celery -A celerywyrm worker --autoscale=10,1 -l info -Q high_priority,medium_priority,low_priority,streams,images,suggested_users,email,connectors,lists,inbox,imports,import_triggered,broadcast,misc
+    command: celery -A celerywyrm worker --pool=gevent --concurrency=1000 -l info -Q high_priority,medium_priority,low_priority,streams,images,suggested_users,email,connectors,lists,inbox,imports,import_triggered,broadcast,misc
     healthcheck:
       test: celery -A celerywyrm status
       interval: 10s
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ main = [
     "bleach==6.1.0",
     "boto3==1.34.74",
     "bw-file-resubmit==0.6.0rc2",
-    "celery==5.3.6",
+    "celery==5.6.2",
     "colorthief==0.2.1",
     "Django==5.2.9",
     "django-celery-beat==2.8.1",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ main = [
     "bleach==6.1.0",
     "boto3==1.34.74",
     "bw-file-resubmit==0.6.0rc2",
-    "celery==5.6.2",
+    "celery[gevent,redis]==5.6.2",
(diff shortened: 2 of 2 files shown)


hughrun:
I'm having trouble understanding the Celery docs - can you explain the changes to the flags in the `celery worker` command?

ilkka-ollakka:
> I'm having trouble understanding the Celery docs - can you explain the changes to the flags in the `celery worker` command?

Sure, thanks for looking this. Main idea is to change celery worker in same way as was done with gunicorn. So instead of each worker running in separate process and increasing memory usage, gevent would run workers inside same process, using same memory allocation.

By default workers used cpu-amount of workers (so memory usage is cpu-count * N) and when changed to autoscale, the max memory usage went from N to 10*N. For my setup, N is around 230M, so celery worker used between 230M and 2.3G of RAM. With gevent pooling, memory usage seems to be around 230M-290M regardless of celery jobs amount.

hughrun:
Looks good. Should we add something to the docs about this somewhere for people who need to scale up?

ilkka-ollakka:
> Looks good. Should we add something to the docs about this somewhere for people who need to scale up?

Ah good point, I'll add PR on docs to mention how to scale this up/down.

hughrun:
@ilkka-ollakka does this look correct? https://github.com/bookwyrm-social/documentation/pull/279

I also notice that there is [a note in the Celery docs](https://docs.celeryq.dev/en/stable/userguide/concurrency/gevent.html) that Python 3.11 causes errors: 

> There is a known issue using python 3.11 and gevent. The issue is documented [here](https://github.com/celery/celery/issues/8425) and addressed in a [gevent issue](https://github.com/gevent/gevent/issues/1985). Upgrading to greenlet 3.0 solves it.There is a known issue using python 3.11 and gevent. The issue is documented [here](https://github.com/celery/celery/issues/8425) and addressed in a [gevent issue](https://github.com/gevent/gevent/issues/1985). Upgrading to greenlet 3.0 solves it.There is a known issue using python 3.11 and gevent. The issue is documented [here](https://github.com/celery/celery/issues/8425) and addressed in a [gevent issue](https://github.com/gevent/gevent/issues/1985). Upgrading to greenlet 3.0 solves it.

We are using 3.11 in our docker file but testing this it seems to work fine. Not sure if you have any ideas/information on that.

ilkka-ollakka:
The python 3.11 gevent issue seemed to have resolve instructions to update greenlet to new enough version at that point, at least what I checked, current requirements are way more recent than what is needed for that fix.

The documentation update looks great.

mouse-reeve:
I applied this change manually to b.s just now to see if it improved how load from csv imports is handled

PR #3882, opened 2026-04-11
caching: set cache expire to 0 in debug mode

ilkka-ollakka:
## Description

Change the caching flow so that nginx listens caching headers from django and django sets expire times to 60 seconds by default and 0 if it is in debug mode.

This should resolve, at least partially, the caching template issue in debug mode, as nginx was also caching the template output for non-logged in users.

- Related Issue #3828 
- Closes #

- [ ] Bug Fix
- [X] Enhancement
- [ ] Refactor

## Documentation

### Tests

I tested this locally in dev-flow and outside of it, but it is possible that I have missed some case of caching fully.

### DIFF
--- a/bookwyrm/settings.py
+++ b/bookwyrm/settings.py
@@ -112,6 +112,7 @@ INSTALLED_APPS = [
 ]
 
 MIDDLEWARE = [
+    "django.middleware.cache.UpdateCacheMiddleware",
     "django.middleware.security.SecurityMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.middleware.locale.LocaleMiddleware",
@@ -125,6 +126,7 @@ MIDDLEWARE = [
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
     "bookwyrm.middleware.FileTooBig",
     "bookwyrm.middleware.ForceLogoutMiddleware",
+    "django.middleware.cache.FetchFromCacheMiddleware",
 ]
 
 ROOT_URLCONF = "bookwyrm.urls"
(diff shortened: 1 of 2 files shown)


hughrun:
This will probably work, but whilst looking at it I realised we already have `django.core.cache.backends.dummy.DummyCache` configured, which [seems to be a better option](https://django.pyth.onl/en/6.0/topics/cache/#dummy-caching-for-dev).

See settings:

https://github.com/bookwyrm-social/bookwyrm/blob/a5d7a0dde124cc11d72819adefc1f1e8735d4f0e/bookwyrm/settings.py#L247-L257


It looks like we set this up for testing, as the only place it's referenced is in the pytest env configuration. So it might be better to do something like this in `settings.py` instead of adding new middleware:

```py
if env.bool("USE_DUMMY_CACHE", False) or env.bool("DEBUG", False):
```

ilkka-ollakka:
> This will probably work, but whilst looking at it I realised we already have `django.core.cache.backends.dummy.DummyCache` configured, which [seems to be a better option](https://django.pyth.onl/en/6.0/topics/cache/#dummy-caching-for-dev).
> 
> See settings:
> 
> https://github.com/bookwyrm-social/bookwyrm/blob/a5d7a0dde124cc11d72819adefc1f1e8735d4f0e/bookwyrm/settings.py#L247-L257
> 
> It looks like we set this up for testing, as the only place it's referenced is in the pytest env configuration. So it might be better to do something like this in `settings.py` instead of adding new middleware:
> 
> ```python
> if env.bool("USE_DUMMY_CACHE", False) or env.bool("DEBUG", False):
> ```

You might be right that Dummy cache should be used in django end in dev-mode. My approach with this PR was mainly to set expire headers that nginx caching would listen. So even if django doesn't cache things, currently nginx unconditionally caches not-logged-in requests by 1 minute.

PR #3883, opened 2026-04-11
replace JS_CACHE with ManifestStaticFilesStorage

ilkka-ollakka:
## Description

Remove use of JS_CACHE cache busting query-arg. Use ManifestStaticFilesStorage (https://docs.djangoproject.com/en/5.2/ref/contrib/staticfiles/#manifeststaticfilesstorage) that adds checksum of file in path, so when file content is changed, the file-name changes and it is 
cached separately by upstream caches.

ManifestStaticFilesStorage also seems to take DEBUG=True into account and doesn't send checksummed file-name in those cases, but plain one.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


ilkka-ollakka:
This only applies to local filestorage, but not sure if js_cache query arg was usefull with s3 or azure storage?

mouse-reeve:
I believe the js_cache predates s3 support, so I doubt there’s any particular value there

PR #3884, opened 2026-04-11
bind gunicorn to ipv6 and pump nginx to 1.30.1

ilkka-ollakka:
## Description

* bind gunicorn to ipv6
* Pump nginx version to 1.30.1 (https://nginx.org/en/CHANGES-1.30) that defaults to keepalive and http/1.1 in upstream connections

Binding to IPv6 relates to enabling ipv6 in docker compose network, nginx/gunicorn uses ula-prefixed ipv6 in connection between containers.

Use of Keepalive and Http/1.1 doesn't bring usually any measureable benefit in low-traffic sites, but in really high traffic deployments might save some tcp handsake rtt worth of time in each request.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

Nginx version changed in docker-compose files.

## Documentation

### Tests

### DIFF
--- a/docker-compose-init_letsencrypt.yml
+++ b/docker-compose-init_letsencrypt.yml
@@ -1,6 +1,6 @@
 services:
   nginx:
-    image: nginx:1.25.2
+    image: nginx:1.30.1
     ports:
       - "80:80"
       - "443:443"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -7,7 +7,7 @@ x-logging:
 
 services:
   nginx:
-    image: nginx:1.28.1
+    image: nginx:1.30.1
     logging: *default-logging
     restart: unless-stopped
     ports:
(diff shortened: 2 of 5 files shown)


hughrun:
I'm wondering if it might be better to upgrade the nginx version to `1.29.7`, and [use the defaults](https://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) (`keepalive 32;` and `proxy_http_version 1.1`)?

What's the advantage of using ipv6 internally?

ilkka-ollakka:
Yes, most likely would be better to upgade to nginx 1.29.x and use new defaults from that. Good point!

For IPv6 advantages internally, there is not really much pro/against using it, other than enabling it to be there. Binding to IPv6 sockets in this case binds to both IPv6 and IPv4 sockets, so both are in use. It is more of 'if someone wants to run IPv6 only networks, it works out of box'.

hughrun:
LGTM

PR #3889, opened 2026-05-10
Add ability to block/mute books

### LINKED ISSUE #2848: Mute individual books
As a user I sometimes encounter a book that is personally disturbing to me, and I would like to not see any reference to it.

Examples might be: It is a book about a topic that is triggering to me; it is a book written by someone I have personal bad experience with; it is a book that *everyone is talking about*; it has artwork that is disturbing to me.

I would like the power to hide any post about a particular book, irrespective of who made the post. This includes:

- Reviews, quotes or comments on a book, if I follow the person who made the post.
- Reviews, quotes or comments on a book, if that post is boosted by someone I follow.
- Comments in response to the above, even if I am mentioned.
- Appearances of the book in other users' libraries. (Unresolved: what about if it's in my library? Is this how I remove a mute?)
- Trending books or mentions of the book on the Discover page.
- Search results. (Unresolved: what if I search for that exact ISBN?)
- Lists (maybe?)

All editions of a book should be included. I'm muting the work, not an edition.

The mute's duration might be configurable. Mute forever, or for a week, or a year, or ...

I should be able to browse a list of muted books so that I can review them.

I should be able to cancel muting of a book.

I should be able to export my muted books for importing into another Bookwyrm instance or sharing with other users.

The mute should apply to me only; posts about the book still needs to be visible to other users on the instance.

No one else should be able to know that I have muted a particular book.

Unresolved: I don't know how this feature request relates to Shelves.

Unresolved: I don't know how this feature request relates to Lists.

Unresolved: I don't know how this feature request relates to Filters.

I don't want to unfollow or block the person who posted a comment about the book. My problem is with the book, not the user. I also don't want to (as the admin of my multi-user instance) delete the book because other users may not have an issue with the book.

**Additional context**
This is a different feature request to muting an author.

This is a different feature request to muting a user.

hughrun:
## Description

Adds the ability to block/mute a Work and all its editions.

This will block the book from appearing in all of:

* Home and Books timelines
* Book shelves
* Discovery directory
* Lists directory
* Individual List pages and their suggestions
* User Activities on user profiles
* Annual Goal roundup page
* Annual Goal progress page
* Series pages
* Author pages

This applies to all of the above for every user, including your own profile, shelves, etc. Statuses that are in the same thread as a status relating to a blocked book are also blocked from appearing.

### Where books are not blocked

1. ~~Search results~~
2. ~~If you go to the exact URL for the book~~
3. On the Blocked Books page, so you can unblock it - this only shows title and authors, with no image

Local search results filter out blocked books, but if a blocked book has been excluded, the user is notified.

### Notes on implementation

I'm very open to suggestions if there is a cleaner way to do this - I had to effectively repeat some filtering code, but maybe there's a smarter way to do it.

In the related issue, Mouse had some concerns about ordering in Lists. This doesn't look to be a problem because the item is still in the list, it just is invisible to the people blocking it. List order can cope if we assign the same order number to two items.

### Screenshots

The _Relationships_ section heading in user settings is now _Safety_, and includes a page for blocked books as well as blocked users:

menu

User who has blocked a book can't see it in a list, but other users can:

list

list_other_user

---

- Closes #2848

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/models/book.py
```diff
@@ -806,9 +806,15 @@ def repair(self):
         self.save(update_fields=["parent_work"], broadcast=False)
 
     @classmethod
-    def viewer_aware_objects(cls, viewer):
+    def viewer_aware_objects(cls, viewer, check_blocked=False):
         """annotate a book query with metadata related to the user"""
         queryset = cls.objects
+
+        if check_blocked and hasattr(viewer, "blocked_books"):
```
mouse-reeve: I think `check_blocked` is too vague an argument name, since this is specifically about blocked books and we also can block users. That said, should this be checked if the type of objects being filtered is a book and applying the filter if so? That might eliminate the need for the argument entirely

If necessary, it seems better to pass a param to explicitly _not_ check blocked

bookwyrm/models/book.py
```diff
@@ -806,9 +806,15 @@ def repair(self):
...
+            blocked = cls.objects.filter(parent_work__in=viewer.blocked_books.all())
+            queryset = queryset.exclude(id__in=blocked)
+            # TODO: could this just be queryset = queryset.exclude(parent_work__in=viewer.blocked_books.all()) ?
```
mouse-reeve: I think it could

bookwyrm/models/status.py
```diff
@@ -276,6 +276,30 @@ def direct_filter(cls, queryset, viewer):
             ~Q(Q(user=viewer) | Q(mention_users=viewer)), privacy="direct"
         )
 
+    @classmethod
+    def safety_filter(cls, viewer, privacy_levels=None):
+        queryset = super().privacy_filter(viewer, privacy_levels=privacy_levels)
+        blocked = viewer.blocked_books.all() if hasattr(viewer, "blocked_books") else []
```
mouse-reeve: this is probably slightly more efficient if you use `viewer.blocked_books.values_list("id", flat=True)` and then filter based on `comment__book__parent_work__in`?

Also do you need the if/else here (and elsewhere)? Shouldn't the viewer always be a `User` with that database field?

mouse-reeve (commented):
This is such an exciting feature! Thank you!

bookwyrm/models/status.py
hughrun: For some reason I thought we might have to deal with users from instances on an older version but now I'm not sure why I would have thought that - probably a hangover from the work on Series.

bookwyrm/models/book.py
hughrun: In fact, we don't need this at all. I was getting myself confused by the fact we have two different methods called `viewer_aware_objects`, on different classes. `cls.objects` here _always_ refers to editions and we always want to check. Much simpler.

mouse-reeve (changes requested):
I'm still able to see blocked books in the series page and the author page. I'm notable not able to see it when I go directly to the books page, which from your description sounds like it isn't the intended behavior.

It seems like this is working as intended, but I'm not sure it _should_ show up in search? 

Also, when I unblock a book, statuses about that book from users that I don't follow show up in my home timeline, so I think there's something amiss about the activitystreams task that handles unblocking

hughrun:
> I'm still able to see blocked books in the series page and the author page. 

Fixed!

> I'm notable not able to see it when I go directly to the books page, which from your description sounds like it isn't the intended behavior.

This was not deliberate, but I've changed my mind on this. Partially because this is coming from `viewer_aware_objects` and will be a pain to separate, but mostly because I think it's correct to block the book completely until it's unblocked.

> It seems like this is working as intended, but I'm not sure it should show up in search?

I also changed my mind on this! I think you're right – we can't really stop it appearing in remote search, but we can do our best for local search. If a blocked book would have been returned, it is now removed from the results but we tell the user that it was excluded:

Screenshot 2026-05-24 at 6 58 39 PM

I'm very open to feedback on this approach and the wording in the message.

> Also, when I unblock a book, statuses about that book from users that I don't follow show up in my home timeline, so I think there's something amiss about the activitystreams task that handles unblocking

Thanks for noticing this. I have given `HomeStream` its own version of `add_blocked_book_statuses`, which I believe has fixed the problem.

mouse-reeve (approved):
LGTM!

PR #3891, opened 2026-05-11
Prevent users being added when signature validated

### LINKED ISSUE #3762: Many deleted federated accounts being created
**Describe the bug**
My instance (outside.ofa.dog) has been operating for a couple of years. In this time, it has accumulated 132,000 deleted users from Mastodon.social, and is continuing to accumulate them at the rate of one every eight or so minutes. Smaller amounts of deleted users are created from other instances.

**To Reproduce**
1. Go to Admin > Users > Federated Community
2. Notice how many pages of deleted users there are.
3. Wait a bit and refresh the page. A few more appear.

**Expected behavior**
Users that we never heard of at our instance do not get created if the first we hear of them is that the remote instance is deleting them.

**Screenshots**
Image

**Instance**
outside.ofa.dog running 0.8.0

**Additional context**
ActivityPub is known to send user deletions to pretty much every instance it knows about, to cast a wide net and ensure that everyone knows about the deletion. Consequently the first time my instance may ever hear about a user is when they are deleted.

mouse said that there is code in verbs.py to handle the Delete activity that should be handling this.

hughrun:
## Description

The inbox method `has_valid_signature` currently uses the default kwarg values. This includes 'save', which defaults to `True`. The consequence is that _every actor who ever sends anything to a BookWyrm inbox is saved to the database_.

This was most apparent for 'Delete' activities, which are sent by Mastodon to every instance it knows about, just in case. BookWyrm instances have therefore been filling up with deleted mastodon users who otherwise were unknown.

This commit:

* sets `save` to `False` in `has_valid_signature`
* provides a management command, `purge_remote_deleted_users` to:
    * report on candidates for purging (remote, deleted, not bookwyrm users, no statuses, no favorites)
    * with a `--confirm` flag, bulk-delete such users from the database
* adjusts a couple of tests that no longer passed because they now rely on a side effect from an async celery task

The command also reports how many users and from which instances they were deleted.

- Related Issue #
- Closes #3762

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

- [x] New functionality should be noted in release notes

### Tests

### DIFF
--- a/bookwyrm/management/commands/purge_remote_deleted_users.py
+++ b/bookwyrm/management/commands/purge_remote_deleted_users.py
@@ -1,0 +1,96 @@
+"""Remove remove delted users from the database"""
+
+from django.core.management.base import BaseCommand
+from django.db.models import Count, F
+from bookwyrm import models
+
+
+class Command(BaseCommand):
+    """command-line options"""
+
+    help = "Remove users that were only added because they were deleted"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--confirm",
+            action="store_true",
+            help="Delete identified users from database",
+        )
+
+    def handle(self, *args, **options):
+        """Find and report on user accounts that can be purged"""
+
+        targets = models.User.objects.filter(
+            local=False,
+            is_deleted=True,
+            is_active=False,
+            bookwyrm_user=False,
+            status=None,
+            favorites=None,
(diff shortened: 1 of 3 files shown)


ilkka-ollakka (approved):
Great finding!

I think code looks solid and makes sense, I tested the management command in dev run and it seemed to do what was expected.

PR #3905, opened 2026-05-17
bw-dev: rework dev command to a prefix

ilkka-ollakka:
## Description

Split commands to production-only (s3 commands), development flow commands (linters, pytest, makemigrations etc) and both.

Both category commands run in production by default, but if prefixed with 'dev', they can be ran against dev environment. For example './bw-dev dev migrate' or './bw-dev dev up'

Modify prod_error function also to check if command goes to production or dev instead of checking DEBUG-flag.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

Some development flow documentation might need recheck, but I think most of them should be correct. I didn't spot right away anything that wouldn't be backward incompatible.

### Tests

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -159,10 +159,10 @@ case "$CMD" in
         $DOCKER_COMPOSE rm -svf
         docker volume rm -f dev-wyrm_media_volume dev-wyrm_pgdata dev-wyrm_redis_activity_data dev-wyrm_redis_broker_data dev-wyrm_static_volume, dev-wyrm_exports_volume
         $DOCKER_COMPOSE up --build -d
-        runweb python manage.py admin_code
+        rundev python manage.py admin_code
         ;;
     makemigrations)
-        runweb python manage.py makemigrations "$@"
+        rundev python manage.py makemigrations "$@"
         ;;
     migrate)
         migrate "$@"
@@ -327,6 +327,7 @@ case "$CMD" in
         echo "Unrecognised command. Try:"
         echo " Development flow commands:"
         echo "    makemessages"
+        echo "    resetdb"
         echo "    compilemessages [locale]"
         echo "    pytest [path]"
         echo "    makemigrations [migration]"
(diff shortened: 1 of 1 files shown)


bw-dev
```diff
@@ -158,14 +154,15 @@ case "$CMD" in
...
+        $DOCKER_COMPOSE rm -svf
         docker volume rm -f dev-wyrm_media_volume dev-wyrm_pgdata dev-wyrm_redis_activity_data dev-wyrm_redis_broker_data dev-wyrm_static_volume, dev-wyrm_exports_volume
-        dev up --build -d
-        rundev python manage.py admin_code
+        $DOCKER_COMPOSE up --build -d
+        runweb python manage.py admin_code
         ;;
     makemigrations)
-        rundev python manage.py makemigrations "$@"
+        runweb python manage.py makemigrations "$@"
```
mouse-reeve: Shouldn't this still be `rundev` since it's a development only command?

bw-dev
```diff
@@ -158,14 +154,15 @@ case "$CMD" in
         $DOCKER_COMPOSE --file docker-compose-upgrade-db.yml run --rm db_update
         ;;
     resetdb)
```
mouse-reeve, line 156: I think this should also be deemed a dev command - it's so destructive that I don't want it to be easy to run on production

mouse-reeve (changes requested):
Overall I think this is a big improvement, just two commands that I believe should use `rundev`

bw-dev
ilkka-ollakka: ah yes, good point. I'll change it back

bw-dev
ilkka-ollakka: yes, I'll update this point also.

mouse-reeve (approved):
Looks good! Thank you so much for all your work on it

Issue #3890, opened 2026-05-10
Privacy and security improvements


hughrun:
This is primarily a place to track privacy and security enhancement needs and requests. It will be updated as we complete work or think of more requirements.

### Users
- [ ] provide some options for [retrospectively making statuses private](https://github.com/bookwyrm-social/bookwyrm/issues/3701)
- [x] [Users should be able to have private profiles](https://github.com/bookwyrm-social/bookwyrm/issues/2748)
- [ ] Provide ability to add privacy settings to [individual books](https://github.com/bookwyrm-social/bookwyrm/issues/3583) in shelves, readthroughs, and lists

### Admins
- [ ] retrospectively [scrub EXIF data](https://github.com/bookwyrm-social/bookwyrm/issues/3522) for legacy images
- [ ] [enable FediBlock API integration](https://github.com/bookwyrm-social/bookwyrm/issues/3413)
- [ ] [Enable admins to prevent specific usernames being chosen](https://github.com/bookwyrm-social/bookwyrm/issues/2619)
- [ ] Enable admins to prevent public posts, lists, etc
- [x] [Enable admins to force all views to require login](https://github.com/bookwyrm-social/bookwyrm/pull/3906) 
- [ ] Enable ["Local Only" posting](https://github.com/bookwyrm-social/bookwyrm/issues/559)
- [ ] Enable ["Local Only" lists](https://github.com/bookwyrm-social/bookwyrm/issues/2831)
- [ ] Enable admins to restrict federation to allow list of instances

### Plumbing/Infrastructure

- [x] [Verify HTTP signatures for GET requests](https://github.com/bookwyrm-social/bookwyrm/pull/3906) (including JSON) as well as for POST ("secure mode" / AUTHORIZED_FETCH)

phanky1:
I'd love to help with the [Users should be able to have private profiles](https://github.com/bookwyrm-social/bookwyrm/issues/2748) if you're not already on it.

hughrun:
Go for it 🙂

Issue #3701, opened 2025-09-27
make everything private (or followers only)


Skivling:
In the scenario that I no longer want my profile to be publicly accessible, it is difficult to properly make everything private.

The easy steps are changing default posting settings, requiring follower confirmation and changing shelf privacy. However, all previous statuses remain public. The UI seems to let them be changed, but due to ActivityPub this likely is not federated and will not actually lead to the outcome one expects. 

Making a private account not show any posts on the profile page would be an improvement, but this does not directly solve federated views of them. If other bookwyrm sites also hid the posts, but it is unlikely that every fediverse platform would be able to.

Are there any workarounds, like migrating the account? I seem to recall that one gets to decide what privacy level to put the new posts as, does this migrate every single status and all database info? 
Then the original account could be deleted and eventually it would disappear from federated servers. The main issue is conversations and post interactions cannot be kept intact.

hughrun:
> The UI seems to let them be changed, but due to ActivityPub this likely is not federated and will not actually lead to the outcome one expects.

This feels like a bug: the nature of ActivityPub is that you _can't retrospectively change status permissions_. Permissions determine to which inboxes the status is sent, and once they're delivered you can't really "undeliver" them. This is the basic trade-off of federated media: you can't be "cancelled" by a single monopoly platform, but also you have less control once statuses are published. So if we're allowing users to think they can change publishing permissions on already-published statuses we should fix that because it's very misleading.

> I seem to recall that one gets to decide what privacy level to put the new posts as, does this migrate every single status and all database info?

Alas this is not the case - importing posts will retain the original permissions. We could maybe adjust that, but I think it might cause some strange effects if people who originally replied can no longer see the conversation, for example.

jaschaurbach:
In fact ActivityPub allows update of visibility updates of a post or a note. 

As it is nature that posts get federated and we do not know the update cycle due to unforseen circumstances (it is always DNS!) we can not promise that the visibility will really change - especially when there is a cached version, etc. 

I am against an update of visibility. Delete is most times working so maybe we should have a "delete and recreate with new visibility and new timestamp" function. (I think BlueSky does this - delete a post and recreate a new post with the fixes)

hughrun:
Delete-and-republish is a great idea.

PR #3906, opened 2026-05-18
add custom security middleware

hughrun:
## Description

This middleware adds a number of security settings for administrators. Note that these are settings in the Federation administration area, not env values.

## `require_signed_get`

Requires signed GET requests in addition to signed POST requests. This is loosely equivalent to Mastodon's "secure mode", except we don't check JSON-LD signatures.

BookWyrm has been signing GET requests for ActivityPub data since 2022, but without this setting we don't check them.

the `.well-known` endpoints are excluded, otherwise user discovery breaks, but you don't really discovery anything about a user from that endpoint.

## `require_login_nearly_everywhere`

This setting effectively adds `login_required` to every page other than the absolute basics required to manage logins, invites, and basic info about an instance. It also hides the "recent books" from the homepage.

`require_login_everywhere` does _not_ apply to API requests, because they can't log in the same way a normal user does. To restrict API calls admins will need to enable `require_signed_get` and be careful about federation. We probably should consider a setting for federation "allow lists" (i.e. the inverse of a block list), but that is out of scope for this PR. The nuclear option is the disable federation, which is already available.

## `block_incoming_search`

Disallows use of the search endpoint by peers.
I don't think it's generally a great idea to do this because it's likely to increase the number of duplicate/orphaned records, but for instances that need to be particularly careful about how easily their collection can be discovered, this may be desirable.

Requires (nearly) all GET 

- Related Issue #3890 #3701
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
~~Not sure what's causing these test errors~~ - tests all pass locally for me.
UPDATE:  I needed to mock `get_activity_stream`

hughrun:
hmm actually I think the search block is a bit wrong too. Will update that.

bookwyrm/middleware/bookwyrm_security.py
```diff
@@ -0,0 +1,156 @@
...
+            SiteSettings.get().require_signed_get
+            and request.method == "GET"
+            and is_api_request(request)
+        ):
+            # if we disabled federation, disallow all API requests
+            # this will include nodeinfo and webfinger etc
+            SiteSettings.raise_federation_disabled()
+
+            # check if this server is on our shitlist
+            raise_is_blocked_user_agent(request)
```
mouse-reeve: Should this check be happening here? It seems like it's relevant in evaluating any api request, regardless of require_signed_get

bookwyrm/middleware/bookwyrm_security.py
```diff
@@ -0,0 +1,156 @@
...
+    r"^/invite/(?P<code>[A-Za-z0-9]+)/?$",
+    r"^/2fa-check/?$",
+]
+
+allowed_api_views = [
+    r"^/\.well-known/webfinger/?$",
+    r"^/\.well-known/nodeinfo/?$",
+    r"^/\.well-known/host-meta/?$",
+    r"^/nodeinfo/2\.0/?$",
+]
```
mouse-reeve: Would it make more sense to create a decorator for these views, rather than listing these out here? Like how `@require_federation` works: https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/decorators.py

bookwyrm/middleware/bookwyrm_security.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+class BookWyrmSecurityChecks:
```
mouse-reeve: I feel like this has too much going on in one function, from a readability perspective. I'd prefer to break it down in some way, whether using decorators would help, or just creating separate middleware for each functionality?

bookwyrm/middleware/bookwyrm_security.py
```diff
@@ -0,0 +1,156 @@
...
+    """lock down incoming access depending on site settings"""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """run before next middleware or view"""
+
+        # block search endpoint if disabled
+        # TODO - actually this should be "^/?search/?$" (i think) and is_api_request
```
mouse-reeve: This TODO looks resolved?

mouse-reeve (commented):
This is a great change! I have some implementation questions:

bookwyrm/middleware/bookwyrm_security.py
hughrun: Yes! TBH I added it here because I was reading about how Mastodon implemented the equivalent, but yes we should be checking every API call for that, and if it's to a user endpoint maybe also checking if the requestor is blocked personally? Though in most cases I expect it will be a server instance actor.

bookwyrm/middleware/bookwyrm_security.py
hughrun: oops!

bookwyrm/middleware/bookwyrm_security.py
hughrun: I already know I agree with this because I split the tests into two test classes. Probably should have realised at that point 😅.

I see what you're saying about decorators and it would be a bit more elegant, but I think it's a dangerous way to implement safety checks on endpoints. If someone makes a new view (e.g. we make Groups AP-aware) and we use decorators, then security on the view becomes "opt in", whereas if we put it in middleware it becomes opt-out. You can forget to exclude a view and that's just annoying, but also probably will be picked up in tests. If you forget to _include_ a view, that's harder to spot.

If you're comfortable with that reasoning, I'll make two changes:

1. make `block_incoming_search` a decorator because it literally only applies to one view
2. split the other two into separate middlewares

mouse-reeve: That absolutely makes sense! I tried to think of a clever way to make an opt-out decorator but didn't come up with anything 🤷

hughrun: Yeah as far as I can tell it's impossible because to make something apply universally you'd use middleware, but middleware applies prior to the view (and therefore the decorator opting it out) being processed.

mouse-reeve:
On the face of it, this all looks good to me (there are a couple trivial linting issues since I made ruff stricter), but I'm not really sure how to test authorized fetch?

hughrun:
Apologies for all the commits, I had some merge conflicts and ended up making a bit of a mess with rebase.

Regarding how to test authorised fetch - you can send a GET request that asks for json but doesn't have signed headers. e.g. using `httpie`:

```sh
http -j https://example123.ngrok-free.app/user/mouse/generatednote/1 # returns json object
# turn on "Require signed GET requests"
http -j https://example123.ngrok-free.app/user/mouse/generatednote/1 # returns 403 error
```

Or you can try to view a `.json` endpoint in a browser and it will fail to load anything

If you want to test it _with_ signed headers, send a follow request to your test site from BookWyrm and/or Mastodon, both of which already sign GET headers by default. If you can successfully follow and get user information, including e.g. shelves and follower counts, that indicates that it works.

mouse-reeve:
Looks good! I'm going to squash marge to circumvent the rebase commits

PR #3910, opened 2026-05-20
fix: List 'edit notes' exposes html

### LINKED ISSUE #3608: List 'edit notes' exposes html
When editing the note added to a book on a list, the raw html is shown:

Image

This is not a good idea for usability as most people don't know html.

leni1:
## Description
This fixes the "Edit notes" feature on Book Lists to no longer show HTML tags when the user tries to edit a note they added to a book list.

- Closes #3608 

- [x] Bug Fix

A migration for the List model that adds the `raw_notes` field.

## Documentation

- [x] No documentation is required as a result of this change.

### Tests


bookwyrm/tests/views/lists/test_list_item.py
```diff
@@ -71,4 +71,5 @@ def test_add_list_item_notes(self):
         self.assertEqual(mock.call_count, 1)
 
         item.refresh_from_db()
-        self.assertEqual(item.notes, "<p>beep boop</p>")
+        self.assertEqual(item.notes, "beep boop")
```
mouse-reeve: It looks like you just have note and raw_notes backwards here

leni1: Thanks :raised_hands: Fixed in [877a1ba](https://github.com/bookwyrm-social/bookwyrm/commit/877a1ba78d677eed3861da85a24a3c68df66ebcc)

bookwyrm/views/list/lists.py
```diff
@@ -45,7 +45,6 @@ def post(self, request):
         # list should not have a group if it is not group curated
         if not book_list.curation == "group":
             book_list.group = None
-        book_list.save()
```
mouse-reeve, line 48: I don't think this line should be deleted?

mouse-reeve (changes requested):
You've got a good start here, but there are a few things left to do:
 - The edit notes field in the list view still shows the HTML compiled notes, rather than `raw_notes`. You'll need to update the [list item django form](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/lists.py#L20) and the [template that is used to edit notes](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/lists/item_notes_field.html#L16) and get them using the correct field
 - When I add a list item from the suggestions sidebar, it saves the `raw_notes` but doesn't generate the HTML version and save it in the `notes` field, so the notes don't show up in the application
 - Creating a list no longer works because of the removed `save`

Make sure you manually go through the steps of creating a list, adding an item, adding notes, and editing notes in your developer environment to confirm that your changes work as expected. The tests help but they don't catch everything, especially when it comes to forms and templates.

Let me know if you need any help or clarification! I'm really glad you've taken it on

bookwyrm/views/list/lists.py
leni1: Good catch. It seems I went overboard in my modifications here. Fixed in [af58a3e](https://github.com/bookwyrm-social/bookwyrm/commit/af58a3e82511e8818dc0606eb15bcd8b93f0f6f3)

leni1:
> You've got a good start here, but there are a few things left to do:
> 
>     * The edit notes field in the list view still shows the HTML compiled notes, rather than `raw_notes`. You'll need to update the [list item django form](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/lists.py#L20) and the [template that is used to edit notes](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/templates/lists/item_notes_field.html#L16) and get them using the correct field

Done in [da76e47](https://github.com/bookwyrm-social/bookwyrm/commit/da76e477cad7af2613716ff72e24b3660d099573). 

>     * When I add a list item from the suggestions sidebar, it saves the `raw_notes` but doesn't generate the HTML version and save it in the `notes` field, so the notes don't show up in the application

Done in [fbd4f6f](https://github.com/bookwyrm-social/bookwyrm/commit/fbd4f6fa6b7082292e5887a1365e1614c4a6b5b7)

>     * Creating a list no longer works because of the removed `save`

Done in [877a1ba](https://github.com/bookwyrm-social/bookwyrm/commit/877a1ba78d677eed3861da85a24a3c68df66ebcc)

> Make sure you manually go through the steps of creating a list, adding an item, adding notes, and editing notes in your developer environment to confirm that your changes work as expected. The tests help but they don't catch everything, especially when it comes to forms and templates.

Good point :raised_hands: I mentally made a note to do this before requesting for a review but I forgot :upside_down_face: 
So the changes work now for books that are newly added (you no longer see the HTML tags when you try to edit the notes on a book) and if you add a book to a list with a note, the book should appear in the list with the note you added with it. But they don't work retrospectively for books added when the tags were saved as part of the note. Not sure how to fix that (I tried setting the note to nothing and re-adding the note afresh but the tags still showed on the book).

> Let me know if you need any help or clarification! I'm really glad you've taken it on

So for the time-being, I am not going to merge `main` just yet because I am not sure if this is complete.
Once it is, then I will merge `main` to update the migration, after which it can be merged.

mouse-reeve:
I prefer to have a different person merge a PR when possible, so I’ll merge once I get a chance to test it and get the workflows going

mouse-reeve (approved):

mouse-reeve:
This looks good! I'm just re-ordering the migration files so it's merge-able. Thank you so much for the fix!

PR #3915, opened 2026-05-22
Adds arbitrary dev command runner to bw-dev

mouse-reeve:
## Description

This allows me to run whatever docker compose command I wan in dev without having to supply the environment files. For example, if I want to restart my local environment, I can run `./bw-dev devcommand restart`.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bw-dev
+++ b/bw-dev
@@ -128,6 +128,9 @@ case "$CMD" in
     down)
         $DOCKER_COMPOSE down
         ;;
+    devcommand)
+        $DOCKER_COMPOSE -f docker-compose.yml -f docker-compose.dev.yml  --env-file .env --env-file .env.dev "$@"
+        ;;
     docker_cache_cleanup)
         echo "Showing current docker storage usage"
         docker system df


hughrun:
I love the principle, but I'm having trouble understanding how this is different to running `./bw-dev dev whatever` after #3905 ?

mouse-reeve:
After #3905 you can't run arbitrary commands with `dev`, just ones that are in `bw-dev` (unless I'm really missing something)

hughrun:
Unless I'm missing something, you can run anything you like with the `dev` prefix except `*media_to_s3` commands. The difference is that some commands _only_ run against dev, and don't need the prefix:

```sh
bookwyrm git:secure-mode ❯ ./bw-dev dev restart
+ case "$CMD" in
+ set +x
Unrecognised command. Try:
 Development flow commands:
    makemessages
    resetdb
    compilemessages [locale]
    pytest [path]
    makemigrations [migration]
    update_locales
    ruff (format and check)
    ruff-format (format code only)
    ruff-check (check code only)
    ruff-fix (auto-fix ruff issues)
    prettier
    eslint
    stylelint
    formatters
    mypy

 Commands that run in prod, but can be prefixed with 'dev' to run in devevelopment environment:
  To run command in dev environment, prefix it with 'dev'
  example './bw-dev dev migrate' or './bw-dev dev up -d
    setup
    up [container]
    down
    service_ports_web
    upgrade_db_version
    initdb
    init_ssl
    migrate [migration]
    create_secrets
    bash
    shell
    dbshell
    restart_celery
    compile_themes
    collectstatic
    build
    docker_cache_cleanup
    clean
    collectstatic_watch
    populate_streams [--stream=<stream name>]
    populate_lists_streams
    populate_suggestions
    generate_thumbnails
    generate_preview_images [--all]
    remove_remote_user_preview_images
    set_cors_to_s3 [cors file]
    runweb [command]
    remove_2fa
    confirm_email

 Command that only operate production
    copy_media_to_s3
    sync_media_to_s3
```

mouse-reeve:
I'm trying to use it with arbitrary docker commands that aren't in the bw-dev file. In this case, when you try `restart`, it gives you a list of bw-dev commands, but what I want is for it to run `docker compose -f docker-compose.yml -f docker-compose.dev.yml --env-file .env --env-file .env.dev restart`

mouse-reeve:
Sorry my brain is mush now, so I don't know if that made sense. But previously, you could run `docker compose <whatever>` locally to do custom docker stuff that isn't encompassed by bw-dev. Now, `docker compose` requires a bunch of flags for env files, which makes it cumbersome to run. `./bw-dev dev <wahtever>` used to work, but since the file has been reworked, it doesn't (but that change is still an overall big win). The command I added just effectively aliases `docker compose -f docker-compose.yml -f docker-compose.dev.yml --env-file .env --env-file .env.dev`

hughrun:
Thanks for the explanation! Makes sense to me but I'd like an opinion from @ilkka-ollakka

ilkka-ollakka:
Yes, it was lost in last change to run other docker compose command to dev than listed. I think the change makes sense and has use, so looks good from my point of view.

ilkka-ollakka (approved):

Issue #3921, opened 2026-05-25
Series books ordered alphabetically rather than numerically


jakkarth:
**Describe the bug**
Viewing a book series sorts the books in order alphabetically, rather than numerically. Thus, you see books 1, 10, 15, 2, 23, etc.

**To Reproduce**
1. Add three books in a series as 1, 2, and 10
2. Click on the series name to view the series
3. See books ordered 1, 10, 2 instead of 1, 2, 10

**Expected behavior**
Order books numerically, including subnumerals like putting 4.5 between 4 and 5, and 10 between 9 and 11 rather than between 1 and 2.

**Additional context**
Possible source of problem: /bookwyrm/views/books/series.py#L38

---

**Desktop (please complete the following information):**
    - OS: linux
    - Browser: firefox
    - Version: 150.0.3

hughrun:
SeriesBook "numbers" (a poor naming choice from me) are deliberately not necessarily `int`s, so we may need to reorder these using a Python method instead of relying on Django's `order_by`.

jakkarth:
Perhaps we could check to see if all of the values parse as floats, and if so apply numeric sorting rather than string? Or separate floatable and non-floatable values, sort them numerically and alphabetically respectively, then append one to the other? Calling it "number" always made sense to me, and I'm curious what other values one might put in that field that aren't numbers.

hughrun:
@jakkarth depending on the source of the data, they can come through as, for example, "Book 1", so we need to allow that.

jakkarth:
Okay. I don't think it would be possible to parse all variations of that. Either of the suggestions I made would handle those by sorting them alphabetically, which isn't ideal, but at least improves the situation for cases where the numbers are in fact numeric.

jakkarth:
Fixed in #3926

PR #3926, opened 2026-05-27
Sort series books numerically

phanky1:
## Description

SeriesBook.series_number is currently a CharField, so ordering is lexicographic. This PR updates ordering so that leading numeric values in series_number are compared numerically. For example, given:
`["10", "2-beta", "1", "Prequel", "2", "4.5", "2-alpha", "1.5-rc"]`

Current ordering: `["1", "10", "1.5-rc", "2", "2-alpha", "2-beta", "4.5", "Prequel"]`
New ordering: `["1", "1.5-rc", "2", "2-alpha", "2-beta", "4.5", "10", "Prequel"]`

- Related Issue: #3921

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -1003,3 +1003,14 @@ class SeriesBook(CollectionItemMixin, BookWyrmModel):
     def raise_not_editable(self, viewer):
         if not viewer.has_perm("bookwyrm.edit_book"):
             raise PermissionDenied()
+
+    @property
+    def natural_sort_key(self):
+        """numeric-aware key for series_number, so '2' sorts before '10'"""
+        value = self.series_number or ""
+        if not value:
+            return float("inf"), ""
+        match = re.match(r"\d+(?:\.\d+)?", value)
+        numeric_prefix = float(match.group()) if match else float("inf")
+        rest = value[match.end() :] if match else value
+        return numeric_prefix, rest
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -1007,8 +1007,8 @@ class SeriesBook(CollectionItemMixin, BookWyrmModel):
     @property
     def natural_sort_key(self):
         """numeric-aware key for series_number, so '2' sorts before '10'"""
-        value = self.series_number or ""
-        if not value:
+        value = self.series_number
+        if value is None or len(value) == 0:
             return float("inf"), ""
         match = re.match(r"\d+(?:\.\d+)?", value)
         numeric_prefix = float(match.group()) if match else float("inf")
--- a/bookwyrm/templates/book/series.html
+++ b/bookwyrm/templates/book/series.html
@@ -76,5 +76,8 @@
 	{% endfor %}
 	</div>
 	{% endwith %}
+    <div>
+		{% include 'snippets/pagination.html' with page=series_books path=request.path %}
+	</div>
 </div>
 {% endblock %}
--- a/bookwyrm/templates/book/series.html
+++ b/bookwyrm/templates/book/series.html
@@ -64,7 +64,7 @@
 	</div>
 
 	<div class="columns is-multiline is-mobile">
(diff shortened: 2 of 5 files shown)


bookwyrm/models/book.py
```diff
@@ -1003,3 +1003,14 @@ def get_remote_id(self):
...
         if not viewer.has_perm("bookwyrm.edit_book"):
             raise PermissionDenied()
+
+    @property
+    def natural_sort_key(self):
+        """numeric-aware key for series_number, so '2' sorts before '10'"""
+        value = self.series_number or ""
+        if not value:
+            return float("inf"), ""
+        match = re.match(r"\d+(?:\.\d+)?", value)
```
jakkarth, line 1013: > This PR updates ordering so that leading numeric values in series_number are compared numerically.

It looks like this will match digits anywhere, not just at the beginning of the value.

bookwyrm/views/books/series.py
```diff
@@ -31,24 +29,15 @@ def get(self, request, series_id, slug=None):
...
-        )
-        for item in items:
-            book = item.book.work.default_edition
-            book_data = {"book": book, "series_number": item.series_number}
-            books.append(book_data)
-            authors = authors.union(item.book.authors.all())
+        items = series.seriesbooks.prefetch_related("book__work", "book__authors").all()
+        series_books = sorted(items, key=lambda sb: sb.natural_sort_key)
 
-        paginated = Paginator(books, PAGE_LENGTH)
```
jakkarth, line 46: Is removal of the pagination functionality here intentional? It seems unrelated to the sorting behaviour.

hughrun: Pagination should be retained, using `series_books`. However, gosh I made a real mess of the original view and template here. The page template is missing the pagination include, so all you get is the first page.

bookwyrm/templates/book/series.html
hughrun, line 77: This was my mistake originally, but we need to add a pagination snippet here:

```html
	<div>
		{% include 'snippets/pagination.html' with page=series_books path=path %}
	</div>
```

bookwyrm/views/books/series.py
```diff
@@ -31,24 +29,15 @@ def get(self, request, series_id, slug=None):
...
-        page = paginated.get_page(request.GET.get("page"))
+        for item in series_books:
+            authors = authors.union(item.book.authors.all())
```
hughrun, line 43: This doesn't work, partially because it recreates the author list on each book, and partially because authors are stored on Editions rather than Works. Again, I believe I got this wrong in the first place, and we don't need a loop. Consider this approach instead:

```py
items = series.seriesbooks.prefetch_related("book__work", "book__work__editions__authors").all()
series_books = sorted(items, key=lambda sb: sb.natural_sort_key)
authors = models.Author.objects.filter(
    id__in=items.values_list("book__work__editions__authors")
)
```

bookwyrm/models/book.py
hughrun: Given the context, I think this is the correct approach. I'd expect "Book 2" to appear after "1" and before "3".

hughrun (changes requested):
@diaxoaine Thanks so much for this!

A few little things need tidying, but mostly that's because you need to clean up the mess I made in the first place.

bookwyrm/models/book.py
phanky1: `re.match` applies the regex at the start of the string, anw I added test cases to cover numbers that aren't at the start

phanky1:
> @diaxoaine Thanks so much for this!
> 
> A few little things need tidying, but mostly that's because you need to clean up the mess I made in the first place.

I added back the pagination.

hughrun (approved):
Thanks for this @diaxoaine, looks great.

PR #3932, opened 2026-06-03
Switch celery pool to threads from gevent

phanky1:
## Description
The broadcast task is using `asyncio`, which can conflict with the `gevent` worker pool because both run on the same single thread. `asyncio.run` creates an event loop while it is running, so if the worker receives another task and accesses the Django ORM, we will see this error:
```
django.core.exceptions.SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
```
This PR switches ActivityPub broadcast fan-out from asyncioto gevent.

- [ ] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests


ilkka-ollakka:
One alternative would be to change from gevent to for example threads in celery worker pool?

phanky1:
> One alternative would be to change from gevent to for example threads in celery worker pool?

yeah thread pool would work with the least changes. But I assume the intention behind gevent pool is to save memory, so i chose to keep gevent pool

ilkka-ollakka:
> yeah thread pool would work with the least changes. But I assume the intention behind gevent pool is to save memory, so i chose to keep gevent pool

Intention with gevent pool was to reduce memory usage with multi-cpu serves compared to pre-fork pool, which is default. But I missed that flow that it caused issues on broadcast when I made the change. Great that you spotted that issue!

I would recommend to check if the issue reproduces with threads pool and if not, change to that, as gevent change is quite recent and there is no clear reasons to stay in that compared to threads-pool for example.

phanky1:
> Intention with gevent pool was to reduce memory usage with multi-cpu serves compared to pre-fork pool, which is default. But I missed that flow that it caused issues on broadcast when I made the change. Great that you spotted that issue!
> 
> I would recommend to check if the issue reproduces with threads pool and if not, change to that, as gevent change is quite recent and there is no clear reasons to stay in that compared to threads-pool for example.

I've verified that thread pool doesn't have this issue, so I switched over to it.
Also I reduced the concurrency from 1000 to 100 because I think that gives us a better balance between DB connection usage and task concurrency

ilkka-ollakka (approved):

PR #3937, opened 2026-06-07
Fix federate reports

hughrun:
## Description

- fixes attached statuses not being sent to external instances
- fixes mastodon `Flag` activities crapping out
- adjusts tests

Essentially what I've done here is:

* allow external connections, which _seems_ to be necessary (?)
* stop local reports broadcasting before the M2M statuses have been attached
* fixes to catch incoming Mastodon `object`s with the reported user listed as the first item.

I don't completely understand why, but mastodon was causing weird issues where the first time a status was reported it wouldn't trigger an `ActivitySerializerError` – I think because we have to do an external request and that overrides any assertions about type? Anyway it would then fail at the point it's added as a status because a user is not a status. 

- Related Issue #
- Closes #

- [x] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests


mouse-reeve (approved):
Thank you _so_ much for getting this one going! I will test more thoroughly on the original PR once it's merged

PR #3941, opened 2026-06-09
Fix importers ignoring shelf column

### LINKED ISSUE #3876: LibraryThing import puts everything onto to-read shelf

### LINKED ISSUE #3887: reading Import seems inconsistnet
I can't for the life of me figure out how to get books to import to the correct shelf

On my last attempt I included only books that i have already read, changed collections to shelf and put every shelf to read (lower case) and still most books are being importated to-read and those that are imported correctly to the read shelf I can't figure out what is differenet so as to copy for the rest. 

attached is last import attempt and origional export from LibraryThing

[read.tsv](https://github.com/user-attachments/files/27397562/read.tsv)

[librarything_fulner_202605050742.tsv](https://github.com/user-attachments/files/27397655/librarything_fulner_202605050742.tsv)

phanky1:
## Description

LibraryThing, OpenReads, and Calibre imports were ignoring `shelf` header from import files, causing books to default to to-read in cases where they should have used other values.

This updates those importers to fall back to the base `Importer.get_shelf` logic when date-based shelf detection does not apply.

- Closes https://github.com/bookwyrm-social/bookwyrm/issues/3887
- Closes https://github.com/bookwyrm-social/bookwyrm/issues/3876

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


mouse-reeve:
This seems like a good and sensible fix, but I'm having a hard time testing it without example files that include a "shelf" field or equivalent. I can see how it works once the parsed export has the "shelf" attribute in its object, but I'm trying to verify that the "shelf" attribute gets created in the first place. Do you have test files that you would be able to share?

phanky1:
> This seems like a good and sensible fix, but I'm having a hard time testing it without example files that include a "shelf" field or equivalent. I can see how it works once the parsed export has the "shelf" attribute in its object, but I'm trying to verify that the "shelf" attribute gets created in the first place. Do you have test files that you would be able to share?

[read.tsv](https://github.com/user-attachments/files/28907329/read.tsv)
You could use the files attached to #3887.

This is how it looks in my dev environment:
image

mouse-reeve:
Gotcha, thank you! Works like a charm :)

PR #3952, opened 2026-06-16
Simplifies PR template

mouse-reeve:
## Description
I think we've erred on the side of being too wordy and detailed in the template. When I open a PR, the comments take up enough space at the top that the place where I would enter my description is not shown, which makes it hard to find. I've removed some lines that I think are self-explanatory (like the comment about what the description section is for), moved the welcome-type message down lower so that it's still present but not the first thing you hit every time you open a PR, and shortened some of the other messaging.

Here's how it looks on my screen currently:
Screenshot 2026-06-16 at 10 02 18 AM

Versus a
Screenshot 2026-06-16 at 10 08 49 AM
fter this change:

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,32 +1,14 @@
-<!--
-Thanks for contributing! This template has some checkboxes that help keep track of what changes go into a release.
-
-To check (tick) a list item, replace the space between square brackets with an x, like this:
-
-- [x] I have checked the box
-
-You can find more information and tips for BookWyrm contributors at https://docs.joinbookwyrm.com/contributing.html
--->
 ## Description
-<!--
-Describe what your pull request does here
--->
 
-
-<!--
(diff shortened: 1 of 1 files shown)


hughrun (approved):
Nice one, looks good.

While you're looking at this, I'm wondering about the tests section. It bothers me that a fourth option isn't represented (I updated existing tests to make them pass with my new functionality), but I don't think it would be particularly helpful as an option. Maybe if someone added/edited tests it's obvious and we don't need them to select anything? The main reason make people look at that section at all is really to remind them they need to think about test coverage.

mouse-reeve:
I think that's more nuance than necessary. We don't merge if tests aren't passing, and it will be clear in the diff if the tests were changed, so imo that section is sufficient as is

PR #3960, opened 2026-06-16
Create mergeable-mixin for book data models

### LINKED ISSUE #562: Allow adding books as an edition when editing their entry
I recently read The History of Standard Oil and logged the main search result. After finishing the book in BookWyrm, I found that the [version of the book I was actually reading](https://bookwyrm.social/book/92639) was available from Open Library, but it wasn't considered one of the [editions](https://bookwyrm.social/book/8284/editions) of the book I logged. I would like to edit the book's entry so that I can switch to having logged the Belt Publishing version I read.

When editing [a book entry](https://bookwyrm.social/book/92639), I would like to be able to associate it to [an edition](https://bookwyrm.social/book/8284/editions). Not sure what the user interaction for this should look like.

**Additional context**

### LINKED ISSUE #1119: Dealing with Duplicate Authors
Currently an author can be registered separately as [Soren Kierkegaard](https://bookwyrm.social/author/6166), [Søren Kierkegaard](https://bookwyrm.social/author/18619), and  [Sören Kierkegaard](https://bookwyrm.social/author/8779), all of which are treated separately and will have books split among them.

Giving users the ability to mark Soren Kierkegaard and Sören Kierkegaard as "aliases" of Søren Kierkegaard would make the cataloguing a lot cleaner.

### LINKED ISSUE #2898: Add UI for merging different works into editions of the same work

### LINKED ISSUE #3490: Merge authors

We see lots of duplicate authors that are actually the same.

The ability to merge authors that show up on a search list.

We end up having to edit all the books so that they point to the "right" author.

### LINKED ISSUE #3493: Let users merge duplicate entries
Frequently you will find duplicate entries of the same book. There is no apparent way as a user to contribute to fixing duplicate entries. All I found for that was for administrators of the instance.

Instance users will typically be the ones to discover duplicate entries. Ideally users can contribute themselves to fixing that problem. Suggestion: add a button either on the page of the book or within the edit book form, titled "Merge duplicate book entries". Users can then enter either the number of the book entry on their instance or a link to it. Where data of the two entries are different, offer to either add them as separate editions or to prioritise the data based on one of the entries. Bonus: let users choose per individual data part, which one to prioritise or even edit them during the merging process. The old entry will then link to the merged entry, so that reviews etc. will still lead to the correct book and show up on its page. Bonus: send notification to users who wrote a review for an entry that was merged, to let them remove their review if it has been maliciously / falsely merged to a different book than what they had reviewed. 

Provide a guide for users to manually edit duplicate entries in a way that will merge them.
Provide a button to notify the administrator of an instance about duplicate entries, which allows to report multiple entries at once.

mouse-reeve:
## Description
Adds a mixin to book/author data models for detecting and merging duplicate entries

- Related Issue #3379
- Closes #3490
- Closes #3493
- Closes #2898
- Closes #1119 
- Closes #562

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/models/book.py
```diff
@@ -89,6 +137,9 @@ class BookDataModel(ObjectMixin, BookWyrmModel):
         null=True,
     )
 
+    pending_merge_date = models.DateTimeField(null=True)
```
mouse-reeve: This might not be needed

bookwyrm/models/book.py
```diff
@@ -89,6 +137,9 @@ class BookDataModel(ObjectMixin, BookWyrmModel):
...
+    prevent_automatic_merge = models.BooleanField(default=False)
```
mouse-reeve: I thought I could add these with the mixin but it didn't seem to pick it up in the migration so I guess not.

bookwyrm/models/mergeable_mixin.py
```diff
@@ -0,0 +1,146 @@
...
+            for value in values:
+                objs = cls.objects.filter(**{field_name: value}).order_by("id")
+                if not objs.exists() or objs.count() <= 1:
+                    continue
+                canonical = objs.first()
+                candidates = objs.exclude(id=canonical.id)
+                candidates.update(pending_merge_target=canonical)
+
+    def find_merge_candidate(self):
+        """look for the first possible duplicate of the current object"""
```
mouse-reeve: I'm not actually sure if this method will be useful. I don't think it will be for automatic merge since that works by scanning the whole model

bookwyrm/models/book.py
mouse-reeve: Nvm it is neeeded

bookwyrm/models/mergeable_mixin.py
hughrun: is this effectively the same as finding all `pending_merge_target`s?

mouse-reeve: Yes, but I wrote it before I realized it made sense to store pending merge targets. I think it's better to use the approach of scanning for duplicates en mass and looking up individual duplicates based on the results of the scan.

bookwyrm/models/book.py
ilkka-ollakka: how so? I think the prevent_automatic_merge-fields aren in migration

mouse-reeve: They are in the migration because they're in the Book model, but if I put them in the mixin, they aren't added. I tried moving them just now and the migration it created deleted the fields.

ilkka-ollakka: I think it just lacked 
```
class Meta:
   abstract = True
``` 
in MergeableMixin class. Adding that kept the migrations when moving pending_merge_date and prevent_automatic_merge to MergeableMixin class.

mouse-reeve: Ah thank you! I wouldn't have thought of that

bookwyrm/models/mergeable_mixin.py
mouse-reeve: Wait I see why it was needed - this line copies over data from the dupe to canonical: https://github.com/bookwyrm-social/bookwyrm/pull/3960/changes/BASE..83cb78505a958a0a00fb1cbc49108c1c13ca38d3#diff-df0735a0ac54cfe0f70835ac949aa5a26881b6614d3cb6c7dc59bb31084f0261R79

bookwyrm/models/book.py
mouse-reeve: Okay I tried this and it still didn't work? I must be missing something obvious.

ilkka-ollakka: I had diff like this and it didn't do any new migrations for me, might that the models.Model is also needed for MergeableMixin:

```
diff --git a/bookwyrm/models/book.py b/bookwyrm/models/book.py
index 399092355..25e06a588 100644
--- a/bookwyrm/models/book.py
+++ b/bookwyrm/models/book.py
@@ -137,9 +137,6 @@ class BookDataModel(ObjectMixin, MergeableMixin, BookWyrmModel):
         null=True,
     )
 
-    pending_merge_date = models.DateTimeField(null=True)
-    prevent_automatic_merge = models.BooleanField(default=False)
-
     @property
     def openlibrary_link(self):
         """generate the url from the openlibrary id"""
diff --git a/bookwyrm/models/mergeable_mixin.py b/bookwyrm/models/mergeable_mixin.py
index 981ebe95d..d01df50c3 100644
--- a/bookwyrm/models/mergeable_mixin.py
+++ b/bookwyrm/models/mergeable_mixin.py
@@ -7,13 +7,22 @@ from typing import Any, Dict
 from typing_extensions import Self
 
 from django.db.models import Count, ManyToManyField, Q
+from django.db import models
 
 from . import fields
 
 
-class MergeableMixin:
+class MergeableMixin(models.Model):
     """A bookwyrm data object that can be deduplicated"""
 
+    pending_merge_date = models.DateTimeField(null=True)
+    prevent_automatic_merge = models.BooleanField(default=False)
+
+    class Meta:
+        """can't initialize this model, that wouldn't make sense"""
+
+        abstract = True
+
     @classmethod
     def deduplication_fields(cls):
         """list all the deduplication fields for a model"""
```

mouse-reeve: Yep, that does it. Thanks!

hughrun:
From #4004:
> I realized while testing this that we need to re-evaluate duplicated when one of them is edited.

So @mouse-reeve this would be somewhere in the edit view for each model, where we'd remove the item from any dupes list and then re-run `mark_merge_candidates`? Or maybe a new method to only check the object we just edited?

mouse-reeve:
I think it should happen at a model level (I started working on it at a view level but now that I think more, model seems better), and it could probably just re-check the duplicates and delete if they no longer match, rather than deleting and then re-checking. Re-checking has a lot of downsides (it's hard to do individually, and expensive to do in aggregate), so I don't want to do it that way

mouse-reeve:
@hughrun what's left here? Just handling multiple notes on/clean deduplication of suggestion list items?

hughrun:
Yes I think that's it. So grouping suggestion list items that are for the same combination of books after a SuggestionList merge (which I think might be possible just within the view rather than having to actually merge items?), and potentially allowing multiple notes in the first place (though that is probably out of scope since this is about merging).

So close!!!

Magess:
Hi, just wanted to know if as a user this means there's a way for me to merge/improve author records if I find duplicates?

hughrun:
@Magess we'll need to add some contextual help on this.

Broadly the answer is "no, but...". How this works is we run an automatic process that finds all records of the same type (Author, Edition, Series) that have a unique identifier in common (ISBN, wikidata ID, etc). Then admins will have the option to either automatically merge those (dangerous) or manually check and merge (slower but safer). Admins can give a particular user a permission level that allows them to do manual merges.

Where ordinary users come in is that you can:

1. block a merge if you can see that they definitely are NOT the same thing
2. edit the book/author/series to add metadata (ISBN, ISNI id, etc) that will match on the other duplicate.

We explored the option to allow users to indicate a merge, but decided (2) would be a better approach because it shows that they really are the same and gives us something to match on.

Magess:
Thanks! Knowing that the right move is to add the same identifier is useful.

PR #3914, opened 2026-05-22
Merge duplicated book data

mouse-reeve:
## Description

Automatically detect and merge duplicate book data objects (authors, editions, works, series). Most of the work is already in place, but this gets the merging going in the application. I'm approaching it as a background service which runs without intervention, but places a notice on the relevant page that a merge is pending (screenshot below). That means that users can't directly trigger or prevent merges, but they should be shown why a merge is pending and what deduplication field that based on, and change that field if needed.

Screenshot 2026-05-22 at 4 58 29 PM

expanded:
Screenshot 2026-05-22 at 4 59 41 PM

Subtasks/todos:
 - [ ] Gracefully migrate any legacy `MergedBook` entries (happily there aren't any on b.s, but it's possible another instance has some)
 - [x] Display pending deduplication status on object page
 - [ ] Handle potentially orphaned works
 - [ ] View and edit works
 - [ ] Admin view to schedule dedupes
 - [ ] Celery task to identify dupes
 - [ ] Celery task to merge dupes

- Related Issue #3379
- Closes #3490
- Closes #3493
- Closes #2898
- Closes #1119 
- Closes #562

## What type of Pull Request is this?

- [x] Bug Fix
- [x] Enhancement
- [ ] Plumbing / Internals / Dependencies
- [ ] Refactor

## Does this PR change settings or dependencies, or break something?

- [ ] This PR changes or adds default settings, configuration, or .env values
- [ ] This PR changes or adds dependencies
- [ ] This PR introduces other breaking changes

## Documentation

- [x] New or amended documentation will be required if this PR is merged
- [ ] I have created a matching pull request in the Documentation repository
- [ ] I intend to create a matching pull request in the Documentation repository after this PR is merged

### Tests

- [ ] My changes do not need new tests
- [x] All tests I have added are passing
- [ ] I have written tests but need help to make them pass
- [ ] I have not written tests and need help to write them

PR #3961, opened 2026-06-16
Automatically detect duplicate book data entries

mouse-reeve:
## Description
Allow admins to run a task that automatically merges duplicate book data

One schedule task will scan the database for duplicates and save the planned canonical version on each of the duplicates using the `pending_merge_target` field. Actually merging the task will happen separately, either automatically or manually.

The banner looks like this:
Screenshot 2026-06-16 at 6 48 07 PM

And when expanded:
Screenshot 2026-06-16 at 6 48 12 PM

- Related Issue #
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


mouse-reeve:
@hughrun I think the easiest way to start the process for a manual merge will be to start from the banner generated by the automated scan? So I'm reducing the scope of this PR to just identifying duplicates, not doing any automatic merging.

hughrun:
This looks pretty good.

I do think we need to also put a banner on the canonical record. It's not necessarily obvious to end users which is which. We already have users editing books to mark things as "duplicates" so it would be good to explicitly note that "yeah we know this has other copies that are the same", and perhaps note that this one is considered the "primary" record.

mouse-reeve:
Good call! I added a basic version of that:

Screenshot 2026-06-19 at 8 36 06 AM

hughrun (approved):
VERY COOL!

I think this is good to merge, though I'm a bit confused about how this relates to #3960 - I guess that would mean merge this into 3960 and then that one is ready as well?

As a small aside - in response to [this comment](https://github.com/bookwyrm-social/bookwyrm/pull/3914#issuecomment-4722770714) something people might like is the ability to mark things as duplicates manually. They probably need a different identifier (`manual_merge_candidate` or something) in the back end, but could be manually checked and merged by admins.

hughrun:
I just found something I hadn't previously considered. What happens if an item both *is* a duplicate and *has* its own duplicates?

Screenshot 2026-06-23 at 7 25 19 PM

I would guess if it is merged first the others will be merged over as candidates? Do we need to test that?

mouse-reeve:
Ah, I guess that will happen if two different identifiers match with two different other objects (like a shared ISBN with one book and a shared OpenLibrary ID with another). It should just merge sequentially -- whichever one is merged first will then be merged into the remaining canonical item. If you navigate to the item that was merged first, you will hit two redirects and end up on the correct page

mouse-reeve:
But yes, adding that to my mental list of things to test in #3967

PR #3967, opened 2026-06-17
Automatically merge duplicates

mouse-reeve:
## Description
This is the section of the work that allows admins to create a task to automatically merge duplicate data.
Remaining work:
 - [x] Tests for tasks
 - [x] Option to prevent automatic merge

The data quality view:
Screenshot 2026-06-18 at 11 08 30 AM

An author with duplicates scheduled to be merged:
Screenshot 2026-06-19 at 8 34 46 AM

And more detailed info about the merge on the pages of the duplicates:
Screenshot 2026-06-19 at 8 29 59 AM

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


mouse-reeve:
This isn't _done_ done but it's ready for review. It is a daisy chain of a pull request that merges into #3914 and #3961, both of which independently are ready for review.

Right now you can schedule scans for duplicate items, schedule merges of duplicate items, and manually prevent items from being merged. When a duplicate is detected and automatic merging is scheduled, it isn't merged until one week from detection to give people a chance to notice if something is amiss. It seems like this works!

I think the option to prevent merges needs more work, but I'd like feedback on what it needs. Who should be able to un-block a merge? Right now you can't un-block, which obviously isn't the final design.

@bookwyrm-social/code-review

bookwyrm/models/mergeable_mixin.py
hughrun, line 156: Why do we save `canonical` here?

hughrun, line 194: This will always create an orphaned `Work` when we delete its last Edition, which is likely to happen often. Could we add a check for this and delete the parent work at the same time? e.g. like in the EditBook view for Series:

https://github.com/bookwyrm-social/bookwyrm/blob/f061b6be6fb8d0fb3dcb23b29f47737b370f4f6b/bookwyrm/views/books/edit_book.py#L101-L108

bookwyrm/templates/settings/data.html
```diff
@@ -95,4 +233,36 @@ <h2 class="title is-4">{% trans "Schedule book data deduplication" %}</h2>
     </div>
 </div>
 
+<section class="block content">
+    <h2>{% trans "Editions with to associated work" %}</h2>
```
hughrun: "Editions with **no** associated work"

bookwyrm/templates/settings/data.html
```diff
@@ -95,4 +233,36 @@ <h2 class="title is-4">{% trans "Schedule book data deduplication" %}</h2>
...
+                <p class="header">{% trans "Broken editions" %}</p>
+                <p class="title is-5">{{ edition_orphan_count|intcomma }}</p>
+            </div>
+        </div>
+    </div>
+</section>
+
+<section class="block content">
+    <h2>{% trans "Orphaned data" %}</h2>
+    <aside class="help">{% trans "Works and authors with no associated books" %}</aside>
```
hughrun: Not sure if you deliberately didn't include Series/SeriesBooks here? Theoretically we shouldn't ever have an orphaned Series normally, but it might happen?

bookwyrm/models/mergeable_mixin.py
hughrun: This fails for `Series` because it doesn't have a `merged_model` attribute.

bookwyrm/templates/settings/data.html
```diff
@@ -95,4 +233,36 @@ <h2 class="title is-4">{% trans "Schedule book data deduplication" %}</h2>
...
+            <div class="notification">
+                <p class="header">{% trans "Broken editions" %}</p>
+                <p class="title is-5">{{ edition_orphan_count|intcomma }}</p>
+            </div>
+        </div>
+    </div>
+</section>
+
+<section class="block content">
+    <h2>{% trans "Orphaned data" %}</h2>
```
hughrun: Is the idea that in future we might auto-cleanup these?

hughrun (changes requested):

bookwyrm/models/mergeable_mixin.py
mouse-reeve: Great question 🤔 I moved this code over from here:https://github.com/bookwyrm-social/bookwyrm/pull/3960/changes#diff-ffdad197341d69624eed89fcfa140fd47de514c7b217b4ee6202e1e7cf1baf71L155 and it's part of the original code that I didn't write.

But I don't see how it does anything

mouse-reeve: That seems like a good call, but I think we will also need to scan for orphaned works, authors, and series.

bookwyrm/templates/settings/data.html
mouse-reeve: Yes, I got ahead of myself in this section, I think it should be future work

bookwyrm/templates/settings/data.html
mouse-reeve: Just an oversight, but I'll leave orphans out of this

mouse-reeve:
I just realized that we're going to run into some problems with suggestion lists. If two works both have suggestion lists and they are merged, I believe the merge will fail because `suggests_for` is a `OneToOneField` between the list and the work. The suggestion lists could be merged, but if both lists have the same work as a `SuggestionListItem`, one person's suggestion would have to be discarded in favor of another person's, since there is also a uniqueness requirement between the list and the suggested work.

The two solutions I can think of for the latter problem are to 1) make the notes field a separate table that can accept notes from multiple users, or 2) make the uniqueness constraint unenforced at the database level and allow duplicates to be created when two suggestions both of which have notes are added during a merge.

The first solution is more complicated, particularly from a UI perspective, but it has some side advantages, like allowing someone to write a note on a suggestion that was added without one. 

@hughrun

hughrun:
@mouse-reeve now that you've laid it out like that, I _absolutely_ think we should go for option 1. I'm not sure whether this is where you were going with it, but it would be great to allow multiple notes on each suggestion.

I hadn't thought about it like this earlier, but currently only the first person to suggest something can explain _why_ they suggested that. Anyone coming after them can merely endorse the suggestion. I think we could show the first note (whoever wrote it) on the book page, and every note on the suggestions page.

On a related point, it now occurs to me that notes on all list items (not just suggestion lists) probably should be able to be reported, moderated, and filtered out if you're blocking the user who wrote the note.

mouse-reeve:
Yes the more I think about it, the more this seems like a worthwhile change regardless of the deduplication implications. And agree on moderation! It definitely should be checking for blocked users at bare minimum..

hughrun:
@mouse-reeve I've been looking into auto-cleanup for related objects. Are you already working on that or are you happy for me to make a suggested PR to merge into this one?

Also happy to work on multiple notes per suggestion if you're not already started on that, since SuggestionList merges get a bit weird otherwise.

mouse-reeve:
Go for it -- I've only thought some thoughts about related objects, not done any actual work on it. The only meaningful thing I realized is that I think when an edition is merged into another edition, if they have different works, the works need to also be merged. Otherwise edition a, b, and c can end up associated with one work, and editions x, y, and z with a different work, and no way to join them.

I'll take a look at multiple notes, I've been holding off on that until #3988 was done, since it changes the suggestions UI quite a bit

mouse-reeve:
@hughrun I think, because we're moving the cleanup of related/orphaned objects to a separate PR, this PR is done?

hughrun:
Do we need some tests?

mouse-reeve:
Good call -- there isn't really that much here that is testable (mostly it's UI that's already checked), but I added some basic tests for the tasks

bookwyrm/views/mixins.py
```diff
@@ -25,3 +29,18 @@ def dispatch(self, request, *args, **kwargs):
...
+class MergeableViewMixin:
+    @method_decorator(login_required, name="dispatch")
+    @method_decorator(
+        permission_required("bookwyrm.edit_book", raise_exception=True), name="dispatch"
+    )
+    def post(self, request, book_id, **kwargs):
+        """Prevent objects from being merged"""
+        obj = get_object_or_404(
+            self.merge_model, id=book_id, pending_merge_target__isnull=False
+        )
```
hughrun: This throws an error on anything that isn't a work or edition because they don't have a `book_id`. I suspect there may be a more elegant way of doing this, but something like the below should work:


```suggestion
    def post(
        self,
        request,
        book_id=None,
        author_id=None,
        list_id=None,
        series_id=None,
        seriesbook_id=None,
        **kwargs,
    ):
        """Prevent objects from being merged"""
        
        id = book_id or author_id or list_id or series_id or seriesbook_id
        obj = get_object_or_404(
            self.merge_model, id=id, pending_merge_target__isnull=False
        )
```

hughrun (changes requested):
Looks good.

Couple of failing tests as previously discussed, and a change needed for the `post` function that stops auto-merging.

I was contemplating whether we need a way for admins to "undo" blocks on auto-merging, but since we now have manual merging as well, I think it's fine as-is.

bookwyrm/views/mixins.py
mouse-reeve: I took the other approach and changed the various object ids to `mergeable_object_id` -- I'm not 100% that it's a better solution, but I don't like enumerating all the types of objects

mouse-reeve:
So, two failing tests are because I used the same task form twice in the same page, which means the id fields are clashing. That was simple to fix with a [form prefix](https://docs.djangoproject.com/en/5.2/ref/forms/api/#django.forms.Form.prefix), but I'm stuck on how to fix the `aria-described-by` field, which [we add manually](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/admin.py#L218). I might need to dig into the django source code to understand how prefixes work, unless you can see something I'm missing?

ilkka-ollakka:
> So, two failing tests are because I used the same task form twice in the same page, which means the id fields are clashing. That was simple to fix with a [form prefix](https://docs.djangoproject.com/en/5.2/ref/forms/api/#django.forms.Form.prefix), but I'm stuck on how to fix the `aria-described-by` field, which [we add manually](https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/forms/admin.py#L218). I might need to dig into the django source code to understand how prefixes work, unless you can see something I'm missing?

For aria-described-by: I think it might be one of those things we could clean-up now that we are in django 5.2, as it should be automaticly handled? (https://docs.djangoproject.com/en/5.1/releases/5.0/#forms)

hughrun:
Oh sweet, good call @ilkka-ollakka! 

@mouse-reeve it appears that if we simply remove the widgets from `IntervalScheduleForm`it automatically gives `aria-describedby` the correct value in each form.

bookwyrm/views/mixins.py
hughrun: I like your solution.

mouse-reeve:
Glorious! Should be fixed now.

bookwyrm/models/mergeable_mixin.py
```diff
@@ -199,3 +227,15 @@ def absorb_data_from(self, other: Self, dry_run=False) -> Dict[str, Any]:
...
         return absorbed_fields
+
+    def merge_related_authors(self, canonical: Self) -> None:
+        """add authors from the candidate onto the canonical"""
+        if isinstance(self, apps.get_model("bookwyrm.Edition")) or isinstance(
+            self, apps.get_model("bookwyrm.Work")
+        ):
+            canonical.authors.add(*self.authors.all())
+
+    def merge_parent(self, canonical: Self, parent: Model) -> None:
```
hughrun:
```suggestion
    def merge_parent(self, canonical: Self, parent: models.Model) -> None:
```

bookwyrm/templates/snippets/merge_banner/work.html
```diff
@@ -0,0 +1,24 @@
+{% extends "snippets/merge_banner/layout.html" %}
+{% load i18n %}
+{% load utilities %}
+{% load humanize %}
+
+{% block content %}
+
+{% blocktrans trimmed with dupe_url=dupe.local_path dupe_title=dupe|book_title %}
+This edition's work been identified as a duplicate of <strong><a href="{{ dupe_url }}">{{ dupe_title }}</a></strong>.
```
hughrun: Sorry I should have picked this up earlier – if the `canonical` work happens to be orphaned, this url will 404 because there's no edition to show in the HTML view.

bookwyrm/models/mergeable_mixin.py
```diff
@@ -199,3 +227,15 @@ def absorb_data_from(self, other: Self, dry_run=False) -> Dict[str, Any]:
...
+        """don't leave childless works lying around"""
+        if parent and not parent.pending_merge_target:
+            parent.merge_into(canonical.parent_work)
```
hughrun, line 246: Oops.

```suggestion
        """don't leave childless works lying around"""
        if parent and not parent.pending_merge_target:
            try:
                parent.merge_into(canonical.parent_work)
            except ValueError:
                # can't merge parent into itself
                pass
```

bookwyrm/views/admin/data_quality.py
```diff
@@ -64,6 +64,26 @@ def schedule_deduplication_scan_task(request):
...
 
+@require_POST
+@permission_required("bookwyrm.edit_instance_settings", raise_exception=True)
+def schedule_deduplication_task(request):
+    """scheduler"""
+    form = forms.IntervalScheduleForm(request.POST)
+    if not form.is_valid():
+        data = data_quality_data()
+        data["merge_form"] = form
+        return TemplateResponse(request, "settings/data.html", data)
```
hughrun:
```suggestion
        return TemplateResponse(request, "settings/manage-data/data.html", data)
```

bookwyrm/views/admin/data_quality.py
hughrun:
```suggestion
form = forms.IntervalScheduleForm(request.POST, prefix="scan")
if not form.is_valid():
data = data_quality_data()
data["scan_form"] = form
return TemplateResponse(request, "settings/manage-data/data.html", data)
```

bookwyrm/views/admin/data_quality.py
hughrun:
```suggestion
    form = forms.IntervalScheduleForm(request.POST, prefix="merge")
    if not form.is_valid():
        data = data_quality_data()
        data["merge_form"] = form
        return TemplateResponse(request, "settings/manage-data/data.html", data)
```

hughrun (changes requested):
Sorry I missed a few things earlier.

bookwyrm/models/mergeable_mixin.py
mouse-reeve: Wouldn't it make more sense to check if the parent is the same, rather than catching ValueError, which could theoretically have a separate cause?

bookwyrm/templates/snippets/merge_banner/work.html
mouse-reeve: What's the right behavior here? It could not link, which might be confusing to the casual user, but maybe it could be clarified in the text? Or should there be a stub page for orphaned works? Should the banner just not be shown?

bookwyrm/models/mergeable_mixin.py
hughrun: yes!

bookwyrm/templates/snippets/merge_banner/work.html
hughrun: hmm, tricky. The reason to show the banner with the link is so that a user can check both records manually, determine whether they're really duplicates, and block the merge if they're not, right? In this case there's no way to do that at all, but even when we have two works with editions, it's hard to tell what's at the work level when looking at what is actually an Edition.

If we have a link, then the canonical record lists the identifiers that match, and allows the user to check anything else. If we don't have a canonical to show them, then what would be most useful is to know what _doesn't_ match. So I think there are three options:

1. **Minimalist**: just show the title instead of a link if canonical is orphaned - "This edition's work has been identified as a duplicate of _Little house on the prairie_ (orphaned work)"
2. **Flipped**: use the "canonical" banner on candidates if the canonical work doesn't have any editions
3. **Special comparison page**: throw to a special page where users can see all the fields on each work - kind of like the manual merge page but only for one candidate and without the power to select fields or trigger the merge, but with the option to stop auto-merge

I think I probably favour option 1 because it's simplest and if the title matches and so does some identifier then it's pretty likely to be a real match - but at least if the title _doesn't_ match a user can determine whether it's just a translated title or it's a mistake. Option 2 might be helpful but also potentially confusing because it would work the opposite to everything else.

mouse-reeve: I think option 1 is sufficient -- works are pretty obscure in the UI and there isn't user-created content that we need to prioritize preserving for them

hughrun (approved):
🎉

PR #3982, opened 2026-06-22
Allow manual merging of mergeable objects

hughrun:
## Description

Initial work to allow instance admins to manually merge a set of candidates, picking fields to use from any merge candidate. This currently isn't really connected to anything but would logically be linked from the candidates listed on the data-quality landing page once the other related PRs are merged.

If a `merge_candidate` has the only instance of a field with a value, that value will be selected by default. Otherwise the field from `canonical` will be selected. `ArrayField`s have every value selected by default, but they can be de-selected. Only fields that would could be changed from the canonical values will be listed.

Select values:

Screenshot 2026-06-22 at 7 41 44 PM

Screenshot 2026-06-22 at 7 42 18 PM

Canonical values are deselected if you select a field from another item:

Screenshot 2026-06-22 at 7 42 23 PM

View a diff of changes before confirming:

Screenshot 2026-06-22 at 7 42 31 PM

I am very interested in any suggestions or reviews! @bookwyrm-social/code-review 

Related PRs: #3960 #3961 #3967 

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
I think maybe the tabs are not obvious enough? Do they need to be labelled differently?

mouse-reeve:
This is very cool!

My first thought is that I don't think this should be part of the admin view. I'm not convinced it should only be available to admins, but even if it is, I think it makes more sense to be part of the items own views, rather than the admin ones. 

That's highlighting that we need to figure out what permission level of user should be able to do manual merges (and if this should be customizable in the instance settings), and how it should be possible to pair items (ie, it's possible to merge any two items by entering their ids, as here, versus only making it possible to merge items that have been identified by the system). I could see a lot of potential for errors in allowing merging of arbitrary items, especially if it was available to all users with editor permissions. I could also see users being frustrated if they were unable to perform merges, and admins unable to keep up with the merges that need to take place.

I think it would be easier to use if it's a two-column view rather than the tabs. I realize that's putting a lot of information on one page, but I find it hard to do a comparison if I can't see both items.

mouse-reeve:
It looks like array fields are shown in the options even if they are identical, and will get added twice if they aren't un-checked on one of the versions

hughrun:
I think I agree in terms of not restricting this to editors.

If I understand the Django permissions system correctly, it's possible to assign custom permissions directly rather than having to bundle them into a user category. In this case I think that might be the way to go, though we would also presumably want to add merge permissions to the admin user type. So a permission like `can_merge`, and then maybe an admin screen where admins can see who has merge permissions, can add and remove them, and possibly see how many merges/what has been merged by each user if we want to allow them to monitor merging behaviour. That would allow _any_ user to be given merge permission if an admin trusts them, whilst not forcing them to give merge permissions to everyone in a given category.

I think I asked about this earlier, but my feeling at this stage is that we should start pretty conservative and only allow merging of objects that are in a `canonical/merge_candidate` relationship. The easiest way to get around that would be to add a matching value in a deduplication field and then it will create the relationship, so it's not a particularly onerous restriction for genuine duplicates.

Regarding the two-column view, that was my original idea. But what I realised is that there is an unlimited number of merge candidates and we don't want to make people merge them two at a time - hence the tabs. This proposed code adds a tab for every merge_candidate, not just comparing two arbitrary items. Here's an example where there are two merge candidates and a canonical:

Screenshot 2026-06-23 at 7 28 53 PM

I think this works _ok_ with the diff for double checking and the ability to go back and adjust it if it's not right. Having said that, I'm not 100% convinced on the _styling_ of the tabs because I don't think it's obvious enough how to interact with them. But I'm open to trying something else if we can get the right balance between clarity and enabling reasonably safe bulk merges.

hughrun:
This might be better, but there's some custom styling overriding the text colour in the selected tab:
Screenshot 2026-06-23 at 8 05 47 PM

mouse-reeve:
What about keeping the canonical pinned on the left-hand column and tabbing through the options in the right-hand column?

mouse-reeve:
Another option might be to aggregate the options for a field from all the duplicates per field. Like if the canonical has the `description` field set, and three dupes, show a list of radio buttons on the right column, one for each of the `description`s provided by the dupes?

hughrun:
I like both of these options but I think "aggregate the options for a field from all the duplicates per field" is probably the way to go. I'll have a play with it. @mouse-reeve would you prefer I change the merge target to `mergable-mixin` or merge in your changes from [mergable-mixin](https://github.com/bookwyrm-social/bookwyrm/tree/mergable-mixin) and point this PR at `main`?

mouse-reeve:
Let’s keep it on mergable-mixin, I feel like we’re close enough on these that it’s better to wait and merge it all together

mouse-reeve:
I didn’t realize merging and deleting the other branch would close this! My bad

PR #3985, opened 2026-06-23
docker-compose: reference by default to builded containers

### LINKED ISSUE #2095: building / publishing generic Docker images

ilkka-ollakka:
## Description

This change allows to develop/run bookwyrm without needing to build all the containers. Building of container is needed only for dev-tools and if you have changes dependencies.

Build is needed only if dependencies are changed when developing, otherwise bookwyrm:main is  good base as it is latest build of main branch.

By default reference to `:latest` containers, which are latest tagged builds, meaning latest releases. This should be sensible default and update to newest version just requires pulling the image.

If people have automatic pull enabled, they might want to pin to specific version instead of latests, to prevent unannounced updates on restart of container.

This might have some rough edges that I didn't anticipate, so please test and comment how it feels and works.

- Related Issue #
- Closes #2095

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

Container images points to bookwyrm container builds instead requiring local builds. Building of release containers is not supported out of the box locally anymore.

## Documentation

- [X] I intend to create a matching pull request in the Documentation repository after this PR is merged

### Tests

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -65,7 +65,8 @@ services:
       - ./certbot/logs:/var/log/letsencrypt
       - ./certbot/data:/var/www/certbot
   db-backup-job:
-    image: ghcr.io/bookwyrm-social/postgres-docker:latest
+    image: ghcr.io/bookwyrm-social/postgres-backup:latest
+    build: postgres-docker
     environment:
       - DEBUG=${DEBUG}
       - PGPASSWORD=${POSTGRES_PASSWORD}
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -98,6 +98,7 @@ services:
   web:
     restart: unless-stopped
     image: ghcr.io/bookwyrm-social/bookwyrm:latest
+    build: .
     env_file: .env
     command: gunicorn bookwyrm.wsgi:application
     volumes:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,4 @@
-x-logging:
-  &default-logging
+x-logging: &default-logging
   driver: "json-file"
   options:
     max-size: "150m"
@@ -66,7 +65,7 @@ services:
       - ./certbot/logs:/var/log/letsencrypt
       - ./certbot/data:/var/www/certbot
   db-backup-job:
-    build: postgres-docker
+    image: ghcr.io/bookwyrm-social/postgres-docker:latest
     environment:
       - DEBUG=${DEBUG}
       - PGPASSWORD=${POSTGRES_PASSWORD}
@@ -96,8 +95,8 @@ services:
     networks:
       - main
   web:
-    build: .
     restart: unless-stopped
+    image: ghcr.io/bookwyrm-social/bookwyrm:latest
     env_file: .env
     command: gunicorn bookwyrm.wsgi:application
     volumes:
@@ -140,8 +139,8 @@ services:
       - main
     restart: unless-stopped
   celery_worker:
+    image: ghcr.io/bookwyrm-social/bookwyrm:latest
     env_file: .env
-    build: .
     networks:
       - main
     command: celery -A celerywyrm worker --pool=threads --concurrency=20 -l info -Q high_priority,medium_priority,low_priority,streams,images,suggested_users,email,connectors,lists,inbox,imports,import_triggered,broadcast,misc
@@ -164,8 +163,8 @@ services:
         condition: service_started
     restart: unless-stopped
   celery_beat:
+    image: ghcr.io/bookwyrm-social/bookwyrm:latest
     env_file: .env
-    build: .
     networks:
       - main
     command: celery -A celerywyrm beat -l INFO --scheduler django_celery_beat.schedulers:DatabaseScheduler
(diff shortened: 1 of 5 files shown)


ilkka-ollakka:
I'm not sure if the supporting local build should be kept for now, to make transition easier. It should be trivial to add similar `build: .` to web container in docker-compose.yml than docker-compose.dev.yml has.

mouse-reeve:
I didn't run into any issues with this, but I'd like to get some other folks to test and think about edge cases, etc @bookwyrm-social/code-review

ilkka-ollakka:
I think it also would make sense to start building dev-container, so developer could just use most recently pulled containers, without needing to keep builds up to date, unless there is some dependency changes done.

hughrun:
I'll take a proper look at this soon.

I would expect we need to keep the ability to do local builds regardless. I'm a bit unclear about dev containers because the point of dev is we're changing the code? Or would there be a different docker config for dev with a shared volume for bookwyrm files?

Re prod containers, the entire production workflow should change. In theory there is no need to use `git` or the repository at all: you copy a compose file, customise an `.env` file, and pull the containers. But for people transitioning from the current arrangement to the new one we're going to need a step by step guide. Naturally we still need to support dockerless - the main change there will be pulling a tagged release instead of pulling the HEAD of `production`.

As a separate concern, we probably need to reconsider how `./bw-dev` is expected to be used, whether a bash script is still the right approach, and whether it might make sense to split it into prod and dev (maybe `./bw` and `./bw-dev`, for example).

In any case, this will require clear documentation and release notes.

ilkka-ollakka:
> I'll take a proper look at this soon.
> 
> I would expect we need to keep the ability to do local builds regardless. I'm a bit unclear about dev containers because the point of dev is we're changing the code? Or would there be a different docker config for dev with a shared volume for bookwyrm files?

Local builds are mostly needed if dependencies are changed, otherwise in dev-flow current code is mounted inside the container. So code-change doesn't require container re-building currently.

But you are right, that it is feasible to keep build possibility so it doesn't break current flow. I'll add that back.

> Re prod containers, the entire production workflow should change. In theory there is no need to use `git` or the repository at all: you copy a compose file, customise an `.env` file, and pull the containers. But for people transitioning from the current arrangement to the new one we're going to need a step by step guide. Naturally we still need to support dockerless - the main change there will be pulling a tagged release instead of pulling the HEAD of `production`.

We could have separate repo with just docker-compose and nginx/anubis config-files. 


> In any case, this will require clear documentation and release notes.

docker-compose.yml
hughrun, line 56: If there is no file at `./anubis/botPolicy.yaml` on the host machine this will create a **directory** at that path. Even if the plan is to put this in a repository with al the required config files I think this should be changed explicitly to a bind:

```yaml
volumes:
  - type: bind
    source: "./anubis/botPolicy.yaml"
    target: "/data/cfg/botPolicy.yaml"
    read_only: true
```

docker-compose.yml
```diff
@@ -66,7 +65,7 @@ services:
       - ./certbot/logs:/var/log/letsencrypt
       - ./certbot/data:/var/www/certbot
   db-backup-job:
-    build: postgres-docker
+    image: ghcr.io/bookwyrm-social/postgres-docker:latest
```
hughrun: This is failing with a permission error for me.

hughrun (commented):
Ok I think I get it now! This is cool.

I tried running this with just a docker-compose and .env file and it looks like it won't need too much more - the anubix config file, optionally `bw-dev`, and it seems one or two nginx configs.

I don't normally run in production so I'm not sure if I just messed it up, but I could only load the nginx default page, not an actual BookWyrm page, though it seemed like everything was running.

ilkka-ollakka:
> Ok I think I get it now! This is cool.
> 
> I tried running this with just a docker-compose and .env file and it looks like it won't need too much more - the anubix config file, optionally `bw-dev`, and it seems one or two nginx configs.
> 
> I don't normally run in production so I'm not sure if I just messed it up, but I could only load the nginx default page, not an actual BookWyrm page, though it seemed like everything was running.

you need most of the nginx-directory files, so better to copy whole nginx-directory so it should work. Those nginx configs are still littlebit convoluted and could do some simplifications.

docker-compose.yml
ilkka-ollakka: I had typo in image name here, fixed.

ilkka-ollakka:
I added workflow to also build dev-tools, as it is easy to forged and miss. It works only after merge and idea is that it is build after bookwyrm:main is build so it uses that as reference.

Also restored the `build: .` lines to docker-compose, to keep better backward compatibility.

PR #3988, opened 2026-06-24
Re-arranges book view

mouse-reeve:
## Description
I quietly launched the suggestions feature on bookwyrm.social and it was immediately obvious that it was completely hidden and almost impossible to find underneath scrolling upon scrolling through reviews. That got me thinking about the information hierarchy and how we're using space in that view, and I tried this to rearrange it all.

Here's the tldr on what I changes:
 - Hid book edition info behind a details panel
 - Hid subjects list behind a details panel
 - Grouped reading activity into tiles
 - Moved suggestions up above review
 - Hid the create review form behind a details panel
 - Tweaked the wording and spacing to be a little tighter

I'm not at all convinced that this is the ideal fix, and I'd love suggestions, but hopefully it's a step in the right direction. I'd like to include more socially relevant information in this view in the future (like which of the people you follow have this book on their shelves), but for now I just wanted to move things around.

Here's a couple examples -
Before:
Screenshot 2026-06-24 at 4 16 32 PM
After:
Screenshot 2026-06-24 at 4 15 38 PM

Before:
Screenshot 2026-06-24 at 4 17 46 PM

And after:
Screenshot 2026-06-24 at 4 18 17 PM

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests


nasamuffin:
I like the suggestions becoming visible - you're right, I didn't know that feature even existed before now. And I do tend to trust bookwyrm.social opinions pretty highly so I think it's a feature I'd use.

However, for what it's worth, having the "also recommended" interceding between the book summary and the reviews for the book that you're looking at the page for makes it feel like an ad, which is unfortunate. Like, "Hey, you're here looking at Goodnight Moon, and we'd love to tell you more about Goodnight Moon, but first, have you heard of Cat in the Hat? We think you'd really like The Mitten." It feels distracting from the purpose of the actual page you're looking at.

For me, I think I'd rather see them on the sidebar under the Lists / Get a Copy sections, which means I don't have to scroll further to get to the reviews for the book I'm actually trying to look at, BUT I have tantalizing little distractions on the sidebar.

You didn't post screenshots with the mockup on mobile screen size, but when I have a look at the current book page on mobile, I think that would put suggestions way at the bottom, which to me is fine - since I think reviews are paginated anyway, I'd still run across them pretty quickly.

Just my 2c - thanks for your work on it, I really like Bookwyrm :)

 - nasamuffin at bookwyrm dot social

nerdteacher:
I really like the 'Your Reading' being made more compact rather than taking up a good chunk of the page (especially if it is something you've read multiple times). Though, I like the edition details being presented as they are in the original. I'm not a fan of having to click through, click through, click through (even if one of those clicks is collapse-able... it makes things feel tedious).

But I also agree that the suggestions feel intrusive and like an advertisement when positioned like this. Something smaller and off to the side would be better, especially as I'm not visiting the book page primarily for suggestions; I'm visiting it primarily for that book (reviewing comments and quotes I saved, looking at other reviews, getting information on it).

So I agree with the suggestions of @nasamuffin.

mouse-reeve:
From @nimloth@todon.eu on Mastodon:

> Many changes I don't really mind (and some I like) except the "suggestions" block being above the reviews block.
> 2 reasons:
> 1) When we visit the page of a book, I guess we want to know more about the book, including (and sometimes especially) reviews. And in a second time, (maybe) get recommendations or if our Follows have read this book.
> 2) Having the "Readers also recommend" block that present feels very much like an ad (not great on Bookwyrm). Or at least like an algorithm (which is ok but you always wonder how the algorithm works in that case). 
> 
> However, I would be ok knowing at a glance which of my Follows have read or noted the book (or having that behind a details panel).
> 
> I like that things are hidden beneath a details panel: especially "write a review" block (because most people are not on the page to write a review).
> Except maybe book edition because I think it's important information for the readers.

fofwisdom:
When I read a quote on fediverse, I get interested in the book and click on the link, but I rarely come across any reviews—since people rarely write them—and I’m always disappointed because I can’t find the quotes I was hoping to see.

mouse-reeve:
> When I read a quote on fediverse, I get interested in the book and click on the link, but I rarely come across any reviews—since people rarely write them—and I’m always disappointed because I can’t find the quotes I was hoping to see.

Oh yes! You've hit on two issues that are out of scope for this PR -- we aren't very good about pulling in reviews from around the fediverse on books, so there aren't as many reviews as there could be, and we also don't display content that isn't reviews. Both of which I'd like to change.

5moufl:
One addition to the other remarks: I feel like the reading activity cards being displayed in a grid is weird (vs something more linear, a list or a timeline)

fofwisdom:
Traditionally, bookstores have promoted books by making cards with comments or quotes written on their covers, and even in the digital age, online bookstores offer services that add captions to images. Sharing comments and quotes and providing easy access to them is a natural practice for everyone.

So I like the “Mouse Says” card under the “Readers Also Recommended” section. The “book description” isn’t a catchphrase, so the first sentence can sometimes come across as dry.

mouse-reeve:
Thank you, this is such useful feedback. I want the UI to emphasize that recommendations are directly from users and not algorithmic, and now that yall mention it I can see how disruptive it is across the middle of the screen. 

It's a little tricky to get the suggestion box narrow enough to work as a sidebar (maybe there's a smart way to do this I overlooked), but here's a mock up:
Screenshot 2026-06-25 at 9 51 35 AM


Once I get further I'll make sure to post mobile examples as well.

mouse-reeve:
> One addition to the other remarks: I feel like the reading activity cards being displayed in a grid is weird (vs something more linear, a list or a timeline)

Legit! It needs to take up less space than the current UI, but I'm also not convinced that this is the fix

viviicat:
For the reading activity changes, maybe it makes sense to have the existing card for the most recent (or only) update, and when there are multiple entries, a sort of colored timeline below. It would be a bar whose start and end corresponds to the first and last events, and is filled in with a color to indicate a rough timespan that reading took place. And it would be an expandable section that would open to the existing vertical list of components for editing/viewing in more detail.

I am thinking something not dissimilar to the "languages" bar used by Github, except one color (interspersed with transparent blocks) and without requiring a label.

nerdteacher:
Could the reading recommendations maybe be... a collapse-able? Perhaps that'd help with the space.

I'm still not a fan of the edition info being put behind one (because that's the actual thing I tend to view book pages for... other than reviews or my comments/quotes), but I definitely don't feel like suggestions should be placed in a person's face with such prominence unless they actively want to engage with them.

mouse-reeve:
@nerdteacher I'm going to work on an alternative to the details panel I'm using exposes some of the info and just abbreviates it if necessary -- I also don't like totally hiding the edition details

mouse-reeve:
Alright, I made a modified version of the `details` panel that shows some initial content in its closed state. I think this makes it a lot more usable, while still keeping the size in check. Do you feel like this is a reasonable compromise, @nerdteacher (and anyone else with feelings about this)? I'd also be curious about what you would want visible by default in the book details section (I personally think ISBN should be added in there? but it would be great to get outside input).

closed state:
Screenshot 2026-07-08 at 10 38 40 AM

expanded state:
Screenshot 2026-07-08 at 10 38 30 AM

nerdteacher:
Ah! This is lovely, @mouse-reeve!

mouse-reeve:
Okay, I got the suggestion list flow working, reviewed the layout with a screen reader, and checked that it still validates with schema.org. I'm not sure why curlylint was unhappy, and we validate the html in the tests, so I just deleted it. This is ready for 👀 @bookwyrm-social/code-review

mouse-reeve:
Here's a screenshot with more content:
Screenshot 2026-07-17 at 4 41 14 PM

prolibre:
Hello @mouse-reeve 

Personally, as I edit and proofread the data for quite a few books, I’d prefer it if, at the very least, the following were displayed: number of pages, language, publication date and ISBN.

I’m making a comment here, but perhaps this deserves a ticket (?): on many recent books, only ‘First published date:’ is mentioned, whilst ‘Published date:’ is left blank… so the date doesn’t appear. It would be useful to display the former if the latter is blank.
So that’s my modest contribution.

PatrickChildersIT:
I personally liked the suggestions being above reviews and scrolling to the right like in the first mockup.
I agree though that it did feel a bit like an advertisement, but I think suggestions will always feel at least a little like an advertisement if the user doesn't choose to look at them.
What about moving it back to where it was, but putting it in a collapsed section, as a way to _opt-in_ to that "advertisement"?

Reading activity can also take up a lot of space like in the first mockup, where there are a lot of entries.

If there's a long list of items i have to scroll down through it is annoying if I already know what the items are, like my own purchase history on a shopping site. Reviews can be endlessly scrollable because I don't know what each next review will say, this work son social media too like Lemmy or Mastodon.

Maybe collapse reading history by default, or provide an entirely different way to look at it. That way I down have to scroll down a time or two just to get to reviews.

Also back to the suggestions, if they stay a long list on the right sidebar as they are and i want to go from the last suggestion to the first review (or the other way) I have to scroll mostly all the way back up. If that's necessary maybe at least provide a way to get the user quickly back to the top.

mouse-reeve:
Reading activities are collapsed by default, just showing the first entry. And right now it only shows three suggestions. I'm trying to make layout of the suggestions emphasize that they are user-provided content, not algorithmic recommendations or ads, by showing the user that suggested it and their thoughts. The right-side works a LOT better than the original version -- the whole flow of the page makes a lot more sense, and the content fits better. If it is enough of a problem to have them visible at all, I'm happy to add an option to hide them, but I think it's a really fun feature and I'm hoping people will give it a shot.

bookwyrm/templates/components/details.html
```diff
@@ -0,0 +1,14 @@
+{% load i18n %}
+<div class="preview-details">
```
hughrun: The "Add read dates" button is flush against the bottom of the reading block, which I think doesn't look great. This would create a little bit of space:
```suggestion
<div class="preview-details mb-2">
```

Image

bookwyrm/templates/book/sections/reviews.html
```diff
@@ -0,0 +1,73 @@
...
+        {% endwith %}
+    </div>
+</details>
+{% endif %}
+<div class="block" id="reviews">
+    {% if request.user.is_authenticated %}
+    {% if user_statuses.review_count or user_statuses.comment_count or user_statuses.quotation_count %}
+    <nav class="tabs">
+        <ul>
+            {% url 'book' book.id book.name|slugify as tab_url %}
```
hughrun:
```suggestion
            {% url 'book' book.id book.title|slugify as tab_url %}
```

hughrun (changes requested):
This absolutely rules.

Couple of little fixes, then I think it's good to go. 

I have some thoughts on potential enhancements but they are out of scope for now.

hughrun:
FYI @prolibre : 

Default (closed/minimised):
Screenshot 2026-07-19 at 3 13 13 PM

With blocks expanded:
Screenshot 2026-07-19 at 3 13 27 PM

Published date uses `date=book.published_date|default:book.first_published_date`

So I think this should already account for all your requests!

hughrun (approved):

prolibre:
Hello @mouse & @hughrun 

I’m just adding a small note here regarding the new layout. For the French translation, I’ve used: "_Postez un avis, un commentaire ou une citation_" (Post a review, a comment or a quote) for the original text “Add your thoughts” (bookwyrm/templates/book/sections/reviews.html:8) because the original isn’t clear enough about the available options :-) 
Perhaps this needs clarifying.

sudo-nano:
These look great, and I'm a fan of the more compact "Your Reading Activity" section. Just one thing: the reading direction (left to right or top to bottom) of the activity items grid seems a touch ambiguous to me. I had to examine the dates to figure out that left to right was the intended reading order. Perhaps putting light divider lines between the rows would make it immediately obvious without being obtrusive.

mouse-reeve:
@sudo-nano That's a good idea! Can you open a separate issue for that? I'm going to forget all about it if it's in the comments on a closed PR

PR #3993, opened 2026-06-25
postgres backup to use pg_dump custom format to reduce size and speed up restore

ilkka-ollakka:
## Description

Use custom format of pg_dump for backups, those are by default compressed. Allows less storage use for backups and more parallel restore.

example restore command with backup created with this flow is:
```
pg_restore --if-exists --clean -j <ncpu-count> --dbname bookwyrm /backups/backup_bookwyrm_2026-06-25.sql
```

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

backup-mode changed to postgresql compressed custom instead of plain sql-file

## Documentation

### Tests

- [X] I have not written tests and need help to write them

I tried the tests written for weed, but I didn't manage to make them pass even without any changes made.

### DIFF
--- a/postgres-docker/backup.sh
+++ b/postgres-docker/backup.sh
@@ -10,5 +10,5 @@ fi
 BACKUP_DB=${PGDATABASE:-bookwyrm}
 BACKUP_USER=${PGUSER:-bookwyrm}
 filename=backup_${BACKUP_DB}_$(date +%F)
-pg_dump -U "${BACKUP_USER}" "${BACKUP_DB}" --file "/backups/$filename.sql"
-info "backup: completed backup of $BACKUP_DB to /backups/$filename.sql"
+pg_dump -U "${BACKUP_USER}" "${BACKUP_DB}" --format=custom --file "/backups/$filename.sql"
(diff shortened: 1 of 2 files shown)


ilkka-ollakka:
For comparision, using compress=zstd in current flow drops the backup size from 745M->229M. Using directory-mode does have small overhead (in test instance backup size from 14M .sql file to 15M directory), but can do backup and restore selective way (only specific tables) and in parallel.

hughrun:
@bookwyrm-social/code-review I don't feel qualified to review this, can someone take a look?

ilkka-ollakka:
Changed the PR to use custom format of pg_dump, makes it easier to move backups and still able to restore in parallel.

MattSmiglarski:
Note that the `pg_restore` command which is currently in the description gives an error (assuming `nproc` is greater than 1):

`pg_restore: error: cannot specify both --single-transaction and multiple jobs`

Suggest updating the description to keep the `-j` and drop the `--single-transaction`, rather than the other way around?

PR #4002, opened 2026-06-27
Fix garbled avatar object

### LINKED ISSUE #3986: Invalid `name` attribute for account avatars
**Describe the bug**

We have discovered that since yesterday, some bookwyrm.social accounts fail to correctly process in Mastodon.

This seems to be caused by incorrect serialization of the `name` attribute of an actor's `icon` `Image` object.

Indeed, `https://bookwyrm.social/user/todrobbins` for instance has the following object as its `icon` attribute:
```json
{
  "type": "Image",
  "url": "https://bookwyrm-social.sfo3.digitaloceanspaces.com/images/avatars/8cf12bc0-9c0a-4975-af96-f3045b458f5e.png",
  "name": {
    "_args": [
      "avatar for todrobbins"
    ],
    "_kw": {}
  },
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    {
      "Hashtag": "as:Hashtag"
    }
  ]
}
```

The way the `@context` is recalled is weird, especially since it includes something it doesn't use, but it's valid.

However, the value of `name` is clearly buggy. It should be a `xsd:string` or `rdf:langString`.

hughrun:
## Description

Well this was an adventure.

Per [Django docs](https://docs.djangoproject.com/en/6.0/topics/i18n/translation/#formatting-strings-format-lazy):

> Python f-strings cannot be used directly with gettext functions because f-string expressions are evaluated before they reach gettext.

Using `or` statements inside gettext also seems to be the likely cause of the weird return value, though I can't pin down a definitive explanation.

Should resolve #3986

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/models/user.py
+++ b/bookwyrm/models/user.py
@@ -229,7 +229,7 @@ class User(OrderedCollectionPageMixin, AbstractUser):
     @property
     def alt_text(self):
         """alt text with username"""
-        return _(f"avatar for {self.localname or self.username}")
+        return _("avatar for {name}").format(name=(self.localname or self.username))
 
     @property
     def display_name(self):
--- a/bookwyrm/tests/models/test_user_model.py
+++ b/bookwyrm/tests/models/test_user_model.py
@@ -1,4 +1,5 @@
 """testing models"""
+
 import pathlib
 
 from unittest.mock import patch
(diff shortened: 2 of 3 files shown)


ClearlyClaire:
Thanks for the quick fix!
I'm a bit surprised at the feature though, what's the benefit of providing such generic alt-text? Especially since it can't be ensured it would be translated to the viewer's language.

hughrun:
@ClearlyClaire this function is shared across multiple templates and views so it's primarily intended for the BookWyrm interface where it's lazily interpreted. So yes it makes less sense in the context of activity json but saves us having a special method just for that.

PR #4016, opened 2026-07-06
Don't ignore boosts from external services

### LINKED ISSUE #4011: Boosting a local status from a remote server fails to create a `Boost` record
**Describe the bug**
Recently I got reports that there are problems with the communication between remote instances (Friendica, Mastodon) and BookWyrm. So I wrote a book review (https://bookwyrm.de/user/qbi/review/30922) and retootet it from my Mastodon account https://freie-re.de/@qbi. I could not find the retoot in my BookWyrm instance. So I started to dig.

As far as I see it this is caused by two separate issues.

1. [`Status.ignore_activity()`](https://github.com/bookwyrm-social/bookwyrm/blob/7472f6e3f9599447194ce5f9538a659073b6ef60/bookwyrm/models/status.py#L129) calls [`resolve_remote_id()`](https://github.com/bookwyrm-social/bookwyrm/blob/7472f6e3f9599447194ce5f9538a659073b6ef60/bookwyrm/models/status.py#L132) without passing `model=`. So no database lookup is done.
The code path https://github.com/bookwyrm-social/bookwyrm/blob/7472f6e3f9599447194ce5f9538a659073b6ef60/bookwyrm/activitypub/base_activity.py#L356 is never entered. This means that every call from `ignore_activity()` skips the the local-lookup shortcut, even when the status exists.
Passing `model=Status` could resolve this issue.
2. The function `get_activitypub_data()` falls back to the default user-agent from `requests` which is `python-requests/a.b.c`. So `is_bookwyrm_request()` doesn't recognize that it comes from a BookWyrm instance. The server returns `type: "Article"` instead of `type: "Review"` and `get_model_from_type("Article")` raises an `ActivitySerializerError`.
The issue in 1. leads to an re-fetch, but `get_activitypub_data()` never sets a user-agent which looks like a BookWyrm user-agent (just the Python request default one). The function `is_bookwyrm_request()` returns `False`because https://github.com/bookwyrm-social/bookwyrm/blob/7472f6e3f9599447194ce5f9538a659073b6ef60/bookwyrm/views/helpers.py#L65 
So in https://github.com/bookwyrm-social/bookwyrm/blob/7472f6e3f9599447194ce5f9538a659073b6ef60/bookwyrm/views/feed.py#L148 it is always `pure=True`.
It would help to set `"User-Agent": settings.USER_AGENT` in `get_activitypub_data()`.

The remote server sees no error. The `inbox` request sends a 200/202 status.

**To Reproduce**
1. Use a BookWyrm review with a known `remote_id`. In my case: https://bookwyrm.de/user/qbi/review/30922
2. Send a Announce activity from a remote ActivityPub actor which already exists in the local database. The activity should target the `remote_id` from the status as `object`.
3. The inbox returns a 200 (or 202) status code, but no boost row is created.

When I try the following within the Django console I get a failure from `result.status`:

```python
from bookwyrm.views.inbox import activity_task

test_json = {
    "id": "https://freie-re.de/users/qbi/statuses/116855536350503086/activity",
    "type": "Announce",
    "actor": "https://freie-re.de/users/qbi",
    "published": "2026-07-03T10:19:39Z",
    "to": ["https://www.w3.org/ns/activitystreams#Public"],
    "cc": [
        "https://bookwyrm.de/user/qbi",
        "https://freie-re.de/users/qbi/followers",
    ],
    "object": "https://bookwyrm.de/user/qbi/review/30922",
}

result = activity_task.apply_async(args=(test_json,))
```
Furthermore a traceback is thrown.

> bookwyrm.activitypub.base_activity.ActivitySerializerError: No model found for activity type "Article"

**Expected behavior**
See a retoot in the web interface or find it in the database.

**Instance**
https://bookwyrm.de/ runs on 0.9.0 (91ea26db16fdf3368d762359436b3ac8860f3ba0) with a self-hosted dockerless setup. It uses Python 3.11.2.

**Additional context**
I saw the issue #3994 from @mouse-reeve which looks similar. But my impression is that this is a separate issue.

mouse-reeve:
## Description
Per @qbi's suggestion, this adds the model to the lookup for whether to ignore a boost, which allows it to complete the lookup without making an HTTP request. It also sets the user agent so that Reviews are detected as type `Review`, not type `Article`

- Related Issue #
- Closes #4011 

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -16,7 +16,7 @@ from bookwyrm import models
 from bookwyrm.connectors import ConnectorException, get_data
 from bookwyrm.models import base_model
 from bookwyrm.signatures import make_signature
-from bookwyrm.settings import DOMAIN, INSTANCE_ACTOR_USERNAME
+from bookwyrm.settings import DOMAIN, INSTANCE_ACTOR_USERNAME, USER_AGENT
 from bookwyrm.tasks import app, MISC
 
 logger = logging.getLogger(__name__)
@@ -432,6 +432,7 @@ def get_activitypub_data(url):
                 "Accept": 'application/ld+json; profile="https://www.w3.org/ns/activitystreams"',
                 "Date": now,
                 "Signature": make_signature("get", sender, url, now),
+                "User-Agent": USER_AGENT,
             },
             timeout=15,
         )
--- a/bookwyrm/models/status.py
+++ b/bookwyrm/models/status.py
@@ -133,6 +133,7 @@ class Status(OrderedCollectionPageMixin, BookWyrmModel):
                 activity.object,
                 get_activity=True,
                 allow_external_connections=allow_external_connections,
+                model=cls,
             )
             if not boosted:
                 # if we can't load the status, definitely ignore it
--- a/bookwyrm/models/status.py
+++ b/bookwyrm/models/status.py
@@ -133,7 +133,7 @@ class Status(OrderedCollectionPageMixin, BookWyrmModel):
                 activity.object,
                 get_activity=True,
                 allow_external_connections=allow_external_connections,
-                model=cls,
+                model=apps.get_model("bookwyrm.Status", require_ready=True),
             )
             if not boosted:
                 # if we can't load the status, definitely ignore it
--- a/bookwyrm/tests/models/test_status_model.py
+++ b/bookwyrm/tests/models/test_status_model.py
@@ -589,23 +589,24 @@ class Status(TestCase):
             to="",
         )
         status = {
-            "id": "http://fish.com/nothing",
-            "type": "Article",
+            "id": "https://fish.com/user/mouse/review/11934130",
+            "type": "Review",
             "published": "2026-07-06T16:05:15.746622+00:00",
-            "attributedTo": "http://fish.com/user/mouse",
-            "content": "<p>review content</p>",
+            "attributedTo": "https://fish.com/user/mouse",
+            "content": "<p>content</p>",
             "to": ["https://www.w3.org/ns/activitystreams#Public"],
-            "cc": ["http://fish.com/user/mouse/followers"],
+            "cc": ["https://fish.com/user/mouse/followers"],
             "replies": {},
+            "tag": [],
+            "attachment": [],
             "sensitive": False,
-            "inReplyToBook": "http://fish.com/book/2265331",
-            "name": 'Review of "Radiant Star": Leckie\'s Study of Provincial Life',
+            "inReplyToBook": "https://fish.com/book/2265331",
+            "name": "Leckie's Study of Provincial Life",
             "@context": [
                 "https://www.w3.org/ns/activitystreams",
                 {"Hashtag": "as:Hashtag"},
             ],
         }
-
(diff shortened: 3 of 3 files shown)


mouse-reeve:
Even after these changes, I'm still getting the error `No model found for activity type "Article"`

qbi:
> Even after these changes, I'm still getting the error `No model found for activity type "Article"`

Maybe a caching issue? Usually nginx caches the request for one minute. You could try to temporarily set `proxy_cache_key "$scheme$proxy_host$request_uri$http_user_agent";` in nginx. 

Hope this helps :-)

qbi:
I looked at the commit https://github.com/bookwyrm-social/bookwyrm/pull/4016/changes/5d7908b439986763aaa16e81f702f61ae8937215 and have the impression that it does some things differently than my suggestion. Could you have a look if my observations are correct?

In the case I described in #4011 `model = cls` leads to `model = Boost`, because this is the class which is called from the method. 

So in the following code block `cls.objects` is `Boost.objects`.

https://github.com/bookwyrm-social/bookwyrm/blob/3804a93cb6cf969dd603d926d0a6cd6ede16f531/bookwyrm/models/activitypub_mixin.py#L122-L127

The SQL query which is built contains an INNER JOIN on `bookwyrm_status` and looks at rows which have an entry in `bookwyrm_boost`. In my specific case the query looks like:

```sql
SELECT "bookwyrm_status"."id", "bookwyrm_status"."created_date", "bookwyrm_status"."updated_date", "bookwyrm_status"."remote_id", "bookwyrm_status"."user_id", "bookwyrm_status"."content", "bookwyrm_status"."raw_content", "bookwyrm_status"."local", "bookwyrm_status"."content_warning", "bookwyrm_status"."privacy", "bookwyrm_status"."sensitive", "bookwyrm_status"."published_date", "bookwyrm_status"."edited_date", "bookwyrm_status"."deleted", "bookwyrm_status"."deleted_date", "bookwyrm_status"."reply_parent_id", "bookwyrm_status"."thread_id", "bookwyrm_status"."ready", "bookwyrm_boost"."status_ptr_id", "bookwyrm_boost"."boosted_status_id" FROM "bookwyrm_boost" INNER JOIN "bookwyrm_status" ON ("bookwyrm_boost"."status_ptr_id" = "bookwyrm_status"."id") WHERE "bookwyrm_status"."remote_id" = https://bookwyrm.de/user/qbi/review/30922 ORDER BY "bookwyrm_status"."published_date" DESC
```

My review has no entry in `bookwyrm_boost`, so the query is always empty. 

Line 124 (`Boost.objects.select_subclasses` in my case) just looks for sub classes, but review is a sister class. 

So as result the code `Boost.find_existing_by_remote_id(review_url)` never finds a local review. 

This means that a re-fetch is done. With your fix the re-fetch works, because the user agent is correct.

If you set `model=Status`  (instead of `model=cls`) we would save the re-fetch. The other behavior should be the same.

In general the current fix works, but always triggers an HTTP re-fetch.

mouse-reeve:
Yes, you're right -- it's always investigating an existing `Status` at that point, not an existing boost. And it's working, caching was indeed the issue :)

PR #4024, opened 2026-07-10
Mergeable suggestion list

mouse-reeve:
## Description
Makes the suggestion list model mergeable, by making the uniqueness constraint enforced at the model level rather than the database level.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/views/admin/data_quality.py
```diff
@@ -94,6 +94,12 @@ def data_quality_data():
         "series_example": models.Series.objects.filter(
             pending_merge_target__isnull=False
         ).first(),
+        "suggestion_list_count": models.SuggestionList.objects.filter(
```
hughrun, line 108: I think we're looking for duplicate `SuggestionListItem`s rather than duplicate lists? In which case we need to know whether a `SuggestionListItem` has both the same `work` and the same `book_list` as another item (i.e. "unique together", except we are not enforcing that at DB level).

I'm not really sure how to check that - I don't think we can use `deduplication_field` because what we need to check is the combination of two fields?

hughrun: Never mind, I have now remembered _why_ we need to merge suggestions lists! Because we might be merging two works that both have suggestions, right?

Though the above still stands, I think. If there's a way to turn a duplicate suggestion into an additional comment & endorsement that would be cool.

bookwyrm/models/list.py
```diff
@@ -65,15 +66,21 @@ class SuggestionList(AbstractList):
         through_fields=("book_list", "work"),
     )
 
-    suggests_for = fields.OneToOneField(
+    suggests_for = fields.ForeignKey(
         "Work",
         on_delete=models.PROTECT,
-        activitypub_field="book",
```
hughrun, line 71: Did you mean to remove this `activitypub_field`?

hughrun (commented):
I think this looks good, pending my query about the activitypub_field. We also seem to have a `summary` for the `SuggestionList` activity but I can't see any obvious source for this.

bookwyrm/views/admin/data_quality.py
mouse-reeve: Yes, you're right about the items. I'm still trying to get my head around the fix for merging items and also allowing multiple users to write notes on the same suggestion. It's also made me realize that when we merge two editions which have different works, we probably need to merge the two works as well -- otherwise the suggestion lists will be orphaned. 

This fix pretty much just prevents suggestion lists from blocking merges, which is important but also the easy part

bookwyrm/models/list.py
mouse-reeve: nope!

PR #4043, opened 2026-07-18
Keep shelf links on the local instance for remote users

### LINKED ISSUE #3685: [Bug] Shelf tabs when viewing a remote user redirect to their original instance

phanky1:
## Description

`Shelf.local_path` was not actually local for remote owners, it returned the origin instance URL, resulting in the shelf tabs redirecting to the remote url.

- Closes https://github.com/bookwyrm-social/bookwyrm/issues/3685

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/models/shelf.py
+++ b/bookwyrm/models/shelf.py
@@ -8,7 +8,6 @@ from django.db import models
 from django.utils import timezone
 
 from bookwyrm import activitypub
-from bookwyrm.settings import BASE_URL
 from bookwyrm.tasks import BROADCAST
 from bookwyrm.utils.db import add_update_fields
 from .activitypub_mixin import CollectionItemMixin, OrderedCollectionMixin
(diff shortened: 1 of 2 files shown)


hughrun (approved):
Thanks for this @phanky1 - LGTM!

For future reference, we no longer need to patch `broadcast_task` in tests - it just hasn't been removed from legacy tests.

PR #4047, opened 2026-07-20
streamline follow buttons

hughrun:
## Description

The "follow from fediverse" button I added a while back is a bit clunky. This changes behaviour so that for logged-in users following from another fediverse account (i.e. not their BookWyrm account) is another option in the dropdown menu rather than a second button.

Screenshot 2026-07-20 at 3 09 57 PM

As an aside, @mouse-reeve is there a particular process to follow if one wants to add another icon to the collection? I was hoping to add an "link opens externally" icon but I couldn't work out how to do it.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests


mouse-reeve:
My increasingly janky workflow for that is to upload the existing icon set into [icomoon's legacy icon editing interface](https://icomoon.io/app/#/select), add the new icon, re-download the icon set, and copy the new css and font files over.

mouse-reeve:
Does this one work? https://github.com/bookwyrm-social/bookwyrm/pull/4050

hughrun:
Looks a bit nicer now:

Logged out:

Screenshot 2026-07-21 at 7 54 17 AM

Logged in:

Screenshot 2026-07-21 at 7 53 52 AM

bookwyrm/templates/snippets/remote_follow_button.html
```diff
@@ -0,0 +1,13 @@
...
+<div class="control">
+    <a
+    class="button is-small is-link {{ class }}"
+    href="{% url 'remote-follow-page' %}?user={{ user.username }}"
+    target="_blank"
+    rel="nofollow noopener noreferrer"
+    onclick="BookWyrm.displayPopUp(`{% url 'remote-follow-page' %}?user={{ user.username }}`, `remoteFollow`); return false;"
+    aria-describedby="remote_follow_warning"
+    >
+    <span>{% trans 'Follow from Fediverse' %}<span class="ml-2 icon icon-share"></span></span>
```
mouse-reeve, line 11: I think we should keep the phrasing the same between the logged in and the logged out versions (the other is "Follow on Fediverse") to be consistent and avoid creating extra translation strings. I think "from" is better than "on" fwiw

bookwyrm/templates/snippets/user_options.html
```diff
@@ -20,6 +20,9 @@
 <li role="menuitem">
     {% include 'snippets/block_button.html' with user=user class="is-fullwidth" blocks=False %}
 </li>
+<li role="menuitem">
```
mouse-reeve, line 23: I feel like this should be in the top part of this list, maybe first on the list. At least I think the positive actions (follow, message) should be together before the block/report actions

mouse-reeve (changes requested):
I like this change! A couple quibbles about placement/wording

bookwyrm/templates/snippets/remote_follow_button.html
hughrun: AFAIK I already changed this 😄

mouse-reeve: You're right -- I must have hit a cached version or just mixed up which branch I was on

mouse-reeve (approved):

PR #4060, opened 2026-07-21
Auto merge related objects

hughrun:
## Description

Initial work to manage related objects when merging.
We don't want to unnecessarily create childless parents when merging, but we should be cautious about deleting or merging related objects too aggressively.

@mouse-reeve this is very much WIP and not ready for merging but I'm sharing now in case you think I'm on the wrong track or there's a better way to approach this.

See also https://decisions.joinbookwyrm.com/d/CjFgmnDW/merging-editions-and-works-what-should-happen-to-authors-and-series-

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/models/mergeable_mixin.py
```diff
@@ -171,14 +171,69 @@ def merge_into(
             related_objs = related_model.objects.filter(**{related_field: self})
             for related_obj in related_objs:
                 try:
-                    setattr(related_obj, related_field, canonical)
-                    related_obj.save()
+                    with transaction.atomic():
```
mouse-reeve, line 174: will this `transaction.atomic` have any effect given that the whole method has the `@transaction.atomic` annotation?

mouse-reeve (commented):
Absolutely looks on the right track to me. I do wonder if we will run into issues calling this function recursively? But it seems like the correct thing to do logically

bookwyrm/models/mergeable_mixin.py
hughrun: Yes I had to add this _because_ it's already atomic. It was erroring because we're saving another object inside an atomic block. Would be grateful for any suggestions if this is the wrong approach!

mouse-reeve: Fascinating! No suggestion, just trying to make sure I understand everything

hughrun:
@mouse-reeve looking for your review on this before i review #3967 since this merges into that, and also I've fixed some tests here that would be breaking there.

FYI there's still two there that fail but I can't work out:

* `test_handle_add_book_to_suggestion_list`
* `test_suggestion_create`

mouse-reeve:
Thank you! I think I fixes those remaining tests in `mergable-mixin` already (https://github.com/bookwyrm-social/bookwyrm/pull/3960/changes/56715132d7b9b1b202bbb5c607c17b6157ba9b88 and https://github.com/bookwyrm-social/bookwyrm/pull/3960/changes/03eb972d26794cea520c13e229c0fb8fc6cfd891)

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
+            additional_authors = [
+                author
+                for author in self.authors.all()
+                if author not in canonical.authors.all()
+            ]
+            for author in additional_authors:
+                canonical.authors.add(author)
+            parent = self.parent_work
+
+            self.delete()
```
mouse-reeve: Are these deletes and saves going to be broadcast? I feel like they should not, because they would propagate as simple deletions rather than actual merged 🤔

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
                     getattr(related_obj, related_field).remove(self)
+                except IntegrityError:
+                    # e.g. a seriesbook when one already exists for that series
+                    continue
+
+        edition_model = apps.get_model("bookwyrm.Edition")
+        work_model = apps.get_model("bookwyrm.Work")
+        suggests_model = apps.get_model("bookwyrm.SuggestionList")
+
+        if self.__class__ == edition_model:
```
mouse-reeve: should this be `isinstance(self, edition_model)`? does it matter?

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
+                canonical.authors.add(author)
+            parent = self.parent_work
+
+            self.delete()
+
+            # if self.parent is now without any child editions, merge it too
+            # unless it is already marked as a merge candidate
+            if parent:
+                parent.refresh_from_db()
+                if not parent.editions.count() and not parent.pending_merge_target:
```
mouse-reeve: Why only if there are no children? Shouldn't you always merge the work, if the editions of the works are duplicate? Unless the edition was associated with the wrong work, I guess?

bookwyrm/models/mergeable_mixin.py
mouse-reeve: Maybe this code would be easier to follow/test if these if blocks were separate methods (like, `merge_related_works`, `merge_related_authors`, `merge_related_suggestion_lists`)?

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
+                author
+                for author in self.authors.all()
+                if author not in canonical.authors.all()
+            ]
+            for author in additional_authors:
+                canonical.authors.add(author)
+
+            # switch over any suggestion lists
+            # deal with duplicate suggestions in the View
+            if lists := suggests_model.objects.filter(
```
mouse-reeve: I think you can just iterate over this without the if statement, and if the queryset is empty, it won't execute the code?

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
+        elif self.__class__ == work_model:
+            # NOTE: Once we are using Contributions
+            # this will need to be amended
+            additional_authors = [
+                author
+                for author in self.authors.all()
+                if author not in canonical.authors.all()
+            ]
+            for author in additional_authors:
+                canonical.authors.add(author)
```
mouse-reeve: I _think_ you can do all of lines 214 to here with something like `canonical.authors.add(*self.authors.all())` without the for loop, and the filtering, since duplicates aren't created in a many to many add (same applies to the previous usage around line 191)

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,11 +171,63 @@ def merge_into(
...
+
+            # switch over any suggestion lists
+            # deal with duplicate suggestions in the View
+            if lists := suggests_model.objects.filter(
+                suggests_for__in=[self.id, canonical.id]
+            ):
+                for list in lists:
+                    if list.suggests_for == self:
+                        list.suggests_for = canonical
+                        list.save()
```
mouse-reeve: Actually couldn't this whole block from 224 to 230 be:
`suggests_model.objects.filter(suggests_for=self).update(suggests_for=canonical)`?

mouse-reeve (commented):
These are thoughts I had just reading the code, I didn't confirm they actually work, so please sanity check me here

hughrun:
> These are thoughts I had just reading the code, I didn't confirm they actually work, so please sanity check me here

Thankyou! Wow looks like I have been a bit sloppy here TBH.

bookwyrm/models/mergeable_mixin.py
hughrun: It should, because it's more pythonic. It's also one of the things that immediately falls out of my brain again every time I remember it exists.

bookwyrm/models/mergeable_mixin.py
hughrun: My worry was about recursively merging works or triggering a merge on the parent work in the middle of merging its editions, but you're right - if the work isn't marked as a merge candidate then we aren't in the middle of merging it.

bookwyrm/models/mergeable_mixin.py
hughrun: Yes, I keep forgetting I can do things like that! Nice one.

bookwyrm/models/mergeable_mixin.py
hughrun: amazing

hughrun:
Ok I think this is much better now, thanks.

I removed the whole suggestionlist section because they're already merged, since they're not m2m - I got mixed up in my head about that.

bookwyrm/models/mergeable_mixin.py
```diff
@@ -170,13 +171,23 @@ def merge_into(
...
                     getattr(related_obj, related_field).remove(self)
+                except IntegrityError:
+                    # e.g. a seriesbook when one already exists for that series
+                    continue
 
-        self.delete()
+        # Well here we are merging some m2m fields after all
+        self.merge_related_authors(canonical)
+        parent = self.parent_work if hasattr(self, "parent_work") else None
+        self.delete(broadcast=False) # don't send a Delete, we're merging
```
mouse-reeve: `Model.delete() got an unexpected keyword argument 'broadcast'` -- I think since this is a true deletion it isn't doing any activitypubity so the param is unnecessary.

mouse-reeve (commented):
This is the only issue I saw. I'm not sure which version in the merge conflict is better, so if you have a stance you should take it.

hughrun:
Sorry git is doing something weird with the new lines on those template files.

PR #4067, opened 2026-07-23
Add a mutex lock on to_model function

### LINKED ISSUE #4062: federated books still creating duplicates
**Describe the bug**
Our old nemesis is back.

This was added recently:

https://bw.heraut.eu/book/166612/s/notre-s%C5%93ur-rabat-joie

But appears twice on b.s:

https://bookwyrm.social/book/2333298/editions

And created an editionless Work:

https://bookwyrm.social/book/2333297.json

**Expected behavior**
No duplicates and no childless works

**Screenshots**

**Instance**

**Additional context**

---

hughrun:
## Description
Adds a [mutex lock](https://en.wikipedia.org/wiki/Lock_(computer_science)) on every object processed by to_model. This should have a negligible impact on most items but will block processing on any activity object where an incoming object with the same remote ID is currently being processed until the first one releases the lock.

This is intended to fix the problem described in #4062 and partially addressed in #3471 where reading activities with statuses cause two near-simultaneous broadcasts referring to the same external edition. This seems to be at the core of our problem with duplicate identical editions, works, and possible series and authors, as well as many childless works.

### caveats

1. I've never done anything with mutex locks before so there might be a better way to do this. This proposal is based on [the Celery docs](https://docs.celeryq.dev/en/stable/tutorials/task-cookbook.html) plus some web searching to understand that example and for different approaches. Maybe it doesn't work the way I think. Maybe it doesn't work at all! ~I'm not sure how to run tests against it~ - suggestions very welcome.
2. ~There is a failing test and I have no idea why it is failing, or whether the failure is directly related to this code, but it might be.~
3. If an incoming object doesn't have a discernable URI it will throw an error
4. If the first item takes longer than ~10~ ~5 minutes to process the lock will release. Note that this is ~10 minutes~ from the point it first starts running `to_model`, not ~10 minutes~ from when it is received and added to a processing queue.

- Related Issue #
- Closes #4062

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/activitypub/base_activity.py
+++ b/bookwyrm/activitypub/base_activity.py
@@ -1,9 +1,13 @@
 """basics for an activitypub serializer"""
 
 from __future__ import annotations
+from contextlib import contextmanager
 from dataclasses import dataclass, fields, MISSING
+from hashlib import md5
 from json import JSONEncoder
 import logging
+import random
+import time
 from typing import Optional, Union, TypeVar, overload, Any
 
 import requests
@@ -15,6 +19,7 @@ from django.utils.http import http_date
 from bookwyrm import models
 from bookwyrm.connectors import ConnectorException, get_data
 from bookwyrm.models import base_model
+from bookwyrm.redis_store import r
 from bookwyrm.signatures import make_signature
 from bookwyrm.settings import (
     DOMAIN,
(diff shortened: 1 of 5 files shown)


bookwyrm/activitypub/base_activity.py
```diff
@@ -41,6 +46,45 @@ def default(self, o):
...
+    """Set a mutex lock on an incoming item to prevent race conditions causing duplicates"""
+
+    # hash the remote id so the key is a reasonable length
+    try:
+        id_hash = md5(object_id.encode("utf-8")).hexdigest()
+    except AttributeError:
+        raise ActivitySerializerError("Incoming object has no identifier")
+    lock_id = f"lock:{id_hash}"
+    # expire after 10 minutes in case something goes wrong
+    expire_seconds = time.monotonic() + 600
```
ilkka-ollakka: i think most of things in bookwyrm are handled in seconds, so maybe 1 minute is enough for default expire?

bookwyrm/activitypub/base_activity.py
```diff
@@ -41,6 +46,45 @@ def default(self, o):
...
+    # try to get the lock until you have it or it has (nearly) expired
+    while time.monotonic() < expire_seconds - 5:
+        # 1: arbitrary value for the key
+        # nx: only set if the key doesn't already exist
+        # ex: expire after this many seconds
+        if lock := r.set(lock_id, "1", nx=True, ex=600):
+            break
+
+        # if we don't have the lock, try again in 5 to 10 seconds
+        sleep_time = random.randint(5, 10)
```
ilkka-ollakka: I would sleep maybe 1-5 seconds for same reason as previous.

ilkka-ollakka:
For failed testings, you should add redis_activity to containers that are started on pytest, or maybe take fakeredis into use on pytest side?

bookwyrm/activitypub/base_activity.py
hughrun: 10 minutes is deliberately generous but it is probably too much. I'm thinking of an incoming item from a server with a slow connection and, say, lots of authors and associated editions. Maybe it should be based on `QUERY_TIMEOUT`? Something like `30 * QUERY_TIMEOUT` would be 150 seconds with the default value.

ilkka-ollakka: yes, I think it could be good idea. But time-values are more on fine-tuning end of the change, so even with current ones we can go forward just fine.

ilkka-ollakka (approved):
I think this looks really good!

PR #4070, opened 2026-07-27
update edit book views and templates (WIP)

hughrun:
## Description

Draft to support Mouse's initial work in `author-table`. Some of my changes here a a bit opinionated so feel free to push back if you think it's the wrong approach. So far:

- fixes some logic in edit book view
- changes contribution type enum to match ONIX codes (this is pre-empting planned work on connectors)
- adds ability to update contribution types in edit book page
- removes free text contribution description on each contribution
- maybe fixes search_vector update?

In general, I'm trying to make this work similarly to the recent changes to series. The tricky bit comes next when we need to look at connector imports - we want to make it seamless so usually it will just import as "author", but where we do have more information (primarily from Inventaire, but also Finna and maybe soon from SRU-based connectors) we should use it. 

I've also noted a couple of `TODO`s where I'm just a bit unclear on the original logic of the CreateBook view in particular. As far as I can see we don't really check for any `deduplication_field` uh, duplicates, but we should? The code comments imply that this is happening and that we might be pulling in authors from an existing parent work, but I don't see where this happens - @mouse-reeve it would be great to either be educated on how this happens, or alternatively we could fix it if it isn't happening.

### TODO 

(by anyone - probably not all in this single PR)

- [ ] add contribution types to author pages
- [ ] add contribution types to series pages
- [ ] update connectors: we should do it like series: check for a match based on identifiers and if there isn't a fuzzy match at all add a new one, if we're not sure bring in as a string with a manual check
- [ ] update Activity objects
- [ ] add tests

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/views/books/edit_book.py
```diff
@@ -132,19 +137,19 @@ class CreateBook(View):
...
         # returns None if no match is found
         form = forms.EditionForm(request.POST, request.FILES)
-        data = {"form": form}
+        data = {"contribution_types": ContributionType, "form": form}
 
         # collect data provided by the work or import item
         parent_work_id = request.POST.get("parent_work")
         authors = None
-        if request.POST.get("authors"):
+        if request.POST.get("authors"): # TODO When would we have authors that aren't 'add_author' in a newly created book?
```
mouse-reeve, line 152: Maybe creating a new edition of an existing work?

mouse-reeve:
Agreed that we should check for deduplication fields. Maybe that's out of scope for this PR, though.

PR #4075, opened 2026-08-01
fix series displaying as stringified json

### LINKED ISSUE #4069: new first-class series concept somehow getting munged later on into raw json?
**Describe the bug**

Books that previously looked fine are now showing a second series string that's the raw json from some part of the book's metadata. Not all books with series, not even sure which cases or not to reproduce. Including screenshot and the raw text in case it helps. Example: instead of showing "culture" (the series name) this string is present: 

```
([{'name': 'Culture series', 'alternative_names': ['Ciclo della Cultura', 'La Cultura', 'Cultura (serie SF)', 'Kültür serisi', 'Kultur-Zyklus', 'Культура', 'Kulturen', 'Cycle de La Culture', '文明系列 (小说)', 'תרבות (סדרת ספרים)', 'Seria o Kulturze', 'Culture series'], 'inventaire_id': 'wd:Q5193682', 'wikidata': 'Q5193682', 'goodreads_key': '49118', 'isfdb': '504'}], #5) 
```

Stringified python object maybe? FWIW these book listings definitely didn't have these bugs when I last interacted with them (definitely would have noticed), so whatever the cause was: it happened between my status updates' timestamps and now.

**Expected behavior**

If a user hand-enter that, then I suppose it's expected behavior, but I suspect more likely this is a programmatic bug of some sort.

**Screenshots**

<details><summary>click to expand for all the screenshots</summary>

Image

Image

Image

Image

Image

</details>

**Instance**

- https://bookwyrm.social/book/27414/edit
- https://bookwyrm.social/book/6036/s/use-of-weapons
- status: https://bookwyrm.social/user/jzacsh/generatednote/12096966#anchor-12096966
- https://bookwyrm.social/user/jzacsh/generatednote/12096997#anchor-12096997

-----

    - Device: [e.g. iPhone6] pixel 9 pro
    - OS: [e.g. iOS8.1] Android 17
    - Browser [e.g. stock browser, safari] Firefox beta
    - Version [e.g. 22]:

```
154.0b2 (Build #2016174575), b7b1868f2e31da0d089ccb869ec773fa5d9fd9ac
GV: 154.0-20260724090549
AS: 154.0
OS: Android 17
```

hughrun:
## Description

In #3747 we added `get_or_create_seriesbook_from_data` and adjusted the Inventaire connector so that we were receiving full Inventaire series objects. This is great except I forgot to ensure that the work `series` value was deleted after we'd processed it. This means any works that were imported from Inventaire now retain the object as their series value.

When that work is imported into another BookWyrm instance it will be processed as a `string` because that's what the Book activity serializer tells it to do. Now we have a stringified object as the `series` value on the work and soon in child editions. Oh no.

### Prospective fix
I have adjusted `get_or_create_seriesbook_from_data` so that we clear the `series` values on both the edition and the work once we no longer need them.

### Retro/prospective fix
I have added a data migration that will run a cleanup process automatically. Any work or edition with a `[` at the start of the `series` entry is checked to see if we can create a Series object with `json.loads`. If we can, we turn it into a proper seriesbook, if we can't load the string we leave it because it must be something else like `Series Name [book 1]`

The data migration should not run on brand new installs that haven't initiated a database yet, and run in all other cases.

- Related Issue #
- Closes #4069

- [x] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests


bookwyrm/migrations/0243_auto_20260730_0926.py
```diff
@@ -0,0 +1,79 @@
...
+    Work = apps.get_model("bookwyrm", "Work")
+    Edition = apps.get_model("bookwyrm", "Edition")
+
+    try:
+        connector_info = Connector.objects.get(identifier="bookwyrm.social")
+        connector = connector_manager.load_connector(connector_info)
+        fixed_works = 0
+        fixed_editions = 0
+
+        works = Work.objects.filter(series__contains="[")
```
ilkka-ollakka: would it be enough if this would be just series__startswith ? Not big issue as there is check if it can be read as json fully.

hughrun: that might be slightly more efficient, yes

ilkka-ollakka:
I think both migrations run in new installations. Data-migration would be just no-op as there wouldn't yet data to migrate.

bookwyrm/migrations/0243_auto_20260730_0926.py
```diff
@@ -0,0 +1,79 @@
...
+        print("Skipping series data fix - connectors not loaded yet")
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('bookwyrm', '0242_alter_seriesbook_book'),
+    ]
+
+    operations = [
+        migrations.RunPython(fix_series),
```
mouse-reeve:
```suggestion
        migrations.RunPython(fix_series, reverse_code=migrations.RunPython.noop),
```

Adding a reverse here will make it possible to migrate backwards past this migration

bookwyrm/migrations/0243_auto_20260730_0926.py
```diff
@@ -0,0 +1,79 @@
...
+
+    try:
+        connector_info = Connector.objects.get(identifier="bookwyrm.social")
+        connector = connector_manager.load_connector(connector_info)
+        fixed_works = 0
+        fixed_editions = 0
+
+        works = Work.objects.filter(series__startswith="[")
+        for work in works:
+            if edition := work.default_edition:
```
mouse-reeve: I hit an error here: `AttributeError: 'Work' object has no attribute 'default_edition'`

mouse-reeve (changes requested):

bookwyrm/migrations/0243_auto_20260730_0926.py
hughrun: hmm. Will fix that but out of interest, how is that possible? Shouldn't migrations have added that attribute to all your Works even if it's empty?

mouse-reeve: I don't know.. I wonder if migrations have less access to attributes than the regular code does?

hughrun: I think this is because data migrations can't access properties set with the `@property` decorator.

mouse-reeve (approved):

PR #4080, opened 2026-08-05
Improved JavaScript

PatrickChildersIT:
## Description
Stumbled upon the js files and thought they could use a touch-up.

Main changes are:
`let` -> `const` is better scoping and occasionally js engines optimize it better

removal of `addRemoveClass` this function was terribly confusing to parse since it did two things, and they were opposite things depending on the optional third parameter. the replacement code is perhaps a bit more verbose but who said that was such a bad thing. at least it's much clearer what is happening with the show/hide activate/deactivate functions or sometimes just replacing it with the equivalent `classList` modification

i removed a bit of code that was still around from before i made the star ratings function differently, so that being vestigial could just get cut right out

Finally, and what started me making changes to begin with was the `autocomplete.js` file. I reimplemented it, same functionality, but now much more concise and understandable just by using a bit of regex.

- [ ] Bug Fix
- [ ] Enhancement
- [x] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/static/js/bookwyrm.js
+++ b/bookwyrm/static/js/bookwyrm.js
@@ -156,13 +156,16 @@ let BookWyrm = new (class {
 
         if (count != currentCount) {
             const wrapper = counter.closest("[data-poll-wrapper]");
-            if(count < 1) {
+
+            if (count < 1) {
                 this.classHide(wrapper);
             } else {
                 this.classShow(wrapper);
             }
+
             counter.innerText = count;
-            if(hasMentions) {
+
+            if (hasMentions) {
                 this.classList.add("is-danger");
             } else {
                 this.classList.remove("is-danger");
@@ -240,7 +243,7 @@ let BookWyrm = new (class {
         if (targetId && !trigger.classList.contains("pulldown-menu")) {
             const target = document.getElementById(targetId);
 
-            if(pressed) {
+            if (pressed) {
                 this.classShow(target);
                 this.classActivate(target);
             } else {
@@ -299,7 +302,7 @@ let BookWyrm = new (class {
         if (targetId) {
             const target = document.getElementById(targetId);
 
-            if(expanded) {
+            if (expanded) {
                 this.classActivate(target);
             } else {
                 this.classDeactivate(target);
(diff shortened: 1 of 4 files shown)


bookwyrm/static/js/bookwyrm.js
```diff
@@ -384,24 +411,23 @@ let BookWyrm = new (class {
...
-            node.classList.add(classname);
-        } else {
-            node.classList.remove(classname);
-        }
+    classHide(node) {
+        node.classList.add("is-hidden");
+    }
+
+    classShow(node) {
+        node.classList.remote("is-hidden");
```
ilkka-ollakka: I think this should be remove?

PatrickChildersIT: 🤦 
Thank you for pointing out my typo.
Even after staring at your comment and the line it still took me about three takes to see it.

mouse-reeve:
Your solution for the autocomplete script is so much better than my original one that I genuinely laughed. Much appreciated!

bookwyrm/static/js/bookwyrm.js
```diff
@@ -384,24 +411,23 @@ let BookWyrm = new (class {
...
-            node.classList.add(classname);
-        } else {
-            node.classList.remove(classname);
-        }
+    classHide(node) {
+        node.classList.add("is-hidden");
+    }
+
+    classShow(node) {
+        node.classList.remove("is-hidden");
```
mouse-reeve, line 419: These don't seem to work for me -- I see the method get called, but the class list on the item that is passed in doesn't change

mouse-reeve (changes requested):
I agree with all your changes philosophically! As it stands, this breaks the boost/fav buttons (and likely others that were previously using `addRemoveClass`)

bookwyrm/static/js/bookwyrm.js
PatrickChildersIT: ah the problem was the dreaded inverted logic
I had `indexOf("is-hidden") == -1` in a variable `isHidden`, but that name is the opposite of what that is checking

if it is -1 it can't find "is-hidden" so it's _not_ hidden. That is only used in the `interact(event)` which is used for the boost/fav buttons and other things

so I just flipped it to `!=` so the variable is named correctly and the `if` statement uses the correct `classShow` or `classHide`

mouse-reeve (approved):

PR #4087, opened 2026-08-08
More type hinting

PatrickChildersIT:
## Description

Went to write some unit tests to improve code coverage and I was just annoyed enough at the lack of type hinting that I got a bit carried away adding some.

Draft PR because before I go and spend more time adding type hints to places I wanted to see if anyone had any issues with how I've gone about it so far.

Does anyone feel strongly about avoiding explicit type hints that could be inferred vs adding them anyway?
Are there any non-stdlib type hints that should be avoided, such as `QuerySet[Model, Any]`?
I'm still learning some parts of Django, so are there any Django-specific nuances that mean I should add type hints differently?

I appreciate any and all feedback!

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/activitystreams.py
+++ b/bookwyrm/activitystreams.py
@@ -8,6 +8,7 @@ from django.utils import timezone
 from opentelemetry import trace
 
 from bookwyrm import models
+from bookwyrm.models.base_model import BookWyrmModel
 from bookwyrm.redis_store import RedisStore, r
 from bookwyrm.tasks import app, STREAMS, IMPORT_TRIGGERED
 from bookwyrm.telemetry import open_telemetry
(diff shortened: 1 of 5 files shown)


mouse-reeve:
This all looks good to me! I'm still not super comfortable with the ins and outs of typing, so I'd like to get a second take from another person (@bookwyrm-social/code-review), but definitely let's consider this ready to review, and open a new PR for any further work.

hughrun:
Looks great so far.

> Does anyone feel strongly about avoiding explicit type hints that could be inferred vs adding them anyway?

I think the intention of adding typing was that eventually everything will include type hints. We just haven't done that because nobody was keen to retrospectively type the entire legacy codebase (but if you're keen, go for it ;) ) I think even if they can be inferred, we should make them explicit so as to catch any bugs before they're created.

> Are there any non-stdlib type hints that should be avoided, such as QuerySet[Model, Any]?

Ideally we'll avoid types with `Any`, but I think once you start going down the path of needing to change a bunch of code in order to have more specific type hints, you're heading in the wrong direction. So my view on this is that something like `QuerySet[Model, Any]` is better than no type hint at all and will catch anything that wants to pass something that isn't a queryset.

> I'm still learning some parts of Django, so are there any Django-specific nuances that mean I should add type hints differently?

Like @mouse-reeve I'm fairly new to Python typing so I'm not sure there's anything in particular to look out for other than what you've already noted in that Django has things like `QuerySet` that need to be accounted for.

hughrun (approved):

PR #4097, opened 2026-08-15
Fix ipblocklist x forwarded for header

### LINKED ISSUE #4095: IPBlocklistMiddleware may not match the real user's IP address
**Describe the bug**

I noticed that IPBlocklistMiddleware finds the user's IP address via REMOTE_ADDR, but since the backend sits behind nginx, that value is nginx's IP rather than the user's. I think we should use X-Forwarded-For.
https://github.com/bookwyrm-social/bookwyrm/blob/89165dc696fea8371a0823a390f0bc2f394841a1/bookwyrm/middleware/ip_middleware.py#L13-L14

timothyjrogers:
## Description

- Closes #4095

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/middleware/ip_middleware.py
+++ b/bookwyrm/middleware/ip_middleware.py
@@ -12,7 +12,7 @@ class IPBlocklistMiddleware:
 
     def __call__(self, request):
         if "x-forwarded-for" in request.headers:
-            address = request.headers.get("x-forwarded-for")
+            address = request.headers.get("x-forwarded-for").split(',')[0]
         else:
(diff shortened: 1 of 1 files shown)


ilkka-ollakka:
Tested this and it works as expected.

We should consider if we should also override the REMOTE_ADDRESS in forwarded-for case, to show right remote address in all the other places too. But that is out of scope for this pr, and not trivial to get right with all the ways it can be spoofed.

Issue #3731, opened 2025-11-14
`quote` property conflicts with FEP-044f


silverpill:
The `quote` property is used by Bookwyrm to specify quoted text in a [quotation](https://github.com/bookwyrm-social/bookwyrm/blob/main/FEDERATION.md#quotation) object (the example might be wrong, I got an object with type `Note`, not `Quotation`). However, this property is also used in [FEP-044f: Consent-respecting quote posts](https://codeberg.org/fediverse/fep/src/branch/main/fep/044f/fep-044f.md).

This may cause interoperability issues if a remote server expects the value of `quote` to be a URI and not HTML.

Bookwyrm version: 0.7.3

Related issue: https://github.com/bookwyrm-social/bookwyrm/issues/3730

hughrun:
I agree that we need to fix our custom activities by adding proper [JSON-LD extension references via `@context`](https://www.w3.org/TR/activitystreams-core/#extensibility)

Just a couple of comments for some background on this:

1. There was a problem with `quote` objects in Mastodon when they first released them, because they published with a `@context` but didn't actually check it when receiving activities, so didn't know what to do with BookWyrm's `quote` objects - that is fixed in the most recent Mastodon release. I'm not throwing too many stones from our glass house here because that's basically what is described in this Issue, but from our end.
2. BookWyrm generally tries to turn custom activity types into a `Note` if sending to a non-BookWyrm server. We can do a better job of this too, by [specifying the `type` as an array with `Note`](https://www.w3.org/TR/activitystreams-core/#object) rather than checking whether it tells us it is a BookWyrm server, as is currently the case.

silverpill:
>we need to fix our custom activities by adding proper JSON-LD extension references via `@context`

This solves the problem for JSON-LD consumers, but they are a minority, and everyone else will still need to deal with the ambiguous property, somehow. However, I think adding Bookwyrm-specific URL to `@context` would be an improvement even for JSON consumers. Currently l search for a `/quotation/` string in the `id` to determine whether it was generated by Bookwyrm or not, which is not reliable.

>We can do a better job of this too, by specifying the type as an array with Note

Type arrays are not widely supported. It would be better to send the real type, this allows other platforms to add proper support for Bookwyrm objects (some may already accept `Quotation`).

An exception could be made for Mastodon which has a hard-coded list of supported types and refuses to extend it.

hughrun:
[...]

PR #4098, opened 2026-08-16
Add JSON-LD namespacing to BookWyrm activities

hughrun:
## Description
Adds explicit json-ld namespacing for BookWyrm activity objects.

This needs to be a two-part process:

* First of all we need to implement "forwards compatible" handling of namespaced types and properties. That is this PR. 
* Then once all BookWyrm instances have upgraded, we can implement a breaking change (i.e. breaking from before the version with the earlier PR) that removes the non-namespaced properties (most problematically, `quote`) and sends extended types as part of an array rather than just the ActivityStreams type. That is in #4099 which will need to wait before being merged.

There are three primary changes here:

1. add a reference to a full BookWyrm vocabulary in the `@context` of most BookWyrm extended types
2. add a _namespaced_ reference to a full BookWyrm vocabulary in the `@context` of BookWyrm extended types that inherit from `Note` or `Article`. This allows for namespaced extended properties on `Note` types (e.g. `bw:quote`), Note that this is _additional_ to the non-namespaced property, because if we just switch it straight over all existing BookWyrm implementations will fail to process these properties
3. add support for receiving activity `type`s as an Array/list - this is supported in the JSON-LD and Activity Streams standards so we really should be able to handle it.

In addition, I've added a feature flag reserved for future use. This will allow admins to, at some point in the future, switch on "strict" checking of BookWyrm types. At the moment we just blindly accept anything with a type that matches a BookWyrm type, but it could be anything. Using this flag will mean we check that at BookWyrm context is being used if we're accepting a BookWyrm type. Since this would immediately break federation if switched on right now (as nobody is including the BookWyrm vocab in the context), it's currently set to `False` and unable to be changed. The intention here is that it will at some point default to `True` for new installs, and also be able to be changed by admins.

The BookWyrm vocabulary is now outlined in the docs at https://docs.joinbookwyrm.com/bookwyrm-vocabulary.html. This is permalinked from [https://www.w3id.org/BookWyrm/ns](https://www.w3id.org/BookWyrm/ns) so you can reference like [`https://www.w3id.org/BookWyrm/ns#quote`](https://www.w3id.org/BookWyrm/ns#quote) and the anchor will be passed through.

- Related Issue #3731 https://github.com/bookwyrm-social/documentation/pull/320
- Closes #

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


bookwyrm/activitypub/base_activity.py
```diff
@@ -32,9 +32,36 @@
 
 logger = logging.getLogger(__name__)
 
-
 TBookWyrmModel = TypeVar("TBookWyrmModel", bound=base_model.BookWyrmModel)
 
+# custom properties that appear on various decendants of Note
+namespaced = [
```
PatrickChildersIT: ideally lists should be plural, maybe `namespaced_properties` ?

also, depending on what style guide you adhere to, module scope constants should be `CAPS_SNAKE_CASE`

bookwyrm/activitypub/base_activity.py
```diff
@@ -104,8 +131,45 @@ def naive_parse(activity_objects, activity_json, serializer=None):
             activity_json["type"] = "PublicKey"
 
         activity_type = activity_json.get("type")
-        if activity_type in ["Question", "Article"]:
+        if isinstance(activity_type, str) and activity_type in ["Question", "Article"]:
```
PatrickChildersIT: is this addition actually necessary? if `activity_type` isn't a string then it surely won't be in the following list of strings that is the second part of the expression

bookwyrm/activitypub/base_activity.py
```diff
@@ -104,8 +131,45 @@ def naive_parse(activity_objects, activity_json, serializer=None):
...
+                raise ActivitySerializerError("No @context provided")
+            # not ActivityPub
+            if "https://www.w3.org/ns/activitystreams" not in activity_json["@context"]:
+                raise ActivitySerializerError("Not an activitystreams activity")
+            # is a non-Note BookWyrm type without BW namespace
+            if activity_type in other_custom_activity_objects and not has_namespace:
+                raise ActivitySerializerError("BookWyrm namespace error")
+            # is decendent of Note without the bw namespace
+            if (
+                len(set(activity_json.keys()) & set(namespaced)) > 0
```
PatrickChildersIT: I believe that `dict_keys` can use `&` intersection, so they don't have to be converted to a set
also empty sets are falsy so no need to use `len`
turning this line into `if( (activity_json.keys() & set(namespaced) ) and not has_bw_namespace):`

bookwyrm/activitypub/base_activity.py
```diff
@@ -187,7 +251,11 @@ def to_model(
             self.to_model. e.g. if this is a Work being dereferenced from
             an incoming Edition
         """
-        model = model or get_model_from_type(self.type)
+        try:
```
PatrickChildersIT: i prefer to avoid using try/except for control flow, asking for forgiveness instead of permission
what would you think about
```python
model = self.type[0] if isinstance(self.type, list) else self.type
```

bookwyrm/models/activitypub_mixin.py
```diff
@@ -238,7 +238,6 @@ def save(
                 return
 
             try:
-                # TODO: here is where we might use an ActivityPub extension instead
                 # do we have a "pure" activitypub version of this for mastodon?
```
PatrickChildersIT, line 241: did you mean to delete this line as well?

bookwyrm/models/site.py
```diff
@@ -115,6 +115,9 @@ class SiteSettings(SiteModel):
 
     field_tracker = FieldTracker(fields=["name", "instance_tagline", "logo"])
 
+    # feature flags
+    use_strict_ap_namespacing = models.BooleanField(default=False)
```
PatrickChildersIT: i try to resist acronyms when possible, the wily things like to add up, maybe `use_strict_activitypub_namespacing` ?

bookwyrm/tests/views/inbox/test_inbox.py
```diff
@@ -174,3 +174,31 @@ def test_create_by_deactivated_user(self, _):
...
+            "content": "<p>I am really enjoying this book</p>",
+            "to": ["https://www.w3.org/ns/activitystreams#Public"],
+            "cc": ["https://example.com/user/mouse/followers"],
+            "replies": {},
+            "summary": "spoilers",
+            "attachment": {},
+            "inReplyToBook": "https://example.com/book/2343050",
+            "@context": "https://www.w3.org/ns/activitystreams",
+        }
+        with patch("bookwyrm.views.inbox.has_valid_signature") as mock_valid:
```
PatrickChildersIT: a decorated `@patch` and two context managers, any reason not to put all three patches as decorators?

bookwyrm/activitypub/base_activity.py
hughrun: hmm yes I think this is an artefact of an earlier approach that I changed my mind about, thanks for the pick up.

bookwyrm/activitypub/base_activity.py
hughrun: ah good call. Though it's `model` if that is passed in so I think it will need to be something like

```py
model = model or (self.type[0] if isinstance(self.type, list) else self.type)
```

bookwyrm/models/activitypub_mixin.py
hughrun: Yes I think I was wrong (I wrote that line originally). I originally thought we could resolve the need to change the content field depending on the target software by using namespaced extensions (basically the point of this PR). But now I've looked into how we'd actually _do_ that, the only real solution would be to add yet another extended field specifically for BookWyrm to use instead of `content`. Which I guess we could do but seems unnecessary give we already have the "pure content" approach.

bookwyrm/tests/views/inbox/test_inbox.py
hughrun: copypasta!

PR #4106, opened 2026-08-21
fix: build the shelf RSS alternate link from the path only

### LINKED ISSUE #4105: `get_full_path` adds endless rss entries to the path.
**Describe the bug**
While researching a performance issue I found some "interesting" entries in my nginx log. There were some `GET` requests like
>  /user/USERNAME/books/read?sort=sort_title/rss/rss/rss/…/rss&page=34/rss/…/rss
The longest entry in my log file contained 380 `/rss` elements.

What I found is that the RSS `<link rel="alternate">` on shelf pages is built with `request.get_full_path`. This includes the query string. So `/rss` is appended *after* the query string. Fetching the resulting link produces a `sort` parameter that ends in `/rss`, and the template then appends `/rss` again. So BookWyrm produces an endless stream of unique URLs. In my case an AI bot (Meta) visited every URL and generated the load.

**To Reproduce**

1. Open a shelf with a sort parameter, e.g. `/user/<user>/books/read?sort=sort_title`
2. Inspect the RSS link in the page head:  `href="/user/<user>/books/read?sort=sort_title/rss"`
3. Request that URL; the link is now  `?sort=sort_title/rss/rss`.
4. GOTO 1.

**Expected behavior**
The RSS link should point to a canonical URL and should be independent from the query string.

**Instance**
- https://bookwyrm.de

yzxcj797:
Fixes #4105.

## Summary

The shelf template built the RSS `<link rel="alternate">` href with `request.get_full_path`, which **includes the query string**, so `/rss` was appended after it:

```
/user/USERNAME/books/read?sort=sort_title/rss
```

Fetching that URL gives the view a `sort` parameter that ends in `/rss`; the template then appends `/rss` again — an unbounded family of unique URLs. The report's nginx log shows requests carrying up to **380** `/rss` segments as a crawler walked the chain.

The fix uses `request.path` (the shelf feed URL itself is query-free — the feed does not consume `sort`), so the alternate link is simply:

```
/user/USERNAME/books/read/rss
```

## Testing

Added `test_shelf_page_rss_link_ignores_query_string` to `bookwyrm/tests/views/shelf/test_shelf.py`: renders a sorted shelf page (`?sort=sort_title`) and asserts the alternate link href is the plain `/user/<name>/books/<shelf>/rss` with no query string.

Note: I could not execute the Django suite locally (the project's settings require a PostgreSQL server, none available in this environment); the test follows the existing patterns in the same file (`test_shelf_page_sorted_shelved` next door) and runs under the project's CI.

### DIFF
--- a/bookwyrm/templates/shelf/shelf.html
+++ b/bookwyrm/templates/shelf/shelf.html
@@ -14,7 +14,7 @@
 
 
 {% block head_links %}
-    <link rel="alternate" type="application/rss+xml" href="{{ request.get_full_path }}/rss" title="{{ user.display_name }} - {{ shelf.name }}" />
+    <link rel="alternate" type="application/rss+xml" href="{{ request.path }}/rss" title="{{ user.display_name }} - {{ shelf.name }}" />
 {% endblock %}
 
 {% block content %}
--- a/bookwyrm/tests/views/shelf/test_shelf.py
+++ b/bookwyrm/tests/views/shelf/test_shelf.py
@@ -146,6 +146,33 @@ class ShelfViews(TestCase):
         validate_html(result.render())
         self.assertEqual(result.status_code, 200)
 
+    def test_shelf_page_rss_link_ignores_query_string(self):
+        """the RSS alternate link must use the path only: with the query
+        string included, /rss lands after ?sort=..., crawlers fetch it as a
+        sort value ending in /rss and the feed link grows /rss forever
+        (#4105)"""
+        view = views.Shelf.as_view()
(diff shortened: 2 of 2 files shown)


mouse-reeve:
Hi there, did you use LLMs in your workflow for this?

qbi:
This really looks like a LLM-PR. I had several of the same type in the past. :face_with_diagonal_mouth:

mouse-reeve:
yeah, I’m hoping the move off of github will reduce this problem

PR #4153, opened 2026-09-10
Even more type hinting

PatrickChildersIT:
## Description

I have added more type hinting, focusing mostly on models and activitypub where the hints are more challenging.

I also renamed some things to make going through the code a bit easier, for instance replacing `r` with `redis_instance` because I dislike single-letter variable names no matter how much others insist upon them. If this is disliked as part of this PR (because it's not strictly type-hinting) then I would be ~more than happy~ ~happy~ ~fine~ unhappy but acquiescent to remove it.

Some instances of using the new union syntax (`Union[TypeA, TypeB]` is equivalent to `TypeA | TypeB`)

serialization and deserialization (especially around activitypub) is woefully bound to stay as `dict[str, Any]` unless considerable effort is made to refactor it into a more solid structure

The change of `FormatChoices` to `FORMAT_CHOICES` is to adhere to PEP-standard casing that module-level constants should be `CAPS_SNAKE_CASE`, so not an important change if someone pushes back on it.

It is worth noting that the best typing for `unfurl_related_field` would be not `Union[A, B]` (either A _or_ B) but `Intersection[A, B]` (type that inherits from both A _and_ B). Such a feature has been talked about for [at least a decade now](https://github.com/python/typing/issues/213), the RFC is tracked [here](https://github.com/CarliJoy/intersection_examples/issues/53), but as of this PR there is no perfect way to represent a type hint that is a subclass of two specified classes. I look for an alternative for far too long. If the function is part of said class one can use `Self`, which I made liberal use of in `fields.py` for each` field_from_activity`. For `unfurl_related_field`, though, a Union is the best I could come up with.

There is a circular dependency trap between 
- `base_activity` 
  - which wants to use **base_model**.BookWyrmModel for conversions in things like `to_model`
- `base_model`
  - which wants to use `RemoteIdField` from `fields`
- `fields`
  - which wants to type hint the overloads to `set_field_from_activity` that takes a `BookWyrmModel` instance as input

In the end I used [ForwardReferences](https://peps.python.org/pep-0484/#forward-references) in `fields` to break the circular dependency, but if the whole repository was type hinted I suspect there would be some more.

I also used some forward references if models A and B are defined in the same file (Book & Edition, for instance) and the type hinting for A wants to use B before B is defined. The alternative to using forward references would be moving around the order of the classes defined in the file, which felt like too much of a change for the nice-to-have that is type hinting.

some types I had to go to the Django source code for, because Django themselves doesn't type hint nearly as well as I would like, and the library that helps by adding in Django type hinting isn't a flawless drop-in solution. Sometimes even that just turns an `Any` into a `dict[str, Any]` when (for example) the "context" object is really just a dictionary that anything can be thrown into.
Also it wasn't documented well in Django's documentation that the default `id` fields on Models, when not provided explicitly, is in fact an integer. You'd think that would be called out somewhere but no.

I also corrected some of my previously incorrect type hints (some ids that I thought were `str` but were `int` because they were default model `id` fields that had been renamed after passing through a function or two). Also some of my corrections were because of multiclassing, I had correctly identified and type hinted for one class, when in fact a parameter needed to be a Union, usually including a mixin.

- Related Issue #
- Closes #

- [ ] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


PatrickChildersIT:
image

_excuse me!?_ this is a _type hint_, CodeQL????

If I make it `ContentFile[bytes]` CodeQL yells that it is _somehow_ a request forgery, but if I make it `ContentFile` without the generic type parameter then `mypy` complains about it. I can't win here, so I'm just going to change it to `Any` 🙄

PatrickChildersIT:
ok I don't know how to make CodeQL happy with this, my change clearly has nothing to do with request forgery, so someone else is going to have to appease the mighty CodeQL

mouse-reeve:
I dismissed it. Thanks for the type hinting work!

bookwyrm/activitypub/base_activity.py
```diff
@@ -356,7 +367,7 @@ def set_related_field(
         item.save()
 
 
-def get_model_from_type(activity_type):
+def get_model_from_type(activity_type: str) -> type[TBookWyrmModel]:
```
mouse-reeve, line 370: why use `TBookWyrmModel` here and `base_model.BookWyrmModel` on line 411 below?

mouse-reeve: I guess I'm also not sure what `type[]` is doing here 🤔

bookwyrm/activitypub/base_activity.py
```diff
@@ -475,15 +486,15 @@ def get_representative():
     )[0]
 
 
-def get_activitypub_data(url):
+def get_activitypub_data(url: str | bytes) -> Any:
```
mouse-reeve, line 489: When does this take bytes?

bookwyrm/models/activitypub_mixin.py
```diff
@@ -92,7 +96,7 @@ def find_existing_by_remote_id(cls, remote_id: str) -> Self:
         return cls.find_existing({"id": remote_id})
 
     @classmethod
-    def find_existing(cls, data):
+    def find_existing(cls, data: dict):
```
mouse-reeve, line 101: Is there a way to indicate the return value here as `cls` (or `None`, but I know how to indicate that)?

bookwyrm/models/activitypub_mixin.py
```diff
@@ -392,7 +403,7 @@ def to_activity(self, **kwargs):
             self.collection_queryset, **kwargs
         ).serialize()
 
-    def delete(self, *args, broadcast=True, **kwargs):
+    def delete(self, *args, broadcast=True, **kwargs) -> None:
```
mouse-reeve:
```suggestion
    def delete(self, *args, broadcast: bool = True, **kwargs) -> None:
```

bookwyrm/models/activitypub_mixin.py
```diff
@@ -341,8 +347,13 @@ def collection_remote_id(self):
         return self.remote_id
 
     def to_ordered_collection(
-        self, queryset, remote_id=None, page=False, collection_only=False, **kwargs
-    ):
+        self,
+        queryset: QuerySet,
+        remote_id: str = None,
+        page=False,
```
mouse-reeve: Should these also be typed `bool`?

mouse-reeve (changes requested):
I agree with your philosophical takes here, and all my comments below are genuine questions. Once I got to `activitypub_mixin.py`, I feel like I may have lost the plot a little -- it seems like there are things in there that still need to be typed?

My humble request is, could you remove all the files after [bookwyrm/models/activitypub_mixin.py](https://github.com/bookwyrm-social/bookwyrm/pull/4153/changes#diff-f99ea5a624a193ccbb8e103f8bf7c884e23f069aba0054766e9a09c966904e53) and open PRs with no more than 5 files changed? I want to make sure I understand and meaningfully review these (especially since I don't have a lot of experience with type hinting, and a diff this long is too hard for me to manage

bookwyrm/activitypub/base_activity.py
PatrickChildersIT: Type hints describe types, but a `Type` is itself an object of type `Type`

if something outputs an `int` it outputs an object of type `int`, but if something outputs a `type[int]` it outputs the `Type` object `int`. to unapologetically cite [the python documentation](https://docs.python.org/3/library/typing.html#the-type-of-class-objects):

```
a = 3         # Has type ``int``
b = int       # Has type ``type[int]``
c = type(a)   # Also has type ``type[int]``
```

and ultimately what is being returned is some class's `activity_serializer` which is a `Type`

PatrickChildersIT: to answer the original question, I think when I wrote this I actually didn't fully understand `TypeVar` bounding, but I read up some more to make sure I could answer accurately and it seems that using the `TypeVar` that is `bound` to `base_model.BookWyrmModel` any type hinting and type checkers preserve subclasses, that if you give it a, say, Edition to a function then `TBookWyrmModel` will know it's an `Edition` back out, but if you just used `base_model.BookWyrmModel` then the type hinting would just accept the base class, losing the fact that you explicitly hinted something as an `Edition`

which means for this file `TBookWyrmModel` should be used when the type hinting is really more about the subclasses, and `base_model.BookWyrmModel` when the type hinting is tell you you're just using the variable as its base class, and the function doesn't care that what it was using was actually a subclass

when I update the PR to remove all the extra model changes (to separate them into their own PRs) I'll do another once-over on this file for this distinction

bookwyrm/activitypub/base_activity.py
PatrickChildersIT: it doesn't, but it could
the first parameter to `requests.get` _can_ accept either string or bytes.

bookwyrm/models/activitypub_mixin.py
PatrickChildersIT: I _believe_ you could do it with a typevar alias bound to the mixin with a forward reference string

```python
TActivitypubMixin = TypeVar('TActivitypubMixin', bound='ActivitypubMixin')
```

and then `cls` would be typed as `cls: type[TActivitypubMixin]`

the _return value_ here, though, are not _types_ but the result of `select_subclasses` (filtered and matched but still) and I believe that function returns instances
based on [this page](https://django-model-utils.readthedocs.io/en/latest/managers.html#inheritancemanager)
which provides bits of example code and says (emphasis mine) "the `select_subclasses()` method on the `InheritanceManager` or any `QuerySet` from it, and **the resulting objects will be instances of** `Restaurant` or `Bar`"

so the return type would probably then just be `TActivitypubMixin`

now, if you might ask how to go a step further and ensure the type checker knows how to associate the static `TActivitypubMixin` bound type with the Django-related filtered and selected model that is the result of `select_subclasses`? I have absolutely. no. idea. Good luck on that one, but I feel there be dragons.

regardless, I hadn't added typing to `cls` becuase I wasn't sure static class variables even _should_ be type hinted. If you feel it is beneficial I'll add it here.

bookwyrm/models/activitypub_mixin.py
PatrickChildersIT: yes. good spot.

PatrickChildersIT (commented):
follow up on some comments before I get to changing the code

bookwyrm/models/activitypub_mixin.py
mouse-reeve: Thanks for explaining! I don't think you need to change this, I'm just making sure I get the reasoning

PatrickChildersIT:
> My humble request is, could you remove all the files after [bookwyrm/models/activitypub_mixin.py](https://github.com/bookwyrm-social/bookwyrm/pull/4153/changes#diff-f99ea5a624a193ccbb8e103f8bf7c884e23f069aba0054766e9a09c966904e53) and open PRs with no more than 5 files changed? I want to make sure I understand and meaningfully review these (especially since I don't have a lot of experience with type hinting, and a diff this long is too hard for me to manage

ok I have whittled this PR down, let me know if it needs to be even smaller

bookwyrm/models/activitypub_mixin.py
```diff
@@ -492,8 +507,8 @@ def to_undo_activity(self):
         ).serialize()
 
 
-def generate_activity(obj):
-    """go through the fields on an object"""
+def generate_activity(obj: ActivitypubMixin) -> dict[str, Any]:
+    """go through the fields on an object"""  # and do what?
```
mouse-reeve, line 511:
```suggestion
    """create a dict from an ActivityPub model object"""
```

mouse-reeve (approved):
This is great, thank you so much for digging in to one of the gnarliest parts of the codebase!

Issue #4155, opened 2026-09-10
A round-up of deduplication bugs


mouse-reeve:
I'm merging duplicates on b.s and finding bugs! Making one issue to catalogue them.

- [ ] Only author and edition are being scanned, not work or series
- [ ] Long wikidata url makes the merge form so long you can't find the button to perform the merge
- [ ] This book is mis-marked and the matching fields can't be edited: https://bookwyrm.social/author/9062/s/pierre-christin
- [ ] The counts are different on the Data Quality view vs the Merge Data view (7,380 authors vs 6,698)
- [ ] Sorting duplicates by ID in the Data Qaulity produces more interesting items first
- [ ] We should let users edit work identifiers
- [ ] ISNI links seem not to work (98% confident this is their error not ours)
- [ ] Maybe we should store multiple openlibrary keys for one item
- [ ] When I merge an author from the Merge Data admin view, I'm redirected to the edition list
- [ ] It would be nice to get an alert when ISBN _doesn't_ match but other identifiers do
- [ ] Error related to covers (https://bookwyrm.social/book/134133/s/hitchhiker-trilogy):
<details>

```
web-1  | Internal Server Error: /settings/manage-data/confirm-manual-merge/edition/26525                                                                                                                
web-1  | Traceback (most recent call last):                                                                                                                                                             
web-1  |   File "/venv/lib/python3.11/site-packages/django/core/handlers/exception.py", line 55, in inner                                                                                               
web-1  |     response = get_response(request)                                                                                                                                                           
web-1  |                ^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                           
web-1  |   File "/venv/lib/python3.11/site-packages/django/core/handlers/base.py", line 197, in _get_response                                                                                           
web-1  |     response = wrapped_callback(request, *callback_args, **callback_kwargs)                                                                                                                    
web-1  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                    
web-1  |   File "/venv/lib/python3.11/site-packages/django/views/decorators/http.py", line 64, in inner                                                                                                 
web-1  |     return func(request, *args, **kwargs)                                                                                                                                                      
web-1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                      
[...]

PR #4160, opened 2026-09-11
Merging fails on duplicates that will appear twice on someone's shelves

mouse-reeve:
## Description
A major cause of errors when merging duplicates is caused by a user who has both versions (canonical and dupe) on the same shelf. The uniqueness constraint of the shelfbook model prevents removing the foreign key reference to the dupe, and thus it cannot be deleted because it of the `ondelete=PROTECT` status of the field. 

This change has the shelf book model cascade rather than protect, which means that the duplicate shelfbook is deleted. Since the only metadata stored on shelfbook is shelved date (other reading dates are ReadThroughs), I don't think this is a problematic deletion in terms of user data.

- Related Issue #4155 
- Closes #

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


mouse-reeve:
Please sanity check me that this is the right fix

hughrun (approved):
I think this looks right. So the original book is deleted, but before that we update the ShelfBook to use the canonical. And if there is already a ShelfBook with that book (i.e. a duplicate ShelfBook I guess), an IntegrityError will be raised but we ignore it, and the ShelfBook with the candidate is deleted via CASCADE. Is that right?

mouse-reeve:
Yeah, that's right

PR #4169, opened 2026-09-12
Display and edit work metadata 

mouse-reeve:
## Description
Show the work's metadata on the list of editions page, and make the work editable.
Screenshot 2026-09-12 at 10 10 38 AM

- Related Issue #4155 
- Closes #

- [ ] Bug Fix
- [x] Enhancement
- [ ] Refactor

## Documentation

### Tests

### DIFF
--- a/bookwyrm/views/books/edit_book.py
+++ b/bookwyrm/views/books/edit_book.py
@@ -39,28 +39,32 @@ class EditBook(View):
     def get(self, request, book_id):
         """info about a book"""
         book = models.Book.objects.select_subclasses().get(id=book_id)
-        if isinstance(book, models.Work):
-            seriesbooks = book.seriesbooks.all()
-        else:
-            seriesbooks = book.parent_work.seriesbooks.all()
+        is_work = isinstance(book, models.Work)
+        seriesbooks = (
+            book.seriesbooks.all() if is_work else book.parent_work.seriesbooks.all()
+        )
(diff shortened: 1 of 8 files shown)


bookwyrm/views/books/edit_book.py
hughrun: Adding a new edition from the work page is showing a blank form.

```suggestion
    data = {
        "book": book, 
        "model": "edition", 
        "form": forms.EditionForm(request.POST)
    }
```

hughrun (changes requested):
This is cool and I like it a lot.

Just wondering about the "Edit book" on the book landing page. Is there anything in the edition form that flows through to the work? I really like the idea of separating these (at least Edition > Work) but the terminology we use is a bit confusing. We say "edit book" but ideally you would only ever be editing an _edition_, unless you're on the work/editions page. Which then makes me wonder if "n editions" should have a clearer indication that it links to a page where you can edit the parent work.

We should keep it simple for people, but also not confusing. e.g. there can be a Libris ID on both the edition and the work, and they are presumably different. Same for Inventaire and some other identifiers. It would be good for people to be clear which one they're editing.

mouse-reeve:
Yeah I think you’re right about clarifying the language around book/work/edition. We can probably also add some help text clarifying the differences mid form

hughrun (approved):
LGTM!

PR #4174, opened 2026-09-14
Fix/remote fetch SSRF

noamaanMulla-03:
## Description

BookWyrm accepts ActivityPub messages from remote servers. To verify a message’s HTTP signature, it first needs to retrieve the claimed sender’s public key from the message’s `actor` URL.

Previously, that URL was fetched before the signature was confirmed. Because an attacker controls the URL in an untrusted message, they could attempt to make the BookWyrm server request a private service reachable only from its own network, such as an internal admin service or container-only endpoint.

The inbox message would still fail signature verification afterwards, but the server-side request would already have happened. Remote connector, image, and import-origin fetches also lacked complete hostname and redirect validation.

The fix introduces a reusable remote-request helper and uses it for ActivityPub profile retrieval, connector data, remote images, and import-origin checks.

Before connecting, the helper:

- allows only HTTP(S) URLs with hostnames;
- rejects direct IP-address URLs and URLs containing credentials;
- resolves hostnames and rejects loopback, private, link-local, and other non-public addresses;
- disables automatic redirects and validates each redirect destination before following it;
- limits redirect chains.
- Validated public DNS results are cached locally in each worker for 30 seconds, with a maximum of 1,000 entries. 

Connector-search requests now reject redirects instead of following them. ActivityPub requests are re-signed for each allowed redirect destination, since the destination URL is part of the HTTP signature.

This prevents BookWyrm from making requests to obvious internal/private destinations supplied by an attacker. Production deployments should also restrict outbound container access to internal and cloud-metadata networks as defence in depth.

Existing remote-request fixtures were updated to provide mocked public DNS results before their mocked HTTP responses are used.

- Related Issue # N/A — GitHub Security alert [GHSA-5m7g-66h6-5cvq — Server-side request forgery](https://github.com/bookwyrm-social/bookwyrm/security/advisories/GHSA-5m7g-66h6-5cvq)
- Closes # N/A

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

- Connector search endpoints that depend on HTTP redirects will no longer be followed. Configure connectors with their final destination URL instead.
- Adds `cachetools==7.1.7` for the bounded per-process DNS-validation cache.

## Documentation

### Tests


bookwyrm/utils/remote_requests.py
```diff
@@ -0,0 +1,103 @@
...
+    params: Optional[dict[str, str]] = None,
+    method: Literal["get", "head"] = "get",
+    validate_url: Callable[[str], None] = validate_remote_url,
+) -> requests.Response:
+    """Request a remote URL, validating the initial URL and every redirect."""
+    current_url = url
+    current_params = params
+
+    for _ in range(MAX_REMOTE_REDIRECTS + 1):
+        validate_url(current_url)
```
ilkka-ollakka, line 102: Might be good to cache this littlebit? I think ttl values are atleast 10-30s.

ilkka-ollakka (commented):
With quick check, looks ok. Didn't yet looked more indepth.

bookwyrm/utils/remote_requests.py
noamaanMulla-03: Good idea, I'll add a per process cache for validated dns lookups with a 30s ttl.

bookwyrm/utils/remote_requests.py
```diff
@@ -12,12 +13,48 @@
 
 
 MAX_REMOTE_REDIRECTS = 3
+REMOTE_URL_CACHE_TTL = 30
+REMOTE_URL_CACHE_MAX_SIZE = 1000
+
+_remote_url_cache: dict[tuple[str, int], tuple[float, set[str]]] = {}
```
ilkka-ollakka: you could use TTLCache directly with timedelta. https://cachetools.readthedocs.io/en/latest/#cachetools.TTLCache

or redis cache with expire.

noamaanMulla-03: Initially went with Redis since it was an existing dependency but had to switch to TTLCache since implementing a 1000 entry limit only to this hostname-validation cache would've had an effect on other dependent services.

ilkka-ollakka (approved):
I think this looks good. Didn't spot anything that would break with connectors.

Would be great for @hughrun or @mouse-reeve to check also, as getaddrinfo might work slightly differenly in OS X than in Linux.

ilkka-ollakka:
Would be nice to cache the dns according to ttl value, but it is somewhat non-trivial and can be done in the future.

bookwyrm/utils/remote_requests.py
```diff
@@ -0,0 +1,131 @@
...
+from typing import Literal, Optional, cast
+from urllib.parse import urljoin, urlsplit
+
+from cachetools import TTLCache
+import requests
+
+from bookwyrm.utils.ip_utils import check_ip_routable
+
+
+MAX_REMOTE_REDIRECTS = 3
```
hughrun: It might be good to allow this to be overridden with a `.env` value just in case a valued node temporarily has extra redirects for some reason and an instance admin wants to maintain contact.

bookwyrm/tests/activitypub/test_base_activity.py
```diff
@@ -46,6 +47,15 @@ def setUpTestData(cls):
         cls.user.save(broadcast=False, update_fields=["remote_id"])
 
     def setUp(self):
+        getaddrinfo = patch(
+            "bookwyrm.utils.remote_requests.socket.getaddrinfo",
+            return_value=[
+                (socket.AF_INET, socket.SOCK_STREAM, 6, "", ("93.184.216.34", 443))
```
hughrun: I realise this is the example IP address used in the python `socket` docs and allegedly the one used by example.com but it appears to be a real address. For testing and docs we should be using [one of the addresses reserved for that](https://en.wikipedia.org/wiki/List_of_reserved_IP_addresses). e.g. `198.51.100.34`

hughrun (changes requested):
This looks great to me. Just a couple of suggestions.

I'm a little unclear what will happen if we're using `localhost` and non-standard ports for testing, however. Outbound requests from localhost seems fine now, but once all the production peers are using this code will that still be the case?

PR #4177, opened 2026-09-14
Revert "compose: add healthcheck to redis instances and wait for them"

mouse-reeve:
Reverts bookwyrm-social/bookwyrm#4092

This change seems to have made it impossible for redis_activity to recover from an OOM error


mouse-reeve:
I suspect the timeout windows were way to short on this? It seems like redis was unable to restart properly and got stuck in a loop

ilkka-ollakka:
I assumed that 3min would be enough, but I agree that it is better to revert for now.

PR #4183, opened 2026-09-19
Fix: `TypeError: Model.save() got an unexpected keyword argument 'name'` wh

### LINKED ISSUE #4182: `TypeError: Model.save() got an unexpected keyword argument 'name'` when sving series with args
**Describe the bug**
I might just be doing something silly but I think there may be a problem with `MergeableMixin` when passing in parameters that aren't on models.Model.

**To Reproduce**
1. Save a series using `Series.save(**data)`
2. Get error: `TypeError: Model.save() got an unexpected keyword argument 'name'`

**Expected behavior**
A newly saved Series

**Screenshots**

**Instance**
dev

**Additional context**
I'm writing a PR to fix the borked series imported from Inventaire.

@mouse-reeve any ideas here?

barleslie-rgb:
### Autonomous Bug Fix

Closes https://github.com/bookwyrm-social/bookwyrm/issues/4182

**Summary of Fix:**
The bug is that when creating a new Series using the `Series.save` method, it is passing in a keyword argument 'name' which is not present in the `models.Model` class. This results in a `TypeError`. The fix is to ensure that only valid keyword arguments are passed to the `save` method, adhering to the model's attribute names.

Automated verification: Python bytecode compilation passed.


mouse-reeve:
Hi there, we don't accept LLM/AI pull requests

Issue #4143, opened 2026-09-08
Historical Book model has duplicate _precision fields, breaking `save()` in data migrations


qbi:
**Describe the bug**
While upgrading to the latest git version I stumbled across some issues. To me this are three different bugs, so I will create three issues (#4144 and #4145). This is the first.

When running the migrate script I ran into the error message:

> multiple assignments to same column "first_published_date_precision"

The reason seems to be the `save()` on a `Book`, `Work` or `Edition`. The historical model carries the `_precision` fields twice. They are mapped to the same database column, so the UPDATE compiler emits a duplicate `SET` clause.

In normal operations nothing breaks, the problems occurs in migrations only.

This issue was introduced with https://github.com/bookwyrm-social/bookwyrm/blob/main/bookwyrm/migrations/0187_partial_publication_dates.py and it blocks future data migrations that need to save these models. 

Here a `CharField` is attached to every partial date field.
https://github.com/bookwyrm-social/bookwyrm/blob/7962f57cee2261998c77346fbea557032e1b372c/bookwyrm/utils/partial_date.py#L232-L245

The migration script records the same as explicit `AddField` operation. 
https://github.com/bookwyrm-social/bookwyrm/blob/7962f57cee2261998c77346fbea557032e1b372c/bookwyrm/migrations/0187_partial_publication_dates.py#L29-L48

On the real class `contribute_to_class()` runs once and produces one field.
On the historical model the field arrives twice: once from the `AddField` state, and again when `contribute_to_class()` runs as Django renders the class from that state. 

I created a workaround by changing https://github.com/bookwyrm-social/bookwyrm/blob/7962f57cee2261998c77346fbea557032e1b372c/bookwyrm/migrations/0243_auto_20260730_0926.py#L39-L41
to
```
Work.objects.filter(pk=work.pk).update(series=None, series_number=None)
```
However this seems more like WFM solution and I'm not sure what the right fix would be.

**Instance**
https://bookwyrm.de/ running on latest git (7962f57ce). 

**Additional context**
Error message after the migration:

```
psycopg.errors.SyntaxError: multiple assignments to same column "first_published_date_precision"

  File "bookwyrm/migrations/0243_auto_20260730_0926.py", line 42, in fix_series
    work.save()
  File ".../django/db/models/base.py", line 1058, in _save_parents
    updated = self._save_table(
  ...
django.db.utils.ProgrammingError: multiple assignments to same column "first_published_date_precision"
```

[...]

PR #4184, opened 2026-09-20
Series fix command

### LINKED ISSUE #4144: Migration 0243 crashes on some instances and repairs nothing on the rest
**Describe the bug**
When running the migration script I ran into two errors:
1. `save()` contained `broadcast=False` which was corrected in 0e07eeb2f7450b40789bdb8827c1d05e24faca3b.
2. When the issue from #4143 is corrected, the migration script runs, but doesn't correct anything. Output was something like "Fixed 0 of 221 works and 0 of 72 editions with suspected stringified JSON series entries".

The script parses the corrupted `series` entries with `json.loads()`, but the data looks more like Python's `repr` (Probably someone used `str(list)` to write to `CharField` instead of json.dumps(list)`):
```
>>> work = Work.objects.filter(series__startswith="[").first()
>>> print(repr(work.series)[:120])
'[{\'name\': \'The Murderbot Diaries\', \'alternative_names\': [\'マーダーボット・ダイアリー・シリーズ\', ...
>>> json.loads(work.series)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)
```

On my instance `json.loads()` failed on 220 of 220 entries. Later I used `ast.literal_eval()` which could parse 216 of 220. The four remaining contained brackets:
```
123227 '[Die Vanitas-Reihe]'
90477  '[Hildur]'
122396 '[The detective Kaga mysteries]'
104396 '[Hildur]'
```

So 216 could be, but were not repaired. The 72 edition all contained brackets.

Fixing the parser alone is not enough. With `ast.literal_eval()` the migration reaches the connector for the first time, and `get_or_create_seriesbook_from_data()` calls `work.save()` on the historical model ([abstract_connector.py:230](https://github.com/bookwyrm-social/bookwyrm/blob/7962f57cee2261998c77346fbea557032e1b372c/bookwyrm/connectors/abstract_connector.py#L230[())), which fails as described in #4143:

```
File "bookwyrm/connectors/abstract_connector.py", line 230, in get_or_create_seriesbook_from_data
    work.save()
django.db.utils.ProgrammingError: multiple assignments to same column "first_published_date_precision"
```

So with `json.loads()` the migration completes and does nothing; with `ast.literal_eval()` it crashes. Both fixes are needed.

I would suggest 
1. using `ast.literal_eval()` instead of `json.loads()`. It should catch `ValueError` and `SyntaxError` instead of `JSONDecodeError`,
2. using `update()` instead of `save()` in childless and orphaned branches,
3. when looking at #4143 it might be good to not hand historical models into the connector
4. create a new migration, because 0243 ran "successful", but some instances have corrupted data and no way to re-run the repair..

I repaired the data on my instance. 

The migration script `0210` seems to have the same issue. I would suggest looking into it.

**Expected behavior**
Successfull and correct finish of the migration script. So the script should repair the 216 corrupted entries.

**Instance**
https://bookwyrm.de/ running on latest git (https://github.com/bookwyrm-social/bookwyrm/commit/7962f57cee2261998c77346fbea557032e1b372c).

### LINKED ISSUE #4145: Connector creates Series records with empty or serialised names
**Describe the bug**
Some Inventaire data carry only `alternative_names`, `inventaire_id` and `wikidata` and no `name`:
```
"[{'alternative_names': ['El mar'], 'inventaire_id': 'wd:Q27780705', 'wikidata': 'Q27780705'}]"
"[{'alternative_names': ['Die Gretchen-Reihe'], 'inventaire_id': 'wd:Q136363528', 'wikidata': 'Q136363528'}]"
```
The code has two branches:
https://github.com/bookwyrm-social/bookwyrm/blob/7962f57cee2261998c77346fbea557032e1b372c/bookwyrm/connectors/abstract_connector.py#L174-L186

When `work.series` is a list of dicts, `models.Series(**data)` is called for every dict. But when there is no `name`, `name` stays empty. When `work.series` is a string, the string is the name. However in the case of #4144  a stringified object is written. So I had one occurrence with a 2000 character long `repr` expression in my data.

Those corrupted data didn't come from the migration (as I wrote in #4144 it did nothing on my instance), but either from some import or another source.

I would suggest to check if a `name` is in the dict. If not, don't create the entry or use `alternative_names`. Furthermore one could check if a string starts with a `[` and is some Python literal then it is not a title. 

**To Reproduce**
Find affected records with:
```python
from bookwyrm import models

print("empty names:", models.Series.objects.filter(name="").count())
for s in models.Series.objects.filter(name=""):
    print("  ", s.pk, "SeriesBooks:", models.SeriesBook.objects.filter(series=s).count())

print("serialised names:", models.Series.objects.filter(name__startswith="[").count())
for s in models.Series.objects.filter(name__startswith="["):
    print("  ", s.pk, repr(s.name)[:100])
```

Out was on my instance:
```
empty names: 3
   161 SeriesBooks: 1
   213 SeriesBooks: 1
   194 SeriesBooks: 1
serialised names: 1
   160 "[{'alternative_names': ['แฮร์รี่ พอตเตอร์', 'هيري پوٽر', 'Harri Potter', ...
```

**Expected behavior**
A `Series` record should always have a usable name or not be created at all.

**Instance**
https://bookwyrm.de/ running on latest git (https://github.com/bookwyrm-social/bookwyrm/commit/7962f57cee2261998c77346fbea557032e1b372c).

hughrun:
## Description

Creates a new admin command to fix stringified JSON series names created by previous bug.

I _think_ this will actually fix it this time. An admin command is a bit more flexible than a data migration but will require admins to actually run it.

This will need to go in the release notes.

- Related Issue #4143 (this is now irrelevant because the migration file has been removed)
- Closes #4144 
- Closes #4145 

- [x] Bug Fix
- [ ] Enhancement
- [ ] Refactor

## Documentation

### Tests


hughrun:
@qbi do you mind taking a look at this?

bookwyrm/management/commands/repair_series.py
```diff
@@ -0,0 +1,273 @@
...
+            return f"ERROR fixing nameless series id: {series.id} error: Can't find series name on Inventaire"
+
+        series.name = data["name"]
+        series.alternative_names = list(
+            set(series.alternative_names + data["alternative_names"])
+        )
+        for field in Series._meta.get_fields():
+            if hasattr(field, "deduplication_field") and field.name in data:
+                setattr(series, field.name, data[field.name])
+        series.save()
```
hughrun: oop, just realised this need `update_fields` set.

bookwyrm/management/commands/repair_series.py
```diff
@@ -0,0 +1,273 @@
...
+        return series
+
+    elif options["all"]:
+        # there is no inventaire_id so we can't just re-fetch the series
+        # try using the first alternative name if we have one
+        if not series.alternative_names:
+            return f"ERROR fixing nameless series id: {series.id} error: No names or inventaire id"
+
+        series.name = series.alternative_names[0]
+        series.save()
```
hughrun: and here too

bookwyrm/management/commands/repair_series.py
```diff
@@ -0,0 +1,278 @@
...
+    if series.inventaire_id not in ["", None]:
+        series_list = load_connector().format_series(
+            keys=[f"https://inventaire.io/entity/{series.inventaire_id}"]
+        )
+        if not series_list or len(series_list) < 1:
+            return f"ERROR fixing nameless series id: {series.id} error: Can't find series data on Inventaire"
+
+        data = series_list[0]  # we only passed in one series key
+        if "name" not in data:  # let's double check!
+            return f"ERROR fixing nameless series id: {series.id} error: Can't find series name on Inventaire"
```
ilkka-ollakka, line 155: might be better to raise ValueError with those texts and handle those in caller. Also typehint for return-values should contain Series as it should return it in all-good-path?

qbi:
> @qbi do you mind taking a look at this?

Will do, but it will take some time (probably not until next week).

